PR #12, opened 2017-10-25
Fix enum values to PascalCase and replace return type of a method returning a stream from RedableStream to Response

amarzavery:


olydis:
~~~ Haskell
> will publish once PR gets merged
~~~

olydis:
> win32-x64 restart

PR #43, opened 2018-03-29
XML generator support

RikkiGibson:
- Adds XML metadata to mappers
- Adds some inline logic related to root lists (if you see `response.ModelType is SequenceType`, that's what that's about). For deserialization this is just an inline check but for serialization we use the method msRest.prepareXMLRootList.
- Calls into msRest.stringifyXML instead of JSON.stringify as appropriate
- Adds a heaping helping of XML tests based on recent changes to autorest.testserver


olydis:
~~~ Haskell
> will publish once PR gets merged
~~~

Issue #40, opened 2018-03-25
Determine what types to use for stream request/response bodies on browser and Node


RikkiGibson:
I noticed the generated return type for getFile() is a `Response` type which requires we compile with the `lib: ["dom", ...]` argument in tsconfig.json. I don't think this is quite right as I can see that the [test case](https://github.com/Azure/autorest.typescript/blob/2cac0b1efd1e4a213d49fed03f11316f486828a0/test/vanilla/AcceptanceTests/acceptanceTests.ts#L1809) requires type asserting the response as `any`.

At a glance it looks like the `"dom"` lib adds all the DOM-only APIs to the global scope. We probably don't want that in an isomorphic lib, right? I'm not going to track that down yet as part of my "make CI pass" work, but we'll want to address it at some point because good server and browser interchangeable APIs for streaming I/O will be needed.

RikkiGibson:
Consider this resolved by Azure/ms-rest-js#60.

Essentially, Blob for the browser, and ReadableStream for node.js.

PR #47, opened 2018-04-09
Browser tests

RikkiGibson:
- Some tests against error messages have been changed because they're not the same between different browsers
- A test related to HTTP redirects is skipped here because it's failing in Chrome--it passes in Safari, so I suspect it might be a CORS-related bug in Chrome.
- I'm currently including shims for `process` and `Buffer` in webpack config. I think this bloats the bundle size so let's consider alternatives here. I fiddled around with the "buffer" npm package quite a bit but seemed to have trouble making it work for whatever reason.
  - Consider this related to the stream work in #40
- We might want to use `ms-rest-js/msRestBundle.js` as a global in webpack so we can test the actual bundle we expect to ship to users. For dev I wasn't doing this because it makes debugging easier to gather up the source files for all the dependent libraries ourselves.
  - IMO we might as well just do this further down the line after we have solutions for all the currently disabled browser tests.
- Going into package.json for ms-rest-js and changing the `main` to `lib/msRest.ts` caused me to get a source map to the original .ts file instead of the compiled .js file. This improved the debugging experience in the browser significantly. I would like to figure out a way to get this level of convenience without having to muck around in all the package.json files so much.
- Need to merge other PRs before this will pass CI:
  - Azure/ms-rest-js#34
  - Azure/autorest.testserver#63

Usage tips:
- When developing, pop open a terminal window and run `./node_modules/.bin/webpack-dev-server` in the background. This will make the process of rerunning the browser tests after changes much faster. When the dev server is running, you can just navigate to `http://localhost:8080` to run the browser tests.
- The tests are passing in Chrome (with a meaningful number of them simply skipped) but several more tests are failing in Safari and Firefox. Would like to get tests passing on Edge as well.


.gulp/gulpfile.iced
```diff
@@ -21,14 +23,34 @@ task 'init', "" ,(done)->
...
   await execute "tsc -p ./test/tsconfig.generated.json", defer _
   done();
 
-task 'test', 'run unit tests', [], (done) ->
+task 'test/nodejs-unit', 'run nodejs unit tests', [], (done) ->
   await execute "mocha", defer _
   done();
 
+task 'test/chrome-unit', 'run browser unit tests', [], (done) ->
+  count = 2;
```
daschult: Yay for magic numbers! (Maybe a quick comment explaining this value might be a useful addition)

.gulp/gulpfile.iced
```diff
@@ -21,14 +23,34 @@ task 'init', "" ,(done)->
...
+    if count == 0
+      runMochaChrome(testServer, webpackDevServer, done)
+
+  runMochaChrome = () ->
+    await execute "./node_modules/.bin/mocha-chrome http://localhost:8080", defer _;
+    testServer.kill();
+    webpackDevServer.kill();
+
+  testServer.stdout.on('data', onData)
+  webpackDevServer.stdout.on('data', onData)
```
daschult: Since you're passing onData to both of these, does that mean that you're just waiting for both of them to finish, and then you're going to runMochaChrome after that?

.gulp/gulpfile.iced
RikkiGibson: Will update this to explain to readers what's going on

.gulp/gulpfile.iced
RikkiGibson: When the server writes to stdout, we assume it's ready to receive HTTP requests.

RikkiGibson:
Closing in favor of #53.

PR #83, opened 2018-05-16
[Do not merge] Monorepo

RikkiGibson:
For now this is basically an experiment to see if we can bundle the ms-rest-js and ms-rest-azure-js projects right into this repo, because the development and publishing of these 3 packages is currently so tightly coupled together. No plans to merge this until we can get together and gauge the cost/benefit as a team.

Right now we have to do a tricky sequence of PRs using GitHub references to check our work as we go. We have to then merge all the interdependent PRs right away and publish the packages (we forget this step sometimes).

This approach takes away our ability to reference the generator in another package.json with a GitHub branch. It takes a long time to build when you do it that way anyway, though. It also takes this feature away for ms-rest-js, which is kind of unfortunate but might be worth the workflow improvements.


RikkiGibson:
Doesn't look like we're moving forward with this--will instead try to focus on continuously publishing ms-rest[-azure]-js and keep it clear for users which versions of the generator and runtime are meant to work together.

PR #95, opened 2018-05-22
Move header parameters and custom headers to runtime

daschult:


src/vanilla/Templates/MethodJsonPipelineTemplate.cshtml
```diff
@@ -10,34 +11,62 @@ try {
...
 <text>
-    httpMethod: "@(Model.HttpMethod.ToString().ToUpper())",
-    requestBodyMapper: requestModelMapper,
+      headerParameters: [
+@foreach (ParameterTS parameter in Model.LogicalParameters.Where(p => p.Location == ParameterLocation.Header))
+{
+<text>
+        {
+          parameterName: "@(parameter.Name)",
+@if (parameter.Name != parameter.SerializedName)
```
RikkiGibson: Can we fail fast in these kinds of cases? (a header parameter that doesn't specify the header name seems like a Swagger problem.)

daschult: This particular line isn't an error case, it's an optimization. In this scenario I'm saying that if the name of the parameter is the same as the name of the header, then don't add the header name because we can just use the parameter name for both.

daschult:
@RikkiGibson @amarzavery Any further comments?

src/vanilla/Templates/MethodJsonPipelineTemplate.cshtml
RikkiGibson: Yeah, just totally misread this yesterday 🤦‍♂️

src/vanilla/Model/ParameterTS.cs
```diff
@@ -47,5 +47,22 @@ public IEnumerable<Core.Model.Property> ComposedProperties
...
         public bool IsLocal => (!IsClientProperty) && !string.IsNullOrWhiteSpace(Name) && !IsConstant;
+
+        [JsonIgnore]
+        public string OperationType
+        {
+            get
+            {
+                string operationType = null;
+
+                IModelType modelType = ModelType;
```
RikkiGibson: minor thing, but I'd remove this local and the blank line on 57

daschult: Done in https://github.com/Azure/autorest.typescript/pull/95/commits/8a706f5e415a9d40e12642b1d3276f6f632cd831.

PR #103, opened 2018-05-23
Move generateClientRequestId to runtime

daschult:


RikkiGibson:
### Size decreased by 0.54% (-29.83 KB)

PR #106, opened 2018-05-24
Move request URL construction to runtime

daschult:


daschult:
Windows validation passed, but it failed on the bundle size reporting step.

PR #108, opened 2018-05-24
Move content-type header assignment to runtime

daschult:


RikkiGibson:
### Size decreased by 1.85% (-36.8 KB)

PR #114, opened 2018-05-25
Stream types

### LINKED ISSUE #60: Add Blob and stream support to generator

RikkiGibson:
Fixes #60. Depends on Azure/ms-rest-js#111.

### DIFF
--- a/src/vanilla/Model/MethodTS.cs
+++ b/src/vanilla/Model/MethodTS.cs
@@ -298,14 +298,17 @@ namespace AutoRest.TypeScript.Model
                 string result = "void";
                 if (ReturnType.Body != null)
                 {
-                    result = ReturnType.Body.TSType(false);
                     // We will return the actual response if the return type is stream.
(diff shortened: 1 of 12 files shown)


src/vanilla/Model/MethodTS.cs
```diff
@@ -298,14 +298,17 @@ public string ReturnTypeTSString
                 string result = "void";
                 if (ReturnType.Body != null)
                 {
-                    result = ReturnType.Body.TSType(false);
                     // We will return the actual response if the return type is stream.
                     // That provides better user experience as customers can use
                     // .text(), .json(), etc. inbuilt methods of the response class
                     // to read the stream.
-                    if (result.Contains("ReadableStream"))
+                    if (ReturnType.Body is PrimaryType pt && pt.KnownPrimaryType == KnownPrimaryType.Stream)
```
daschult: This condition seems to be duplicated a few times. Can you extract it somewhere and give it a name?

PR #120, opened 2018-05-25
Refactor operation generation, add TypeScript DSL classes

daschult:
Resolves https://github.com/Azure/autorest.typescript/issues/123.
The bulk of this PR is copying the DSL classes that I wrote during my TypeSpec phase. I find these constructs to be very useful because I can focus on generating the content of the TypeScript code without worrying about indentation, commas, parentheses, curly brackets, or square brackets (content over presentation). This also allows our generated code to get rid of a bunch of unnecessary double quotes and commas.
I also refactored the MethodJsonPipelineTemplate and MethodStreamPipelineTemplates. The code that sends the Operation request was identical between these two templates, so I extracted it to MethodTS and then used the DSL to create the TypeScript code.


test/azure/generated/AzureCompositeModelClient/azureCompositeModel.ts
```diff
@@ -144,17 +144,16 @@ class AzureCompositeModel extends msRestAzure.AzureServiceClient {
...
     try {
       operationRes = await client.sendOperationRequest(
         httpRequest,
         {
           arguments: {
             "this.subscriptionId": this.subscriptionId,
-            "resourceGroupName": resourceGroupName,
-            "apiVersion": apiVersion,
-            "this.acceptLanguage": this.acceptLanguage,
+            resourceGroupName: resourceGroupName,
```
RikkiGibson: Probably not for this PR, but one to keep in your back pocket: is it possible to generate merely `{ resourceGroupName }` instead of `{ resourceGroupName: resourceGroupName }` when the expression is a valid identifier and equal to the key? The space savings could add up.

daschult: That's definitely possible. Let me see if I can get it done here.

RikkiGibson:
The CI failure seems bogus to me.

RikkiGibson:
Pull request validation is enabled for the autorest.typescript Windows build in aka.ms/autorest-ci, so I'm really not sure.

daschult:
Resolves https://github.com/Azure/autorest.typescript/issues/123.

RikkiGibson:
Zero bytes of size change seems...surprising.

PR #128, opened 2018-05-30
Move basePath from OperationSpec.baseUrl to OperationSpec.path

daschult:
Resolves https://github.com/Azure/ms-rest-js/issues/113.


src/vanilla/Model/MethodTS.cs
```diff
@@ -59,6 +59,45 @@ public override Parameter Add(Parameter item)
             return parameter;
         }
 
+        [JsonIgnore]
+        public string SchemeHostAndPort => (CodeModel as CodeModelTS).SchemeHostAndPort;
```
RikkiGibson: If this expected to never fail I prefer a cast. Possibly a CodeModelTS property to contain that?

RikkiGibson:
Is this going to satisfy both the use cases for URL/basePath? Or is this just part of that work?

RikkiGibson:
Same CI error!

```sh
        Swagger LRO Happy BAT
          should work with Post200WithPayload:
      Uncaught AssertionError: expected '1' to be 1
```

Are you able to repro this locally @daschult ?

src/vanilla/Model/MethodTS.cs
daschult: Fair enough.

daschult:
@RikkiGibson No, all of my tests pass fine. This is weird.

daschult:
@RikkiGibson Yes, this should solve all of our basePath problems.

daschult:
@RikkiGibson Just repro-ed the failed test. Update the `@microsoft.azure/autorest.testserver` dependency to  2.5.12. Fails like a charm. I'll look into this.

RikkiGibson:
https://github.com/Azure/autorest.testserver/commit/4d11c867f46397a92a8497b7f8f7a7b726faff47

RikkiGibson:
I would just change the `1` to `"1"`.

RikkiGibson:
### Size decreased by 0.02% (-501 B)

PR #130, opened 2018-05-30
Run browser tests in CI

RikkiGibson:
I think it might be this simple. Otherwise just need to add a chrome install step to the Windows build definition.

### DIFF
--- a/.gulp/gulpfile.iced
+++ b/.gulp/gulpfile.iced
@@ -23,7 +23,7 @@ task 'init', "" ,(done)->
   done()
 
 # Run language-specific tests:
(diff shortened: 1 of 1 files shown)


RikkiGibson:
Interesting:
```sh
 [0m  1) typescript
        Swagger BAT
          Files Client
            browser should correctly deserialize large streams:
 [0m[31m     Error: Timeout of 50000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.[0m[90m
```
Will probably put this one back on the back burner.

PR #131, opened 2018-05-30
Disable colors in CI output for mocha

RikkiGibson:
VSTS puts in lots of noise characters instead of colors, so I'm turning mocha's colors off for CI.

Note I'm not adding this to `mocha.opts` because I still want a plain old `mocha` invocation to have colors.


RikkiGibson:

Issue #118, opened 2018-05-25
Support requiring a single OperationGroup in isolation


RikkiGibson:
- Generate `FooServiceClientContext` as a supertype of `FooServiceClient` (or some other workable variation)
- Export OperationGroups and have them take a `FooServiceClientParams`
- Ensure that a project that requires only one operation group produces a bundle of only that operation group and essential dependencies, instead of bundling every operation group.

PR #132, opened 2018-05-31
Support require() for a single operation group

RikkiGibson:
This breaks the generated ServiceClient up into two parts:

```js
import * as Models from "ms-rest-js";
import * as Mappers from "ms-rest-js";
export class FooServiceContext extends msRest.ServiceClient {
  baseUri?: string;
  serializer: msRest.Serializer;
  // ... other service client properties needed to run any operation
}
```
and
```js
import * as operations from "./operations";
export class FooService extends FooServiceContext {
  // Operation groups
  bars: Bars;
  bazs: Bazs;
  // Operations directly on the service client
}
```

This enables users to directly create the operation group they're interested in, and avoid bundling the other operation groups:
```ts
// Old way
import { AutoRestHttpInfrastructureTestService } from "./generated/Http/autoRestHttpInfrastructureTestService";

const client = new AutoRestHttpInfrastructureTestService();
const successOps = client.httpSuccess;
successOps.get200()
  .then(res => console.log(res))
  .catch(err => console.error(err));

// New way
import { AutoRestHttpInfrastructureTestServiceContext } from "./generated/Http/autoRestHttpInfrastructureTestServiceContext";
import { HttpSuccess } from "../vanilla/generated/Http/operations/HttpSuccess";

const context = new AutoRestHttpInfrastructureTestServiceContext();
const successOps = new HttpSuccess(context);
successOps.get200()
  .then(res => console.log(res))
  .catch(err => console.error(err));
```

In the case of `AutoRestHttpInfrastructureTestServiceContext`, the bundle size is cut from 300 KB to ~150 KB (possibly as low as 130 KB when I finish fiddling with ES tree shaking, probably in another PR.)

We may want to look into restructuring the way we pass mappers to the serializer so that we can leverage tree shaking to include only the mappers used in the operations the user is actually bundling.


RikkiGibson:
When requiring only LoadBalancers from azure-arm-network, this reduces the bundle size from 879 KB to 260 KB. @amarzavery

RikkiGibson:
### Size increased by 0.38% (7.39 KB)

RikkiGibson:
Too bad about the size increase but I think will be worth it for our Portal users.

daschult:
Definitely. We'll find other ways to bring the size down.

RikkiGibson:
Fixes #118

PR #134, opened 2018-05-31
Add unit tests for autorest.typescript

daschult:
Resolves https://github.com/Azure/autorest.typescript/issues/133.


src/DSL/TSBuilder.cs
```diff
@@ -89,6 +89,10 @@ public void Insert(int index, string text)
                         position.CharactersAfterPreviousPosition += text.Length;
                         break;
                     }
+                    else
+                    {
+                        positionIndex += position.CharactersAfterPreviousPosition;
+                    }
```
daschult: This is the bug that I was looking for. I found it by adding unit tests.

RikkiGibson:

Issue #94, opened 2018-05-22
Generate operation parameter mappers


daschult:
These mappers should replace the OperationParameterType enum.

PR #138, opened 2018-05-31
Add parameter mappers to OperationSpec

daschult:
Resolves https://github.com/Azure/autorest.typescript/issues/94.


src/vanilla/ClientModelExtensions.cs
```diff
@@ -630,314 +671,246 @@ public static string ConstructMapper(this IModelType type, string serializedName
...
-            {
-                builder.AppendLine($"xmlName: '{xmlName}',");
-            }
-
-            if (!isCaseSensitive)
+            value.Object(mapper =>
             {
-                serializedName = serializedName.ToLower();
-            }
+                bool wroteXmlName = isXML && !string.IsNullOrEmpty(xmlName) && xmlName != serializedName;
```
RikkiGibson: should this be called `shouldWriteXmlName` or similar?

test/azure/generated/AzureResource/models/mappers.ts
```diff
@@ -14,205 +14,179 @@ export const CloudError = CloudErrorMapper;
 export const BaseResource = BaseResourceMapper;
 
 export const ErrorModel = {
-  required: false,
-  serializedName: 'Error',
+  serializedName: "Error",
```
RikkiGibson: does the linter prefer double quotes? I seem to recall our ms-rest-js and autorest.typescript linters having contradictory settings.

src/vanilla/ClientModelExtensions.cs
daschult: Changed to `addXmlNameParameterValue`. This variable really keeps track of whether we should add the xmlName parameter that is passed to this function or whether we should use the xmlName property on the property variable.

test/azure/generated/AzureResource/models/mappers.ts
daschult: That sounds familiar, but I'm just going to go with whatever settings we use in the ms-rest-js repository. If we start getting linter errors in autorest.typescript, then we should update the autorest.typescript linter rules to be the same as ms-rest-js.

RikkiGibson:
The bundle size check is failing due to "stdout maxbuffer exceeded". Will fiddle with that in another PR, probably.

PR #140, opened 2018-06-02
Import helpers

RikkiGibson:
- Adds importHelpers: true to the base tsconfig. We don't ship this to users, so it doesn't really matter as much.
- Adds importHelpers: true to tsconfig.json template and tslib to package.json template. This will affect how the user's code bundles. In the ARM network client it seemed to make a pretty negligible difference (0.5% of a 800~ KB bundle). Perhaps the generated code doesn't make as use of as many language features that require helpers to be generated by tsc?
- Changes the target of the base tsconfig.json to es6 to match our tsconfig.json template. What target should we really have here?

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "through2-parallel": "^0.1.3",
     "ts-loader": "^4.1.0",
     "ts-node": "^5.0.1",
+    "tslib": "^1.9.2",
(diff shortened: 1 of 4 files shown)


RikkiGibson:
### Size increased by 2.10% (40.94 KB)

PR #149, opened 2018-06-04
No Buffer

RikkiGibson:
Follow up to Azure/ms-rest-js#124

### DIFF
--- a/test/azure/generated/AzureCompositeModelClient/operations/primitive.ts
+++ b/test/azure/generated/AzureCompositeModelClient/operations/primitive.ts
@@ -2068,8 +2068,8 @@ export class Primitive {
     let field = (options && options.field !== undefined) ? options.field : undefined;
     // Validate
     try {
-      if (field && !Buffer.isBuffer(field)) {
-        throw new Error('field must be of type buffer.');
+      if (field && !(field instanceof Uint8Array)) {
+        throw new Error('field must be of type uint8array.');
       }
       if (this.client.acceptLanguage !== null && this.client.acceptLanguage !== undefined && typeof this.client.acceptLanguage.valueOf() !== 'string') {
         throw new Error('this.client.acceptLanguage must be of type string.');
--- a/test/vanilla/acceptanceTests.ts
+++ b/test/vanilla/acceptanceTests.ts
@@ -60,6 +60,14 @@ const readStreamCountBytes = async function (stream: NodeJS.ReadableStream): Pro
   });
 }
 
+function stringToByteArray(str: string): Uint8Array {
+  if (msRest.isNode) {
+    return Buffer.from(str, "utf-8");
+  } else {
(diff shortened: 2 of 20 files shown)


test/azure/generated/AzureCompositeModelClient/operations/primitive.ts
```diff
@@ -2068,8 +2068,8 @@ export class Primitive {
     let field = (options && options.field !== undefined) ? options.field : undefined;
     // Validate
     try {
-      if (field && !Buffer.isBuffer(field)) {
-        throw new Error('field must be of type buffer.');
+      if (field && !(field instanceof Uint8Array)) {
+        throw new Error('field must be of type uint8array.');
```
daschult: Should the casing be the same here as in the previous line (Uint8Array)?

test/vanilla/complexTypesTests.ts
```diff
@@ -379,22 +394,22 @@ describe('typescript', function () {
...
+        testClient.polymorphism.putValidMissingRequired(getBadfish(), function (error, result) {
           should.exist(error);
           error.message.should.containEql('birthday');
           error.message.should.containEql('cannot be null or undefined');
           done();
         });
       });
 
-      var rawSalmon: AutoRestComplexTestServiceModels.SmartSalmon = {
+      var getRawSalmon: () => AutoRestComplexTestServiceModels.SmartSalmon = () => ({
```
daschult: Why did this change?

test/vanilla/acceptanceTests.ts
```diff
@@ -489,7 +510,13 @@ describe('typescript', function () {
       it('should get invalid byte value', function (done) {
         testClient.byteModel.getInvalid(function (error, result) {
           should.not.exist(error);
-          assert.deepEqual(result, new Buffer(':::SWAGGER::::', 'base64'));
+          // Base64 decoding of the string ':::SWAGGER::::'
+          const expected = new Uint8Array([73, 96, 6, 24, 68]);
```
RikkiGibson: This smells funny--will want to check on what the real intent of this test is and maybe adjust accordingly

test/vanilla/complexTypesTests.ts
RikkiGibson: I wanted to be able to delete properties from this without worrying about the state of it in subsequent usages (e.g. doing a put after the checking the result of the get)

test/azure/generated/AzureCompositeModelClient/operations/primitive.ts
RikkiGibson: That seems best, but we probably have a .ToLower() in this generator step for some reason.

RikkiGibson:
### Size decreased by 0.50% (-10.01 KB)

RikkiGibson:
### Size decreased by 0.99% (-19.82 KB)

PR #155, opened 2018-06-06
Remove leading dash from PackageJson.cshtml

RikkiGibson:
Somehow I added a leading dash into this file by accident, and probably since it just looks like a line marker in a git diff it was overlooked. Let's fix it.


RikkiGibson:
bundle-size failed, but it seems to work locally--don't want to mess around with it right now, so I'm just merging

PR #160, opened 2018-06-07
Mappers closure

RikkiGibson:
Related to Azure/ms-rest-js#126

- Adds serializer to individual operation groups
- Performs closure over model mappers at generation time to include the minimal number of model mappers in the resulting bundle

In the azure-arm-network/networkInterfaces scenario, this reduces the size of the bundled mappers from about 140 kB to about 70 kB.

Seems like a significant improvement, but I wonder how much overlap there is between model mappers in different operation groups--are we going to wind up making users redownload the same mappers multiple times? I think proper use of webpack's splitting feature might be able to prevent that from being a problem.


src/vanilla/Model/MethodGroupTS.cs
```diff
@@ -26,6 +26,74 @@ protected MethodGroupTS(string name): base(name)
...
+                    int initialCount = modelNames.Count;
+                    foreach (Method method in Methods)
+                    {
+                        if (method.Body != null)
+                        {
+                            ModelsClosure(modelNames, method.Body.ModelType);
+                        }
+                        foreach (Response response in method.Responses.Values)
+                        {
+                            ModelsClosure(modelNames, response.Body);
```
daschult: Can you rename `ModelsClosure()`? This method is really about collecting all of the model type names, but that isn't communicated to me with the name `ModelsClosure`.

src/vanilla/Model/MethodGroupTS.cs
```diff
@@ -26,6 +26,74 @@ protected MethodGroupTS(string name): base(name)
         [JsonIgnore]
         public IEnumerable<MethodTS> MethodTemplateModels => Methods.Cast<MethodTS>();
 
+        public ISet<string> OperationModelNames
+        {
+            get
+            {
+                ISet<string> modelNames = new HashSet<string>();
+                while (true)
```
daschult: What's the purpose of looping here? Generally if you have a stack of ModelTypes that you're iterating through (and adding to while you iterate), then you need a loop to check if the stack is empty. It looks like you're doing that graph traversal recursively, however, which indicates to me that you don't need a loop here.

src/vanilla/Templates/MethodGroupTemplate.cshtml
```diff
@@ -11,14 +11,18 @@ import * as msRest from "ms-rest-js";
 {
 @:import * as Models from "../models";
 }
-import * as Mappers from "../models/mappers";
+import * as Mappers from "../models/@(Model.TypeName.ToCamelCase())Mappers";
```
daschult: Use MethodGroupTS property here.

src/vanilla/Templates/MethodGroupTemplate.cshtml
```diff
@@ -11,14 +11,18 @@ import * as msRest from "ms-rest-js";
...
 import { @(Model.CodeModelTS.ContextName) } from "../@(Model.CodeModelTS.ContextName.ToCamelCase())";
 @EmptyLine
 const WebResource = msRest.WebResource;
 @EmptyLine
 /** Class representing a @(Model.TypeName). */
 export class @(Model.TypeName) {
   private readonly client: @(Model.CodeModelTS.ContextName);
+@{
+  var serializerParams = Model.CodeModel.ShouldGenerateXmlSerialization ? "Mappers, true" : "Mappers";
```
daschult: I've seen this a few times now, too. Can you encapsulate it as a property/method on MethodGroupTS too?

src/vanilla/Templates/MethodGroupTemplate.cshtml
```diff
@@ -11,14 +11,18 @@ import * as msRest from "ms-rest-js";
...
+}
+  private readonly serializer = new msRest.Serializer(@serializerParams);
```
daschult: This is probably the same as the Azure template version. Can you extract it to MethodGroupTS as a `GenerateSerializerMemberVariable()` (or something like that) method?

src/vanilla/Templates/ServiceClientContextTemplate.cshtml
```diff
@@ -37,7 +36,7 @@ export class @(Model.ContextName) extends msRest.ServiceClient {
     }
 }
   baseUri: string;
-  serializer: msRest.Serializer;
+  serializer?: msRest.Serializer;
```
daschult: Same as above. Is this still necessary?

src/vanilla/Model/MethodGroupTS.cs
RikkiGibson: We want to ensure that if any polymorphic subtypes were discovered, that we 1. add any unknown property types to the closure, and 2. add any polymorphic subtypes of those newly discovered types to the closure.

Expressing this recursively (I think) would involve factoring out another method which consumes the initial size of the set, then performs another recursive call if the size of the set changed.

```cs
void FindModelNames(ISet<string> names, int initialSize, CompositeType model)
{
  ModelsClosure(names, model);
  if (initialSize != names.Count) FindModelNames(names, names.Count, model);
}
```

I think this is a little less clear than the `while (true) { ... }`. I might be missing something about this algorithm or how to simplify it, though! I think I would also be alright with making this explicitly stack-based instead for some kind of stylistic consistency.

src/vanilla/Templates/ServiceClientContextTemplate.cshtml
RikkiGibson: Turns out the answer is no.

RikkiGibson:
@daschult I've made most of the suggested changes. Let me know if my POV on the FindReferencedModelNames thing is acceptable or if you want me to refactor.

RikkiGibson:
Also note that the `serializer` property on the ServiceClient template is only needed if there are any methods present directly on that ServiceClient.

RikkiGibson:
`bundle-size` failed--so CI succeeded, I guess! (Will fix that check ASAP.)

PR #161, opened 2018-06-09
No structural equals

### LINKED ISSUE #159: Fix bundle-size check
For some reason `webpack -p` has started failing in CI, but passing when run locally. I'll need to track this down, probably first by logging stdout/stderr upon an exec failure.

RikkiGibson:
This fixes a pathological behavior in the azure-arm generator where generating a network client would take 7+ minutes due to insane amounts of JSON serialization/comparision through IModelType.StructurallyEquals. Now it takes about 1 minute on my PC.

While I was at it I fixed the bundlesize check and fixed some build errors in the tests related to recent ms-rest-js changes.

Fixes #159

### DIFF
--- a/.scripts/bundlesize.js
+++ b/.scripts/bundlesize.js
@@ -10,49 +10,63 @@ const filesize = require("filesize");
 
 const { GitHubCiClient } = require("./github");
 const {
-    repo,
-    githubToken,
-    pr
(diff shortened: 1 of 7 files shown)


RikkiGibson:
  ### Size decreased by 0.00% (-68 B)

PR #163, opened 2018-06-11
Move parameter unflattening and ungrouping to runtime

daschult:
Resolves https://github.com/Azure/autorest.typescript/issues/157.
Resolves https://github.com/Azure/autorest.typescript/issues/156.
There is still a little bit of code remaining that deals with ungrouping, but this is actually parameter validation. I'll take care of this next.


RikkiGibson:
  ### Size increased by 0.31% (6.22 KB)

PR #166, opened 2018-06-14
Move validation to runtime

daschult:


RikkiGibson:
  ### Size decreased by 5.37% (-107.83 KB)

unittests/DSL/TSBuilderTests.cs
```diff
@@ -180,5 +180,36 @@ public void InsertNewLine()
             builder.InsertNewLine(0);
             Assert.Equal("\n", builder.ToString());
         }
+
+        [Fact]
+        public void AddPropertyToParentObjectWhileBuildingChildObject()
+        {
+            TSBuilder builder = new TSBuilder();
+            Exception exception = null;
```
RikkiGibson: Consider using the `Exception exception = Assert.Throws(() => builder.Object(...))` pattern

daschult: I didn't know that Assert.Throws() returned the validated exception. I'll change it to that.

RikkiGibson:
  ### Size decreased by 5.37% (-107.83 KB)

PR #168, opened 2018-06-15
Use AutoRest.common as a submodule

fearthecowboy:
same as the js one 💃


RikkiGibson:

RikkiGibson:
I think `git submodule update --init --recursive` is the command we need @daschult

PR #169, opened 2018-06-15
Enable chrome tests in CI

### LINKED ISSUE #129: Run browser tests in CI
Since we don't seem to be using Linux CI for now, and I believe Windows was where Chrome testing was actually working, let's try and add it back, parallelizing the tests to reduce the time taken.

RikkiGibson:
Fixes #129


RikkiGibson:
I suspect Chrome is not really shutting down/letting go of file handles before we begin the regeneration process. I'd rather not do something hacky like just regenerate before running Chrome, but otherwise I'm not sure how to make the tests behave properly in CI (run a script that finds and kills Chrome? :-1:)

RikkiGibson:
Will solve in a different PR

PR #171, opened 2018-06-18
Add TS_NODE_TYPE_CHECK environment variable before automated tests

daschult:
If we don't add this environment variable (TS_NODE_TYPE_CHECK = "Y"), then when ts-node runs our tests it won't find the const enum values (complains about not being able to find Csv on an undefined value). This problem can be fixed by setting the environment variable before running ts-node (as I've done in this PR), or to get rid of ts-node and just compile the typescript code using tsc. I figured everyone would prefer this route. The only problem is that now running "mocha" from the command line outside of gulp, npm, or vscode won't work unless you've previously set the TS_NODE_TYPE_CHECK environment variable.

### DIFF
--- a/.gulp/gulpfile.iced
+++ b/.gulp/gulpfile.iced
@@ -7,6 +7,7 @@ Tasks "regeneration"
 Tasks "publishing"
 
 Install "child_process"
+Install "process"
 
 # ==============================================================================
 # Settings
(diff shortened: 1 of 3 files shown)


RikkiGibson:
How do you find it affects the runtime of `mocha`?

daschult:
I haven't found a noticeable difference in test run times.

RikkiGibson:
I'm wondering if maybe the QueryCollectionFormat type should change to a union of literal strings, just because it seems difficult for any consumer of the generated code through ts-node to have to figure this out.

RikkiGibson:
Or just revert it to a non-const enum.

daschult:
I'd prefer to revert it to a non-const enum. I know that saves us some space, but I think we might find other ways to save space rather than this particular case.

RikkiGibson:
  ### Size increased by 0.16% (3.03 KB)

RikkiGibson:
BTW, Basarat's documentation on const enums implies we can also fix the behavior by turning `preserveConstEnums` back to true in ms-rest-js: https://basarat.gitbooks.io/typescript/content/docs/enums.html#const-enums

We still get the extra code size of the enum *declaration*, but for bundling use cases, it might(?) get removed anyway, and it fixes the ts-node usage.

PR #174, opened 2018-06-18
Support headerCollectionPrefix

### LINKED ISSUE #167: Add support for x-ms-meta

RikkiGibson:
Depends on Azure/ms-rest-js#139
Fixes #167


README.md
```diff
@@ -9,8 +9,8 @@ npm install -g autorest
 # Compatability
 This AutoRest extension generates TypeScript code that is compatible with:
 ```
-"ms-rest-azure-js": "~0.6.75",
-"ms-rest-js": "~0.10.268"
+"ms-rest-azure-js": "/Users/rigibson/src/ms-rest-azure-js",
```
RikkiGibson: revert once new ms-rest-js is published

src/vanilla/Model/CompositeTypeTS.cs
```diff
@@ -6,6 +6,7 @@
 using System.Linq;
 using AutoRest.Core.Model;
 using AutoRest.Core.Utilities;
+using AutoRest.Extensions;
```
RikkiGibson: revert

src/vanilla/TransformerTS.cs
```diff
@@ -22,15 +22,48 @@ public override CodeModelTS TransformCodeModel(CodeModel cm)
...
             CreateModelTypeForOptionalClientProperties(codeModel);
             CreateModelTypesForOptionalMethodParameters(codeModel);
             AddEnumTypesToCodeModel(codeModel);
+
             return codeModel;
         }
 
+        public void TransformHeaderCollectionPropertyTypes(CodeModelTS cm)
+        {
+            foreach (var modelType in cm.HeaderTypes)
```
RikkiGibson: I found this transformation was only necessary for the method parameters, not the response headers, so I may delete this. It's strange because I don't remember having to do something like this in Java for method parameters. I tried updating the modeler again but it doesn't really seem to have an effect.

RikkiGibson: It seems like for whatever reason the AutoRest modeler interprets header parameters with the HeaderCollectionPrefix extension as being exactly the type indicated in the Swagger--not a dictionary whose values are of the indicated type. Since these header parameters can also have constraints applied to them, we have to do some contortions to make it so the constraints are applied to the dictionary values, not just the keys.

RikkiGibson:

PR #179, opened 2018-06-22
Don't collect types within an error type

### LINKED ISSUE #178: Regenerating DNS client results in reference to nonexistent CloudErrorBody
Reported from Azure Portal. Probably due to something around the way I'm doing closure over the models--will check today.

RikkiGibson:
Fixes #178.

Would like to dive in a little and investigate why the original code resulted in a CloudErrorBody but not the code which generates all mappers using essentially `CodeModel.ModelTypes.Concat(CodeModel.HeaderTypes)`. After we confirm working behavior in some of the main ARM services, we may wish to just merge this as it will prevent our Portal users from needing to manually modify the generated code.

### DIFF
--- a/src/vanilla/Model/MethodGroupTS.cs
+++ b/src/vanilla/Model/MethodGroupTS.cs
@@ -55,7 +55,11 @@ namespace AutoRest.TypeScript.Model
                     }
                     if (method.DefaultResponse != null)
                     {
-                        CollectReferencedModelNames(modelNames, method.DefaultResponse.Body);
+                        // We don't want to collect types contained within the error type
+                        if (method.DefaultResponse.Body is CompositeType)
+                        {
+                            modelNames.Add(method.DefaultResponse.Body.Name);
+                        }
                         CollectReferencedModelNames(modelNames, method.DefaultResponse.Headers);
                     }
                 }
--- a/test/vanilla/acceptanceTests.ts
+++ b/test/vanilla/acceptanceTests.ts
@@ -1980,6 +1980,9 @@ describe('typescript', function () {
               ev.loadedBytes.should.be.a.Number;
             }
           });
+          if (response.blobBody) {
+            await response.blobBody();
+          }
           assert(uploadNotified);
           assert(downloadNotified);
         });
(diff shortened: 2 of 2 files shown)


RikkiGibson:

RikkiGibson:
Looking into it a bit, I see the following in the DNS preview Swagger which we happened to be using for generation:

```json
    "CloudError": {
      "type": "object",
      "properties": {
        "error": {
          "$ref": "#/definitions/CloudErrorBody",
          "description": "The error message body"
        }
      },
      "x-ms-external": true,
      "description": "An error message"
    },
    "CloudErrorBody": {
      "type": "object",
      "properties": {
        "code": {
          "type": "string",
          "description": "The error code"
        },
        "message": {
          "type": "string",
          "description": "A description of what caused the error"
        },
        "target": {
          "type": "string",
          "description": "The target resource of the error message"
        },
        "details": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/CloudErrorBody"
          },
          "description": "Extra error information"
        }
      },
      "x-ms-external": true,
      "description": "The body of an error message"
    }
  },
```

x-ms-external tells AutoRest to not include this type in CodeModel.ModelTypes but just assume it is present. @amarzavery @shahabhijeet is there possibly a requirement from ARM that we start exporting this type from the Azure runtime so it's available for services that use it?

It seems like we already ignore the CloudError type from the Swagger and replace it with our own, so I may just move forward with the workaround in this PR.

RikkiGibson:
Incidentally going to fix some minor things in browser tests based on Azure/ms-rest-js#148

RikkiGibson:
  ### Size increased by 0.00% (59 B)

RikkiGibson:
Been thinking on this for a little while--maybe we should just somehow hunt down the x-ms-external models and always generate interfaces and mappers for them, since we don't care how many copies of the same interface there are across multiple libraries.

Otherwise the "robust" thing to do with the external models is to compare the entire structure of the x-ms-external model at generation time with what we believe we have available in the runtime.

RikkiGibson:
Just gonna merge this even though we are likely to need a better solution in the future.

PR #181, opened 2018-06-25
Request content type

### LINKED ISSUE #180: 'Consumes' field on swagger definition is ignored after v2.0.245 for Get/Delete requests
Before the version 2.0.245 the following was generated for my swagger definition.
```  typescript
    ...
    // Create HTTP transport objects
    const httpRequest = new WebResource();
    // Set Headers
    httpRequest.headers.set("Content-Type", "application/json; charset=utf-8");
    ...
```
So the generated package was specifying Content-Type as application/json for all of the types of requests (GET, PUT, etc.).

Since 2.0.245, autorest.typescript seems to only depend on the autorest schema passed to sendOperationRequest function when it comes to the 'Content-Type'. Eg:
``` typescript
...
operationRes = await this.sendOperationRequest(
        httpRequest,
        operationArguments,
        {
          httpMethod: "POST",
          baseUrl: this.baseUri,
          path: "api/1.0/Controller/Endpoint",
          ...
          requestBody: {
            ...
          },
          contentType: "application/json; charset=utf-8",
          responses: {
           ...
          },
          serializer: this.serializer
        });
...
```
This currently works fine for Patch, Put and Post as far as I can tell but doesn't work for Get and Delete.
For Get and Delete the content type is not set in this autorest schema passed to sendOperationRequest, which makes request coming from a browser default to 'application/x-www-form-urlencoded'. This fails on the server side (415) because controllers are explicitly specified to consume 'application/json'. 

I have checked the swagger definition generated by the server-side. All endpoints contain: 
```JSON
...
"consumes": [ "application/json"],
...
```

Am I missing something? Is there a workaround for this other than downgrading?

RikkiGibson:
Fixes #180


RikkiGibson:
The request content type is overwriting the operation header parameter `contentType`. So I think a runtime change is needed here.. I think the old way must have had the operation parameter "win" over the RequestContentType from the AutoRest model.

RikkiGibson:
The repetition of `contentType: "application/json; charset=utf-8"` is slightly bothersome to me..but at least it should gzip well.

RikkiGibson:
  ### Size increased by 1.76% (23.66 KB)

RikkiGibson:
  ### Size increased by 1.78% (23.66 KB)

RikkiGibson:
@daschult were you planning on doing any work to include a "base" OperationSpec object in the operation group class or something that contains client-level parameters, the serializer, etc? I would really like to just have this serve as a ride-along in such an object.

e.g.

```js
  const baseOperationSpec = {
    return {
      clientParam: this.clientParam,
      otherParam: this.otherParam,
      requestContentType: ...,
      serializer: ...
    }
  }

  const myOperationSpec = {
    ...baseOperationSpec,
    localParam
  }
```

daschult:
@RikkiGibson No, I wasn't planning on doing that, because I want the OperationSpec objects to be static immutable data objects. As soon as you start referencing a `this` pointer, that can't hold true anymore.
The way I was going to solve this that when the operation argument values are being retrieved, the service client would first check the arguments that are passed to called function, and if it didn't find anything, then it would check its client properties. That way we could remove the client parameters from the OperationArguments.

RikkiGibson:
OK, seems like we can achieve the same thing that way.

RikkiGibson:
This now depends on Azure/ms-rest-js#178

RikkiGibson:
[master is also failing CI](https://travis-ci.org/Azure/autorest.typescript/builds/408736808) in Travis which I suspect is due to recent ms-rest-azure-js changes.

The Windows CI build seems to be passing which might explain how it slipped through the cracks. What could be causing a Linux-specific failure?

daschult:
@RikkiGibson I'll figure out how to fix the autorest.typescript LRO tests. I assumed that the specification I was following would be correct for the swagger tests, but I guess I did something wrong. I'll take care of that today.

Issue #28, opened 2018-03-20
Reduce the size of the generated sdk and the underlying dependencies (ms-rest-js and ms-rest-azure-js)


daschult:
Depends on https://github.com/Azure/autorest.typescript/issues/66.
Depends on https://github.com/Azure/ms-rest-js/issues/97.
> **Estimate: ____(weeks)**
- Investigate if we can find out a way to not have wrapper methods for callback/promise and push all of this to a generic method in the runtime. This generic method gets the method signature of the method that was called, gets the parameters that were actually called on that method, it also gets other information that is required to make a successful request and what it should do upon receiving a response from the pipeline. 
- Another area of improvement is to move away from moment as a dependency. We use it only for validating, [de]serializing Durations. Currently ms-rest-js is around 109 KB in size of which 59KB is just moment. If we can rip out the pieces needed for Duration and create a separate library for that, then we are game. That should considerably bring the sdk size down.

XiaoningLiu:
For the callback/promise, aws-js SDK provides a runtime switch. We can have a look for reference.

https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/using-promises.html

RikkiGibson:
It might actually make a measurable difference in bundle size to simplify some validations in generated code to smaller equivalents: e.g. `foo === null || foo === undefined || typeof foo !== 'string'` can be reduced to `typeof foo !== 'string'` (because `typeof null === 'object'` and `typeof undefined === 'undefined'`)

PR #183, opened 2018-06-25
Move success response deserialization to runtime

daschult:
Resolves https://github.com/Azure/ms-rest-js/issues/97.
Resolves https://github.com/Azure/autorest.typescript/issues/28.

There's still a tiny bit of response deserialization generated for some very specific scenarios (such as a HEAD request that wants to return a boolean), but I'm going to leave those for another day.


src/DSL/TSTryBlock.cs
```diff
@@ -12,7 +12,7 @@ public TSTryBlock(TSBuilder builder)
         {
         }
 
-        public TSTryBlock Catch(string errorName, Action<TSBlock> catchAction)
+        public void Catch(string errorName, Action<TSBlock> catchAction)
```
RikkiGibson: no need for `finally`, basically?

daschult: Yeah, until we need it, I'm going to pretend that doesn't exist.

test/azure/generated/AzureCompositeModelClient/azureCompositeModel.ts
```diff
@@ -90,21 +89,17 @@ class AzureCompositeModel extends AzureCompositeModelContext {
...
-          "this.subscriptionId": this.subscriptionId,
-          resourceGroupName,
-          apiVersion,
-          "this.acceptLanguage": this.acceptLanguage
-        },
-        options);
       operationRes = await this.sendOperationRequest(
-        httpRequest,
-        operationArguments,
+        msRest.createOperationArguments(
```
RikkiGibson: I think this is more readable.

PR #196, opened 2018-06-28
Support x-nullable

RikkiGibson:
Needs manual testing

Related to Azure/ms-rest-js#159

### DIFF
--- a/src/vanilla/ClientModelExtensions.cs
+++ b/src/vanilla/ClientModelExtensions.cs
@@ -682,6 +682,18 @@ namespace AutoRest.TypeScript
                 mapper.BooleanProperty("required", true);
             }
 
+            if (parameter?.IsXNullable != null)
+            {
+                if (parameter.IsXNullable.Value)
+                {
+                    mapper.BooleanProperty("nullable", true);
+                }
+                else
+                {
+                    mapper.BooleanProperty("nullable", false);
+                }
+            }
+
             if (isReadOnly)
             {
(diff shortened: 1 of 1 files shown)


RikkiGibson:
  ### Size increased by 1.31% (17.43 KB)

src/vanilla/ClientModelExtensions.cs
```diff
@@ -682,6 +682,18 @@ public static void ConstructMapper(TSObject mapper, IModelType type, string seri
...
+            if (parameter?.IsXNullable != null)
+            {
+                if (parameter.IsXNullable.Value)
+                {
+                    mapper.BooleanProperty("nullable", true);
+                }
+                else
+                {
+                    mapper.BooleanProperty("nullable", false);
+                }
```
daschult: This should probably be reduced to just the true scenario. If `nullable` is false, then we shouldn't include it in the mapper because not being included defaults to falsy, right?

RikkiGibson: true, false and undefined are actually all treated differently for nullable. See https://github.com/RikkiGibson/ms-rest-js/blob/13da6db58bca7b109fad4888310a39a4c3a87798/lib/serializer.ts#L89

RikkiGibson: Good now @daschult?

RikkiGibson:
I have *no idea* where this extra 17 KB is coming from! But I have to assume it's not related to this PR.

PR #202, opened 2018-07-11
Get client properties from ServiceClient instead of OperationArguments

daschult:
Resolves https://github.com/Azure/autorest.typescript/issues/189.
Depends on https://github.com/Azure/ms-rest-js/pull/171.


RikkiGibson:
  ### Size decreased by 1.12% (-14.93 KB)

src/vanilla/Model/ParameterTS.cs
```diff
@@ -1,51 +0,0 @@
...
-            Name.OnGet += value =>
-                IsClientProperty
-                    ? false == Method?.MethodGroup?.IsCodeModelMethodGroup
-                        ? $"this.client.{ClientProperty.Name}"
-                        : $"this.{ClientProperty.Name}"
-                    : CodeNamer.Instance.GetParameterName(value);
-        }
-
-        [JsonIgnore]
-        public IEnumerable<Core.Model.Property> ComposedProperties
```
RikkiGibson: This property didn't make sense, right?

daschult: It didn't really make sense, but more importantly it wasn't used anywhere. That's the real reason I removed it.

PR #204, opened 2018-07-12
Polymorphic interfaces

RikkiGibson:
Fixes Azure/azure-sdk-for-node#2960

We're not 100% sure this is the approach we want, but we have played with a bunch of different possible representations and this so far seems to have the best editor experience.

### DIFF
--- a/src/vanilla/Model/CompositeTypeTS.cs
+++ b/src/vanilla/Model/CompositeTypeTS.cs
@@ -33,18 +33,18 @@ namespace AutoRest.TypeScript.Model
             return result;
         }
 
-        private IReadOnlyList<CompositeTypeTS> polymorphicUnionCases;
+        private IReadOnlyList<CompositeTypeTS> immediatePolymorphicSubtypes;
 
         /// <summary>Contains the immediate polymorphic children of this class.</summary>
-        public IReadOnlyList<CompositeTypeTS> PolymorphicUnionCases
+        public IReadOnlyList<CompositeTypeTS> ImmediatePolymorphicSubtypes
         {
             get
             {
-                if (polymorphicUnionCases == null)
+                if (immediatePolymorphicSubtypes == null)
                 {
                     if (BaseIsPolymorphic)
                     {
-                        polymorphicUnionCases =
+                        immediatePolymorphicSubtypes =
                             CodeModel.ModelTypes
                                 .Where(m => m.BaseModelType == this)
                                 .Cast<CompositeTypeTS>()
@@ -53,15 +53,15 @@ namespace AutoRest.TypeScript.Model
                     }
                     else
                     {
-                        polymorphicUnionCases = new List<CompositeTypeTS>().AsReadOnly();
+                        immediatePolymorphicSubtypes = new List<CompositeTypeTS>().AsReadOnly();
                     }
                 }
-                return polymorphicUnionCases;
+                return immediatePolymorphicSubtypes;
             }
         }
 
         /// <summary>The name to use when referring to this type in a parameter, return type or property definition.</summary>
-        public string UnionTypeName => PolymorphicUnionCases.Any() ? Name + "Union" : Name.ToString();
+        public string UnionTypeName => ImmediatePolymorphicSubtypes.Any() ? Name + "Union" : Name.ToString();
 
         public string NameAsFileName => Name.EqualsIgnoreCase("index") ? "IndexModelType" : (string)Name;
 
@@ -285,7 +285,7 @@ namespace AutoRest.TypeScript.Model
         {
             get
             {
-                IEnumerable<string> cases = new[] { this.Name.ToString() }.Concat(PolymorphicUnionCases.Select(m => m.UnionTypeName));
+                IEnumerable<string> cases = new[] { this.Name.ToString() }.Concat(ImmediatePolymorphicSubtypes.Select(m => m.UnionTypeName));
                 return $"export type {Name}Union = {string.Join(" | ", cases)};";
             }
         }
--- a/src/vanilla/Model/CompositeTypeTS.cs
+++ b/src/vanilla/Model/CompositeTypeTS.cs
@@ -33,6 +33,35 @@ namespace AutoRest.TypeScript.Model
             return result;
         }
 
+        private List<CompositeTypeTS> polymorphicUnionCases;
+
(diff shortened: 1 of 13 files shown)


test/azure/generated/AzureCompositeModelClient/models/index.ts
```diff
@@ -184,11 +184,20 @@ export interface Siamese extends Cat {
...
+
 /**
  * @interface
  * An interface representing Fish.
  */
 export interface Fish {
+  /**
+   * @member {string} fishtype Polymorphic Discriminator
+   */
+  fishtype: "Fish";
```
RikkiGibson: I inserted a .OrderBy in the generator to make the discriminator always the first property in the polymorphic interface. I think this is a better user experience.

RikkiGibson:

src/vanilla/Model/CompositeTypeTS.cs
```diff
@@ -33,6 +33,35 @@ public override Property Add(Property item)
...
+                    {
+                        polymorphicUnionCases = new List<CompositeTypeTS>() { };
+                    }
+                }
+                return polymorphicUnionCases;
+            }
+        }
+
+        /// <summary>The name to use when referring to this type in a parameter, return type or property definition.</summary>
+        public string ReferenceName => PolymorphicUnionCases.Any() ? Name + "Union" : Name.ToString();
```
daschult: Maybe call this `UnionTypeName`? It's basically saying that if this type is part of union, then use this type's name with the "Union" suffix. Otherwise, just use this type's name.

src/vanilla/Model/CompositeTypeTS.cs
```diff
@@ -33,6 +33,35 @@ public override Property Add(Property item)
...
+            {
+                if (polymorphicUnionCases == null)
+                {
+                    if (BaseIsPolymorphic)
+                    {
+                        polymorphicUnionCases =
+                            CodeModel.ModelTypes
+                                .Where(m => m.BaseModelType == this)
+                                .Cast<CompositeTypeTS>()
+                                .ToList();
```
daschult: So this only looks for types that derive from this type, right? This won't include this type?

src/vanilla/Model/CompositeTypeTS.cs
```diff
@@ -33,6 +33,35 @@ public override Property Add(Property item)
             return result;
         }
 
+        private List<CompositeTypeTS> polymorphicUnionCases;
```
daschult: Why is this a List? Is it going to be modified after it is initially created?

src/vanilla/Model/CompositeTypeTS.cs
```diff
@@ -33,6 +33,35 @@ public override Property Add(Property item)
...
+
+        /// <summary>Contains the immediate polymorphic children of this class.</summary>
+        public List<CompositeTypeTS> PolymorphicUnionCases
```
daschult: Why is this a List? Are you expecting the caller of this property to add or remove values from this List?

src/vanilla/Model/CompositeTypeTS.cs
RikkiGibson: Might make sense. I thought the name conveyed the purpose: when referencing this type in an annotation, use this ReferenceName, and when declaring it or referencing it in a mapper, use Name. But it's a little contrived no matter what you do.

src/vanilla/Model/CompositeTypeTS.cs
RikkiGibson: Yeah, this is so you can `type FishUnion = Fish | SalmonUnion | SharkUnion`-- by treating "this type" as a special case, you can make sure you reference the interface instead of the tagged union.

src/vanilla/Model/CompositeTypeTS.cs
RikkiGibson: I'd like for it to be immutable, computed lazily on the first get and then cached. This is used in `ReferenceName` to decide whether to refer to the type as `FooUnion` or just `Foo` based on whether any polymorphic subtypes exist.

src/vanilla/Model/CompositeTypeTS.cs
RikkiGibson: Think this would be better named... `ImmediatePolymorphicSubtypes` or something..

src/vanilla/Model/CompositeTypeTS.cs
daschult: I get that, but why is it a List and not an Array or an IEnumerable?

src/vanilla/Model/CompositeTypeTS.cs
daschult: Same as above. Why isn't this an IEnumerable?

src/vanilla/Model/CompositeTypeTS.cs
RikkiGibson: Not IEnumerable because I want it to be unambiguous that "work" isn't being done upon enumeration.

daschult: If that's the case, then why not an array? Also, why does the laziness of the work matter to you? Isn't your only requirement that it be a sequence of values?

RikkiGibson: It can be an array. We can convey even more that you're not meant to overwrite values in the array by using `IReadOnlyList`. I just want to avoid repeated execution of the query through multiple usages of `ReferenceName`.

RikkiGibson:
Updated a few things based on feedback--see what you think. I'm not adamantly against the PolymorphicUnionCases (which I maybe should rename) being array or IEnumerable, just want to avoid a trap I've gotten into in the past where an expensive query I enumerated multiple times slowed things down.

RikkiGibson:

RikkiGibson:
After speaking with TS team it seems like this is probably a reasonable design--we could consider factoring out `AnimalBase` with all properties except the discriminator, but users who look at these generated interfaces will have to contend with all these contortions around it--Animal, AnimalUnion, AnimalBase... etc.

Keeping that to a minimum is probably more desirable, and it sounds like we might be able to get a fix in the TS go-to-definition behavior that makes it so the candidate symbols are further constrained by a tagged union discriminator, which pretty much eliminates our remaining concerns about this approach.

RikkiGibson:
Rebased to ensure compatibility with recent changes

PR #207, opened 2018-07-12
Travis config for bundle size check

### LINKED ISSUE #129: Run browser tests in CI

RikkiGibson:
Fixes #129


RikkiGibson:
Thank you @bsiegel!

PR #208, opened 2018-07-13
No await

### LINKED ISSUE #186: Replace async/await with then() calls in generated operation methods

RikkiGibson:
Fixes #186

### DIFF
--- a/src/DSL/TSBuilder.cs
+++ b/src/DSL/TSBuilder.cs
@@ -468,6 +468,16 @@ namespace AutoRest.TypeScript.DSL
             Text("undefined");
         }
 
+        public void Lambda(string paramName, Action<TSBlock> lambdaBodyAction)
+        {
+            Line($"{paramName} => {{");
+            Indent(() =>
+            {
+                lambdaBodyAction.Invoke(new TSBlock(this));
+            });
+            Line($"}}");
+        }
+
         /// <summary>
         /// Invoke the provided action in order to produce a value in this TSBuilder.
         /// </summary>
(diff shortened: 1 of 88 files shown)


src/DSL/TSBuilder.cs
```diff
@@ -468,6 +468,16 @@ public void Undefined()
...
         }
 
+        public void Lambda(string paramName, Action<TSBlock> lambdaBodyAction)
+        {
+            Line($"{paramName} => {{");
+            Indent(() =>
+            {
+                lambdaBodyAction.Invoke(new TSBlock(this));
+            });
+            Line($"}}");
```
daschult: Since a lambda is an expression (as opposed to a statement), the last output should be a `Text()` call, not a `Line()`.

src/DSL/TSBuilder.cs
```diff
@@ -503,6 +513,16 @@ public void AsyncMethod(string methodName, string returnType, string parameterLi
...
+        {
+            Line($"{methodName}({parameterList}): {returnType} {{");
+            Indent(() =>
+            {
+                methodBodyAction.Invoke(new TSBlock(this));
+            });
+            Line($"}}");
+        }
+
         public void AsyncMethod(string methodName, string returnType, Action<TSParameterList> parameterListAction, Action<TSBlock> methodBodyAction)
```
daschult: With this change, are we still using AsyncMethod anywhere? If not, please remove it.

RikkiGibson:
Made the suggested changes in 44aa377. Will rebase after the optional parameters PR is merged. All good?

PR #210, opened 2018-07-14
Reduce size of body overloads

RikkiGibson:
Depends on Azure/ms-rest-js#172
Resolves https://github.com/Azure/autorest.typescript/issues/192.

### DIFF
--- a/src/vanilla/Model/CodeModelTS.cs
+++ b/src/vanilla/Model/CodeModelTS.cs
@@ -361,7 +361,7 @@ namespace AutoRest.TypeScript.Model
 
         public virtual string PackageDependencies()
         {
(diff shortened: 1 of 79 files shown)


src/vanilla/Model/CodeModelTS.cs
```diff
@@ -361,7 +361,7 @@ public virtual string ConstructRuntimeImportForModelIndex()
 
         public virtual string PackageDependencies()
         {
-            return "\"ms-rest-js\": \"~0.15.357\"";
+            return "\"ms-rest-js\": \"/Users/rigibson/src/ms-rest-js\"";
```
RikkiGibson: Will fix after publishing new ms-rest-js

RikkiGibson:
Merging these changes as part of #208.

PR #211, opened 2018-07-16
Flatten OperationArguments and remove optional parameter local variables

daschult:
Resolves https://github.com/Azure/autorest.typescript/issues/188.


src/vanilla/Model/MethodTS.cs
```diff
@@ -569,39 +532,28 @@ public string GenerateWithHttpOperationResponseMethod(string emptyLine)
             return builder.ToString();
         }
 
-        public void GenerateOperationArguments(TSArgumentList operationArguments)
+        public void GenerateOperationArguments(TSObject operationArguments)
         {
-            operationArguments.Object(obj =>
+            ParameterTransformations transformations = GetParameterTransformations();
+            Action<string, string> addArgument = (string operationArgumentName, string operationArgumentValue) =>
```
RikkiGibson: Suggest using a local function for this

RikkiGibson:
```cs
void addArgument(string operationArgumentName, string operationArgumentValue)
{
    ...
}
```

src/vanilla/Model/ParameterTransformations.cs
```diff
@@ -13,6 +13,10 @@ namespace AutoRest.TypeScript.Model
     /// </summary>
     public class ParameterTransformations
     {
+        /// <summary>
+        /// The names of input parameters that are in the options parameter.
+        /// </summary>
+        private readonly ISet<string> inputParametersInOptions;
```
RikkiGibson: Does the parameter itself not know this based on `.Required`?

daschult: `.Required` doesn't quite cover whether or not a parameter is included in the `options` parameter. [This function](https://github.com/Azure/autorest.typescript/pull/211/files#diff-a0945a4f62f5a6e958b85ad35db7086eR756) was added to share the logic for what determines whether a parameter is in `options` or not.

test/azure/generated/AzureCompositeModelClient/operations/arrayModel.ts
```diff
@@ -400,7 +401,10 @@ const putValidOperationSpec: msRest.OperationSpec = {
   ],
   requestBody: {
     parameterPath: {
-      arrayProperty: "arrayProperty"
+      arrayProperty: [
```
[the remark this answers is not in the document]
daschult: That's what I'm relying on.

src/vanilla/Model/ParameterTransformations.cs
RikkiGibson: Makes sense--things seem OK as is, then.

.gulp/gulpfile.iced
```diff
@@ -62,13 +62,13 @@ task 'test/nodejs-unit', 'run nodejs unit tests', [], (done) ->
   await execute "#{basefolder}/node_modules/.bin/mocha", defer _
   done();
 
-task 'testci/chrome-unit', '', [], (done) ->
+task 'testci/chrome-unit', '', ['testci/nodejs-unit'], (done) ->
```
RikkiGibson: the CI jobs run the tests in parallel in different containers so I suggest not having a relationship between these tasks

.gulp/gulpfile.iced
```diff
@@ -62,13 +62,13 @@ task 'test/nodejs-unit', 'run nodejs unit tests', [], (done) ->
...
 
-task 'testci/chrome-unit', '', [], (done) ->
+task 'testci/chrome-unit', '', ['testci/nodejs-unit'], (done) ->
   global.verbose = true
   await run "test/chrome-unit", defer _
   await execute "node #{basefolder}/.scripts/coverage", defer _
   done()
 
-task 'test/chrome-unit', 'run browser unit tests', [], (done) ->
+task 'test/chrome-unit', 'run browser unit tests', ['test/nodejs-unit'], (done) ->
```
RikkiGibson: Ordering these is probably for the best..

.gulp/gulpfile.iced
daschult: Removed in https://github.com/Azure/autorest.typescript/pull/211/commits/65d262eca39c20e4affd29ec9f696aa810c2cf41.

src/vanilla/Model/MethodTS.cs
daschult: Done in https://github.com/Azure/autorest.typescript/pull/211/commits/65d262eca39c20e4affd29ec9f696aa810c2cf41.

RikkiGibson:
Now that we run `npm install` without `--ignore-scripts` in the bundlesize check, a dotnet build occurs and it's necessary to modify .travis.yml to install the libunwind8 prerequisite.

You can either pass the `--ignore-scripts` flag or modify the .travis.yml to add the dotnet prerequisites:

```
    - env: CHECK_NAME="Bundle Size"
      install:
          # dotnet 2.0 prereq
        - sudo apt-get install libunwind8
        - npm install
      script: npm run bundle-size
```

This adds about a minute to the job.

.scripts/bundlesize.js
```diff
@@ -23,26 +20,52 @@ function execVerbose(script) {
...
-  dependencies.refreshNodeModules();
-  const headSize = await getBundleSize();
-
-  const change = (headSize / baseSize) - 1;
-  const percentChange = (Math.abs(change) * 100).toFixed(2) + '%';
-
+/**
+ * Output a message to the console.
+ */
+function outputMessage(message) {
```
RikkiGibson: Note that ci-sidecar will only consume the first `---output` block--subsequent blocks will be ignored

PR #217, opened 2018-07-18
Fix XML tests that were failing

daschult:
I just noticed that these tests were failing on master on my machine.


daschult:
Oh, fun. Travis behaves differently than my machine. I'll see if I can figure out a way to make both machine types happy with a PR to the testserver repository.

daschult:
And now my machine is working without these changes... weird... As long as the tests pass, I don't really care too much, though.

RikkiGibson:
I fixed them in another branch--deleting and reinstalling the testserver should give a consistent behavior with Travis. I also made us start using a fixed version of the testserver.

PR #218, opened 2018-07-19
Add model-enum-as-union doc

RikkiGibson:


README.md
```diff
@@ -56,6 +56,21 @@ autorest --typescript \
   --generate-metadata=true
 ```
 
+### Generate enums as unions
+By default, a Swagger enum is translated into a TypeScript enum, like so:
+```ts
+enum Colors {
+  Redcolor = 'red color',
```
daschult: Capital C?

RikkiGibson: Can change it but it's copy pasted from the actual gen code

PR #219, opened 2018-07-20
Parameter mappers

### LINKED ISSUE #185: Extract parameter mappers into parameterMappers.ts file

RikkiGibson:
Fixes #185
Depends on Azure/ms-rest-js#176

Mostly working, but could be refined a bit more-- for instance, body parameters don't have their mappers included in the parameters module. 

- [x] Need to `import * as Parameters` conditionally based on whether any parameters are actually being used.

- [x] Verified that `import * as Parameters` still allows tree shaking as long as you only access properties on `Parameters` and don't pass the overall object as an argument to something.

**arm-network size goes from 791 KB down to 677 KB from this change**. The size of each operation group dropped by about 50% according to source-map-explorer, but source-map-explorer tends to overestimate the sizes of modules.


src/vanilla/Model/MethodTS.cs
```diff
@@ -570,17 +570,9 @@ public void GenerateOperationSpec(TSObject operationSpec)
             Parameter[] logicalParameters = LogicalParameters.ToArray();
             ParameterTransformations parameterTransformations = GetParameterTransformations();
```
daschult: Please move this variable down closer to where it is used now.

test/azure/generated/AzureCompositeModelClient/models/parameters.ts
```diff
@@ -0,0 +1,78 @@
...
+ * license information.
+ *
+ * Code generated by Microsoft (R) AutoRest Code Generator.
+ * Changes may cause incorrect behavior and will be lost if the code is
+ * regenerated.
+ */
+
+import * as msRest from "ms-rest-js";
+
+export const subscriptionId0: msRest.OperationURLParameter = {
```
daschult: Can you sort this list alphabetically? That would definitely help with code reviews when parameters change and it will also help us to compare the differences between `subscriptionId0` and `subscriptionId1`.

RikkiGibson: Yeah, great idea.

RikkiGibson:
Yeah, the test clients I think slightly increased in size on average while the arm-network client size went down by 15%.

src/vanilla/Model/MethodGroupTS.cs
```diff
@@ -29,6 +29,14 @@ protected MethodGroupTS(string name): base(name)
 
         public string MappersModuleName => TypeName.ToCamelCase() + "Mappers";
 
+        public string Imports
```
RikkiGibson: I need to make use of this. The plan is to throw the logic in here to decide which import statements to emit--I think there's an existing method around MethodGroup or CodeModel that I should just edit or copy, though.

PR #222, opened 2018-07-20
Mapper element type

RikkiGibson:
Removes the serializedName from array and dictionary element types.

I think we can save a little space this way and still create good error messages over in ms-rest-js by checking the className of the element type.


RikkiGibson:
Saves about 12 KB in arm-network monolithic bundle

RikkiGibson:
This shouldn't type-check, and yet it does. At least while `serializedName` is still required in `Mapper`.

test/azure/generated/AzureCompositeModelClient/models/mappers.ts
```diff
@@ -9,11 +9,12 @@
  */
 
 import {CloudErrorMapper, BaseResourceMapper } from "ms-rest-azure-js";
+import * as msRest from "ms-rest-js";
```
daschult: Isn't it better for webpack if you did `import { CompositeMapper} from "ms-rest-js";`?

RikkiGibson: It's possible it will save webpack time when tree shaking, but webpack is still able to shake `import * as msRest` imports, as long as you don't pass the `msRest` object off to a function or something like that (like we do with `Mappers`)

I think I was mainly thinking about name collisions--even though we're realistically already in a bit of a precarious spot with that.

PR #223, opened 2018-07-23
Enum union bug fix

RikkiGibson:
Just failed to assign this generator setting in azure mode. Maybe we should have a method for initializing the settings on the CodeModel so the copy/paste isn't necessary.


src/vanilla/CodeGeneratorTS.cs
```diff
@@ -25,6 +25,17 @@ public class CodeGeneratorTS : CodeGenerator
 
         public override string UsageInstructions => $"The {ClientRuntimePackage} or higher npm package is required to execute the generated code.";
 
+
+        protected void InitializeSettings(CodeModelTS codeModel)
```
daschult: I'd recommend replacing the body of this method with this logic: https://github.com/Azure/autorest.nodejs/blob/master/src/Program.cs#L88-L113

daschult: That will solve this problem forever.

RikkiGibson: This method isn't doing quite the same thing as that one, but I've copied over the reflection-based approach in 42de528b.

PR #227, opened 2018-07-26
Update metadata generation

### LINKED ISSUE #225: Generated webpack config does not allow "test" or "node_modules" anywhere in path
Line 19 of the generated webpack.config.js has this: exclude: /(node_modules|test)/
This causes the webpack run to fail if "test" or "node_modules" is anywhere in the parent folder structure (e.g. C:\buildandtest\{folderContainingTypescript} or C:\whatever\testament\youwant\{folderContainingTypescript})

RikkiGibson:
Fixes #225

- Change formatting in PackageJson.cshtml so that npm won't modify it upon `npm install` or similar
- Update dependency versions in PackageJson.cshtml
- Add tsconfig.webpack.json to target ES5 and use ESM syntax for compatibility and space
- Update and simplify WebpackConfig.cshtml
- Add jsdoc `@type` to WebpackConfig.cshtml which enables type checking for the webpack config without having to pull in ts-node and use webpack.config.ts :rocket:
- Adds `regenerate-metadata` and `test/metadata` tasks so that regressions and issues with metadata generation can be caught
Should there be automated tests for this? Maybe a `gulp regenerate-metadata` task that creates a small SDK for vanilla and Azure, installs dependencies and tries to bundle it?


daschult:
In autorest.java we had separate gulp tasks for generating specific test swaggers with certain flags turned on. Those generations took about a second each (which were done in parallel so they ended up taking about a second total), and I found them valuable.

PR #229, opened 2018-07-27
Fix tests that were broken after LRO work

daschult:


test/azure/lro.ts
```diff
@@ -28,294 +29,186 @@ describe('typescript', function () {
     var testClient = new AutoRestLongRunningOperationTestService(credentials, baseUri, clientOptions);
     var product: AutoRestLongRunningOperationTestServiceModels.Product = { location: 'West US' };
 
-    it('should work with Put201CreatingSucceeded200', function (done) {
-      testClient.lROs.put201CreatingSucceeded200({ product: product }, function (error, result) {
-        should.not.exist(error);
-        done();
-      });
+    it('should work with Put201CreatingSucceeded200', async () => {
```
RikkiGibson: IMO don't go out of your way to change `function` syntax to `() => ...` syntax--especially because mocha tests depend on the dynamic `this` to make calls like `.timeout` and `.skip` on a per-test or conditional basis.

RikkiGibson: However, thank you for updating these to use Promises! This tends to make the test fail in a timely manner unlike the callbacks where we throw AssertionErrors into the aether instead of calling `done(err)`

test/azure/lro.ts
```diff
@@ -368,206 +261,131 @@ describe('typescript', function () {
 
     //TODO: Port more C# test case over after 4103936 Fix exception type
```
RikkiGibson: Is this still relevant?

test/azuremetadata/generated/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "AutoRestLongRunningOperationTestService Library with typescript type definitions for node.js and browser.",
   "version": "",
   "dependencies": {
-    "ms-rest-azure-js": "~0.11.115"
+    "ms-rest-azure-js": "C:/Users/daschult/Sources/ms-rest-azure-js"
```
RikkiGibson: Waiting on a new publish job?

RikkiGibson:
```
  1) typescript
       Swagger LRO Happy BAT
         should work with DeleteAsyncNoHeaderInRetry:
     TypeError: Cannot read property 'status' of undefined
      at retry (node_modules/ms-rest-js/dist/lib/policies/exponentialRetryPolicy.js:105:52)
      at /home/travis/build/Azure/autorest.typescript/node_modules/ms-rest-js/dist/lib/policies/exponentialRetryPolicy.js:46:46
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:188:7)
```

Showing in nodejs but not chrome. I recall finding this problem late last week.. Very strange.

daschult:
Yup, I'll look further into that test failure, and if I can't figure it out then I'll just skip it and move onto something else.

PR #230, opened 2018-08-01
Export context

RikkiGibson:
Exports the context and operations directly in the client class so that the context pattern can be used from an external module.

### DIFF
--- a/src/DSL/TSExport.cs
+++ b/src/DSL/TSExport.cs
@@ -1,0 +1,89 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+//
+
+using System;
+using System.Collections.Generic;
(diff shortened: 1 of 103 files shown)


src/DSL/TSExport.cs
```diff
@@ -0,0 +1,89 @@
...
+
+namespace AutoRest.TypeScript.DSL
+{
+    /// <summary>
+    /// A TypeScript DSL representation for an ESM export block.
+    /// </summary>
+    public class TSExport : IDisposable
+    {
+        private readonly TSBuilder builder;
+        private string propertyBeingConstructed;
```
RikkiGibson: this was copy/pasted and should be deleted

RikkiGibson:
We may choose to produce a separate index file down the line---as long as it exports all this same stuff, there won't be a breaking change.

PR #247, opened 2018-08-15
Disable client side validation

### LINKED ISSUE #243: AutoRest should provide a way to disable client-side validation
Flag doesn't work. If we are using autorest-typescript for building API test's, we don't want the client-side validation included.

RikkiGibson:
Resolves #243 by simply not emitting the constraints.

We may wish to generate additional tsdoc that describes the constraints that we check at runtime, but if we do that it will be part of a separate task.


src/vanilla/ClientModelExtensions.cs
```diff
@@ -727,7 +728,8 @@ public static void ConstructMapper(TSObject mapper, IModelType type, string seri
 
             void applyConstraints(TSObject obj)
             {
-                if (constraints != null && constraints.Any())
+                bool useClientSideValidation = (bool) Settings.Instance.CustomSettings["ClientSideValidation"];
```
daschult: Can you add this to GeneratorSettingsTS? That way we don't have to use the string literal.

RikkiGibson:
325cd87 shows the difference

shabfactor:
@RikkiGibson  so do i need to install latest version to make it work for TS?

RikkiGibson:
Yes, please try generating with `@microsoft.azure/autorest.typescript@preview` and install `ms-rest-js@preview` in your project

shabfactor:
@RikkiGibson  when i tried to install `ms-rest-js@preview` it did ask for the version to install, which one i should be installing?

RikkiGibson:
Please try 0.19.380

DavidNorena:
is this documented ?

PR #248, opened 2018-08-16
Better response types

### LINKED ISSUE #237: HttpOperationResponse properties should be required based on the operation
Right now, properties like parsedBody, readableStreamBody, etc. are always all optional. This is bad for the user experience because they don't know which properties to expect to be present or absent.

We should make the properties known to be present in a certain operation's response required, and ideally remove those properties which we know will be absent.
- If a response headers type is defined, the parsedHeaders should be required.
- If a response body type is defined, the parsedBody should be required.
- In a streaming scenario, the blob and Readable properties should be present and maybe required.

RikkiGibson:
Resolves #237

Depends on Azure/ms-rest-js#191

- Idea: The docs for the generated body properties could pull from the docs for the model type used as the body. Then when someone just types `response.parsedBody` in the IDE they might see something more meaningful.
  - Shelving this, will maybe experiment with it down the line.


src/azure/Model/CodeModelTSa.cs
```diff
@@ -77,7 +77,7 @@ public override string ConstructRuntimeImportForModelIndex()
 
         public override string PackageDependencies()
         {
-            return "\"ms-rest-azure-js\": \"~0.13.130\"";
+            return "\"ms-rest-azure-js\": \"/Users/rigibson/src/ms-rest-azure-js\"";
```
daschult: Run `npm run preview`?

src/vanilla/Model/MethodTS.cs
```diff
@@ -45,21 +45,40 @@ public enum MethodFlavor
...
+            get
+            {
+                string responseName = HttpOperationResponseName;
+                return responseName == baseHttpOperationResponseName
+                    ? baseHttpOperationResponseName
+                    : "Models." + responseName;
+            }
+        }
+
+        public bool HasCustomHttpResponseType => HttpOperationResponseName != baseHttpOperationResponseName;
```
daschult: This property's value is very close to the logic on [line 551](https://github.com/Azure/autorest.typescript/pull/248/files#diff-a0945a4f62f5a6e958b85ad35db7086eR551,) but slightly different. Can you somehow point out the difference between these two expressions (either with a new property or a comment)?

src/vanilla/Model/MethodTS.cs
```diff
@@ -738,6 +763,49 @@ private static void GenerateRequestParameterPath(TSObject parent, string propert
             }
         }
 
+        public string GenerateResponseType(string emptyLine)
+        {
+            TSBuilder builder = new TSBuilder();
+            builder.ExportInterface(HttpOperationResponseName, "msRest.BaseHttpOperationResponse", iface =>
```
daschult: Use the constant defined above instead of the string literal for "msRest.BaseHttpOperationResponse".

src/azure/Model/CodeModelTSa.cs
RikkiGibson: Once the dependent PRs for ms-rest-js (and maybe ms-rest-azure-js) are merged

src/vanilla/Model/MethodTS.cs
RikkiGibson: I can just fix line 551 to use this property. The difference in practice is nil.

daschult: That's not true, though, right? I thought that the property didn't include the `msRest.` prefix?

RikkiGibson: The base response is always referenced as `msRest.BaseHttpOperationResponse`. The custom responses are what get prefixed at the usage site.

RikkiGibson:
BTW: The return types on the `begin` methods are basically inaccurate because they indicate (for instance) that they'll contain the resource in the body when that might not be true in practice. I'm assuming you're overhauling this anyway so I'm leaving it alone.

PR #253, opened 2018-08-17
Remove 'git clean -xdf' from metadata tests

daschult:
I was consistently running into errors on a clean copy of the master branch whenever I ran `gulp test` with errors about webpack or other various problems. If I remove these two lines, all of the problems go away.


RikkiGibson:
So strange. I'd like to follow up on what was happening here sometime as the git clean does help with the `npm run local` scenario.

PR #255, opened 2018-08-22
Update LRO methods

daschult:
Update generated LRO methods so that we no longer output a wrapper method for begin* methods.
Modified the behavior of generated LRO methods so that they now use the returned LROPoller.pollUntilFinished() method.


src/azure/Model/MethodTSa.cs
```diff
@@ -1,13 +1,14 @@
...
 
-using System.Linq;
-using System.Net;
 using AutoRest.Core.Model;
+using AutoRest.Core.Utilities;
 using AutoRest.Extensions.Azure;
 using AutoRest.TypeScript.DSL;
 using AutoRest.TypeScript.Model;
 using Newtonsoft.Json;
+using System.Linq;
```
kpajdzik: Do we have a convention to keep System.* usings first?

RikkiGibson:
Was there more work pending on this?

daschult:
@RikkiGibson Yup. I just finished up the deserialization work that you asked me about last week.

PR #258, opened 2018-08-23
Date as string

### LINKED ISSUE #251: Accept strings for date parameters
Some services need to produce date strings in formats other than what is produced by [Date.toISOString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString).

Let's consider widening the Date parameter types to be `Date | string`. In the case where a string is provided, we will just write it to the wire without modification.

RikkiGibson:
Resolves #251

Introduces a `--model-date-as-string` flag which makes the generator produce `string` instead of `Date` throughout.


src/vanilla/TransformerTS.cs
```diff
@@ -33,9 +33,61 @@ public override CodeModelTS TransformCodeModel(CodeModel cm)
...
 
+            if (codeModel.ModelDateAsString)
+            {
+                ConvertDateToString(codeModel);
+            }
+
             return codeModel;
         }
 
+        private void ConvertDateToString(CodeModelTS codeModel)
```
RikkiGibson: Unfortunately this proved the most expedient approach, as passing the CodeModel down through all the TSType, TSProperty etc methods from templates was unreasonably difficult (`IModelType.CodeModel` is not populated and can't be generically populated because there's no setter on the interface)

RikkiGibson:
It's just the same node unit test failure so will just merge

test/date-as-string/dateAsStringTests.ts
```diff
@@ -0,0 +1,17 @@
+import * as assert from "assert";
+import { AutoRestDateTestService } from "./generated/BodyDate/autoRestDateTestService";
+
+const testClient = new AutoRestDateTestService();
+describe("--model-date-as-string=false", function () {
```
RikkiGibson: this was misnamed, will want to correct when we get a chance

RikkiGibson: should be `--model-date-as-string=true`

PR #259, opened 2018-08-27
Fix bundle size script

RikkiGibson:
I found this silently broke when the dependencies script changed, probably because we didn't use the `import` syntax in TypeScript.

This should do the trick, and also succeed at reporting non-webpack errors. `npm i` should be sufficient in this environment as the branches being compared should not depend on any local versions of packages.


RikkiGibson:
Looks like the Regenerate Azure job had a transient error.

PR #263, opened 2018-08-28
Response flattening

### LINKED ISSUE #241: Flatten parsedBody and parsedHeaders into a single response type
It would be very preferable to be able to ship *just one* method for each "normal" (non-paging, non-
LRO) Swagger operation which makes all the properties defined on the response easily accessible without searching through different parsedHeaders, parsedBody etc properties.

After discussing with @XiaoningLiu and @JeffreyRichter we've considered a design which *flattens* the properties specifically defined in the Swagger, a bit like the following:

```ts
interface HttpResponse {
  request: WebResource;
  status: number;
  headers: HttpHeaders;
  // maybe include body properties? genuinely not sure
}


// generated per-operation for each operation containing a headers definition.
interface MyOperationResponse {
  // contains the raw response data not defined in the swagger
  response: HttpResponse;

  myCustomHeader1: string;
  myCustomHeader2: string; // etc with all defined response headers

  // this requires duplication of the body properties and tsdocs, but might help us disambiguate common names between response headers and body properties.
  myBodyProperty1: number;
  myBodyProperty2: Date // etc with all defined body properties

  // in the stream case, we would just stick the readableStreamBody and blobBody right on here.
}
```

Alternatively, since response header definitions are not shared between operations, we could define the response like so:

```ts
type MyOperationResponse = {
  // the special .response property
  response: HttpResponse;
  // all headers definitions
  // if a header has the same name as a body property,
  // we will have to come up with an unambiguous name for the header here
  myCustomHeader1: string;
  myCustomHeader2: string;
} & Models.MyBodyType // could even say "extends" here if it's easier for users to understand
```

RikkiGibson:
Resolves #241


test/azure/generated/AzureCompositeModelClient/azureCompositeModel.ts
```diff
@@ -85,13 +85,18 @@ class AzureCompositeModel extends AzureCompositeModelContext {
    *
    * @reject {Error|ServiceError} The error object.
    */
-  listWithHttpOperationResponse(resourceGroupName: string, options?: msRest.RequestOptionsBase): Promise<Models.ListResponse> {
+  list(resourceGroupName: string): Promise<Models.ListResponse>;
+  list(resourceGroupName: string, options: msRest.RequestOptionsBase): Promise<Models.ListResponse>;
+  list(resourceGroupName: string, callback: msRest.ServiceCallback<Models.CatalogArray>): void;
```
RikkiGibson: remaining design decision: should the ServiceCallback receive only the original body, or should it get the new flattened response value? Keeping in mind that the HttpOperationResponse is still available by defining additional parameters on the callback.

RikkiGibson:
The preview script does weird stuff in my environment:

1. it seems to start up node REPLs that I have to ctrl+c out of to make the script continue running.
2. ms-rest[-azure]-js's preview scripts have a type-checking error so I have to modify package.json to pass `-T` to ts-node. (can also just fix the type errors, but yeah.)

Will try and address these when I get a chance.

test/azure/generated/AzureCompositeModelClient/azureCompositeModel.ts
RikkiGibson: I'll take it that you guys are basically on board with preserving the current ServiceCallback behavior? @amarzavery

RikkiGibson:
Down the line we can decide to make tweaks like using `& BodyType` instead of inlining the BodyType properties and it won't be a breaking change. For now I feel it's better for readability to avoid the indirection and only use the intersection types for either:

1. very frequently reused base types
2. object types that can only be represented with intersection

PR #264, opened 2018-08-29
Date time as string

### LINKED ISSUE #251: Accept strings for date parameters

RikkiGibson:
Resolves #251 (again).


README.md
```diff
@@ -71,20 +71,22 @@ type Colors = 'red color' | 'green-color' | 'blue_color';
...
-### Generate dates as strings
-Swagger strings can have the format `"date"`, `"date-time"` or `"date-time-rfc1123"`, which causes AutoRest to model the parameter as JavaScript Date.
+### Generate date-times as strings
+Swagger strings can have the format `"date-time"`, which causes AutoRest to model the parameter as JavaScript Date.
 ```ts
   getDate(): Promise<Date>;
 ```
 
-If you pass `--model-date-as-string=true` to the generator, it will instead model and treat the string as a plain old `string` in the interfaces and at runtime.
+If you pass `--model-date-time-as-string=true` to the generator, it will instead model and treat the string as a plain old `string` in the interfaces and at runtime.
```
amarzavery: Please mention that this will affect ISO8601 format of the date string. That makes it super clear in the readme.

RikkiGibson: "If you pass `--model-date-time-as-string=true` to the generator, it will instead
model and treat the ISO 8601 formatted date-time as a plain old `string` in the interfaces and at runtime."

good?

PR #266, opened 2018-08-29
Readonly doc

### LINKED ISSUE #240: readonly properties in an interface should mention in their description that this property will not be serialized.
properties marked as `"readOnly": true` in a swagger spec indicate that this property will not be sent on the wire. It can only be populated by the server. This information should be appended to the description of every readonly property in the model interface. 

When a property on an interface is marked as `readonly` in TS, it means that the value of the property cannot be changed once set. While creating an object of a given type, one can provide the value for the readonly property. In fact, TS would complain if the readonly property is not optional and is not provided by the customer. However, when the sdk serializes the object, it will strip this property. Letting the customer know about this in the description of the property can be valuable.

```ts
interface Foo {
  /**
   * @property {string} x Describes the identifier for Foo.
   * 
   * **NOTE:** This property will not be serialized. It can only be
   * populated by the server.
   */
  readonly id: string;
  y: string;
}
let x1: Foo = {
  y: "",
  id: "some-id"
};

console.log(x1);
```

### LINKED ISSUE #246: Support marking response headers as required
Right now, [Swagger response headers](https://swagger.io/specification/v2/#headersObject) don't appear to support a `required` attribute like model properties do. We interpret this by creating headers models where all the properties are optional. @fearthecowboy may be able to comment on the tradeoffs here.

This results in a worse user experience because every header appears like it could be absent, even if the spec authors want to guarantee that a given response will always include certain headers.

Can we change the interpretation to instead see all response headers as required/definitely present? Perhaps always including default values for response headers in mappers would help?

RikkiGibson:
Resolves #240
Resolves #246

### DIFF
--- a/src/vanilla/Model/MethodTS.cs
+++ b/src/vanilla/Model/MethodTS.cs
@@ -774,7 +774,8 @@ namespace AutoRest.TypeScript.Model
                         foreach (var property in ((CompositeTypeTS)ReturnType.Headers).Properties)
                         {
                             iface.DocumentationComment(property.Documentation);
-                            iface.Property(property.Name, property.ModelType.TSType(inModelsModule: true));
+                            // hard-coding response headers non-optional because Swagger itself can't specify whether or not they're required.
+                            iface.Property(property.Name, property.ModelType.TSType(inModelsModule: true), optional: false);
                         }
                     }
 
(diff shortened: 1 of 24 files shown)


src/vanilla/Model/MethodTS.cs
```diff
@@ -774,7 +774,8 @@ public string GenerateResponseType(string emptyLine)
                         foreach (var property in ((CompositeTypeTS)ReturnType.Headers).Properties)
                         {
                             iface.DocumentationComment(property.Documentation);
-                            iface.Property(property.Name, property.ModelType.TSType(inModelsModule: true));
+                            // hard-coding response headers non-optional because Swagger itself can't specify whether or not they're required.
+                            iface.Property(property.Name, property.ModelType.TSType(inModelsModule: true), optional: false);
```
RikkiGibson: will want to do this when generating headers model interfaces as well and will maybe solve that issue at the same time

PR #267, opened 2018-08-29
Automatically rename conflicting options properties

### LINKED ISSUE #245: REST API parameters shouldn't conflict with RequestOptionsBase properties
All of our generated options objects inherit from [RequestOptionsBase](https://github.com/Azure/ms-rest-js/blob/master/lib/webResource.ts#L407), adding the optional REST API parameters as additional properties on the interface.

Azure Storage has a `timeout` query parameter which overrides the `timeout` in RequestOptionsBase. We will need some way to make these properties distinct at runtime--right now Storage is basically forced to use the same value for both the client-side timeout and the query parameter.

RikkiGibson:
Resolves #245

Would be nice to have an autorest.testserver case to verify this. Maybe we can locally turn the parameter grouping off for the paging test?

Also, this might be difficult to maintain as the ServiceClientOptions and RequestOptionsBase continue to evolve :/


RikkiGibson:
In practice this results in a `timeout` property on the options for the client-side timeout, and a `timeoutParameter` for the parameter described in the Swagger.

After thinking it over a bit, I feel these names are good enough.
- `timeout` is an HTTP client property, equivalent to the well-known XMLHttpRequest.timeout and the axios timeout properties.
- `timeoutParameter` is a parameter to the REST API.

src/vanilla/TransformerTS.cs
```diff
@@ -318,25 +318,56 @@ public void PerformParameterMapping(CodeModelTS cm)
...
                 // We could end up having a property that is required but has a default value based on the above condition. If so then make it optional.
-                optionalProperitesOnClient.Where(p => p.IsRequired && !string.IsNullOrEmpty(p.DefaultValue)).ForEach(prop => prop.IsRequired = false);
-                modelType.AddRange(optionalProperitesOnClient);
+                optionalPropertiesOnClient.Where(p => p.IsRequired && !string.IsNullOrEmpty(p.DefaultValue)).ForEach(prop => prop.IsRequired = false);
+
+                foreach (var prop in optionalPropertiesOnClient)
+                {
+                    if (reservedClientPropertyNames.Contains(prop.Name.ToCamelCase()))
+                    {
+                        prop.Name += "Property";
```
RikkiGibson: This might be introducing a problem where the property was renamed on the options object but not on the client itself. Needs a test case, maybe needs to be handled in a separate PR.

RikkiGibson:
I added a test fixture which demonstrates that the automatic renaming works in the cases we care about so far. I think keeping up on automatic renaming of method and client parameters will be a bit of whack-a-mole in practice.

src/vanilla/TransformerTS.cs
RikkiGibson: It looks like there's no bug (for now). Reference semantics will wreck us when the opportunity presents, though.

PR #272, opened 2018-09-05
Multi api

### LINKED ISSUE #265: Multi-api generation
- Generate private package.json, tsconfig.json, tsconfig.es.json, etc. when `--multiapi` flag is set
- Generate project root when some kind of flag we define like `--multiapi-root` is set

RikkiGibson:
Resolves #265 

I wanted to get rid of the ApiVersions setting but I found it too error-prone to list only the folders corresponding to another API version. (maybe searching for package.json files in subfolders will do the trick?)

I would like to add a fixture in here for multi-api but it will take a bit of legwork--particularly, will need to introduce a readme.md config for it.

See https://github.com/RikkiGibson/azure-rest-api-specs/blob/ts-multi-api/specification/storage/resource-manager/readme.typescript.md for an example config file for multi api.


test/multiapi/multiapiTests.ts
```diff
@@ -0,0 +1,15 @@
+import * as assert from "assert";
+
+import * as vDefault from "./generated";
```
RikkiGibson: I'm concerned that these tests will require a multiapi build before they pass. The dependencies in gulp are just gonna start looking a bit wacky.

RikkiGibson:
@daschult please have a look at your convenience

src/vanilla/Templates/PackageJson.cshtml
```diff
@@ -34,6 +43,6 @@
     "url": "@(Model.BugsUrl)"
   },
   "scripts": {
-    "build": "tsc && webpack"
+    "build": "tsc @(string.IsNullOrEmpty(Model.DefaultApiVersion) ? "" : "-b") && webpack"
```
daschult: What does "-b" do? It's not listed on the [TypeScript Compiler Options](https://www.typescriptlang.org/docs/handbook/compiler-options.html).

RikkiGibson: It stands for --bleeding-edge (no, it's short for [--build](https://www.typescriptlang.org/docs/handbook/project-references.html), and it really should be documented.)

PR #275, opened 2018-09-06
Restructure accessing settings within generator

daschult:


src/Program.cs
```diff
@@ -149,52 +153,44 @@ protected override async Task<bool> ProcessInternal()
...
-                codeModel = plugin.Transformer.TransformCodeModel(codeModel);
-                if (await GetValue<bool?>("sample-generation") ?? false)
-                {
-                    plugin.CodeGenerator.GenerateSamples(codeModel).GetAwaiter().GetResult();
-                }
-                else
-                {
-                    plugin.CodeGenerator.Generate(codeModel).GetAwaiter().GetResult();
-                }
+                string[] inputFiles = await GetValue<string[]>("input-file") ?? new[] { "" };
```
RikkiGibson: why an array containing empty string instead of an empty array?

daschult: I brought this change from the previous code, and I think they might've done it so that they could use `new[]` instead of `new string[0]`. I agree that it should be an empty array, though.

daschult:
The only failure is for bundle size, so I'm going to merge this through and see if I get the same failures with my next PR.

RikkiGibson:
I'm going to deal with the bundle size thing.

PR #276, opened 2018-09-06
Pull in a few more settings from autorest.nodejs

daschult:


daschult:
Just bundle size errors, so I'm going to merge this.

PR #277, opened 2018-09-06
Metadata and bundle size fixes

RikkiGibson:
- ts-loader doesn't yet support ts projects that use the new project references feature, so instead using the esm outputs as inputs to webpack
- Added a "prepare" script to build the generated project

I really wish I understood why the multiapi case puts the outputs in `./cjs/lib/file.js` and the metadata case puts the outputs in `./cjs/file.js`. I've pretty much overdosed on typescript configuration so I'm just gonna let it be for now.

Metadata tests just tend to bring up corner cases--to test the custom module resolution via the inner package.json files, we have to run an npm install in the multiapi folder before bundling.

### DIFF
--- a/.gulp/gulpfile.iced
+++ b/.gulp/gulpfile.iced
@@ -63,19 +63,16 @@ task 'test/metadata', '', ['test/vanilla-metadata', 'test/azure-metadata'], (don
 task 'test/vanilla-metadata', '', [], (done) ->
   cwd = "#{basefolder}/test/metadata/generated"
   await execute "npm install", { cwd: cwd, silent: false }, defer _
-  await execute "npm run build", { cwd: cwd, silent: false }, defer _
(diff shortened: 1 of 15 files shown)


RikkiGibson:
@daschult I think the merge may have went awry--errors due to the unknown property `DefaultApiVersion`.

daschult:
Yeah, I'll take care of it.

daschult:
@RikkiGibson I fixed the problems that I knew about, but the problems I'm seeing now are back to webpack errors, which are your specialty.

RikkiGibson:
So, as long as the base branch has a bad bundlesize script in it, we won't be able to see it working in CI. Bizarrely, we have to merge this while CI is red in order to get CI green again.

I'll merge this and use another PR to verify things are good again.

PR #280, opened 2018-09-10
Additional properties

### LINKED ISSUE #232: Handle additionalProperties: true
It appears there is an `additional-properties.json` in the testserver which we haven't generated a client for. Let's do that, implement the tests, and check the nodejs runtime/generator for any special handling that might have been added there but didn't make it over here.

RikkiGibson:
Resolves #232


src/vanilla/ClientModelExtensions.cs
```diff
@@ -896,6 +893,32 @@ void applyConstraints(TSObject obj)
                             }
                         });
                     }
+
+                    if (composite.AdditionalProperties != null)
+                    {
+                        typeObject.ObjectProperty("additionalProperties", additionalProperties =>
+                        {
+
```
daschult: Remove extra whitespace.

test/azuremetadata/generated/Lro/README.md
```diff
@@ -1,80 +1,80 @@
...
-    </script>
-  </head>
-  <body>
-  </body>
-</html>
-```
-
-# Related projects
- - [Microsoft Azure SDK for Javascript](https://github.com/Azure/azure-sdk-for-js)
+# Microsoft Azure SDK for isomorphic javascript - AutoRestLongRunningOperationTestService
```
RikkiGibson: Wish all this line endings junk wasn't in here. But I don't believe meaningful changes happened.

src/vanilla/Model/CompositeTypeTS.cs
```diff
@@ -67,6 +67,38 @@ public IReadOnlyList<CompositeTypeTS> ImmediatePolymorphicSubtypes
...
+                var type = AdditionalProperties.TSType(true);
+                if (type != "any")
+                {
+                    result += $"The value of an unknown property MUST be of type \"{type}\". Due to valid TS constraints " +
+                        $"we have modeled this as a union of `{type} | any`.";
+                }
+                else
+                {
+                    result += "The value of an unknown property can be of \"any\" type.";
+                }
```
daschult: Is it possible for AdditionalProperties to have a description property specified? If so, then I'd add that description first.

RikkiGibson: The AdditionalProperties property is created by the modeler. It has the generated description "Unmatched properties from the message are deserialized this collection".

It doesn't appear to be possible to add a description for the `"additionalProperties"` in Swagger, so I think I'm content with this approach.

Issue #278, opened 2018-09-06
100% testserver coverage


RikkiGibson:
https://github.com/Azure/autorest.typescript/search?q=TODO+coverage

Related to Azure/ms-rest-js#227

PR #284, opened 2018-09-12
Increase coverage

RikkiGibson:
Related to #278


test/azure/lro.ts
```diff
@@ -187,6 +187,21 @@ describe('typescript', function () {
       await testClient.lROs.post202NoRetry204({ product: product });
     });
 
+    // TODO coverage
+    it.skip('should work with LROPostDoubleHeadersFinalLocationGet', async () => {
+      await testClient.lROs.postDoubleHeadersFinalLocationGet({ product });
```
daschult: Do these methods have return values? If so, can you check them? I understand that this currently just shows that the test doesn't throw an error, but it would be nice if they had some sort of expected result as well.

test/vanilla/complexTypesTests.ts
```diff
@@ -473,6 +473,28 @@ describe('typescript', function () {
...
+
+        const picture = (response.siblings[1] as AutoRestComplexTestServiceModels.Sawshark).picture;
+        delete (response.siblings[1] as AutoRestComplexTestServiceModels.Sawshark).picture;
+        should.exist(picture);
+        picture.length.should.equal(5);
+        picture[0].should.equal(255);
+        picture[1].should.equal(255);
+        picture[2].should.equal(255);
+        picture[3].should.equal(255);
+        picture[4].should.equal(254);
```
daschult: Can this just be replaced by an `assert.deepStrictEqual([255, 255, 255, 255, 254], picture)`?

test/vanilla/extensibleEnumsTests.ts
```diff
@@ -0,0 +1,37 @@
...
+import { PetStoreInc } from "./generated/ExtensibleEnum/petStoreInc";
+import { DaysOfWeekExtensibleEnum, IntEnum } from "./generated/ExtensibleEnum/models";
+
+import * as assert from "assert";
+
+const testClient = new PetStoreInc();
+
+describe("Extensible enums", function () {
+  it("should handle expectedEnum", async function () {
+    await testClient.pet.getByPetId("tommy");
```
daschult: Check return value, please.

test/vanilla/complexTypesTests.ts
RikkiGibson: Nope :( deep equality comparison on buffer types just doesn't work the way you'd hope.

test/azure/lro.ts
RikkiGibson: I won't spend more time on the skipped tests for now-- this call on line 192 does not terminate--but whoever properly implements this test should check the return value if applicable

test/vanilla/extensibleEnumsTests.ts
RikkiGibson: Will take care of this one, then merge

PR #289, opened 2018-09-17
Automatically determine package.json version from existing package.js…

daschult:
…on or from NPM

### DIFF
--- a/src/vanilla/GeneratorSettingsTS.cs
+++ b/src/vanilla/GeneratorSettingsTS.cs
@@ -2,7 +2,16 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 //
 
+using AutoRest.Core;
 using AutoRest.Core.Extensibility;
+using AutoRest.Core.Logging;
+using Newtonsoft.Json.Linq;
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Text;
 
 namespace AutoRest.TypeScript
 {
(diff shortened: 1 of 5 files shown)


src/vanilla/GeneratorSettingsTS.cs
```diff
@@ -86,5 +95,209 @@ public class GeneratorSettingsTS : IGeneratorSettings
...
+                            }
+                            else
+                            {
+                                PackageVersion = packageJsonVersion;
+                                Log(Category.Information, $"Got version \"{PackageVersion}\" for package \"{PackageName}\" from existing package.json file.");
+                            }
+                        }
+                        catch (Exception e)
+                        {
+                            Log(Category.Error, $"{e.GetType().Name} - {e.Message}");
```
RikkiGibson: stack trace?

RikkiGibson: Guess there's not that large a net to catch it in?

src/vanilla/GeneratorSettingsTS.cs
```diff
@@ -86,5 +95,209 @@ public class GeneratorSettingsTS : IGeneratorSettings
...
+                    }
+                    Log(Category.Information, $"Package doesn't exist on NPM, so setting its initial version to \"{newPackageVersion}\".");
+                    PackageVersion = newPackageVersion;
+                }
+            }
+        }
+
+        private static string GetNPMPackageVersion(string packageName)
+        {
+            string filePath = ResolveFilePath("npm");
```
RikkiGibson: Does it really not work to just use `"npm"` as the file name?

PR #292, opened 2018-09-20
Check the value of LRO extension instead of just its existence

daschult:


daschult:
Surprisingly, `false`: https://github.com/Azure/azure-rest-api-specs/blob/master/specification/network/resource-manager/Microsoft.Network/stable/2018-08-01/networkProfile.json#L82

Issue #290, opened 2018-09-18
Doc comments on overloads


RikkiGibson:
We generate our set of TS overloads like so:

```ts
  /**
   * Lists all the storage accounts available under the subscription. Note that storage keys are not
   * returned; use the ListKeys operation for this.
   *
   * @param {RequestOptionsBase} [options] Optional Parameters.
   *
   * @returns {Promise} A promise is returned
   *
   * @resolve {HttpOperationResponse} The deserialized result object.
   *
   * @reject {Error|ServiceError} The error object.
   */
  list(): Promise<Models.StorageAccountsListResponse>;
  list(options: msRest.RequestOptionsBase): Promise<Models.StorageAccountsListResponse>;
  list(callback: msRest.ServiceCallback<Models.StorageAccountListResult>): void;
  list(options: msRest.RequestOptionsBase, callback: msRest.ServiceCallback<Models.StorageAccountListResult>): void;
  list(options?: msRest.RequestOptionsBase, callback?: msRest.ServiceCallback<Models.StorageAccountListResult>): Promise<Models.StorageAccountsListResponse> {
    return this.client.sendOperationRequest(
      {
        options
      },
      listOperationSpec,
      callback) as Promise<Models.StorageAccountsListResponse>;
  }
```

The trouble is that the doc comment is only associated with the "top" overload. If I pass an options object, I won't see the doc comment in the editor any more.

We may just want to generate the doc comments again, but it's a lot of noise in the generated code

RikkiGibson:
Also, I believe `@resolve` and `@reject` have not yet been officially adopted by JsDoc.

RikkiGibson:
@daschult do you think you could make time for this while working on reducing redundancy in the doc comments?

PR #316, opened 2018-09-27
Refactoring and add comment to each overload

daschult:
Resolves https://github.com/Azure/autorest.typescript/issues/290.

This got a little bigger than I had originally planned. This change restructures how we generate method comments, especially in the context of method overloads.


src/DSL/TSParameter.cs
```diff
@@ -0,0 +1,47 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// 
+
+namespace AutoRest.TypeScript.DSL
+{
+    /// <summary>
+    /// A data class that contains the details needed to describe a TypeScript parameter.
+    /// </summary>
+    public class TSParameter
```
RikkiGibson: I think you should reconsider this path unless you're positive it's the way you want-- it's starting to get beyond "fancy stringbuilders" toward "a data model you have to contend with and maintain along with the AutoRest model"-- I took away from our Java experience that the first thing is excellent while the second thing winds up being harder to maintain and understand.

RikkiGibson: Figuring out "why is the property value this way in the TSParameter" means traversing additional indirections compared to using the AutoRest Parameter type directly with the TSBuilder.

test/azure/generated/AzureCompositeModelClient/azureCompositeModel.ts
```diff
@@ -47,24 +45,27 @@ class AzureCompositeModel extends AzureCompositeModelContext {
...
-   * @returns A promise is returned
+   * @param [options] The optional parameters
+   * @returns Promise<Models.ListResponse>
+   */
+  list(resourceGroupName: string, options?: msRest.RequestOptionsBase): Promise<Models.ListResponse>;
+  /**
+   * @param resourceGroupName Resource Group ID.
+   * @param callback The callback
    */
-  list(resourceGroupName: string): Promise<Models.ListResponse>;
```
RikkiGibson: :+1: to removing these

test/azure/generated/AzureCompositeModelClient/azureCompositeModel.ts
```diff
@@ -47,24 +45,27 @@ class AzureCompositeModel extends AzureCompositeModelContext {
...
    * @param resourceGroupName Resource Group ID.
-   *
-   * @param [options] Optional Parameters.
-   *
-   * @returns A promise is returned
+   * @param [options] The optional parameters
+   * @returns Promise<Models.ListResponse>
+   */
+  list(resourceGroupName: string, options?: msRest.RequestOptionsBase): Promise<Models.ListResponse>;
+  /**
```
RikkiGibson: What's the value added for the doc comments with no description on the overloads, e.g. in the editor or in docs.microsoft.com?

RikkiGibson: I don't think I'd want to copy the doc around because it can bloat the .d.ts file, just wondering what documenting the parameters specifically does.

daschult: The value is that we don't duplicate the description in every overload. The idea is that in your editor the description is shown while you're choosing a method, but you don't need to see the description again after you've chosen the method. At that point you only need to see the parameters and the return type.

src/DSL/TSBuilder.cs
```diff
@@ -17,12 +17,17 @@ public class TSBuilder
...
 
+        private readonly int commentWordWrapWidth;
         private readonly StringBuilder contents = new StringBuilder();
         private readonly StringBuilder linePrefix = new StringBuilder();
         private readonly List<TSPosition> positions = new List<TSPosition>();
 
-        public const int multiLineCommentWordWrapWidth = 100;
+        public TSBuilder(int? commentWordWrapWidth = defaultCommentWordWrapWidth)
+        {
+            this.commentWordWrapWidth = commentWordWrapWidth ?? defaultCommentWordWrapWidth;
```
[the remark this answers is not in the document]
daschult: I guess I was trying to make this constructor really safe, but on second thought that's kind of stupid. I'll change it. Thanks for pointing this out.

src/DSL/TSParameter.cs
daschult: I tried to remove this and found very quickly that it was not as easy as I thought it would be. We have a lot of special logic around converting the Parameter's IModelType to its generated string representation that made it cumbersome for me to say, "This is the type I want to generate."
I understand your apprehension at seeing me introduce things that remind you of the work I did in the Java generator. I have definitely learned where I went wrong in that situation, and I'd be happy to talk to you about how this is a different scenario. I don't think a PR is the best place to have that conversation.
TLDR: I'm going to leave TSParameter as I think it helps the generation scenario, but I do appreciate the reminder/warning that my efforts in the Java generator didn't end up as well as I had hoped. :)

PR #317, opened 2018-09-28
Update MethodTS.HttpResponseName to look for duplicated ModelType names

daschult:


src/vanilla/Model/CodeModelTS.cs
```diff
@@ -92,12 +92,18 @@ public string BasePath
...
+        private IEnumerable<string> allModelTypeNames;
+        [JsonIgnore]
+        public IEnumerable<string> AllModelTypeNames
         {
-            return ModelTypes.Any();
+            get
+            {
+                if (allModelTypeNames == null)
+                {
+                    allModelTypeNames = AllModelTypes.Select((CompositeType modelType) => modelType.Name.ToString()).ToHashSet();
```
RikkiGibson: Why HashSet?

RikkiGibson: Risk of dupes?

test/azure/generated/Lro/models/index.ts
```diff
@@ -3075,8 +3075,7 @@ export type LROsPutAsyncSubResourceResponse = SubProduct & {
 };
 
 /**
- * Contains response data for the deleteProvisioning202Accepted200Succeeded
```
RikkiGibson: Did comment wrapping break? Or is this expected?

src/vanilla/Model/CodeModelTS.cs
daschult: Not because of possible duplicates. I went with a set because I was hoping that would be the best for performance. It should be a constant-time lookup compared to a sorted array's logarithmic time and a non-sorted array's linear time.

RikkiGibson: Oh, because the .Contains will direct to the HashSet implementation. Got it.

test/azure/generated/Lro/models/index.ts
daschult: Comment wrapping didn't break. Previously these comments were being created inside of Templates using the WrapComment() method. This PR changes them to be created by TSBuilder.DocumentationComment(), which uses a 100 column wrap width, instead of the 80 that WrapComment() uses.

PR #319, opened 2018-10-01
Package per api version

### LINKED ISSUE #265: Multi-api generation

RikkiGibson:
Resolves #265

remaining:
- [x] Generate a table of API versions with links in the main README.md
- [x] Fix multiapi fixture build (need npm link/script to create symlink/lerna)?


src/vanilla/GeneratorSettingsTS.cs
```diff
@@ -58,17 +58,26 @@ public class GeneratorSettingsTS : IGeneratorSettings
...
-        /// The path to the "default" API version, e.g. 2016-04-01.
-        /// The presence of this property indicates that we are generating
-        /// the multi-api "root" artifacts, and not any TS source files.
+        /// The NPM package name of the "default" API version package.
+        /// e.g. "@azure/arm-storage-2018-03-01-preview"
+        /// This package will be used as a dependency in the "alias" package.
+        /// e.g. "@azure/arm-storage"
         /// </summary>
-        public string DefaultApiVersion { get; set; }
+        public string DefaultApiVersionPackage { get; set; }
```
daschult: If you want to name this "alias", then I'd recommend changing the variable so that "alias" is in the name, such as `AliasPackageAPIVersion`.

src/vanilla/GeneratorSettingsTS.cs
```diff
@@ -58,17 +58,26 @@ public class GeneratorSettingsTS : IGeneratorSettings
...
+
+        /// <summary>The version of the package referenced in <see cref="DefaultApiVersionPackage" />.</summary>
+        public string AliasedNpmVersion { get; set; }
 
         /// <summary>
         /// All API version subfolders present in this package.
         /// </summary>
         public string[] ApiVersions { get; set; }
```
daschult: This shouldn't be necessary anymore, right?

src/vanilla/GeneratorSettingsTS.cs
```diff
@@ -58,17 +58,26 @@ public class GeneratorSettingsTS : IGeneratorSettings
...
+        /// The NPM package name of the "default" API version package.
+        /// e.g. "@azure/arm-storage-2018-03-01-preview"
+        /// This package will be used as a dependency in the "alias" package.
+        /// e.g. "@azure/arm-storage"
         /// </summary>
-        public string DefaultApiVersion { get; set; }
+        public string DefaultApiVersionPackage { get; set; }
+
+        /// <summary>The version of the package referenced in <see cref="DefaultApiVersionPackage" />.</summary>
+        public string AliasedNpmVersion { get; set; }
```
daschult: We should be able to calculate this from the `AliasPackageAPIVersion` property, right?

src/vanilla/GeneratorSettingsTS.cs
```diff
@@ -43,7 +43,7 @@ public class GeneratorSettingsTS : IGeneratorSettings
         public string SourceCodeFolderPath { get; set; } = "lib";
 
         /// <summary>
-        /// The name of the npm package.
+        /// The name of the npm package, e.g. "@azure/arm-storage".
         /// </summary>
         public string PackageName { get; set; }
```
daschult: I'd recommend renaming this to `PackageNamePrefix`. It isn't the package's name anymore, or at least in the majority of cases it isn't.

src/vanilla/Model/CodeModelTS.cs
```diff
@@ -30,6 +30,15 @@ public class CodeModelTS : CodeModel
...
         public string ContextName => Name + "Context";
 
+        [JsonIgnore]
+        public string ApiVersionPackageName => Settings.PackageName + "-" + Settings.ApiVersion;
+
+        /// <summary>
+        /// The value to use in the "name" property of package.json.
+        /// </summary>
+        [JsonIgnore]
+        public string CurrentNpmPackageName => Settings.Multiapi && string.IsNullOrEmpty(Settings.DefaultApiVersionPackage) ? ApiVersionPackageName : Settings.PackageName;
```
daschult: I'd recommend renaming this to `PackageName`.

src/vanilla/Model/CodeModelTS.cs
```diff
@@ -30,6 +30,15 @@ public class CodeModelTS : CodeModel
         [JsonIgnore]
         public string ContextName => Name + "Context";
 
+        [JsonIgnore]
+        public string ApiVersionPackageName => Settings.PackageName + "-" + Settings.ApiVersion;
```
daschult: Can't you remove this and just use the next property (`PackageName`)?

src/vanilla/Model/CodeModelTS.cs
```diff
@@ -393,7 +402,13 @@ public virtual string ConstructRuntimeImportForModelIndex()
 
         public virtual string PackageDependencies()
         {
-            return "\"ms-rest-js\": \"~0.22.422\"";
+            string deps = "\"ms-rest-js\": \"~0.22.422\"";
+            if (!string.IsNullOrEmpty(Settings.DefaultApiVersionPackage))
+            {
+                string version = Settings.AliasedNpmVersion ?? "^1.0.0";
```
daschult: In GeneratorSettingsTS we call NPM to get the latest version number of different packages. I'd recommend doing that here. If the configuration has an `AliasedNpmVersion`, then use that, otherwise use whatever is latest in NPM. If that doesn't exist, then default to "^1.0.0".

src/vanilla/GeneratorSettingsTS.cs
RikkiGibson: I think it will be necessary for the readme.md table.

src/vanilla/Model/CodeModelTS.cs
RikkiGibson: No

src/vanilla/GeneratorSettingsTS.cs
RikkiGibson: `bool MultiapiLatest`

src/vanilla/Model/CodeModelTS.cs
RikkiGibson: I can try to initialize the AliasedNpmVersion at the same time as the PackageVersion

src/vanilla/GeneratorSettingsTS.cs
RikkiGibson: Same discussion as https://github.com/Azure/autorest.typescript/pull/319#discussion_r221762722

RikkiGibson:
FYI: the multi-api fixture is making general node resolution start to get pretty weird. I had to modify mocha.opts to only look in `test/*/*.ts` and `test/*.ts` to avoid it trying to parse `.d.ts` files. So now files containing unit tests in deeper folders won't be found. Would really like a cleaner solution if possible.

fearthecowboy:
## success (version: 2.0.540)

Issue #318, opened 2018-10-01
Multipart/formdata codegen bug


R0nd:
Consider an API method with the following description:
```json
"/ControllerName": {
  "post": {
    "tags": ["ControllerName"],
    "operationId": "ControllerName_Create",
    "consumes": ["multipart/form-data"],
    "produces": ["application/json", "text/json", "text/html", "application/xml", "text/xml"],
    "parameters": [{ "name": "file", "in": "formData", "description": "...", "required": true, "type": "file" }],
    "responses": {
      "201": { "description": "...", "schema": { "$ref": "#/definitions/Void" } },
      "400": { "description": "...", "x-ms-error-response": true }
    }
  }
}
```

The operation specification generated for it is as follows:
```typescript
const createOperationSpec: msRest.OperationSpec = {
  httpMethod: "POST",
  path: "ControllerName",
  formDataParameters: [
    Parameters.file
  ],
  contentType: "multipart/form-data",
  contentType: "multipart/form-data",
  responses: {
    201: {
      bodyMapper: {
        serializedName: "parsedResponse",
        type: {
          name: "Object"
        }
      }
    },
    400: {},
    default: {}
  },
  serializer
};
```

The generated code is incorrect, as the object contains duplicate `contentType` fields.
I have traced the bug to lines 426-446 in MethodTS.cs, where my method definition meets both conditions (both has a formData parameter and has contentType defined on it).

RikkiGibson:
I wonder if we should prefer the user's specified `"consumes"` value to any implicitly added content types like `"multipart/form-data"`. If the operation contains formdata parameters but doesn't have the right content type, I think the runtime won't do the right thing.

RikkiGibson:
Any opinions on this @daschult @amarzavery

daschult:
Yes, we should definitely prefer the user's specified `"consumes"` value. I'll fix this right now.

RikkiGibson:
@R0nd please try generating again with the flag `--use=@microsoft.azure.autorest.typescript@2.0.556`

R0nd:
I can confirm the issue is fixed, though the package name in the autorest flag should be `--use=@microsoft.azure/autorest.typescript@2.0.556` in case anyone references this conversation.

RikkiGibson:
Thanks for spotting the typo :)

PR #322, opened 2018-10-02
Only add one contentType property

daschult:
Resolves https://github.com/Azure/autorest.typescript/issues/318.


RikkiGibson:
Did you do any ad-hoc testing for this?

RikkiGibson:
This is probably good to go--I still have to fix bundlesize

PR #323, opened 2018-10-02
Fix local, preview, and latest scripts

daschult:
Running `npm run local` and `npm run preview` were failing for me locally. This fixed the problems.


RikkiGibson:
Looks like some multiapi stuff wasn't regenerated

daschult:
Only CI failure is bundle-size. I'm good with that.

PR #324, opened 2018-10-02
Rollup

RikkiGibson:
Produces a rollup umd bundle and a minified bundle in the dist/ folder. This bundle works in both node and browser.

I wasn't able to get rollup-plugin-uglify or rollup-plugin-terser to work, so I'm just minifying using the uglifyjs CLI directly.


RikkiGibson:
Deleted some lockfiles by accident--will need to add them back to make the lerna fixture work.

src/vanilla/Model/CodeModelTS.cs
```diff
@@ -39,6 +40,28 @@ public class CodeModelTS : CodeModel
...
+        }
+
+        public string BundleVarName
+        {
+            get
+            {
+                string packageName = PackageName ?? "bundle";
+                int slashIndex = packageName.IndexOf("/");
+                string bundleVarName = packageName.Substring(slashIndex + 1).ToPascalCase();
+                return bundleVarName;
```
daschult: This is almost identical to `BundleFileName`. Can you change this to just be `return BundleFilename.ToPascalCase();`?

test/multiapi/packages/multiapi-test-2017-10-01/rollup.config.js
```diff
@@ -0,0 +1,31 @@
+import nodeResolve from "rollup-plugin-node-resolve";
```
RikkiGibson: @bterlson this is an example of one of our generated rollup configs

test/metadata/generated/BodyComplex/package.json
```diff
@@ -32,7 +35,7 @@
     "url": "https://github.com/azure/azure-sdk-for-js/issues"
   },
   "scripts": {
-    "build": "tsc && tsc -p tsconfig.esm.json && webpack",
+    "build": "tsc && rollup -c rollup.config.js && uglifyjs ./dist/bundle.js -c -m -o ./dist/bundle.min.js",
```
RikkiGibson: Does it provide enough value for the browser use cases to include a .min.js within the package?

RikkiGibson:
Moving from webpack to rollup reduced the size of arm-network latest from **704 KB** -> **524 KB** (!)

I know when webpack bundles arm-network it has *lots* of tiny functions that return individual modules, and rollup's major difference is that it inlines the modules instead of wrapping them in functions. But I definitely wasn't expecting savings that dramatic.

RikkiGibson:
Also, the linecount reduction appears to simply be from the fact that webpack pulls in a lot of stuff that's pretty much gone away with the move to rollup.

RikkiGibson:
new: 
- Automatically creates namespace objects in rollup config based on package scope e.g. `@azure/foo` becomes the var Azure.Foo
- minor whitespace fixes

test/metadata/generated/BodyComplex/package.json
bterlson: Yes, this is a must. Customers using CDNs like jsdelivr etc. which make the npm registry available as a CDN will appreciate minified assets (even though if they really care about code size they should bundle).

test/multiapi/packages/multiapi-test-2017-10-01/rollup.config.js
bterlson: Looks very nice! But do you need the commonjs plugin?

test/metadata/generated/BodyComplex/package.json
RikkiGibson: Makes sense--we will definitely stick with shipping the .min.js

RikkiGibson:
Looks like all the code we actually bundle uses esm syntax. Only the code under `esm/` and `tslib` are bundled.

We actually don't need the commonjs plugin. I'll remove it just for simplicity.

I'm also not positive if the json plugin is really needed. When I was doing the initial manual authoring of rollup.config.js over in azure-sdk-for-js, I was getting build errors pointing to .json files in dependencies that were only solved by adding the json plugin, so I'll probably keep that.

bterlson:
Curious where importing .json is used, can you give me a pointer?

RikkiGibson:
Looks like only ms-rest-js needed it. Using the following rollup.config.js (which is not merged into master or anything over in ms-rest-js, sorry):
```js
import nodeResolve from "rollup-plugin-node-resolve";
import commonjs from "rollup-plugin-commonjs";
/**
 * @type {import('rollup').RollupFileOptions}
 */
const config = {
  input: './es/lib/msRest.js',
  external: ["axios", "xml2js"],
  output: {
    file: "./dist/msRest.js",
    format: "cjs",
    name: "Microsoft.Rest",
    sourcemap: true,
  },
  plugins: [
    nodeResolve({ module: true }),
    commonjs()
  ]
};
export default config;
```

I get this error:
```sh
$ rollup -c rollup.config.js

./es/lib/msRest.js → ./dist/msRest.js...
[!] Error: Unexpected token
node_modules/tough-cookie/package.json (2:9)
1: {
2:   "_from": "tough-cookie@^2.4.3",
            ^
3:   "_id": "tough-cookie@2.4.3",
4:   "_inBundle": false,
```
which is solved by adding the dependency on the json plugin.

Actually, I was hoping to enlist you to help figure out what we need to do to rollup-ify ms-rest-js, as we've hit snags related to the [pkg.browser field](https://github.com/azure/ms-rest-js/tree/master/package.json#L31) being treated differently in rollup than in webpack. We use [NormalModuleReplacementPlugin](https://github.com/azure/ms-rest-js/tree/master/webpack.config.ts#L14) for bundling within the ms-rest-js repo and I don't see a great alternative to that.

RikkiGibson:
BTW-- I'd like to revamp sample generation, but it's basically blocked by the initial publish of ms-rest-browserauth.

I think I'd like to ship a v0.1.0 of ms-rest-browserauth, with docs/guide but no sample app yet, so that we can unblock the sample generation for the SDKs themselves,

fearthecowboy:
## success (version: 2.0.559)

PR #325, opened 2018-10-02
Remove previously committed package-lock.json files

daschult:


RikkiGibson:
npm doesn't seem to produce these files when run in CI, and then lerna gets cranky in CI :/

daschult:
Eh, I'll think about this later. Not worth my time right now.

PR #330, opened 2018-10-05
Exclude CloudErrorBody from OperationModelNames

daschult:


unittests/AssertEx.cs
```diff
@@ -86,5 +86,39 @@ public static void EqualLines(IEnumerable<string> expected, IEnumerable<string>
...
             }
         }
+
+        /// <summary>
+        /// Assert that the two IEnumerables have equal contents in the same order.
+        /// </summary>
+        /// <typeparam name="T">The type of the values contained by the IEnumerables.</typeparam>
+        /// <param name="expected">The expected IEnumerable.</param>
+        /// <param name="actual">The actual IEnumerable.</param>
+        public static void HaveEqualContents<T>(IEnumerable<T> expected, IEnumerable<T> actual)
```
RikkiGibson: any concerns about ordering when OperationModelNames I think returns set (or did you add ordering)?

daschult: I didn't add ordering, but if/when it comes up, we can deal with it then. I'm not too worried about it, especially since this is test code.

PR #332, opened 2018-10-12
Update scripts to use non-greedy regular expressions and to use caret…

daschult:
… instead of tilde


.scripts/dependencies.ts
```diff
@@ -222,15 +222,15 @@ function updateGeneratedPackageDependencyVersion(codeFilePath: string, dependenc
       codeFilePath,
       codeFileContents,
       dependencyName,
-      new RegExp(`\\\\"${dependencyName}\\\\": \\\\"(.*)\\\\"`),
+      new RegExp(`\\\\"${dependencyName}\\\\": \\\\"(.*?)\\\\"`),
```
kpajdzik: Does it mean the version can be empty?

daschult: I suppose it could. That's okay with me though.

PR #339, opened 2018-10-19
Fix pattern constraint forward slash escaping

daschult:


src/vanilla/ClientModelExtensions.cs
```diff
@@ -175,9 +175,32 @@ public static bool IsDictionaryContainingDateKind(this IModelType type)
             return tsType;
         }
 
-        private static string CreatePatternConstraintValue(string constraintValue)
+        internal static string CreatePatternConstraintValue(string constraintValue)
```
kpajdzik:
```suggestion
        internal static string CreateRegexPatternConstraintValue(string constraintValue)
```

daschult: Done.

PR #355, opened 2018-11-01
Fix package user agent telemetry generation

kpajdzik:
Related:
https://github.com/Azure/ms-rest-azure-js/pull/74
https://github.com/Azure/ms-rest-js/pull/254


src/azure/Model/CodeModelTSa.cs
```diff
@@ -71,7 +71,7 @@ public override string ConstructRuntimeImportForModelIndex()
 
         public override string PackageDependencies()
         {
-            return "\"ms-rest-azure-js\": \"^1.0.176\"" + ",\n" + base.PackageDependencies();
+            return "\"ms-rest-azure-js\": \"file:/Users/kamip/Repos/ms-rest-azure-js\"" + ",\n" + base.PackageDependencies();
```
kpajdzik: To fix

src/azure/Templates/AzureServiceClientContextTemplate.cshtml
```diff
@@ -38,6 +38,8 @@ export class @(Model.ContextName) extends msRestAzure.AzureServiceClient {
     if (!options) {
       options = {};
     }
+    const defaultUserAgent = msRest.getDefaultUserAgentValue();
+    options.userAgent = `${packageName}/${packageVersion} ${defaultUserAgent}`;
```
daschult: This should only be done if `options.userAgent == undefined`, right?

src/vanilla/Templates/ServiceClientContextTemplate.cshtml
```diff
@@ -39,6 +39,8 @@ export class @(Model.ContextName) extends msRest.ServiceClient {
     if (!options) {
       options = {};
     }
+    const defaultUserAgent = msRest.getDefaultUserAgentValue();
+    options.userAgent = `${packageName}/${packageVersion} ${defaultUserAgent}`;
```
daschult: This should only be done if `options.userAgent == undefined`, right?

Issue #367, opened 2018-11-07
Add section to generated README.md files for installing authentication packages


daschult:
See https://github.com/Azure/azure-sdk-for-js/issues/415 for more details.

Issue #366, opened 2018-11-07
Unify README.md file generation


daschult:
Merge the templates used to generate README.md files into a single template.
See https://github.com/Azure/azure-sdk-for-js/issues/415 for more details.

Issue #341, opened 2018-10-19
Create MarkdownBuilder


daschult:
This would help with generating our README.md files.

billytrend:
Randomly came across this issue: not sure if you'd find this useful: https://github.com/ts-common/azure-openapi-markdown. We use it in portal for readme manipulation and serialisation. Lmk if you want to chat about it!

daschult:
Thanks for thinking of us, but this isn't quite what we need. We need a way to generate Markdown files from our C#-based generator. The package you've linked is written in TypeScript, so it won't quite work for our needs.

billytrend:
ohh fair, also yeah you're probably just creating them from scratch ? We needed to parse -> manipulate -> serialise which was a bit more involved

Issue #340, opened 2018-10-19
Create HTMLBuilder


daschult:
This would help with building the browser samples.

PR #370, opened 2018-11-09
Unify readme generation processes and add auth installation instructions

daschult:
This PR merges our README.md templates into a single builder function. It also introduces MarkdownBuilder, HTMLBuilder, and JSBuilder classes. The TSBuilder class now also inherits from JSBuilder since TypeScript is an extension of JavaScript.

Resolves https://github.com/Azure/autorest.typescript/issues/367.
Resolves https://github.com/Azure/autorest.typescript/issues/366.
Resolves https://github.com/Azure/autorest.typescript/issues/341.
Resolves https://github.com/Azure/autorest.typescript/issues/340.


src/DSL/Builder.cs
```diff
@@ -0,0 +1,307 @@
...
+using System.Text;
+
+namespace AutoRest.TypeScript.DSL
+{
+    public class Builder : IBuilder
+    {
+        private const string newLine = "\n";
+        private const string singleIndent = "  ";
+
+        private readonly StringBuilder contents = new StringBuilder();
```
sergey-shandar: Instead of using `StringBuilder` we should use generators (`IEnumerable` and `yield return`) but I'm afraid it's a bigger task. `IEnumerable` should speed up the generation and reduce memory consumption.

PR #385, opened 2018-11-16
Add sourcemaps plugin to generated rollup configuration

daschult:
Resolves https://github.com/Azure/autorest.typescript/issues/378.


src/DSL/JSBuilder.cs
```diff
@@ -283,7 +283,12 @@ public void Object(Action<JSObject> action = null)
         /// <param name="text">The text to double-quote and add to this JSBuilder.</param>
         public void QuotedString(string text)
         {
-            Text($"\"{text}\"");
+            char quote = '"';
+            if (!string.IsNullOrEmpty(text) && (text.Contains('\n') || text.Contains('"')))
```
kpajdzik: Do we have a method to escape text in JS?

src/vanilla/Model/CodeModelTS.cs
```diff
@@ -983,5 +983,50 @@ private void GenerateRelatedProjects(MarkdownBuilder builder)
                 builder.List("[Microsoft Azure SDK for Javascript](https://github.com/Azure/azure-sdk-for-js)");
             });
         }
+
+        public string GenerateRollupConfig()
```
kpajdzik: Does rollup config change per package?

PR #395, opened 2018-11-29
Remove preview script and update generated markdown block names

daschult:


daschult:
@kpajdzik This seems to be blocked by the ms-rest-js deserialization bug we found today.

PR #400, opened 2018-12-04
Change default generated package version from 0.1.0 to 1.0.0

daschult:


src/vanilla/GeneratorSettingsTS.cs
```diff
@@ -20,7 +20,7 @@ public class GeneratorSettingsTS : IGeneratorSettings
         /// <summary>
         /// The version that packages that haven't been published to NPM before will be generated with.
         /// </summary>
-        public const string newPackageVersion = "0.1.0";
+        public const string newPackageVersion = "1.0.0";
```
sergey-shandar: You may use "1.0.1" at least for existing packages because we unpublished some packages with version "1.0.0"

daschult: I'd rather leave this at 1.0.0 and just react to those packages that were unpublished.

PR #425, opened 2019-01-25
Add generating "autoPublish" package.json property

kpajdzik:


src/vanilla/Model/CodeModelTS.cs
```diff
@@ -252,6 +252,19 @@ public string ClientPrefix
...
         }
 
+        public bool AutoPublish
+        {
+            get
+            {
+                if (Settings.AutoPublish.HasValue) {
+                    return Settings.AutoPublish.Value;
+                } else {
+                    bool isGeneratedForAzureSdkForJs = Settings.OutputFolder.Contains(defaultGitHubRepositoryName);
```
daschult: Please extract this expression to a function and use the function in the HomePageUrl property too.

kpajdzik: What do you mean?
```
if (value == null) {
    return value;
} else {
    const defaultValue = getDefaultValue();
    return defaultValue;
}
```

?

PR #431, opened 2019-02-12
Add option to override ServiceClientOptions

kpajdzik:


test/custom-service-client-options/generated/Xml/lib/autoRestSwaggerBATXMLServiceContext.ts
```diff
@@ -0,0 +1,47 @@
...
+    if (!options) {
+      options = {};
+    }
+
+    if (!options.userAgent) {
+      const defaultUserAgent = msRest.getDefaultUserAgentValue();
+      options.userAgent = `${packageName}/${packageVersion} ${defaultUserAgent}`;
+    }
+
+    if (true) {
```
daschult: Not a big fan of this. ;)

test/custom-service-client-options/generated/Xml/lib/autoRestSwaggerBATXMLServiceContext.ts
```diff
@@ -0,0 +1,47 @@
...
+      options = {
+        ...options,
+        "noRetryPolicy": "true",
```
daschult: Definitely agree that this should be reduced to a boolean instead of a string. We may need to have a mapping in the generator that knows the types of the ServiceClientOptions properties. If a value is provided on the command line but there isn't a known property in the mapping, then I'd say default it to a string.

PR #435, opened 2019-02-26
Extract constructor generation to builders

kpajdzik:


src/vanilla/Model/CodeModelTS.cs
```diff
@@ -391,6 +394,110 @@ public string RequiredConstructorParametersTS
             }
         }
 
+        public virtual string GenerateConstructor(string superParameterList, Action<TSBlock> guardChecks = null, Action<TSBlock> implementation = null)
```
daschult: Please introduce the difference between a parameter and an argument. It will make things much clearer.
```suggestion
        public virtual string GenerateConstructor(string superArgumentList, Action<TSBlock> guardChecks = null, Action<TSBlock> implementation = null)
```

src/vanilla/Model/CodeModelTS.cs
```diff
@@ -391,6 +394,110 @@ public string RequiredConstructorParametersTS
...
+            return this.GenerateConstructor(superParameterList, null, GenerateOperationInstantiating);
+        }
+
+        public virtual string GenerateContextConstructor(string emptyLine)
+        {
+            Action<TSBlock> guardCheck = block =>
+            {
+                foreach (var param in RequiredParameters)
+                {
+                    block.If($"{param.Name} === null || {param.Name} === undefined", then =>
```
daschult:
```suggestion
                    block.If($"{param.Name} == undefined", then =>
```

src/vanilla/Model/CodeModelTS.cs
```diff
@@ -391,6 +394,110 @@ public string RequiredConstructorParametersTS
...
+                    {
+                        then.Line($"throw new Error(\"\'{param.Name}\' cannot be null.\");");
```
daschult: This might be a good chance to introduce a Throw() function.

src/vanilla/Model/CodeModelTS.cs
```diff
@@ -391,6 +394,110 @@ public string RequiredConstructorParametersTS
...
+                }
+
+                if (RequiredParameters.Any())
+                {
+                    block.Line(emptyLine);
+                }
+
+                block.If("!options", then =>
+                {
+                    then.Line("options = {};");
```
daschult: Replace this with an `AssignmentObject("options")` call. If AssignmentObject() doesn't already exist, then I recommend creating it.

src/vanilla/Model/CodeModelTS.cs
```diff
@@ -391,6 +394,110 @@ public string RequiredConstructorParametersTS
...
+                    {
+                        block.If($"options.{property.Name} !== null && options.{property.Name} !== undefined", then =>
+                        {
+                            block.ThisAssignment(property.Name, $"options.{property.Name}");
+                        });
+                    }
+                }
+            };
+
+            return this.GenerateConstructor(superParameters, guardCheck, implementation);
```
daschult: Rather than have two Actions in this argument list, I'd prefer to see `guardCheck` go into `implementation` and just have a single Action. Then you could inline the action and it would look more like you were building the code here.
Was there a reason you chose to go this route over a single action?

src/vanilla/Model/CodeModelTS.cs
```diff
@@ -391,6 +394,110 @@ public string RequiredConstructorParametersTS
...
+        }
+
+        protected virtual string GetServiceClientOptionsName()
+        {
+            return ServiceClientOptions;
+        }
+
+        private void GenerateOperationInstantiating(TSBlock block)
```
daschult:
```suggestion
        private void GenerateOperationGroupInitialization(TSBlock block)
```

src/vanilla/Model/CodeModelTS.cs
kpajdzik: so the `super(...)` call goes between those two, that was the level of abstraction I introduced to make it generic enough to work for both Context and Client

kpajdzik:
Sample regenerated package: https://github.com/Azure/azure-sdk-for-js/pull/1348

PR #439, opened 2019-04-09
 Change generated source code directory from lib to src 

kpajdzik:


imdreamrunner:
This is a breaking change if the client doesn't specify `--source-code-folder-path`.
The README.md is not updated about this change either.

Maybe we should either update the README or set default back to `lib`?

amarzavery:
Thanks for letting us know. We will update README.md

PR #447, opened 2019-06-11
Add --use-azure-core flag to enable use of @azure/core-http

daviwil:
This change adds a new `--use-azure-core` flag that will cause `@azure/ms-rest-js` and `@azure/ms-rest-azure-js` to be replaced by `@azure/core-http` and `@azure/core-lro` (https://github.com/Azure/azure-sdk-for-js/pull/3584) in the generated output.

I didn't change every hard-coded usage of `msRestJs` and `msRestAzureJs` with the new variables because it would require far too many changes than necessary to use the different library names.

Please let me know if there's anything else I need to change!

/cc @kpajdzik @daschult @amarzavery

### DIFF
--- a/src/vanilla/Model/PrimaryTypeTS.cs
+++ b/src/vanilla/Model/PrimaryTypeTS.cs
@@ -58,7 +58,10 @@ namespace AutoRest.TypeScript.Model
                         return "Object";
 
                     case KnownPrimaryType.Credentials:
-                        return "msRest.ServiceClientCredentials";
+                        return "msRest.ServiceClientCredentials" + 
+                            ((CodeModel as CodeModelTS).Settings.UseAzureCore ?? false
+                                ? " | msRest.TokenCredential"
+                                : string.Empty);
                 }
                 throw new NotImplementedException($"Primary type {KnownPrimaryType} is not implemented in {GetType().Name}");
             }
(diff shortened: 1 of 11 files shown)


src/vanilla/Model/PrimaryTypeTS.cs
```diff
@@ -58,7 +58,10 @@ public virtual string ImplementationName
                         return "Object";
 
                     case KnownPrimaryType.Credentials:
-                        return "msRest.ServiceClientCredentials";
+                        return "msRest.ServiceClientCredentials" + 
+                            ((CodeModel as CodeModelTS).Settings.UseAzureCore ?? false
+                                ? " | msRest.TokenCredential"
+                                : string.Empty);
```
daschult: I noticed this code was duplicated from above. Please extract into a function so that the logic stays in one place.

daviwil:
Thanks for the review @daschult!  @amarzavery also recommended just find/replacing everything to replace `ms-rest-js` with `core-http` so I've pushed a branch called [`azure-core`](https://github.com/Azure/autorest.typescript/tree/azure-core) where I can start working on that while we're transitioning some libraries over to `core-http`.  Once we're feeling confident with the new module then I'll create a PR to merge over the changes.

PR #451, opened 2019-07-10
Update code generation to use @azure/core-http and @azure/core-arm

daviwil:
This change updates `autorest.typescript` to use the new `@azure/core-http` (formerly `@azure/ms-rest-js`) and `@azure/core-arm` (formerly `@azure/ms-rest-azure-js`) libraries as the runtime.  I've also changed how package metadata (mainly `package.json` dependencies and scripts) and Rollup config is generated to conform to the new standards of `azure-sdk-for-js`.

I'd certainly appreciate any feedback you have, especially anything extra I can do to make sure everything is running smoothly!

Worth noting to people unfamiliar with this repo: almost all of of the code under subfolders of the `tests/` path (the bulk of this PR) is all generated and then tested using files at the top level of the `tests/` path.  I wouldn't bother reviewing those unless you're just really interested in seeing the same changes over and over :)


src/vanilla/Model/CodeModelTS.cs
```diff
@@ -1167,7 +1166,7 @@ public string GenerateRollupConfig()
                 });
                 config.ArrayProperty("plugins", plugins =>
                 {
-                    plugins.Text("nodeResolve({ module: true })");
```
bterlson: mainfields should include `main`. Probably want ['module', 'main'] (and ['browser', 'module', 'main'] for the browser config).

daviwil: Thanks for catching that!  Since we only use `module` and `main` in the generated `package.json`, I've set this field to contain those two entry points and regenerated the test packages to reflect it.

daviwil:
I'm working on PR https://github.com/Azure/azure-sdk-for-js/pull/4237 in tandem to regenerate `@azure/arm-compute` using these changes.

daviwil:
Updated this PR to revert some of the changes based on the plan described in this comment: https://github.com/Azure/azure-sdk-for-js/pull/4237#issuecomment-510600408.  In short, we're holding off on changing the `package.json` scripts to integrate with the new data plane CI in `azure-sdk-for-js` so that we don't disrupt the current pipeline and release process for management plane libraries.

daviwil:
I'll merge this after Azure/azure-sdk-for-js#4279 gets merged.  Once I merge this PR, who should I ask to help get the update shipped?

amarzavery:
AFAIK, once the PR is merged to master, the bot should automatically publish the new version. However, it wouldn't be tagged as latest on npm. One needs to do that manually so that everyone gets the new version by default. 
@kpajdzik  @daschult - Can you confirm?

There is also one more place in autorest (some markdown file) needs to be updated to point to the latest version of autorest.typescript. Otherwise, autorest by default uses an old version of the generator. /cc @fearthecowboy  can shed some light on this.

ramya-rao-a:
To be on the safer side, I'd prefer to have the version here have the "preview" in it and not mark this as "latest" on npm

daviwil:
Just pushed a couple of changes:

- Bumped the version to `5.0.0-preview.1` to be a little more sure that users won't pick this up by mistake
- Swapped the order of `TokenCredential` and `ServiceClientCredentials` in generated `ServiceClient`s to imply the priority use of `TokenCredential` and also make `core-http` incompatibilities easier to spot.  Previously the compiler would complain that a `ServiceClientCredentials` implementor didn't match `TokenCredential` because it was the last type in the union.  This obscured the real problem, that there was a structural incompatibility issue between `ServiceClientCredentials` in `ms-rest-js` and `core-http`.

This PR is ready to be merged once CI is green!

PR #461, opened 2019-08-02
install min version of @azure/ms-rest-nodeauth text in README.md

amarzavery:


test/azuremetadata/generated/Lro/README.md
```diff
@@ -19,8 +19,9 @@ npm install
 
 ##### Install @azure/ms-rest-nodeauth
 
+- Please install minimum version of `"@azure/ms-rest-nodeauth": "^3.0.0"`.
 ```bash
-npm install @azure/ms-rest-nodeauth
+npm install @azure/ms-rest-nodeauth@"^3.0.0"
```
amarzavery: Have verified that specifying the range after `@` actually works while doing npm install.

Issue #464, opened 2019-09-05
Generated code doesn't honor errors defined in swagger


XiaoningLiu:
Hi, this is from storage JS SDK. Related issue is https://github.com/Azure/azure-sdk-for-js/issues/4999.

Storage SDKs use generator to generate what we called protocol layer code. When error happens, for example, Azure Storage REST API returns 500/404 status code in HTTP responses. @azure/ms-rest-js deserialization will throw `RestError` directly, and storage SDK pops error to the top. But this seems incorrect, as the thrown `RestError` should be further deseriazed according to the response headers and bodies, just like swagger defines.

![image](https://user-images.githubusercontent.com/6889160/64310323-3ae9e680-cfd2-11e9-806b-2bf681621f43.png)

For example, when getting properties from non exist blob.

```javascript
    try {
      await blobClient.getProperties();
    } catch (err) {
      console.log(err);
    }
```

![image](https://user-images.githubusercontent.com/6889160/64310126-8a7be280-cfd1-11e9-94aa-c9eeb1ca65fd.png)

We need to retrieve the raw headers to get storage error code by `err.response.headers.get("x-ms-error-code")`.

![image](https://user-images.githubusercontent.com/6889160/64310235-f1010080-cfd1-11e9-8e3a-fafd4cba6c33.png)

The generated code doesn't return errors defined in swagger.

The error thrown should parser error responses and fits the swagger definition correctly. For example, parser `ErrorCode` from `x-ms-error-code` header value. The swagger defines how to parser Azure Storage errors when returned status code is not in the expected list.

![image](https://user-images.githubusercontent.com/6889160/64310323-3ae9e680-cfd2-11e9-806b-2bf681621f43.png)

XiaoningLiu:
@amarzavery Can you take a look?

Another issue is that, the `errorCode` property is generated into wrong place. Just like above swagger samples, the `errorCode` is part of "default" reponses instead of "200". However, in the generated code, `errorCode` is part of "200"'s response models.

For example, we have a RESTful API `Get Blob Properties`. When status code is 200, the generated code will normally return a `BlobGetPropertiesResponse` model includes `BlobGetPropertiesHeaders` . However, `errorCode` is in  the generated `BlobGetPropertiesHeaders` model. 

The trick thing is, when status code is 200, `errorCode` will always be undefined. When status code is not 200, like 500. A `RestError` will be thrown. Caller won't access to the `errorCode` forever.

[...]

PR #469, opened 2019-09-19
Fix for default with headers

sarangan12:
Related to https://github.com/Azure/autorest.typescript/issues/464 & https://github.com/Azure/azure-sdk-for-js/issues/4999


test/xml/generated/Xml/models/xmlMappers.ts
```diff
@@ -20,6 +20,8 @@ export {
   ContainerProperties,
   CorsRule,
   ErrorModel,
+  JSONInput,
```
daviwil: Do you have any idea why all of the `JSONInput` and `JSONOutput` code is now being generated?  I don't see a relevant spec change.  Perhaps this is coming from a standard Swagger spec file from another repo?

sarangan12: Yes. It is from autorest.testserver

daviwil: Thanks!  Looks like it might have come from this PR: https://github.com/Azure/autorest.testserver/pull/87.  PR looks good, ship it!

PR #525, opened 2019-12-10
Use package-lock.json to ensure dependency versions on CI

daviwil:
This change helps unblock https://github.com/Azure/ms-rest-js/pull/377 which is currently failing due to a lack of `package-lock.json` files in this repository when testing dependent libraries.  The fix is to remove `package-lock.json` from `.gitignore` and check in the up-to-date `package-lock.json` files that get generated for the corresponding `package.json` files in this repo.

### DIFF
--- a/test/multiapi/packages/multiapi-test-2017-10-01/package-lock.json
+++ b/test/multiapi/packages/multiapi-test-2017-10-01/package-lock.json
@@ -1,0 +1,493 @@
+{
+	"name": "@azure/multiapi-test-2017-10-01",
+	"version": "1.0.0",
+	"lockfileVersion": 1,
+	"requires": true,
+	"dependencies": {
+		"@azure/abort-controller": {
+			"version": "1.0.1",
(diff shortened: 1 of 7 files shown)


daviwil:
`Check_Everything` failed because I haven't bumped the package version (this repo auto-ships).  This is a build-only change so there's no need to bump and release right now.

test/multiapi/packages/multiapi-test-2017-10-01/package-lock.json
```diff
@@ -0,0 +1,493 @@
+{
```
xirzec: do you need to check in these test package-lock.json files? It seems like you could tweak your .gitignore to ignore all lockfiles under test?

daviwil: This is actually the reason why the `ms-rest-js` repo's build is failing: `npm ci` is being used by Lerna when building this test folder and since the `package-lock.json` isn't there the install command fails.

Issue #503, opened 2019-12-04
Move integration tests to use TestServer v1


joheredi:
Here's: the repo: https://github.com/Azure/autorest.testserver

Estimate: [4 hours]

joheredi:
Merged changes

PR #534, opened 2019-12-13
[v6] Enable testserver-v1, small fix on default value and required params

joheredi:
This PR enables and sets by default integration tests against TestServer v1. All languages agreed to use v1 because the new test server still has some issues which can be confused with library bugs. So to avoid unnecessary friction moving integration tests to v1 while keeping the option to use the new version through `npm run integration-test:new`.

This PR also include small fixes for:
- Honor default value instead of forcing it to `""`  when null or undefined
- Updated parameter isRequired so that Constants are marked as non-required.


joheredi:
Fixes #503

package.json
```diff
@@ -6,9 +6,12 @@
         "start": "node ./dist/src/main.js",
         "test": "npm-run-all unit-test integration-test",
         "unit-test": "mocha -r ts-node/register './test/unit/**/*spec.ts'",
-        "integration-test": "npm-run-all start-test-server -p generate-bodystring generate-bodycomplex -s integration-test:alone stop-test-server",
+        "integration-test": "start-server-and-test start-test-server:v1 http://localhost:3000 generate-and-test",
+        "integration-test:new": "npm-run-all start-test-server -p generate-bodystring generate-bodycomplex -s integration-test:alone stop-test-server",
```
daviwil: I like the side-by-side approach!

PR #535, opened 2019-12-17
[v6] Full BodyComplex

joheredi:
Changes to handle complex schemas and pass all BodyComplex tests.
Also switching to prettier printWidth=80 to match default value

@daviwil I'd like to hear if you have any better ideas on how to handle additionaProperties, since I'm not sure that is the right/scalable way

### DIFF
--- a/src/generators/mappersGenerator.ts
+++ b/src/generators/mappersGenerator.ts
@@ -11,7 +11,7 @@ import {
   CompositeMapperType
 } from "@azure/core-http";
 import { ModelProperties } from "../transforms/mapperTransforms";
-import { keys, isEmpty } from "lodash";
+import { keys, isEmpty, isString } from "lodash";
 
 export function generateMappers(
   clientDetails: ClientDetails,
@@ -53,7 +53,10 @@ export function generateMappers(
         isExported: true,
         declarations: [
           {
-            name: normalizeName(mapper.serializedName, NameType.Class),
+            name: normalizeName(
+              compositeMapper.type.className || "",
+              NameType.Class
+            ),
             type: "coreHttp.CompositeMapper",
             initializer: writer => writeMapper(writer, mapper)
           }
@@ -89,17 +92,36 @@ export function generateMappers(
 const writeMapper = (writer: CodeBlockWriter, mapper: Mapper) => {
   const parents = extractParents(mapper);
   const { type, ...restMapper } = mapper;
-  const { modelProperties, ...restType } = type as CompositeMapperType;
+  const {
+    modelProperties,
+    polymorphicDiscriminator,
+    ...restType
+  } = type as CompositeMapperType;
+  const isReferenceDicriminator = isString(polymorphicDiscriminator);
   writer.block(() => {
     writeObjectProps(restMapper, writer)
       .write("type:")
       .block(() => {
         writeObjectProps(restType, writer)
-          .write("modelProperties:")
-          .block(() => {
-            writeParentMappers(parents, writer);
-            writeObjectProps(modelProperties, writer);
-          });
+          .conditionalWrite(
+            !!polymorphicDiscriminator,
+            "polymorphicDiscriminator:"
+          )
+          .conditionalWrite(
+            polymorphicDiscriminator && isReferenceDicriminator,
+            `${polymorphicDiscriminator as any},`
+          );
+        !isReferenceDicriminator &&
+          polymorphicDiscriminator &&
+          writer
+            .block(() => {
+              writeObjectProps(polymorphicDiscriminator, writer);
+            })
+            .write(",");
+        writer.write("modelProperties:").block(() => {
+          writeParentMappers(parents, writer);
+          writeObjectProps(modelProperties, writer);
+        });
       });
   });
 };
--- a/src/generators/mappersGenerator.ts
+++ b/src/generators/mappersGenerator.ts
@@ -38,7 +38,7 @@ export function generateMappers(
         const { uberParent, className } = compositeMapper.type;
         if (!uberParent || !className) {
           throw new Error(
-            `Expected CompositeMapper with polymorphicDiscriminator to specify uberParent property uberParent: ${uberParent} className: ${className}`
+            `Expected CompositeMapper with polymorphicDiscriminator to specify uberParent property (uberParent: ${uberParent}, className: ${className})`
           );
         }
 
--- a/src/generators/mappersGenerator.ts
+++ b/src/generators/mappersGenerator.ts
@@ -2,9 +2,16 @@
 // Licensed under the MIT License.
 
 import { ClientDetails } from "../models/clientDetails";
-import { Project, VariableDeclarationKind } from "ts-morph";
+import { Project, VariableDeclarationKind, CodeBlockWriter } from "ts-morph";
 import { NameType, normalizeName } from "../utils/nameUtils";
-import { MapperType } from "@azure/core-http";
+import {
+  MapperType,
+  Mapper,
+  CompositeMapper,
+  CompositeMapperType
+} from "@azure/core-http";
+import { ModelProperties } from "../transforms/mapperTransforms";
+import { keys, isEmpty } from "lodash";
 
 export function generateMappers(
   clientDetails: ClientDetails,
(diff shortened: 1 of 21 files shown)


src/generators/mappersGenerator.ts
```diff
@@ -21,29 +28,132 @@ export function generateMappers(
...
+                writer.write(`'${key}': ${discriminators[key]},`);
+              });
+            });
+          }
+        }
+      ]
+    });
 }
+
+const writeMapper = (writer: CodeBlockWriter, mapper: Mapper) => {
```
joheredi: @daviwil before, I was just doing a `JSON.stringify(mapper)` however since we need to spread the properties of the parent mapper and reference other mappers I moved to a WriterFunction. Do you know if there is a better/more elegant way to accomplish this?

i.e.

```typescript
export const Goblinshark: coreHttp.CompositeMapper = {
  serializedName: "goblin",
  type: {
    name: "Composite",
    className: "Goblinshark",
    uberParent: "Fish",
    polymorphicDiscriminator: Fish.type.polymorphicDiscriminator, // <- This
    modelProperties: {
      ...Shark.type.modelProperties, // <- This also
      jawsize: { type: { name: "Number" }, serializedName: "jawsize" },
      color: {
        type: { name: "String" },
        serializedName: "color",
        defaultValue: "gray"
      }
    }
  }
};
```

src/transforms/mapperTransforms.ts
```diff
@@ -44,7 +46,9 @@ const pipe = (
   ...fns: Array<(pipelineValue: PipelineValue) => PipelineValue>
 ) => (x: PipelineValue) => fns.reduce((v, f) => (!v.isHandled ? f(v) : v), x);
 
-type ModelProperties = { [propertyName: string]: Mapper };
+let uberParents: string[] = [];
```
joheredi: I think this is a temporary solution, I don't love having this global array here. I think maybe we can have a set of "indexes" as part of ClientDetails that we can pass around as needed, this uberParents being one. I think we can index/store other stuff we are already processing to avoid double processing and increase performance. @daviwil  what do you think?

These improvements will probably come later once we have more functionality in or perf becomes an issue

src/transforms/mapperTransforms.ts
```diff
@@ -169,18 +185,75 @@ function transformPrimitiveMapper(pipelineValue: PipelineValue) {
...
+    {
+      name: MapperType.Composite,
+      className,
+      modelProperties,
+      ...(discriminatorValue && {
+        uberParent,
+        polymorphicDiscriminator: `${uberParent}.type.polymorphicDiscriminator`
+      }),
+      ...(discriminator && {
+        // TODO need to handle uber parent in a more appropiate way
```
joheredi: Ignore this comment, will remove it in the next iteration... Now using the uberParents array

src/generators/mappersGenerator.ts
```diff
@@ -21,29 +28,132 @@ export function generateMappers(
...
 
+  for (const mapper of clientDetails.mappers) {
     if (mapper.type.name === MapperType.Composite && mapper.serializedName) {
+      const compositeMapper = mapper as CompositeMapper;
+
+      if (compositeMapper.type.polymorphicDiscriminator) {
+        const { uberParent, className } = compositeMapper.type;
+        if (!uberParent || !className) {
+          throw new Error(
+            `Expected CompositeMapper with polymorphicDiscriminator to specify uberParent property uberParent: ${uberParent} className: ${className}`
```
daviwil: Not super important, just adding parentheses to make the error message a little clearer
```suggestion
            `Expected CompositeMapper with polymorphicDiscriminator to specify uberParent property (uberParent: ${uberParent}, className: ${className})`
```

src/generators/mappersGenerator.ts
```diff
@@ -21,29 +28,132 @@ export function generateMappers(
...
+          );
+        }
+
+        if (uberParent === className) {
+          discriminators[className] = className;
```
daviwil: Should `serializedName` be used as the key here, or only when the `uberParent` is used?

src/generators/mappersGenerator.ts
```diff
@@ -21,29 +28,132 @@ export function generateMappers(
...
-      "$1:"
-    );
+  let discriminators: { [key: string]: string } = {};
 
+  for (const mapper of clientDetails.mappers) {
     if (mapper.type.name === MapperType.Composite && mapper.serializedName) {
+      const compositeMapper = mapper as CompositeMapper;
+
+      if (compositeMapper.type.polymorphicDiscriminator) {
+        const { uberParent, className } = compositeMapper.type;
```
daviwil: Do you mind leaving a short comment here to explain the uberParent and discriminator concepts?

src/generators/mappersGenerator.ts
```diff
@@ -21,29 +28,132 @@ export function generateMappers(
...
+        }
+      }
+
       mappersFile.addVariableStatement({
         isExported: true,
         declarations: [
           {
-            name: normalizeName(mapper.serializedName, NameType.Class),
+            name: normalizeName(
+              compositeMapper.type.className || "",
```
daviwil: What happens when name is assigned to an empty string?

src/generators/mappersGenerator.ts
```diff
@@ -21,29 +28,132 @@ export function generateMappers(
...
+
+function writeParentMappers(parents: string[], writer: CodeBlockWriter) {
+  (parents || []).forEach(parent => {
+    writer.write(`...${parent}.type.modelProperties,`);
+  });
+}
+
+function writeObjectProps(obj: any, writer: CodeBlockWriter) {
+  let currentPos = writer;
+  keys(obj).map(key => {
```
daviwil: Technically a `forEach` is more appropriate in this case even though JavaScript's `map` evaluates immediately too
```suggestion
  keys(obj).forEach(key => {
```

src/generators/mappersGenerator.ts
```diff
@@ -21,29 +28,132 @@ export function generateMappers(
...
+    currentPos = currentPos.writeLine(`${key}: ${JSON.stringify(obj[key])},`);
+  });
+
+  return currentPos;
+}
+
+function extractParents(mapper: Mapper) {
+  // TODO: We may need to create a MapperDetails of some sort which containes
```
daviwil: Ideally we would do that but it seems OK for now.  Maybe file an issue so that it gets done later?

src/transforms/mapperTransforms.ts
daviwil: Is this something that could be generated and stored at the time we build `ClientDetails` if we also create a list of `MapperDetails` at the same time?  What I wonder is if we could build all the metadata we need in one step before the generation phase so that the generator just has to spit out code based on what is in the `*Details` objects.

At any rate, it'd be nice to have an issue to track the need for a better solution here, whatever it ends up being

src/transforms/mapperTransforms.ts
```diff
@@ -169,18 +185,88 @@ function transformPrimitiveMapper(pipelineValue: PipelineValue) {
...
+  return immediateParents.some(p => p.type === SchemaType.Dictionary)
+    ? {
+        type: {
+          name: MapperType.Object
+        }
+      }
+    : undefined;
+}
+
+function checkUberParent(objectSchema: ObjectSchema) {
```
daviwil: Is it possible you could change this to `isUberParent` and then move the caching step to `transformObjectMapper` which seems to be the only place this gets used?  Or maybe I'm missing a use case?

test/unit/transforms/mapperTransforms.spec.ts
```diff
@@ -132,6 +136,7 @@ describe("transformObjectMapper", () => {
       serializedName: objectName,
       type: {
         name: MapperType.Composite,
+        className: getMapperClassName(schema),
```
daviwil: Maybe we need a string here to express what the `className` should be?  I think this would always pass because it's the same result as what get's put into the mapper, right?

src/generators/mappersGenerator.ts
joheredi: Yes good catch it should always be serializedName

src/generators/mappersGenerator.ts
joheredi: I have added a check to throw an exception if no ClassName was found

test/unit/transforms/mapperTransforms.spec.ts
joheredi: Yeah you're right adding the string

src/transforms/mapperTransforms.ts
joheredi: Makes sense I filed #538 to improve this

PR #536, opened 2019-12-18
Fix naming issue with x-ms-code-generation-settings name

### LINKED ISSUE #531: if x-ms-code-generation-settings name starts with a lowercase letter, codegen will fail
image
As you can see this swagger file has x-ms-code-generation-settings name starts with a lowercase letter.
after the code generation process, 
It will report error like 
`Namespace '"azure-sdk-for-js/sdk/softwareplan/arm-softwareplan/src/models/index"' has no exported member 'softwareplanClientOptions'.ts`
image
and the actual export Options is `SoftwareplanOptions`

I think the problem lies in https://github.com/Azure/autorest.typescript/blob/master/src/vanilla/TransformerTS.cs#L341 here.

joheredi:
Fixes #531

**Problem**
Autorest assumes that client name is Pascal case, so then `x-ms-code-generation-settings.name` is not it fails to reference the correct Models.

**Fix**
Override CodeModel.Name to always be PascalCase

Also upgrading core-http to 1.0.0 and adding "dom" to tsconfig


test/vanilla/generated/RequiredOptional/operations/explicit.ts
```diff
@@ -31,28 +31,28 @@ export class Explicit {
    * request is sent.
    * @param bodyParameter
    * @param [options] The optional parameters
-   * @returns Promise<Models.ExplicitPostRequiredIntegerParameterResponse>
+   * @returns Promise<coreHttp.RestResponse>
    */
-  postRequiredIntegerParameter(bodyParameter: number, options?: coreHttp.RequestOptionsBase): Promise<Models.ExplicitPostRequiredIntegerParameterResponse>;
+  postRequiredIntegerParameter(bodyParameter: number, options?: coreHttp.RequestOptionsBase): Promise<coreHttp.RestResponse>;
```
joheredi: Changes picked up from regenerating with latest changes in RequiredOptional: https://github.com/Azure/autorest.testserver/commit/e0d8dcad0f06f45ad6ec4416e45b326a139a63ff#diff-a0a97e774df31ce51de76b4a05cea0b7

README.md
```diff
@@ -10,7 +10,7 @@ npm install -g autorest
 This AutoRest extension generates TypeScript code that is compatible with:
 ```
 "@azure/core-arm": "^1.0.0-preview.1",
-"@azure/core-http": "^1.0.0-preview.1"
```
daviwil: Maybe this should go in the core-http bump PR?  Not really a big deal since it's just a README comment.

joheredi: Good catch thanks!

test/vanilla/generated/RequiredOptional/models/index.ts
```diff
@@ -189,303 +189,3 @@ export interface ExplicitPostOptionalArrayPropertyOptionalParams extends coreHtt
 export interface ExplicitPostOptionalArrayHeaderOptionalParams extends coreHttp.RequestOptionsBase {
   headerParameter?: string[];
 }
-
-/**
- * Contains response data for the getRequiredPath operation.
- */
-export type ImplicitGetRequiredPathResponse = ErrorModel & {
```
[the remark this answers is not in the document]
joheredi: Yes, they changed the swagger for OptionsBase so that it defines a success response. Before it only had a default one making all responses an error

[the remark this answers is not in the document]
joheredi: https://github.com/Azure/autorest.testserver/pull/92/files

PR #548, opened 2020-01-16
Rename to @autorest/typescript and prepare for dev releases

daviwil:
This change renames `@microsoft.azure/autorest.typescript` to `@autorest/typescript` and prepares the repo for shipping development releases to GitHub Releases on every successful commit to `master`.  

The advantage of this release automation is that it allows us to run pre-release versions of `@autorest/typescript` with the regression testing tool against `@autorest/core` and `@autorest/modelerfour`, ensuring that changes to the pipeline or modeler don't break our code generator.  This script is *not hooked up yet*; I'll do that after this PR is merged.

We'll only be publishing dev builds to GitHub Releases (for now) because AutoRest v3 can easily pull a specific version of a version from there without a corresponding npm release.  It will keep our npm page free of confusing version cruft.

**TODO**
- [x] Decide whether we should reset the version to `1.0.0` or forge ahead with `6.0.0`
- [ ] Test the release automation (after merging since I don't want to publish anything until we've agreed on the version!)

### DIFF
--- a/.scripts/publish-dev-release.yml
+++ b/.scripts/publish-dev-release.yml
@@ -1,7 +1,8 @@
 # Publishes a development release of @autorest/typescript for
 # any commit to the master branch.
 
-trigger: master
+trigger:
+  - v6
 
 pool:
   vmImage: 'ubuntu-latest'
--- a/.scripts/publish-dev-release.yml
+++ b/.scripts/publish-dev-release.yml
@@ -1,0 +1,25 @@
+# Publishes a development release of @autorest/typescript for
+# any commit to the master branch.
+
+trigger: master
+
+pool:
+  vmImage: 'ubuntu-latest'
+
+steps:
+- task: NodeTool@0
+  inputs:
+    versionSpec: '10.x'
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,7 @@
 {
-    "name": "@microsoft.azure/autorest.typescript",
-    "version": "6.0.0-preview.1",
+    "name": "@autorest/typescript",
+    "version": "6.0.0",
+    "private": true,
```
daviwil: Setting the package to `private` until we're ready to publish on npm.  Happy to remove it if folks feel it is unnecessary.

.scripts/publish-dev-release.yml
```diff
@@ -0,0 +1,25 @@
+# Publishes a development release of @autorest/typescript for
+# any commit to the master branch.
+
+trigger: master
```
joheredi: Does this need to be the `v6` branch?

daviwil: Crap, yes, thank you!

PR #549, opened 2020-01-17
Generate models types for Operation Requests and Responses

joheredi:
Picking up @daviwil PR (Azure/autorest.typescript.v3/pull/15) which generates the necessary model types for operation requests and responses and hooks them up to the `operationsGenerator` so that they get used in the method signatures that get generated.

Also including support for representing Polymorphic parameters and inheritance and some refactoring for better code reuse

P.S. sorry for the big PR :flushed:

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,9 +15,9 @@
         "start-test-server:v1": "ts-node test/integration/testserver-v1/index.ts",
         "stop-test-server": "stop-autorest-testserver",
         "debug": "node --inspect-brk ./dist/src/main.js",
-        "generate-bodystring": "autorest-beta --typescript --output-folder=./test/integration/generated/bodyString --use=. --title=BodyStringClient --input-file=node_modules/@microsoft.azure/autorest.testserver/swagger/body-string.json --package-name=bodyString --package-version=1.0.0-preview1",
-        "generate-bodycomplex": "autorest-beta --typescript --output-folder=./test/integration/generated/bodyComplex --use=. --title=BodyComplexClient --input-file=node_modules/@microsoft.azure/autorest.testserver/swagger/body-complex.json --package-name=bodyString --package-version=1.0.0-preview1",
-        "generate-url": "autorest-beta --typescript --output-folder=./test/integration/generated/url --use=. --title=UrlClient --input-file=node_modules/@microsoft.azure/autorest.testserver/swagger/url.json --package-name=url --package-version=1.0.0-preview1"
+        "generate-bodystring": "autorest-beta --pipeline-model:v3 --typescript --output-folder=./test/integration/generated/bodyString --use=. --title=BodyStringClient --input-file=node_modules/@microsoft.azure/autorest.testserver/swagger/body-string.json --package-name=bodyString --package-version=1.0.0-preview1",
+        "generate-bodycomplex": "autorest-beta --pipeline-model:v3 --typescript --output-folder=./test/integration/generated/bodyComplex --use=. --title=BodyComplexClient --input-file=node_modules/@microsoft.azure/autorest.testserver/swagger/body-complex.json --package-name=bodyString --package-version=1.0.0-preview1",
+        "generate-url": "autorest-beta --pipeline-model:v3 --typescript --output-folder=./test/integration/generated/url --use=. --title=UrlClient --input-file=node_modules/@microsoft.azure/autorest.testserver/swagger/url.json --package-name=url --package-version=1.0.0-preview1"
     },
     "files": [
         "dist/**",
@@ -41,7 +41,7 @@
     },
     "devDependencies": {
         "@autorest/test-server": "^3.0.26",
-        "@microsoft.azure/autorest.testserver": "^2.10.7",
+        "@microsoft.azure/autorest.testserver": "^2.10.9",
         "@types/express": "^4.17.2",
         "@types/js-yaml": "3.12.1",
         "@types/mocha": "5.2.7",
--- a/package.json
+++ b/package.json
@@ -15,9 +15,9 @@
         "start-test-server:v1": "ts-node test/integration/testserver-v1/index.ts",
         "stop-test-server": "stop-autorest-testserver",
         "debug": "node --inspect-brk ./dist/src/main.js",
-        "generate-bodystring": "autorest-beta --version:3.0.6184 --use:@autorest/modelerfour@4.2.108 --typescript --output-folder=./test/integration/generated/bodyString --use=. --title=BodyStringClient --input-file=node_modules/@microsoft.azure/autorest.testserver/swagger/body-string.json --package-name=bodyString --package-version=1.0.0-preview1",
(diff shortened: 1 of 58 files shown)


package.json
```diff
@@ -14,9 +14,9 @@
...
-        "generate-bodycomplex": "autorest-beta --version:3.0.6179 --use:@autorest/modelerfour@4.2.99 --typescript --output-folder=./test/integration/generated/bodyComplex --use=. --title=BodyComplexClient --input-file=node_modules/@microsoft.azure/autorest.testserver/swagger/body-complex.json --package-name=bodyString --package-version=1.0.0-preview1",
-        "generate-url": "autorest-beta --version:3.0.6179 --use:@autorest/modelerfour@4.2.99 --typescript --output-folder=./test/integration/generated/url --use=. --title=UrlClient --input-file=node_modules/@microsoft.azure/autorest.testserver/swagger/url.json --package-name=url --package-version=1.0.0-preview1"
+        "generate-bodystring": "autorest-beta --version:3.0.6184 --use:@autorest/modelerfour@4.2.108 --typescript --output-folder=./test/integration/generated/bodyString --use=. --title=BodyStringClient --input-file=node_modules/@microsoft.azure/autorest.testserver/swagger/body-string.json --package-name=bodyString --package-version=1.0.0-preview1",
```
joheredi: We need a better way to do this (in a follow-up PR). We may need to use gulp as these are becoming more complex, any thoughts?

test/integration/generated/bodyString/src/operations/string.ts
```diff
@@ -365,19 +387,21 @@ export class String {
     stringBody: Uint8Array,
     options?: coreHttp.RequestOptionsBase,
     callback?: coreHttp.ServiceCallback<any>
```
joheredi: Do we still need to support the callback pattern? The other day @ramya-rao-a mentioned that we may no longer need to. Any thoughts?
/cc @chradek @daviwil

daviwil: Since it's easy to interact with promises via callbacks without async/await, I'd say the old-style callback pattern shouldn't be needed any longer.  However, it'd be good to get @bterlson's opinion on that from an API guidelines perspective.

chradek: Agreed that I think we should be able to drop callback support. The only reason I see to continue supporting it is to help users migrate their existing code, but it's easy to 'callbackify' a promise if needed.

bterlson: No callbacks if we have promises. Get rid of it if possible :)

joheredi: Cool! I will get rid of them in this PR :) so this
image


Will turn into this
image

joheredi: Remove the callback parameter

package.json
joheredi: Moved versions to the Readme.md

src/generators/clientFileGenerator.ts
```diff
@@ -65,23 +71,25 @@ export function generateClient(clientDetails: ClientDetails, project: Project) {
...
+      ...requiredParams.map(p => ({
+        name: p.name,
+        hasQuestionToken: !p.required,
+        type: p.typeDetails.typeName
+      })),
       {
         name: "options",
         hasQuestionToken: true,
-        type: "any" // TODO Use the right type `Models.${clientDetails.className}Options`
+        type: `coreHttp.ServiceClientOptions`
```
daviwil: Should this be referring to the client-specific generated options type or is that not being generated yet?

src/generators/modelsGenerator.ts
```diff
@@ -16,31 +37,332 @@ export function generateModels(clientDetails: ClientDetails, project: Project) {
...
 
-  for (const union of clientDetails.unions) {
-    modelsIndexFile.addTypeAlias({
-      name: union.name,
-      docs: [union.description],
-      isExported: true,
-      type: union.values.join(" | "),
-      trailingTrivia: writer => writer.newLine()
+/**
+ * This function writes all UnionTypes, these types represent the optios a request can use for a Polymorphic parameter
```
daviwil:
```suggestion
 * This function writes all UnionTypes, these types represent the options a request can use for a Polymorphic parameter
```

src/generators/modelsGenerator.ts
```diff
@@ -16,31 +37,332 @@ export function generateModels(clientDetails: ClientDetails, project: Project) {
...
+
+  modelsIndexFile.addInterface({
+    name: `${operationGroupName}${operationName}OptionalParams`,
+    docs: ["Optional parameters."],
+    isExported: true,
+    extends: ["coreHttp.RequestOptionsBase"],
+    properties: optionalParams.map<PropertySignatureStructure>(p => ({
+      name: p.name,
+      hasQuestionToken: !p.required,
+      type: p.typeDetails.typeName || "TODO NO TYPE",
```
daviwil: Is this always going to be an error case or just something we need to track adding later?

src/generators/modelsGenerator.ts
```diff
@@ -16,31 +37,332 @@ export function generateModels(clientDetails: ClientDetails, project: Project) {
...
+  properties: PropertySignatureStructure[]
+): PropertySignatureStructure[] {
+  const discriminator = (model as PolymorphicObjectDetails).discriminator;
+  if (!discriminator) {
+    return properties;
+  }
+
+  const discProps = keys(discriminator).map<PropertySignatureStructure>(
+    key => ({
+      docs: [`Polymorphic discriminator`],
```
daviwil: I suppose a discriminator doesn't specify its own documentation string?  I don't know what the model does in this case, but seems like it'd have to be a normal property on the object which would have some meaning to the user

test/integration/generated/bodyComplex/src/operations/array.ts
```diff
@@ -62,63 +47,24 @@ export class Array {
...
-    );
+      putValidOperationSpec
+    ) as Promise<coreHttp.RestResponse>;
   }
 
   /**
    * Get complex types with array property which is empty
    * @param options The options parameters.
    */
-  getEmpty(options?: coreHttp.RequestOptionsBase): Promise<any>;
```
daviwil: Nice to get rid of these!

src/generators/clientFileGenerator.ts
joheredi: You're right we need to generate this type. Will send a followup PR for this

src/generators/modelsGenerator.ts
joheredi: Forgot to remove this. it should be handled when populating typeDetails

src/generators/modelsGenerator.ts
joheredi: In the current samples the description field is empty. I'll dig a bit more and probably we can honor the discriminator description if present or have this as fallback

PR #552, opened 2020-01-24
Support Parametrized Url

joheredi:
These changes include:
* handling the $host global parameter
* Adding Uri parameters to the pathParameters in the operation spec
* Acceptance tests for custom-baseUrl scenarios


src/transforms/parameterTransforms.ts
```diff
@@ -151,11 +165,14 @@ const isClientImplementation = (parameter: Parameter) =>
   parameter.implementation === ImplementationLocation.Client;
 
 function getParameterLocation(parameter: Parameter): ParameterLocation {
-  if (!parameter.protocol.http) {
+  const originalLocaltion = (parameter.protocol.http || { in: undefined }).in;
```
chradek: Nit: The "in: undefined" isn't necessary since "in" will already be undefined by being absent.
```suggestion
  const originalLocaltion = (parameter.protocol.http || {}).in;
```

Side-note: I want optional chaining from TypeScript 3.7...

test/integration/generated/bodyComplex/src/bodyComplexClientContext.ts
```diff
@@ -33,9 +34,20 @@ export class BodyComplexClientContext extends coreHttp.ServiceClient {
...
-    this.baseUri = options.baseUri || this.baseUri || "http://localhost:3000";
     this.requestContentType = "application/json; charset=utf-8";
+
+    // Assigning values to Constant parameters
     this.$host = "http://localhost:3000";
     this.apiVersion = "2016-02-29";
+
+    this.baseUri = options.baseUri || this.baseUri || "{$host}";
+
+    // Overriding values provided in the options parameters, if provided
```
chradek: Nit: This comment makes it sound like you're overriding the values provided by the user via the `options` parameters, when you're actually using the ones provided by the user instead.

Maybe the comment could be changed to something like?:
> Replacing parameter defaults with user-provided parameters.

test/integration/generated/bodyComplex/src/models/parameters.ts
```diff
@@ -9,13 +9,14 @@
 import * as coreHttp from "@azure/core-http";
 import * as Mappers from "../models/mappers";
 
-export const $host: coreHttp.OperationParameter = {
+export const $host: coreHttp.OperationQueryParameter = {
```
chradek: It does seem odd to see something like "$host" as a query parameter :)

src/transforms/operationTransforms.ts
```diff
@@ -54,7 +54,7 @@ export function transformOperationSpec(
 export function extractHttpDetails({ path, method }: OperationRequestDetails) {
   return {
     // TODO: Revisit how we should handle {$host}
```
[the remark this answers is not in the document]
joheredi: Nope, thanks for catching it :)

src/generators/parametersGenerator.ts
```diff
@@ -95,28 +86,35 @@ function getMapperDefaultValue(mapper: Mapper) {
...
-  )}, ${mapperSourceCode}${collectionFormatSourceCode}}`;
+  )}, ${mapperSourceCode}${collectionFormatSourceCode}${skipEncodingCode}}`;
 }
 
 function getParameterType(location: ParameterLocation) {
   switch (location) {
     case ParameterLocation.Path:
       return "coreHttp.OperationURLParameter";
     case ParameterLocation.Query:
+    case ParameterLocation.Uri:
```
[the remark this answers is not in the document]
joheredi: Yeah you are right, this should be in an `OperationUrlParameter` not an `OperationQueryParameter`

src/generators/clientContextFileGenerator.ts
```diff
@@ -1,26 +1,35 @@
...
 export function generateClientContext(
   clientDetails: ClientDetails,
   packageDetails: PackageDetails,
   project: Project
 ) {
+  const clientParams = clientDetails.parameters.filter(
+    param => param.implementationLocation === ImplementationLocation.Client
+  );
+  const requiredParams = getRequiredParameters(clientParams);
+  const hasCredentials = !!clientDetails.options.addCredentials;
```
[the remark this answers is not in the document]
daviwil: `--add-credentials` is an optional parameter because it adds a `TokenCredential` parameter to the generated `ServiceClient` and some generated clients don't want or need it.  However, I think the ones that don't want it are fewer than the ones that do, so maybe we should consider inverting the command line argument at some point

src/transforms/parameterTransforms.ts
joheredi: Thanks for the suggestion, I'm upgrading to 3.7 and using optional chaining :smile:

test/integration/generated/bodyComplex/src/models/parameters.ts
joheredi: It is now a Url parameter :smile: thanks for catching this

src/generators/clientContextFileGenerator.ts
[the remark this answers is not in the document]
joheredi: We get these when users pass the `--add-credentials` flag to autorest i.e. `autorest ... --add-credentials`. 

APIs that don't have authentication (such as test server scenarios so far) don't need this.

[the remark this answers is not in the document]
chradek: Does that mean those generated clients don't need any credentials at all? Or do they take a different kind that we don't know about and can't generate?

src/transforms/urlTransforms.ts
```diff
@@ -0,0 +1,34 @@
...
+  let baseUrl: string | undefined = "";
+  let isCustom = false;
+
+  const $host = (codeModel.globalParameters || []).find(p => {
+    const { serializedName } = getLanguageMetadata(p.language);
+    serializedName === "$host";
+  });
+
+  if (!$host) {
+    const { request } = codeModel.operationGroups[0].operations[0];
```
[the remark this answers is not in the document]
joheredi: We're assuming that all request have the same uri (python is doing this as well). However we may want to explore supporting multi-baseUrl in the future if we need to. I have filed #553 to track this work. Meanwhile I have added a comment about the assumption

src/generators/clientContextFileGenerator.ts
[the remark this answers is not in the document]
joheredi: I think so, although I think you can still pass Authentication headers through the options parameter to authenticate 
```ts
const client = new BodyComplexClient({customHeaders: {Authorization: "<AUTH_TOKEN>"}});
```

I may be wrong so please correct me if I'm wrong @daviwil

package.json
```diff
@@ -59,7 +60,7 @@
         "npm-run-all": "^4.1.5",
         "start-server-and-test": "^1.10.6",
         "ts-node": "^8.5.2",
-        "typescript": "~3.6.2",
+        "typescript": "^3.7.5",
```
chradek: This makes me so happy :)

src/generators/clientFileGenerator.ts
```diff
@@ -26,6 +26,11 @@ export function generateClient(clientDetails: ClientDetails, project: Project) {
     }
   );
 
+  clientFile.addImportDeclaration({
```
chradek: What is this import needed for? Looking at the generated files it doesn't look like `coreHttp` is being used where this is included.

joheredi: Hmm not needed I think I was experimenting with something and didn't remove. I'll take this out

PR #554, opened 2020-01-28
Client Operations

joheredi:
This PR contains the following work:

- Add operations as client properties when there is only one OperationGroup. Otherwise, keep the current behavior of adding a class for each operation group.

- Support `source-code-folder-path` flag to specify a custom source folder path

### DIFF
--- a/src/generators/clientFileGenerator.ts
+++ b/src/generators/clientFileGenerator.ts
@@ -27,7 +27,7 @@ export function generateClient(clientDetails: ClientDetails, project: Project) {
   const mappersName = getMappersName(clientDetails.className);
   const clientContextClassName = `${clientDetails.className}Context`;
 
-  // A client has inline operations when there is only one operation group
+  // A client has inline operations when it has a toplevel operation group
   const hasInlineOperations = clientDetails.operationGroups.some(
     og => og.isTopLevel
   );
--- a/src/generators/clientFileGenerator.ts
+++ b/src/generators/clientFileGenerator.ts
@@ -18,6 +18,9 @@ import {
 import { ParameterDetails } from "../models/parameterDetails";
 import { ImplementationLocation, SchemaType } from "@azure-tools/codemodel";
 import { getCredentialsParameter } from "./utils/parameterUtils";
+import { OperationGroupDetails } from "../models/operationDetails";
+
+type OperationDeclarationDetails = { name: string; typeName: string };
 
 export function generateClient(clientDetails: ClientDetails, project: Project) {
   const modelsName = getModelsName(clientDetails.className);
@@ -25,7 +28,10 @@ export function generateClient(clientDetails: ClientDetails, project: Project) {
   const clientContextClassName = `${clientDetails.className}Context`;
 
   // A client has inline operations when there is only one operation group
-  const hasInlineOperations = clientDetails.operationGroups.length === 1;
+  const hasInlineOperations = clientDetails.operationGroups.some(
+    og => og.isTopLevel
+  );
+
   const hasCredentials = !!clientDetails.options.addCredentials;
 
   const clientFile = project.createSourceFile(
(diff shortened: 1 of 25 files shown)


chradek:
Is it considered a breaking change to add a new operation group to an existing spec? I see the benefit to inlining the operations if there is only a single operation group but want to understand how this affects forwards-compatibility.

What would happen if another operation group is added to the spec, would all the operations be inlined and also have groups?

joheredi:
> Is it considered a breaking change to add a new operation group to an existing spec? I see the benefit to inlining the operations if there is only a single operation group but want to understand how this affects forwards-compatibility.
> 
> What would happen if another operation group is added to the spec, would all the operations be inlined and also have groups?

@chradek, adding a new OperationgGroup would be a breaking change for the generated SDK but probably not for the API. Changing the name I think would be a breaking change in both places.

If a new operation group is added to the spec, the next time the SDK is generated, the client will not have any inline operations. Each operation group will now have its class with its operations

We could probably have a flag to opt-in/out of operation in-lining for future proofing. So users generating an SDK who expect the API to keep adding OGs can opt out so that adding new ones won't be a breaking change. What do you think?

chradek:
Yeah, that was my concern, that updating the spec in a legal way would cause a breaking change in the SDK.

Is there some sort of default/null operation group for top-level operations? If you take the `custom-baseUrl.json` swagger file as an example, the operationId has `paths_` defined, which I _think_ is how the operation group is created:
https://github.com/Azure/autorest.testserver/blob/db336be702f88f1f285fce9e3ad6898eb0f9d8fa/swagger/custom-baseUrl.json#L31

I'd imagine then that if the operationId was `getEmpty` instead of `paths_getEmpty`, it should be a top-level operation, otherwise it should be on the `paths` group.

joheredi:
> Yeah, that was my concern, that updating the spec in a legal way would cause a breaking change in the SDK.
> 
> Is there some sort of default/null operation group for top-level operations? If you take the `custom-baseUrl.json` swagger file as an example, the operationId has `paths_` defined, which I _think_ is how the operation group is created:
> https://github.com/Azure/autorest.testserver/blob/db336be702f88f1f285fce9e3ad6898eb0f9d8fa/swagger/custom-baseUrl.json#L31
> 
> I'd imagine then that if the operationId was `getEmpty` instead of `paths_getEmpty`, it should be a top-level operation, otherwise it should be on the `paths` group.
> 
> Edit: Where I'm going with this is that I'd expect that whatever was set in the `operationId` would eventually be propagated to the operation group name (maybe that's an empty string/null for top-level parameters). If an operation group is included in an `operationId`, I'd want to honor that even if there was only one operation group since the service might have forwards-compatibility in mind. Ideally this would be something we could write a linter to warn on so a swagger author can make the call.

@chradek, the previous version of autorest was doing this when there was only one. However I agree with you that we should honor the operationIds and only inline toplevel operations. Please let me know what you think about the latest change I did

Thanks for bringing this up, this will provide an improved user experience by reducing unexpected braking changes

chradek:
Ah, I didn't realize that the previous version was doing that already, but I'm glad we're open to changes :)

src/generators/clientFileGenerator.ts
```diff
@@ -18,14 +18,20 @@ import {
...
+import { OperationGroupDetails } from "../models/operationDetails";
+
+type OperationDeclarationDetails = { name: string; typeName: string };
 
 export function generateClient(clientDetails: ClientDetails, project: Project) {
   const modelsName = getModelsName(clientDetails.className);
   const mappersName = getMappersName(clientDetails.className);
   const clientContextClassName = `${clientDetails.className}Context`;
 
   // A client has inline operations when there is only one operation group
```
chradek: This comment might need some updating now :)

src/transforms/parameterTransforms.ts
```diff
@@ -41,7 +42,8 @@ export function transformParameters(codeModel: CodeModel): ParameterDetails[] {
 const extractOperationParameters = (codeModel: CodeModel) =>
   codeModel.operationGroups.reduce<OperationParameterDetails[]>((acc, og) => {
     // TODO: Probably want to inline operations in client when there is only one operation group (#551)
```
chradek: Can this TODO be removed now?

src/generators/operationGenerator.ts
```diff
@@ -38,14 +38,20 @@ export function generateOperations(
   clientDetails: ClientDetails,
   project: Project
 ): void {
+  if (clientDetails.operationGroups.length === 1) {
```
chradek: Does this premise still ring true if there are no top-level operation groups?

PR #557, opened 2020-01-31
Multiple clean up fixes for JSDoc and parameter handling

joheredi:
This PR includes:
- Wrap long comments to maximum width
- Simplify optional parameter handling in ClientContext
- Make credentials parameter a "Synthetic" parameter next to SWAGGER parameters to avoid having special hanlding for it
- Add missing JSDoc to Client, and ClientContext

### DIFF
--- a/src/generators/clientFileGenerator.ts
+++ b/src/generators/clientFileGenerator.ts
@@ -17,8 +17,8 @@ import {
 } from "../utils/nameUtils";
 import { ParameterDetails } from "../models/parameterDetails";
 import { ImplementationLocation, SchemaType } from "@azure-tools/codemodel";
-import { getCredentialsParameter } from "./utils/parameterUtils";
 import { OperationGroupDetails } from "../models/operationDetails";
+import { wrapString } from "./utils/stringUtils";
 
 type OperationDeclarationDetails = { name: string; typeName: string };
 
@@ -78,7 +78,7 @@ export function generateClient(clientDetails: ClientDetails, project: Project) {
     extends: clientContextClassName
   });
 
-  writeConstructor(clientDetails, clientClass, hasCredentials);
+  writeConstructor(clientDetails, clientClass);
   writeClientOperations(clientFile, clientClass, clientDetails);
 
   clientFile.addExportDeclaration({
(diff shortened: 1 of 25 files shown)


src/generators/clientFileGenerator.ts
```diff
@@ -111,13 +110,17 @@ function writeConstructor(
       param.schemaType !== SchemaType.Constant
   );
 
+  const docs = [
+    `Initializes a new instance of the ${clientDetails.className} class.`,
+    ...requiredParams.map(p => wrapString(`@param ${p.name} ${p.description}`)),
```
daviwil: Since this is done in multiple places, it might even make sense to have a `formatJsDocParam` function that takes in a parameter and returns the wrapped doc string.

src/transforms/parameterTransforms.ts
```diff
@@ -24,18 +24,50 @@ import { isEqual, isNil } from "lodash";
...
+  description:
+    "Subscription credentials which uniquely identify client subscription.",
+  name: "credentials",
+  serializedName: "credentials",
+  location: ParameterLocation.None,
+  required: true,
+  parameterPath: "credentials",
+  mapper: "any",
+  isGlobal: true,
+  //parameter: new Parameter("credentials", "Subscription credentials which uniquely identify client subscription.", new ObjectSchema("credentials", ""))
```
daviwil: No longer needed?

src/typescriptGenerator.ts
```diff
@@ -28,7 +28,8 @@ const prettierTypeScriptOptions: prettier.Options = {
   printWidth: 80,
   semi: true,
   singleQuote: false,
-  tabWidth: 2
+  tabWidth: 2,
+  proseWrap: "always"
```
daviwil: Maybe unneeded if it didn't help with comments?

test/integration/generated/bodyComplex/src/operations/array.ts
```diff
@@ -69,7 +70,8 @@ export class Array {
 
   /**
    * Put complex types with array property which is empty
-   * @param complexBody Please put an array with 4 items: "1, 2, 3, 4", "", null, "&S#$(*Y", "The quick brown fox jumps over the lazy dog"
+   * @param complexBody Please put an array with 4 items: "1, 2, 3, 4", "", null, "&S#$(*Y", "The quick
+   * brown fox jumps over the lazy dog"
```
daviwil: I guess it'd be difficult to have the wrapper also indent subsequent lines to be equal to the starting position of the docstring in the first line

joheredi: We can pass it a param indentationType which we could use to decide what kind of indentation to have. For instance IndentationType.JSDocParam would be `"      "`

Issue #512, opened 2019-12-04
Support for Header


joheredi:
Full support for header swagger
https://github.com/Azure/autorest.testserver/blob/master/swagger/header.json

Needs:

- Support for header parameters

Estimate: [2 days]

PR #559, opened 2020-02-06
[v6] Xml support

### LINKED ISSUE #522: Support XML swaggers
https://github.com/Azure/autorest.testserver/blob/master/swagger/xml-service.json
https://github.com/Azure/autorest.testserver/blob/master/swagger/xms-error-responses.json

Estimate [3 days]

joheredi:
- [x] Adds isXML boolean parameter to serializer
- [x] Adds isXML property to operation specs that expect XML content
- [x] Adds xml metadata to mappers when request/responses may contain XML content
- [x] Integration test for testserver xmlservice 

Coming in a follow up PR:
- [ ] Support xml in headers response. Need to implement support for response headers #512 which will come in the next PR

Fixes #522


src/generators/mappersGenerator.ts
```diff
@@ -124,14 +127,34 @@ function writeMapper(writer: CodeBlockWriter, mapper: Mapper) {
               writeObjectProps(polymorphicDiscriminator, writer);
             })
             .write(",");
-        writer.write("modelProperties:").block(() => {
-          writeParentMappers(parents, writer);
-          writeObjectProps(modelProperties, writer);
-        });
+
+        const hasParents = !!parents && !!parents.length;
+        const hasModelPropertoes =
```
joheredi: Typo

src/generators/operationGenerator.ts
```diff
@@ -106,11 +107,16 @@ function buildSpec(spec: OperationSpecDetails): string {
...
+  const contentType =
+    spec.requestBody && isXML
+      ? "contentType: 'application/xml; charset=utf-8',"
+      : "";
   return `{ path: "${spec.path}", httpMethod: "${
     spec.httpMethod
   }", responses: {${responses.join(
     ", "
-  )}},${requestBody}${queryParams}${urlParams}${headerParams}serializer
+  )}},${requestBody}${queryParams}${urlParams}${headerParams}${isXML}${contentType}serializer
```
joheredi: Need to cleanup this, probably a better idea to use ts-morph writers instead of building the strings this way. Will explore this in a separate PR filed #560

daviwil: It's not the worst way to do it since Prettier will break up the lines correctly anyway.  There's probably better ways to spend your time for now :)

src/generators/mappersGenerator.ts
```diff
@@ -105,7 +106,9 @@ function writeMapper(writer: CodeBlockWriter, mapper: Mapper) {
   } = type as CompositeMapperType;
   const isReferenceDicriminator = isString(polymorphicDiscriminator);
   writer.block(() => {
-    writeObjectProps(restMapper, writer)
+    const { defaultValue, ...mapperProps } = restMapper;
```
daviwil: Some of these sections are starting to have enough code that it'd be helpful to start inserting comments here and there explaining what the different sections do.  Maybe even some simple example code in comments to show what the output looks like?

This isn't super important for now but it'll be useful in the long term when different people are trying to maintain this code.

src/transforms/mapperTransforms.ts
```diff
@@ -129,18 +134,20 @@ interface PartialMapperType {
 function buildMapper(
   schema: Schema,
   type: PartialMapperType,
-  options?: EntityOptions
+  options: EntityOptions = {}
```
daviwil: :+1:

src/utils/nameUtils.ts
```diff
@@ -55,9 +55,15 @@ function getCasingConvention(nameType: NameType) {
 
 /**
  * TODO: Improve this function to handle cases such as TEST -> test. Current basic implementation
- * results in TEST -> tEST
+ * results in TEST -> test or Test (depending on the CasingConvention). We should switch to relay
+ * on Modeler four namer for this once it is stable
```
daviwil: :+1:

src/generators/mappersGenerator.ts
```diff
@@ -105,7 +106,9 @@ function writeMapper(writer: CodeBlockWriter, mapper: Mapper) {
...
+    writeDefaultValue(writer, defaultValue, restType);
```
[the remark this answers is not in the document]
joheredi: Default value is not specific to XML-based services. We were just not handling this before this change. The reason I started handling this here is because some types need some special stuff... For example you can have a ByteArray with a default value = "". However we cannot just generate the mapper with defaultValue: "", we need to convert "" to  `new Uint8Array("")`.

package.json
```diff
@@ -8,8 +8,8 @@
         "test": "npm-run-all unit-test integration-test",
         "unit-test": "mocha -r ts-node/register './test/unit/**/*spec.ts'",
         "integration-test": "start-server-and-test start-test-server:v1 http://localhost:3000 generate-and-test",
-        "integration-test:new": "npm-run-all start-test-server -p generate-bodystring generate-bodycomplex -s integration-test:alone stop-test-server",
-        "generate-and-test": "npm-run-all -s build -p generate-bodystring generate-bodycomplex generate-url generate-customurl -s integration-test:alone",
+        "integration-test:new": "npm-run-all start-test-server generate-and-test integration-test:alone stop-test-server",
```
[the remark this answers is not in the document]
joheredi: Yes :)

src/transforms/operationTransforms.ts
```diff
@@ -163,23 +166,23 @@ export function extractSchemaResponses(responses: OperationResponseDetails[]) {
...
 
 export function transformOperationResponse(
-  response: SchemaResponse
+  response: SchemaResponse | Response
 ): OperationResponseDetails {
   let isError =
     !!response.extensions && !!response.extensions["x-ms-error-response"];
 
-  if (!response.schema) {
+  if (!(response as SchemaResponse).schema) {
```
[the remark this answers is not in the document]
joheredi: Awesome thanks for the tip, I'm integrating it

joheredi:
> I have some small comments that shouldn't be considered blockers. I don't quite fully understand all the changes (were some changes not XML specific?), but don't see anything obviously wrong!

Some of the changes were not completely scoped to XML, I had to fix a couple bugs such as the one with defaultValue (comment above), and I also did a few refactors in the generators, switching from building the code with string interpolation to writer functions.

src/generators/mappersGenerator.ts
```diff
@@ -95,43 +97,105 @@ export function generateMappers(
...
+  modelProperties: ModelPropertiesType = {}
+) {
+  if (isEmpty(parents) && isEmpty(modelProperties)) {
+    return writer;
+  }
+
+  return writer.write("modelProperties:").block(() => {
+    // We'll first inherit from the parents and then write
+    // its own mappers, it will look like this when generated
+    // modelProperties: {
```
chradek: Thanks for adding these comments. Helps a lot!

src/transforms/operationTransforms.ts
```diff
@@ -242,36 +255,64 @@ export function transformOperation(
...
   const metadata = getLanguageMetadata(operationGroup.language);
   const isTopLevel = !metadata.name;
   const name = normalizeName(
     metadata.name || TOPLEVEL_OPERATIONGROUP,
     NameType.Property
   );
+
+  const operations = await Promise.all(
+    operationGroup.operations.map(
+      async operation => await transformOperation(operation, name)
```
chradek: Minor but I don't think async/await is needed here since you're doing a Promise.all. Similar general feedback that the returned promise from tranformOperation would be unwrapped by await then rewrapped in a new promise.

This change should work since transformOperation returns a promise
```suggestion
      operation => transformOperation(operation, name)
```

PR #561, opened 2020-02-07
Generate header mappers and models

joheredi:
Generating Mapper and Models for response headers, this is the first part of the work to fully support response headers. There will be 2 follow up PRs to this one (1) to correctly generate ResponseTypes when responses contain Headers. (2) wire up the types with the Operations and OperationSpecs and Integration tests


src/utils/headersToSchema.ts
```diff
@@ -0,0 +1,20 @@
...
+  if (!headers || !headers.length) {
+    return undefined;
+  }
+  const headersSchema = new ObjectSchema(`${operationFullName}Headers`, "");
+
+  headers.forEach(({ header, schema }) => {
+    if (!headersSchema.properties) {
+      headersSchema.properties = [];
+    }
+    headersSchema.properties.push(new Property(header, "", schema));
```
daviwil: Can headers have a description?

joheredi: I think so, I'll extract it from the schema and add it, also above

PR #562, opened 2020-02-10
[v6] Complete Response Type Generation

joheredi:
This PR completes the work to generate correctly Response Types for operation, previously we were only using Body for generating the types, with this PR we also take Headers information to get the right Response type

- Update operationTransforms to also extract response Headers information
- Update modelsGenerator to generate responses as intersection types when there are Headers
- Update operationsGenerator to include headersMappers in the OperationSpec

### DIFF
--- a/src/generators/modelsGenerator.ts
+++ b/src/generators/modelsGenerator.ts
@@ -17,12 +17,14 @@ import {
   PolymorphicObjectDetails,
   ObjectDetails,
   PropertyDetails,
-  PropertyKind,
-  TypeDetails
+  PropertyKind
 } from "../models/modelDetails";
 import { normalizeName, NameType } from "../utils/nameUtils";
 import { filterOperationParameters } from "./utils/parameterUtils";
-import { OperationDetails } from "../models/operationDetails";
+import {
+  OperationDetails,
+  OperationResponseDetails
+} from "../models/operationDetails";
 import { ParameterDetails } from "../models/parameterDetails";
 
 export function generateModels(clientDetails: ClientDetails, project: Project) {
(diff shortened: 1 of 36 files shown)


src/generators/modelsGenerator.ts
```diff
@@ -71,113 +75,192 @@ const writeOperationModels = (
...
+
+type IntersectionTypeParameters = [
+  WriterFunctionOrValue,
+  WriterFunctionOrValue,
+  ...WriterFunctionOrValue[]
+];
 
-  const isComposite = typeDetails.kind === PropertyKind.Composite;
+/**
+ * This function vuilds the type to represent an Operation response, taking the response headers and body
```
daviwil:
```suggestion
 * This function builds the type to represent an Operation response, taking the response headers and body
```

src/generators/modelsGenerator.ts
```diff
@@ -71,113 +75,192 @@ const writeOperationModels = (
...
 
-  return isComposite
-    ? Writers.intersectionType(bodyTypeName, innerTypeWriter)
+  let intersectionTypes: WriterFunctionOrValue[] = [innerTypeWriter];
+  bodyProperties?.intersectionType &&
+    intersectionTypes.unshift(bodyProperties.intersectionType);
+  headersProperties?.intersectionType &&
+    intersectionTypes.unshift(headersProperties.intersectionType);
+
+  /**
```
daviwil: Thanks for the helpful comment!

src/generators/modelsGenerator.ts
```diff
@@ -9,21 +9,25 @@ import {
...
 } from "../models/modelDetails";
 import { normalizeName, NameType } from "../utils/nameUtils";
 import { filterOperationParameters } from "./utils/parameterUtils";
-import { OperationDetails } from "../models/operationDetails";
+import {
+  OperationDetails,
+  OperationResponseDetails
+} from "../models/operationDetails";
 import { ParameterDetails } from "../models/parameterDetails";
+import { type } from "os";
```
chradek: I think this is an unused import.

src/generators/modelsGenerator.ts
```diff
@@ -71,113 +75,192 @@ const writeOperationModels = (
...
 ) {
   const responseName = `${operationType.typeName}Response`;
 
   responses
-    .filter(r => !r.isError && !!r.bodyMapper)
-    .forEach(response => {
+    // Filter responses that are not marked as errors and that have either body or headers
+    .filter(
+      ({ isError, mappers }) =>
+        !isError && (!!mappers.bodyMapper || !!mappers.headersMapper)
```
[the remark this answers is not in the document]
joheredi: Yeah I agree, no need for the double !

src/generators/modelsGenerator.ts
joheredi: Fixed

Issue #556, opened 2020-01-31
[v6] Generate type for options


joheredi:
Previous version seems to be using `coreHttp.ServiceClientOptions`

PR #566, opened 2020-02-15
[v6] Multiple fixes

joheredi:
This PR contains a few fixes from discussions during our review:

* Switch options type from OperationRequestBase with OperationBase
   * Add a synthetic `endpoint` client parameter. This was required since we can no longer cheat using OperationRequestBase indexer
* Add missing options types in Client and ClientContext
* Remove clientContext properties that are not needed in serialization
* Remove useless docstring for interfaces, if SWAGGER doesn't include a description drop it
* Name client level operations after the Client rather than with the "topLevelOperation" constant

### DIFF
--- a/src/generators/clientContextFileGenerator.ts
+++ b/src/generators/clientContextFileGenerator.ts
@@ -14,7 +14,7 @@ import { ClientDetails } from "../models/clientDetails";
 import { PackageDetails } from "../models/packageDetails";
 import { ParameterDetails } from "../models/parameterDetails";
 import { ImplementationLocation, SchemaType } from "@azure-tools/codemodel";
-import { BaseUrlDetails } from "../transforms/urlTransforms";
+import { EndpointDetails } from "../transforms/urlTransforms";
 import { formatJsDocParam } from "./utils/parameterUtils";
 
 export function generateClientContext(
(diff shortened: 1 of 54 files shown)


joheredi:
Fixes #558, #556

src/generators/clientContextFileGenerator.ts
```diff
@@ -87,31 +90,34 @@ function writeClassProperties(
...
   contextClass.addProperties(
-    clientParams.map(param => {
-      return {
-        name: param.name,
-        type: param.typeDetails.typeName,
-        hasQuestionToken: !param.required
-      } as PropertyDeclarationStructure;
-    })
+    clientParams
+      .filter(p => !p.isSynthetic)
```
chradek: What is a synthetic parameter?

src/utils/nameUtils.ts
```diff
@@ -83,11 +82,11 @@ function getNameParts(name: string) {
 
 export function getOperationFullName(
   operationGroup: OperationGroup,
-  operation: Operation
+  operation: Operation,
+  clientName: string
 ) {
   const groupName = normalizeName(
-    getLanguageMetadata(operationGroup.language).name ||
-      TOPLEVEL_OPERATIONGROUP,
```
chradek: Is this constant still needed, or can it be removed from the constants file?

joheredi: Yeah we can remove it from the file, good catch!

src/generators/clientContextFileGenerator.ts
joheredi: I've been using Synthetic parameters when they are not defined in the Swagger file, for example, I'm injecting the Credentials parameter as a Synthetic one.

The reason for doing this is to just add the parameter and let it flow through the transform /generation process so we don't need to it handle specially in the layers above. 

Before having Credentials as synthetic, we used to have a specific check on the Generator layer to write the code for the credentials parameter.

chradek: Thanks for that explanation!

PR #573, opened 2020-02-20
Paging

joheredi:
## Implementation Ideas:

- ModelerFour includes paging metadata to operations in operation.language.default.paging
- There are 4 main scenarios we would like to handle for paging
  1 Service returns a relative URL to get the next page
  2 Service returns an absolute URL to get the next page
  3 Service returns a continuation token
  4 Service returns any of the previous, plus an operationId

  - This operationId is the operation to call for getting the next page

- From the operation definition, we can infer if the paging operation is of type (3) vs (1) or (2)
- Operations of type (3) will have a parameter with the same name as the nextLink name
- Operations of type (1) and (2) won't have a parameter with the same name as nextLinkOperation

1.  Handling Operations of type (1) and (2)

    - We can handle these types of operations by modifying a bit the way its OperationSpec is generated
      - This operation spec will be very similar to the original, with 2 differences:
        1 Have the Path parametrized, instead of the actual PATH in the operation specification. This parametrized path will take the "original" value as defaultValue
        2 Add a Path parameter to the urlParameters array in the Operation spec
      - This way we just need to call the same method with the optional pathParameter whenever we get a nextLink
      - \***\*Note\*\*** that core-http seems to be able to extract host and path with no issues. This probably means that we need no special handling between (1) and (2), if the parametrized path is an absolute or relative URL core-http seems to figure it out.

```typescript
// This would be de "Default" generated spec. See below for an example of the modifications described above
const getMultiplePagesOperationSpec: coreHttp.OperationSpec = {
  path: "paging/multiple",
  httpMethod: "GET",
  responses: {
    200: {
      bodyMapper: Mappers.ProductResult
    },
    default: {}
  },
  urlParameters: [Parameters.$host],
  headerParameters: [
    Parameters.clientRequestId,
    Parameters.maxresults,
    Parameters.timeout
  ],
  serializer
};

// This is the modified spec
const getMultiplePagesOperationSpec: coreHttp.OperationSpec = {
  path: "{path}", // Made this parametrized
  httpMethod: "GET",
  responses: {
    200: {
      bodyMapper: Mappers.ProductResult
    },
    default: {}
  },
  urlParameters: [Parameters.$host, Parameters.path /*Added this guy*/],
  headerParameters: [
    Parameters.clientRequestId,
    Parameters.maxresults,
    Parameters.timeout
  ],
  serializer
};

// In parameters.ts
export const path: coreHttp.OperationURLParameter = {
  parameterPath: ["options", "path"],
  mapper: {
    serializedName: "path",
    required: true,
    defaultValue: "http://localhost:3000/paging/multiple",
    type: {
      name: "String"
    }
  },
  skipEncoding: true
};
```

2.  Handling Operations of type (3)

    - These are the operations that return a ContinuationToken instead of a link.
    - This operations typically have a parameter with the same name of the nextLink
    - To get the next page, we just need to call the same function passing nextLink

3.  Handling Operations of Type (4)

    - These operations have a reference to another operation to get the next page
    - We should be able to generate the operation function so that instead of calling itself to get the next page, it calls the specified operation

### Manual Prototyping
- [x] Prototype Async Iterator functions
   - Mostly prototyped this by hand in the generated paging.ts file to use as reference for generation
- [x] Validate initial prototyped function with 2 basic integration tests 
### Actual Generation
- [x] Inject Path parameter and generate parameter
- [x] Generate Operation spec with parametrized path
- [ ] Generate Iterator functions


package.json
```diff
@@ -9,18 +9,19 @@
         "unit-test": "mocha -r ts-node/register './test/unit/**/*spec.ts'",
         "integration-test": "start-server-and-test start-test-server:v1 http://localhost:3000 generate-and-test",
         "integration-test:new": "npm-run-all start-test-server generate-and-test integration-test:alone stop-test-server",
-        "generate-and-test": "npm-run-all -s build -p generate-bodystring generate-bodycomplex generate-url generate-customurl generate-xmlservice generate-header -s integration-test:alone",
+        "generate-and-test": "npm-run-all -s build -p generate-bodystring generate-bodycomplex generate-url generate-customurl generate-xmlservice generate-header generate-paging -s integration-test:alone",
         "integration-test:alone": "mocha -r ts-node/register './test/integration/**/*spec.ts'",
         "start-test-server": "ts-node test/utils/start-server.ts",
-        "start-test-server:v1": "ts-node test/integration/testserver-v1/index.ts",
+        "start-test-server:v1": "start-autorest-express node",
```
joheredi: I had to change this since the way we were doing it through our own express, was not setting up the server correctly. Specifically we were getting invalid "nextLink" in the responses because the port was not set. for example `http://localhost:0/paging/multiple` instead `http://localhost:3000/paging/multiple`

daviwil:
Superceded by https://github.com/Azure/autorest.typescript/pull/577.

PR #582, opened 2020-03-17
[v6] Adds support for operations with multiple content types

chradek:
This PR adds support for operations that consume multiple content-types.

To test, I built a client from the [media_types.json](https://github.com/Azure/autorest.testserver/blob/master/swagger/media_types.json) swagger file. I still need to add support for multiple requests (method signatures) before this can be considered done.

As part of this work I needed to add support for `BinarySchema` as well. I took a look at what was generated for the storage-blob upload method and saw that the binary parameter was the `coreHttp.HttpRequestBody` type, so that's what I decided to go with here.

I also added the test server test coverage report. We now generate a client from [report.json](https://github.com/Azure/autorest.testserver/blob/master/swagger/report.json) and call its operations after all the other integration tests have been run. This is accomplished by making it the last integration test that's ran (by giving the file a `zzz` prefix). In the future, we can push the reports generated by this test to the centralized autorest testing dashboard as part of our CI.

TODO: 
- [x] Support multiple requests per operation


chradek:
I've updated to include multi-request support. Now both tests pass for mediaTypes.

Rather than going with method overloads, I made the operation optional parameter a type alias for a union of possible optional parameters. I went with this approach so that I wouldn't have to expose to users a different interface for each overload, they can just reference the type alias.

The operation body for methods that support multiple media types uses the `contentType` that's passed in by the user to determine how to serialize the http body. This code feels the most fragile since I have to find the contentType parameter from an operation request and get the list of allowed values from it. This seems safe _for now_ since currently we'll only see multiple requests per operation if multiple media types are supported.

test/integration/generated/mediaTypes/src/models/index.ts
```diff
@@ -0,0 +1,101 @@
...
+export type ContentType =
+  | "application/pdf"
+  | "image/jpeg"
+  | "image/png"
+  | "image/tiff";
+
+/**
+ * Optional parameters.
+ */
+export interface MediaTypesClientAnalyzeBody$binaryOptionalParams
```
joheredi: Just curious, why do we add '$' before binary? Is that part of a convention?

chradek: I wanted to add some special character to protect against the off-chance scenario where there's also a `MediaTypesClientAnalyzeBodybinary` operation that would cause a name conflict. I chose `$` since I figured no methods would contain that character, but could have gone with underscore instead.

joheredi: Makes sense thanks for clarifying

PR #592, opened 2020-03-20
[v6] respects x-ms-client-name in mappers

chradek:
Also updates modelerfour and fixes a bug where ClientContext classes were storing `credentials` in a public member that didn't exist. For now I just created a blacklist to hold parameters that shouldn't be exposed, but I think longer-term we could add a field to `ParameterDetails` to determine which parameters should be exposed on the context.

The credentials fix is needed or else TypeScript won't compile the generated code.

### DIFF
--- a/src/generators/clientContextFileGenerator.ts
+++ b/src/generators/clientContextFileGenerator.ts
@@ -246,8 +246,8 @@ function getRequiredParamChecks(requiredParameters: ParameterDetails[]) {
 }
 
 function getRequiredParamAssignments(requiredParameters: ParameterDetails[]) {
-  const blacklistClientParameters = ["credentials"];
+  const disallowedClientParameters = ["credentials"];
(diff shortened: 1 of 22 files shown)


src/generators/clientContextFileGenerator.ts
```diff
@@ -246,5 +246,8 @@ function getRequiredParamChecks(requiredParameters: ParameterDetails[]) {
 }
 
 function getRequiredParamAssignments(requiredParameters: ParameterDetails[]) {
-  return requiredParameters.map(({ name }) => `this.${name} = ${name};`);
+  const blacklistClientParameters = ["credentials"];
```
daviwil: Hate to be the word police, but could we use a different name like "disallowedClientParameters"?

chradek: Updated!

PR #593, opened 2020-03-21
[v6] Improve Dictionary support

joheredi:
Changes needed to fully support Dictionary:

* Make response type an intersection with the Dictionary type, instead of having it as body property.
   * This is to accommodate core-http giving us back a response in this shape
* Use named imports in Operations instead of importing the whole Models module
* Support AnySchema
* Add test coverage for Dictionary scenarios ( This will put us at 66% 😄)
image


src/generators/modelsGenerator.ts
```diff
@@ -146,13 +146,15 @@ function getBodyProperties({
     return;
   }
 
-  const isComposite = bodyType.kind === PropertyKind.Composite;
+  const hasBodyProperty =
+    bodyType.kind !== PropertyKind.Composite &&
+    bodyType.kind !== PropertyKind.Dictionary;
```
chradek: Ooo, so does this mean I just have to add the array kind as well for bodyArray?

src/utils/nameUtils.ts
```diff
@@ -21,6 +22,20 @@ export function guardReservedNames(name: string): string {
   return ReservedModelNames.indexOf(name) > -1 ? `${name}Model` : name;
 }
 
+/**
+ * Returns a normalized Type name, this is, the type name capitalized when needed.
+ * Otherwise, return the original typename, for example primitives "string", etc. don' need capitalization
```
chradek: Nit to show I am paying attention:
```suggestion
 * Otherwise, return the original typename, for example primitives "string", etc. don't need capitalization.
```

test/integration/bodyDictionary.spec.ts
```diff
@@ -0,0 +1,496 @@
...
+    assert.lengthOf(Object.keys(result), 0);
+  });
+
+  it("should put an empty dictionary", async () => {
+    const result = await client.dictionary.putEmpty({});
+    assert.equal(result._response.status, 200);
+  });
+
+  it("should get null dictionary", async () => {
+    const result = await client.dictionary.getNull();
```
chradek: If you were to `console.log(result)`, would it appear as though it were an empty object? That's what I was seeing with arrays (well, empty array instead of object).

test/integration/bodyDictionary.spec.ts
```diff
@@ -0,0 +1,496 @@
+import { assert } from "chai";
```
chradek: Now I feel bad because I used `expect` on my last tests (I think I copied that from another test). Do you prefer `assert`? Happy to use that going forward.

test/integration/bodyDictionary.spec.ts
joheredi: Yup it would be {}

test/integration/bodyDictionary.spec.ts
joheredi: I had been using expect, but started using assert in this one since it looks like it would be the approach will be taking in the SDKs repo. I don't have a strong opinion about it, but if we'll be using assert in the SDKs it would probably be easier for everyone to use it as well here. I'll have to go back and convert other tests to assert

chradek:
I assumed `Models` was broken apart for tree-shaking as well, but now realize that models only contains type info, which doesn't appear in the JS that gets tree-shaken anyway.

joheredi:
> Overall this looks good, but I have a question about the named imports work. What does it accomplish for us other than just not requiring `Models.` everywhere in the generated code? Is the aim to enable better tree-shaking of SDK code? I'm just wondering if the effort to track the imported models is worth the trouble, but I can certainly be convinced if it will provide downstream benefits to users.

The main motivation behind moving to named import is to make our generation of dictionary types easier. We have been adding the "Models." or "Parameters." prefix in the "Generator" layer, keeping the transform layer agnostic of that. This worked fine since the types we had been generating before were either primitive ("number", "string", etc) or composite (model name), which made it very easy to add the prefix on where to find the model if needed for composite.

However when we start generating dictionaries, once the transform layer says that the type is 

```typescript 

{[key: string]: ModelName}

```
it is not as easy to add `Models.` prefix to ModelName anymore. At that point we had a few options, the 2 that I evaluated were:

a) Use named imports so we don't need to add the Models prefix anymore
b) Add extra metadata to the typename so that the Generators layer is able to build the dictionary type with the right model to reference through adding `Models.` prefix to the model name 

I went with a) since I thought we'd also benefit from better treeshaking, and also I remember APIExtractor had issues with `import *`

I'm open to pursue b) or another option of you guys think it'd be better

chradek:
You've got a reasonable justification to me, I don't have an objection.

daviwil:
Thanks for the explanation!  It totally makes sense.  I think the approach you've gone with is sensible in that case so you have my full support :fist_raised:

PR #598, opened 2020-03-25
[v6] improve request multi-mediatype support

chradek:
This PR is an attempt to improve on the multi media-type support on operation requests.

This fixes an issue where the `contentType` contains a single value and is a `ConstantSchema` type.

This also gets rid of the type union for optional parameters that accepted an optional `contentType`. The contentType needs to be specified before specifying other parameters for intellisense to be helpful, so I made it a required parameter if an operation supports multiple media types.

Since `contentType` is now a required parameter, this PR generated an operation overload for each media type. Here's an example generated from the media_types.json swagger file:

```typescript
 /**
   * @param contentType Upload file type
   */
  analyzeBody(
    contentType: ContentType,
  /**
   */
  analyzeBody(
  /**
   * @param contentType Upload file type
   */
  analyzeBody(
    contentType: ContentType | "application/json",
    options?:
      | MediaTypesClientAnalyzeBody$binaryOptionalParams
      | MediaTypesClientAnalyzeBody$jsonOptionalParams
  ): Promise<MediaTypesClientAnalyzeBodyResponse> {
```

I'd love to get some feedback on this approach. The one issue I see with this approach is that if an operation supports a single media type, then adds support for another one in the future, the contentType parameter gets added as a required parameter which would be a breaking change. I don't know if this is a real issue though since I don't know if adding additional media types is actually allowed once a service has released.

If the above is a concern I might be able to work around this by checking if the modeled contentType parameter is the `json` media type and remove the contentType.

/cc @bterlson @jeremymeng


src/generators/operationGenerator.ts
```diff
@@ -286,23 +291,47 @@ type ParameterWithDescription = OptionalKind<
...
+
+  const operationRequests = operation.requests;
+  const overloadParameterDeclarations: ParameterWithDescription[][] = [];
+  const hasMultipleOverloads = Boolean(operationRequests.length > 1);
+
+  for (const request of operationRequests) {
+    const requestMediaType = request.mediaType;
+    // filter our parameters that belong to a different media type
+    const requestParameters = operationParameters.filter(param => {
+      if (!param.targetMediaType) {
```
joheredi: If you want to make this filter shorter, you can have
```typescript
.filter(({targetMediaType}) => !targetMediaType || requestMediaType === targetMediaType)
```

chradek: Shorter and more readable! Thanks, I made this change.

src/transforms/extensions.ts
```diff
@@ -50,15 +48,8 @@ function normalizeMultipleContentTypes(codeModel: CodeModel) {
...
-            !parameter.protocol.http &&
-            !parameterMetadata.serializedName &&
-            parameterMetadata.name.toLowerCase() === "contenttype"
-          ) {
-            parameterMetadata.serializedName = "Content-Type";
-            const http = new Protocol();
-            http.in = ParameterLocation.Header;
-            parameter.protocol.http = http;
+          if (parameterMetadata.name.toLowerCase() === "contenttype") {
+            parameter.required = true;
```
daviwil: Nice workaround for the modelerfour issue, I'm planning to fix this so that it's no longer necessary 😅

src/generators/operationGenerator.ts
```diff
@@ -286,23 +291,44 @@ type ParameterWithDescription = OptionalKind<
...
-      parameters,
-      operation
-    ).map<ParameterWithDescription>(param => {
+  const operationParameters = filterOperationParameters(parameters, operation, {
+    includeContentType: true
+  });
+
+  const operationRequests = operation.requests;
+  const overloadParameterDeclarations: ParameterWithDescription[][] = [];
+  const hasMultipleOverloads = Boolean(operationRequests.length > 1);
```
daviwil: What does `Boolean()` do?  Coerce to boolean somehow?

src/generators/operationGenerator.ts
```diff
@@ -286,23 +291,44 @@ type ParameterWithDescription = OptionalKind<
...
+
+  for (const request of operationRequests) {
+    const requestMediaType = request.mediaType;
+    // filter our parameters that belong to a different media type
```
daviwil:
```suggestion
    // filter out parameters that belong to a different media type
```

src/generators/operationGenerator.ts
```diff
@@ -321,39 +347,141 @@ export function writeOperations(
...
+  return { overloadParameterDeclarations, baseMethodParameters };
+}
+
+/**
+ * Given a list of operation parameter declarations per overload,
+ * returns a list of the parameter declarations that should appear
+ * in the operation's base signature.
+ *
+ * If `overloadParameterDeclarations` contains the parameter declarations for
+ * just a single overload, then thie return value will be the same as the 1st
```
daviwil:
```suggestion
 * just a single overload, then the return value will be the same as the 1st
```

src/generators/operationGenerator.ts
```diff
@@ -321,39 +347,141 @@ export function writeOperations(
...
+ * element in `overloadParameterDeclarations`.
+ * @param overloadParameterDeclarations
+ */
+function getBaseMethodParameterDeclarations(
+  overloadParameterDeclarations: ParameterWithDescription[][]
+): ParameterWithDescription[] {
+  const baseMethodParameters: ParameterWithDescription[] = [];
+  // Need to know which overload has the most parameters to set out upper bound.
```
daviwil:
```suggestion
  // Need to know which overload has the most parameters to set our upper bound.
```

src/transforms/extensions.ts
chradek: You'll only make it required if there are multiple media types though right? :)

test/integration/generated/mediaTypes/src/mediaTypesClient.ts
```diff
@@ -27,26 +29,51 @@ class MediaTypesClient extends MediaTypesClientContext {
...
-    options?: MediaTypesClientAnalyzeBodyOptionalParams
+    contentType: ContentType,
+    options?: MediaTypesClientAnalyzeBody$binaryOptionalParams
+  ): Promise<MediaTypesClientAnalyzeBodyResponse>;
+  /**
+   * Analyze body, that could be different media types.
+   * @param contentType Body Parameter content-type
+   * @param options The options parameters.
+   */
+  analyzeBody(
```
daviwil: Oh, one thing I forgot: Why is the second overload needed?  Is `application/json` not covered by the `ContentType` enum?

src/generators/operationGenerator.ts
chradek: Yes! But I don't need it and will remove.

src/transforms/extensions.ts
daviwil: Actually, I think the issue I'm thinking of is not related to Content-Type parameters, but body parameters when there are multiple requests.  Those body parameters are not carrying forward `required` when they should be.

test/integration/generated/mediaTypes/src/mediaTypesClient.ts
chradek: Correct. ContentType contains the content-types that are binary media types. e.g. application/pdf.
This actually works well for us because one overload accepts the binary content-types (and the request body would accept coreHttp.HttpRequestBody), and the other overload accepts just application/json (so the body parameter is a structured object).

src/transforms/extensions.ts
chradek: Ah right, Brian mentioned that. That will make the overload signatures look much better because the optional parameter will be the same type in all overloads (the body is currently the only thing in the optional type which is why we need multiple versions of it today).

src/generators/operationGenerator.ts
bterlson: Yep, functionally equivalent to `!!` but more clear IMO. Both call the `ToBoolean` abstract operation.

PR #599, opened 2020-03-26
[v6] Full support for parameter grouping

joheredi:
Typescript generator groups optional parameters by default, without x-ms-parameter-grouping being specified,

This PR completes support for parameter grouping while continuing to support our default optional parameters grouping. There are 3 main scenarios

1) An operation specifies `x-ms-parameter-grouping` on any of its parameters
   - We skip the default optional parameters grouping and only group parameters marked with  `x-ms-parameter-grouping`.
   - An `options: coreHttp.OperationOptions` parameter is added to the operation

2) An operation **doesn't** specify `x-ms-parameter-grouping` on any of its parameters.
   - We retain our previous behavior of grouping only optional parameters

3) Autorest.Modelerfour has `group-parameters` flag set to false
   - Same as (2)


test/integration/header.spec.ts
```diff
@@ -181,7 +181,7 @@ describe("typescript", function() {
 
       it("should send and receive duration type headers", async function() {
         const duration = "P123DT22H14M12.011S";
-        await testClient.header.paramDuration("valid", duration);
+        await testClient.header.paramDuration(duration, "valid");
```
chradek: General comment but this is a good example. Why did the positions of these parameters change? Is this the position we would have had with autorest.typescript v4/v5? Main reason I ask was because we got some feedback asking that when possible, we keep the positional arguments in the same order to make it easier for customers to upgrade SDKs.

joheredi: Yeah, I don't love this change of positions. This happened because I had to make sure that optional parameters always show up after required ones, however it also affected other parameters.

I'll think on a way to keep these as consistent with the old version as possible. If you have any suggestions I'd love to hear them

src/generators/operationGenerator.ts
```diff
@@ -339,24 +435,39 @@ function getOperationParameterSignatures(
...
         description: param.description,
         type: typeName,
         hasQuestionToken: !param.required
       };
-    });
 
-    // add optional parameter
-    const optionalParameter = getOptionsParameter(
+      // Make sure required parameters are added before optional
+      return param.required ? [newParameter, ...acc] : [...acc, newParameter];
```
chradek: In order to preserve ordering, we could find the last required parameter already in the list and add the next required element after it. Something like:
```suggestion
      // Make sure required parameters are added before optional
      function findLastRequiredParamIndex(
        params: ParameterWithDescription[]
      ): number {
        for (let i = params.length; i--; ) {
          if (!params[i].hasQuestionToken) {
            return i;
          }
        }
        return -1;
      }
      const newParameterPosition = param.required
        ? findLastRequiredParamIndex(acc) + 1
        : acc.length;
      acc.splice(newParameterPosition, 0, newParameter);
      return acc;
```

src/generators/operationGenerator.ts
```diff
@@ -339,24 +435,39 @@ function getOperationParameterSignatures(
...
       importedModels,
-      {
-        mediaType: hasMultipleOverloads ? requestMediaType : undefined
-      }
+      parameterDeclarations
     );
-    parameterDeclarations.push(optionalParameter);
+
+    // Only use our optional parameter grouping if the parameter grouping extension
+    // is not used in the swagger, if so, honor the extension grouping.
```
chradek: I'm a little confused by this comment. Are we saying that if parameter grouping is enabled for any parameter in an operation, we won't put any non-grouped optional parameters in our options bag? I'm wondering if we would still want to put optional parameters in the `options` object if they aren't part of a parameter group.

Let's take this Swagger example:
```js
"/some/{pathParam1}/{pathParam2}": {
  "operationId": "Update",
  "post": {
    "parameters": [
    {
        "name": "headerParam",
        "in": "header",
        "type": "string",
        "required": false,
        "x-ms-parameter-grouping": {
          "name": "custom-parameter-group"
        }
    },
    {
        "name": "pathParam",
        "in": "path",
        "type": "string",
        "required": true,
        "x-ms-parameter-grouping": {
          "name": "custom-parameter-group"
        }
    },
    {
        "name": "headerParam2",
        "in": "path",
        "type": "string",
        "required": false,
    }]
  }
}
```
In this case, would we get a function like:
```ts
interface CustomParameterGroup {
  headerParam?: string;
  pathParam: string;
}
interface UpdateOptions extends OperationOptions {
  headerParam2?: string
}
client.update({headerParam, pathParam}: CustomParameterGroup, options);
```

joheredi: Thanks for the comment @chradek, this is very reasonable. As we discussed offline we'll proceed with:

* Keep doing our optional parameter grouping, even when parameter grouping extension is used
* If a group is optional, we'll put it in the Options bag

src/generators/operationGenerator.ts
joheredi: Thanks!!! I've integrated your suggestion :smile:

test/integration/header.spec.ts
joheredi: I've integrated your suggestion below to fix this, thanks!

src/generators/operationGenerator.ts
```diff
@@ -198,6 +199,67 @@ function buildMapper(
...
+/**
+ * This function gets the parameter groups specified in the swagger
+ * by the parameter grouping extension x-ms-parameter-grouping
+ */
+function getGroupedParameters(
+  operation: OperationDetails,
+  parameters: ParameterDetails[],
+  importedModels: Set<string>
+): ParameterWithDescription[] {
+  let parameterGroups: Parameter[] = [];
```
chradek: Nit: prefer const over let
```suggestion
  const parameterGroups: Parameter[] = [];
```

src/generators/modelsGenerator.ts
```diff
@@ -75,29 +77,57 @@ const writeOperationModels = (
...
       );
       writeResponseTypes(operation, modelsIndexFile);
     });
   });
 
+/**
+ * Decides whether or not to include the Options parameter model.
+ *
+ * We'll only include this model when the operation doesn't have
+ * any grouped parameters through x-ms-parameter-grouping.
```
chradek: Is this comment still accurate?

joheredi: Thanks for catching this!

PR #601, opened 2020-03-30
[v6] adds support for operation overloads with non-matching parameters

chradek:
The updated version of modelerfour does not guarantee the same value for `required` for body parameters.

Previously, we assumed that across media types, body parameters would be either consistently required, or consistently not required. Modelerfour will now always treat binary request bodies as required, even if the request body is modeled as not required in the swagger.

The [media_types.json](https://github.com/Azure/autorest.testserver/blob/75c1296f7123abbbc750067df043970cea95d7f2/swagger/media_types.json#L14-L40) swagger has a good example of this, where the JSON request body parameter is not required.

After these code changes, operation overloads no longer have to have the same parameter ordering, as long as they each have a `contentType` parameter (we set `always-create-content-type-parameter: true` to ensure this is the case). Using the swagger from above, here's an example of what our overloads now look like when parameter differ
```ts
/**
   * @param contentType Upload file type
   * @param input Input parameter.
   */
  analyzeBody(
    contentType: ContentType,
    input: coreHttp.HttpRequestBody,
  /**
   */
  analyzeBody(
```

For the base operation that implements all the overloads, I decided to use a spread argument where the type is a union of tuples. Each tuple maps to a single overload.

```ts
/**
   * @param args Includes all the parameters for this operation.
   */
  analyzeBody(
    ...args:
      | [
          ContentType,
          coreHttp.HttpRequestBody,
          MediaTypesClientAnalyzeBody$binaryOptionalParams?
        ]
      | ["application/json", MediaTypesClientAnalyzeBody$jsonOptionalParams?]
  ): Promise<MediaTypesClientAnalyzeBodyResponse>
```

TypeScript is smart enough to properly resolve types in tuples if we can treat one of the values as a discriminator type. In our case, we can always use the `contentType` argument as our discriminator, after which the rest of the types are resolved:

```ts
if (
      args[0] === "application/pdf" ||
      args[0] === "image/jpeg" ||
      args[0] === "image/png" ||
      args[0] === "image/tiff"
    ) {
      operationSpec = analyzeBody$binaryOperationSpec;
      operationArguments = {
        contentType: args[0],
        input: args[1],
        options: args[2]
      };
    } else if (args[0] === "application/json") {
      operationSpec = analyzeBody$jsonOperationSpec;
      operationArguments = {
        contentType: args[0],
        options: args[1]
      };
    } else {
      throw new TypeError(
        `"contentType" must be a valid value but instead was "${args[0]}".`
      );
    }
```

### DIFF
--- a/src/generators/operationGenerator.ts
+++ b/src/generators/operationGenerator.ts
@@ -381,46 +381,29 @@ function getOperationParameterSignatures(
 function getBaseMethodParameterDeclarations(
   overloadParameterDeclarations: ParameterWithDescription[][]
 ): ParameterWithDescription[] {
-  const baseMethodParameters: ParameterWithDescription[] = [];
-  // Need to know which overload has the most parameters to set our upper bound.
-  const maxOverloadSize = Math.max(
-    ...overloadParameterDeclarations.map(params => params.length)
-  );
-  for (let i = 0; i < maxOverloadSize; i++) {
-    // attempt to combine types
-    const declarations: ParameterWithDescription[] = [];
-    for (const overloadParameterDeclaration of overloadParameterDeclarations) {
-      if (overloadParameterDeclaration[i]) {
-        declarations.push(overloadParameterDeclaration[i]);
-      }
-    }
-
-    const declaration = declarations.reduce(
-      (prevDeclaration, curDeclaration) => {
-        prevDeclaration = { ...prevDeclaration };
-        if (curDeclaration.name !== prevDeclaration.name) {
-          // Currently we only generate overloads if an operation supports multiple media types.
-          // Since contentType is always required and the 1st parameter, parameter names/ordering
-          // shouldn't change.
-          // In order to support more variance in parameter naming/ordering, we'll need to be able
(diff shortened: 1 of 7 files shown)


src/generators/operationGenerator.ts
```diff
@@ -381,46 +381,29 @@ function getOperationParameterSignatures(
...
+
+  const baseMethodArg: ParameterWithDescription = {
+    name: "args",
+    isRestParameter: true,
+    description: "Includes all the parameters for this operation.",
+    type: overloadParameterDeclarations
+      .map(overloadParams => {
+        return `[ ${overloadParams
+          .map(param => {
+            return param.hasQuestionToken ? `${param.type}?` : param.type;
```
joheredi: not important - you could remove the braces and return inside this map

chradek: I was able to get it to fit on a single line by changing `param` to `p` as well, so made the change.

chradek:
It was a fun change to work on :smile:
Thanks for the review!

PR #602, opened 2020-03-30
[v6] fixes inherited additionalProperties and odata

### LINKED ISSUE #574: [v6] Support odata response
While generating mappers for the paging.json test server swagger file, I encountered an issue with how we generate the serializedName for odata fields in the response. We need to add an escaped backslash before the period in the field name.

Here's a working example from the old autorest:
https://github.com/Azure/autorest.typescript/blob/e719bdb2922e1c1ba21ec5494f45bcb2ea72488b/test/azure/generated/Paging/models/mappers.ts#L244

Currently we don't do anything special and generate a serializedName that looks like `odata.nextLink`.

chradek:
Fixes #574 

This PR adds tests for [additionalProperties.json](https://github.com/Azure/autorest.testserver/blob/master/swagger/additionalProperties.json). I came across 2 issues while adding tests:
1. Mappers that inherited properties from a parent mapper weren't inheriting the `additionalProperties` field.
2. odata fields weren't being serialized properly, which was previously reported in #574.

This PR adds fixes for both issues. For the odata fix, I borrowed the logic from autorest.typescript v5.  It'd be great if modelerfour could escape the `.` in odata fields but I don't know how that might impact other autorest extensions.


src/transforms/mapperTransforms.ts
```diff
@@ -251,10 +251,8 @@ function getXmlMetadata(
 }
 
 // TODO: Make sure this is the correct way to handle additionalProperties
```
daviwil: Is this comment necessary anymore?

chradek: I'm assuming not now that all the tests pass! Removed.

PR #606, opened 2020-04-01
[v6] Implement LRO outline and BodyPolling strategy

joheredi:
Initial implementation of LRO support. This PR includes the outlines for the whole LRO implementation plus the BodyPolling strategy. 

After implementing the rest of the strategies, the plan is to integrate LRO to code generation by including the lro folder in the generated output, and referencing when x-ms-long-running-operation extension is used

Coming up in follow up PRs:
* Azure-AsyncOperation strategy
* Location strategy
* Integrate with Code Generation

### DIFF
--- a/src/lro/bodyPollingStrategy.ts
+++ b/src/lro/bodyPollingStrategy.ts
@@ -1,0 +1,47 @@
+import { LROStrategy, BaseResult, LROOperationState } from "./models";
+import { OperationSpec } from "@azure/core-http";
+
+const terminalStates = ["succeeded", "failed", "canceled", "cancelled"];
+
+/**
+ * Creates a polling strategy based on BodyPolling which uses the provisioning state
+ * from the result to determine the current operation state
+ */
+export function createBodyPollingStrategy<TResult extends BaseResult>({
+  lastOperation: operation,
+  sendOperation
+}: LROOperationState<TResult>): LROStrategy<TResult> {
+  let lastOperation = { ...operation };
+  return {
+    isTerminal: () => {
+      // If provisioning state is missing, default to Success
+      const provisioningState = (
+        lastOperation.result.provisioningState ||
+        lastOperation.result.properties?.provisioningState ||
+        "succeeded"
+      ).toLowerCase();
+
+      return terminalStates.includes(provisioningState);
+    },
+    sendFinalRequest: () => {
+      // BodyPolling doesn't require a final get so return the lastOperation
+      return Promise.resolve(lastOperation);
+    },
+    poll: async () => {
+      // When doing BodyPolling, we need to poll to the original url with a
+      // GET http method
+      const pollingSpec: OperationSpec = {
+        ...lastOperation.spec,
+        httpMethod: "GET"
+      };
+
+      // Execute the polling operation
+      const result = await sendOperation(lastOperation.args, pollingSpec);
+
+      // Update lastOperation
+      lastOperation.spec = pollingSpec;
+      lastOperation.result = result;
+      return lastOperation;
+    }
+  };
+}
--- a/src/lro/bodyPollingStrategy.ts
+++ b/src/lro/bodyPollingStrategy.ts
@@ -38,8 +38,7 @@ export function createBodyPollingStrategy<TResult extends BaseResult>({
       // Execute the polling operation
       const result = await sendOperation(lastOperation.args, pollingSpec);
 
-      // Update lastOperation
-      lastOperation.spec = pollingSpec;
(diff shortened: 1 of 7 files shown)


src/lro/models.ts
```diff
@@ -0,0 +1,38 @@
+import { OperationArguments, OperationSpec } from "@azure/core-http";
+import { PollOperationState, PollOperation } from "@azure/core-lro";
+
+export interface BaseResult {
+  location?: string;
+  operationLocation?: string;
+  azureAsyncOperation?: string;
+  provisioningState?: string;
+  properties?: {
+    provisioningState?: string;
```
chradek, line 10: Is `provisioningState` an example specifically for our `lro.json` swagger, or is it well-defined as something that will always be present when we attempt the "BodyPolling" strategy?

src/lro/bodyPollingStrategy.ts
```diff
@@ -0,0 +1,47 @@
...
+    poll: async () => {
+      // When doing BodyPolling, we need to poll to the original url with a
+      // GET http method
+      const pollingSpec: OperationSpec = {
+        ...lastOperation.spec,
+        httpMethod: "GET"
+      };
+
+      // Execute the polling operation
+      const result = await sendOperation(lastOperation.args, pollingSpec);
```
chradek, line 39: So if the service doesn't support making a `GET` request with the same url/arguments this will throw an error. That seems like the right thing to do. Will that stop any further polling from happening or can you still continue to poll?

src/lro/lroPoller.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    this.intervalInMs = intervalInMs;
+  }
+
+  /**
+   * The method used by the poller to wait before attempting to update its operation.
+   */
+  async delay(): Promise<void> {
+    delay(this.intervalInMs);
+  }
```
chradek, line 66: Does `delay` already return a promise? If so this could be updated to the following for better perf/debugability:
```suggestion
  delay(): Promise<void> {
    return delay(this.intervalInMs);
  }
```
I think either way right now since `delay(this.intervalInMs)` is non-blocking, your wrapper `delay` is going to return faster than expected.

src/lro/operation.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  if (result.azureAsyncOperation) {
+    throw new Error("Azure-AsyncOperation strategy is not yet implemented");
+  }
+
+  if (result.location) {
+    throw new Error("Location strategy is not yet implemented");
+  }
+
+  // TODO: Should we include here other methods like Patch?
```
chradek, line 100: Which methods did we include in autorest.typescript v5/what do other SDKs support? I could see supporting it for anything that's not `HEAD` or `GET`. (Even `DELETE` might return a status if it takes a long time to delete something)

src/lro/models.ts
joheredi: provisioningState is standard for BodyPolling strategy, I think it is a property regularly used for Azure Resource management libraries.

src/lro/bodyPollingStrategy.ts
joheredi: From the implementation of the base Poller in @azure/core-lro, I think it will reject the promise so the polling will stop.

This is the where the base poller handles errors: https://github.com/Azure/azure-sdk-for-js/blob/beef206545ceca025e2114af4690a15eb8ee6bb7/sdk/core/core-lro/src/poller.ts#L323


If we want any kind of retry I a retry policy needs to be provided to coreHttp

src/lro/lroPoller.ts
joheredi: Oops good catch, I missed that. I'm removing the async and returning coreHttp's delay

src/lro/operation.ts
joheredi: It looks like V5 also includes PATCH through core-arm ([code](https://github.com/Azure/azure-sdk-for-js/blob/beef206545ceca025e2114af4690a15eb8ee6bb7/sdk/core/core-arm/src/lroPollStrategy.ts#L298)).

Python seems to only do PUT ([code](https://github.com/Azure/msrestazure-for-python/blob/698eb06bcbf9fda4c7ae04d929f7b5b955fcbfed/msrestazure/polling/arm_polling.py#L334))

I'll keep PUT and PATCH for now

src/lro/models.ts
chradek: Gotcha. Actually really glad to hear that there’s a standard field being used for this case!

PR #607, opened 2020-04-02
[v6] Lro generation

joheredi:
Generating LRO swagger, includes:

- Changes in Operation body when LRO extension is used
- Include LRO Poller in SDK
- Chose success responses when many responses are defined. Only for LRO

Next PRs to follow:
- Implement Azure-AsyncOperation strategy
- Implement Location strategy


src/transforms/operationTransforms.ts
```diff
@@ -257,6 +257,9 @@ export async function transformOperation(
     usedModels: [typeName]
   };
 
+  const isLRO: boolean =
+    operation.extensions && operation.extensions["x-ms-long-running-operation"];
```
chradek, line 261: Does this need to be wrapped in a Boolean?
```suggestion
    Boolean(operation.extensions && operation.extensions["x-ms-long-running-operation"]);
```
Otherwise if either item were undefined, wouldn't `isLRO` be undefined too?

src/generators/LROGenerator.ts
```diff
@@ -0,0 +1,47 @@
...
+  return new Promise((resolve, reject) => {
+    readFile(filePath, "utf8", (error, data) => {
+      if (error) {
+        reject(error);
+      } else {
+        resolve(data);
+      }
+    });
+  });
+}
```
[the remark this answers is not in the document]
joheredi: 😮 that is great! Thanks Chris I'll use that instead

src/transforms/operationTransforms.ts
joheredi: updated, thanks!

PR #611, opened 2020-04-08
[v6] Complete support for LRO

joheredi:
**Summary of changes in this PR:**
   * New LRO pipeline Policy that extracts LRO specific information such as operationState, status, location headers
   * Selective deserialization for LRO operation, so that we only deserialize final responses. 
      * This is to work around LRO's swagger design of not requiring polling operations to be defined
   * Refactor BodyPollingStrategy
      * To use lroData added in the LRO Policy
      * To have a better distinction between current and initial operations
   * Azure-AsyncOperation strategy
      * Polling strategy to follow when Azure-AsyncOperation or Operation-Location headers are present in the initial operation
   * Location strategy
      * Polling strategy to follow when Location header is present in the initial operation (But not Azure-AsyncOperation or Operation-Location)
   * Pass through strategy
      * Default strategy to use when the initial response doesn't meet the criteria for other strategies. This strategy will just return the original response
   * Add Integration tests
 
**More info about the changes**
LRO Generator:
   * Pick all files from LRO folder
ClientContextGenerator:
   * Create default policies from options to allow overriding default policy options
      * i.e. retry interval
   * Add LRO policy to default policies

Operation Generator:
   * When the operation group contains any LRO opertaion, adds a getOperationOptions private function which adds shouldDeserialize to options and sends it down to coreHttp. We'll only deserialize final LRO responses
   * LRO operations pass shouldDeserialize as an OperationOption
   * Converting OperationOptions to RequestOptionsBase before sending them down to coreHttp.sendOperation

Operation Transform:
   * Extract LRO options


src/lro/lroPoller.ts
```diff
@@ -68,3 +97,48 @@ export class LROPoller<TResult extends BaseResult> extends Poller<
...
+function getPollingStrategy<TResult extends BaseResult>(
+  initialOperation: LROOperationStep<TResult>,
+  sendOperationFn: SendOperationFn<TResult>,
+  finalStateVia?: FinalStateVia
+) {
+  const lroData = initialOperation.result._lroData;
+
+  if (!lroData) {
+    throw new Error(
+      "Expected lroData to be present for PollingStrategy discovery"
```
chradek, line 122: This error eventually gets surfaced to the customer using our generated SDK right? I wonder if this should be a `RestError` then so we can include the request/response on the error for better troubleshooting.

src/lro/passthroughStrategy.ts
```diff
@@ -0,0 +1,23 @@
...
+  return {
+    isTerminal: () => {
+      return true;
+    },
+    sendFinalRequest: () => {
+      // BodyPolling doesn't require a final get so return the lastOperation
+      return Promise.resolve(initialOperation);
+    },
+    poll: async () => {
+      throw new Error("Passthrough strategy should never poll");
```
chradek, line 20: I'm a little fuzzy on this part. Does `poll` get called as a result of a method invocation the user can make? I'm wondering if it makes sense to have `poll` return the initialOperation as well, or be a no-op, rather than throwing an error.

joheredi: This poll function is internal to our LRO handling, users cannot call this directly, we'll decide if it is called or not in the update method in operation.ts. 

Core-LRO provides an interface with a poll function which is different to this one, core-lro poll would work with the "operation" state, it checks if the operation is complete by looking at the state.isCompleted, if it is completed it does nothing. If it is not completed, then it calls the **update** method we define. There we decide based on the strategy what needs to happen.

Passthrough strategy isTerminal will always be true so no poll should ever be called, I think that if we return the initial operation we may risk running into an infinite loop if a bug is introduced in passtrough's isTerminal and returns false

src/generators/operationGenerator.ts
```diff
@@ -109,6 +109,26 @@ function generateOperation(
...
+      { name: "finalStateVia", type: "string", hasQuestionToken: true }
+    ],
+    returnType: `coreHttp.RequestOptionsBase`,
+    statements: `
+    const operationOptions: coreHttp.OperationOptions = options || {};
+    operationOptions.requestOptions = {
+      ...operationOptions.requestOptions,
+      shouldDeserialize: shouldDeserializeLRO(finalStateVia),
+    };
+    return coreHttp.operationOptionsToRequestOptionsBase(operationOptions);
```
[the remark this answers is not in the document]
joheredi: I think I missed converting to RequestOptionsBase when I made the switch from RequestOptionsBase to OperationOptions.

Since RequestOptionsBase has an indexer it wouldn't complain about us not converting it, but I think SendOperation  doesn't understand how to flatten OperationOptions and would silently fail to apply the right options

src/generators/LROGenerator.ts
```diff
@@ -11,25 +11,21 @@ export async function generateLROFiles(
...
   }
 
-  const lroFiles = [
-    { name: "bodyPollingStrategy", file: "./src/lro/bodyPollingStrategy.ts" },
-    { name: "lroPoller", file: "./src/lro/lroPoller.ts" },
-    { name: "models", file: "./src/lro/models.ts" },
-    { name: "operation", file: "./src/lro/operation.ts" }
-  ];
+  const lroDir = "./src/lro";
+  const lroFiles = await promises.readdir(lroDir);
```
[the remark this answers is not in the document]
joheredi: Thanks for catching this, I've fixed it. I had to jump a bit more since autorest.core would pickup the already build autorest.typescript and __dirname would be under /dist

src/lro/lroPolicy.ts
```diff
@@ -0,0 +1,81 @@
+import {
```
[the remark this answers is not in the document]
joheredi: I agree, in general we should find a good home for everything in the LRO folder. Whether it is splitting the modules into different existing packages like core-http and core-lro. Or creating new package for azure specific stuff like this to be used in autorest generated code.

I have filed #613 to track this, I'll get to it right after this PR

src/lro/lroPolicy.ts
```diff
@@ -0,0 +1,81 @@
...
+
+class LROPolicy extends BaseRequestPolicy {
+  constructor(nextPolicy: RequestPolicy, options: RequestPolicyOptions) {
+    super(nextPolicy, options);
+  }
+
+  public async sendRequest(
+    webResource: WebResource
+  ): Promise<HttpOperationResponse> {
+    webResource.operationSpec = injectMissingResponses(
```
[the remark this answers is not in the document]
joheredi: We can, and I think I like it better in the generated operationSpecs than at runtime in the pipeline. I'm making the change. Thanks for suggesting this!

src/generators/clientContextFileGenerator.ts
```diff
@@ -141,7 +161,16 @@ const writeStatements = (lines: string[], shouldAddBlankLine = false) => (
...
 
-function writeDefaultOptions(hasCredentials: boolean) {
+function writeDefaultOptions(hasCredentials: boolean, hasLRO: boolean) {
+  const policies = ` 
+  const defaultPipelines = coreHttp.createPipelineFromOptions(options)
+      .requestPolicyFactories as coreHttp.RequestPolicyFactory[];
+
+  options = {
+      ...options,
+      requestPolicyFactories: [lroPolicy(), ...defaultPipelines]
```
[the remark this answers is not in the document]
joheredi: Yes, however it will be no-op, since I'm moving the the extra response injection to the generation step.

src/lro/lroPoller.ts
joheredi: Make sense, made this a RestError, thanks for the suggestion

package-lock.json
```diff
@@ -113,24 +113,24 @@
             "integrity": "sha512-kmv8CGrPfN9SwMwrkiBK9VTQYxdFQEGe0BmQk+M8io56P9KNzpAxcWE/1fxJj7uouwN4kXF0BHW8DNlgx+wtCg=="
```
chradek, line 113: Sorry, you may need regenerate the package-lock.json after rebasing or merging from master now that I've merged the TimeSchema PR. That also updated the package-lock.

src/lro/constants.ts
```diff
@@ -0,0 +1 @@
+export const terminalStates = ["succeeded", "failed", "canceled", "cancelled"];
```
chradek, line 1: That's kind of interesting that we support both spellings for the past tense of `cancel` :smile:

src/lro/requestUtils.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  if (initialOperationInfo.requestMethod === "DELETE") {
+    return true;
+  }
+
+  if (
+    initialOperationInfo.requestMethod === "PUT" &&
+    finalStateVia &&
+    finalStateVia.toLowerCase() === "azure-asyncOperation"
```
chradek, line 67:
```suggestion
    finalStateVia.toLowerCase() === "azure-asyncoperation"
```
I think this change is needed since `finalStateVia` will always be all lowercase.

src/lro/constants.ts
joheredi: I included it since I saw it in the v5 version. I thought it wouldn't hurt to have 😄

package-lock.json
joheredi: done!

src/lro/requestUtils.ts
joheredi: oh good catch!!!

package-lock.json
joheredi: re-grenerated

PR #616, opened 2020-04-14
[v5] Keep parents when generating PageCompositeTypes

### LINKED ISSUE #615: [v5] PageCompositeType generation drops base model
When generating a PageCompsiteType interface, extending from parents defined in swagger is omitted.

For example in [this swagger](https://github.com/Azure/azure-rest-api-specs/blob/0c0880833da9d695140d140071aea579fdc3da38/specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/stable/2019-06-15/bms.json#L3610-L3627) `ResourceList` is dropped when generating `JobResourceList`

```typescript
export interface JobResourceList extends Array<JobResource> {
}
```

Where it should generate the following in order to honor the swagger definition
```typescript
export interface JobResourceList extends Array<JobResource>, JobResourceList {
}
```
See Issue Azure/azure-sdk-for-js/issues/7630 for more details.

joheredi:
Fixes #615 

Make sure that the interface also extends the base model when generating PageCompositeTypes.

For example in [this swagger](https://github.com/Azure/azure-rest-api-specs/blob/0c0880833da9d695140d140071aea579fdc3da38/specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/stable/2019-06-15/bms.json#L3610-L3627) `ResourceList` is dropped when generating `JobResourceList`

```typescript
export interface JobResourceList extends Array<JobResource> {
}
```

Where it should generate the following in order to honor the swagger definition
```typescript
export interface JobResourceList extends Array<JobResource>, JobResourceList {
}
```

@daviwil, @sarangan12  - let me know if you think this is a good approach for fixing this issue

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft.azure/autorest.typescript",
-  "version": "5.0.2",
+  "version": "5.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft.azure/autorest.typescript",
-  "version": "5.0.2",
+  "version": "5.1.0",
   "description": "The typescript extension for classic generators in AutoRest.",
   "scripts": {
     "autorest": "autorest",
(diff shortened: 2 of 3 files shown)


chradek:
So just to be clear, in the issue posted to the azure sdk repo, were you able to see nextLink defined at runtime? It wasn't clear to me whether `nextLink` was just missing from the TypeScript interface, or if it wasn't showing up in the response as well.

Side note: CI should pass if you update the package version.

PR #622, opened 2020-04-16
Generate supported services and add clean tree validation

joheredi:


test/utils/test-swagger-gen.ts
```diff
@@ -1,154 +1,254 @@
 import { exec } from "child_process";
+import { join as joinPath } from "path";
 
 interface SwaggerConfig {
   swagger: string;
   clientName: string;
   packageName: string;
+  category?: Category;
```
chradek, line 8: It looks like this should be required. If you do that ts-node won't complain about category possibly being undefined in the `if (!categories.has(category)) {` check on line 270.
```suggestion
  category: Category;
```

joheredi: Made it required, thanks Chris!

PR #625, opened 2020-04-23
[v6] Naming fixes

joheredi:
- Add more reserved words
- Provide more context info to normalizeName to get better prefixes when find a reserved name
- Make sure to guard for reserved names when normalizing in Transforms
- Use getOperationFullname function to avoid inconsistencies
- Update generated files and tests


test/integration/generated/azureSpecialProperties/src/operations/xMsClientRequestId.ts
```diff
@@ -30,13 +30,15 @@ export class XMsClientRequestId {
    * 9C4D50EE-2D56-4CD3-8152-34347DC9F2B0.
    * @param options The options parameters.
    */
-  get(options?: coreHttp.OperationOptions): Promise<coreHttp.RestResponse> {
+  getOperation(
```
daviwil, line 33: I see a lot of suffixing in places where it might not be semantically necessary, should we try to minimize that so that the operation (and operation group) names are less complicated?

test/integration/generated/bodyString/src/operations/stringOperations.ts
```diff
@@ -0,0 +1,434 @@
+/*
+ * Copyright (c) Microsoft Corporation.
```
daviwil, line 2: Looks like new files are being added for the renamed operation groups, you might need to delete the leftover files with the old name

src/transforms/operationTransforms.ts
```diff
@@ -372,23 +397,35 @@ export function transformOperationGroups(
   );
 }
 
+function getOperationGroupName(
+  operationGroup: OperationGroup,
+  clientName: string = ""
+) {
+  const { name } = getLanguageMetadata(operationGroup.language);
+
+  return normalizeName(name || clientName, NameType.OperationGroup, true);
```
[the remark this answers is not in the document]
joheredi: I think we get the correct name from modelerfour (x-ms-client-name vs name). Here I'm trying to make sure that it has the right casing and if it is a Typescript reserved name add a suffix

test/integration/generated/azureSpecialProperties/src/models/mappers.ts
```diff
@@ -52,11 +52,11 @@ export const HeaderCustomNamedRequestIdHeaders: coreHttp.CompositeMapper = {
   }
 };
 
-export const HeaderCustomNamedRequestIdParamGroupingHeaders: coreHttp.CompositeMapper = {
-  serializedName: "header_customNamedRequestIdParamGroupingHeaders",
+export const HeaderCustomnamedrequestidparamgroupingHeaders: coreHttp.CompositeMapper = {
+  serializedName: "header_customnamedrequestidparamgroupingHeaders",
```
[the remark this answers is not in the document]
joheredi: Thanks for catching these un intended casing change

test/integration/generated/azureSpecialProperties/src/operations/xMsClientRequestId.ts
joheredi: I minimized it by nor suffixing operations. However I had to keep suffixing OperationGroups

src/utils/nameUtils.ts
```diff
@@ -4,7 +4,70 @@ import { Operation, OperationGroup } from "@azure-tools/codemodel";
 import { getLanguageMetadata } from "./languageHelpers";
 import { TypeDetails, PropertyKind } from "../models/modelDetails";
 
-const ReservedModelNames = ["Error", "Date"];
+const ReservedModelNames = [
```
chradek, line 7: Should we be using one set of rules for all models?

I can see one set of rules for things that can be 'new'd up. For example, OperationGroups can be instantiated directly by the user, so you need to guard against names like `Date`, `String`, etc. 

However, properties can be pretty much anything. You can have an interface that looks like this:
```ts
interface Foo {
  string: string;
}
```
and that works just fine. Likewise for parameters:
```ts
function(string: string) {}
```
However, in the parameters case you definitely don't want to allow any reserved words (and that's probably true for properties too so that params/properties stay consistent across operation inputs and outputs.

Side-note: Can you make this list ordered alphabetically? It'll be easier to add things or check what's in it then.

joheredi: Alright, thanks @daviwil and @chradek for your feedback around this. I agree with you and here's what I'm thinking:

* Have metadata on the reserved words to indicate in which cases we want special handling. So that we just maintain a single list of reserved words but can choose case by case if we need to suffix the name.

* File, operations and properties, don't need suffixing as @chradek mentioned above, so I have skipped a suffix in those cases.

* OperationGroups, Parameters, Classes and Interfaces(for consistency) would need to be guarded against a subset of reserved words, such as Error, String, Date, Boolean, etc
   * OperationGroups will get Operations suffix.
   * Parameters would get Param suffix
   * Classes and Interfaces will get Model suffix

test/integration/generated/additionalProperties/src/models/mappers.ts
```diff
@@ -9,7 +9,6 @@
 import * as coreHttp from "@azure/core-http";
 
 export const PetAPTrue: coreHttp.CompositeMapper = {
-  serializedName: "PetAPTrue",
```
joheredi, line 12: This is an intended change. We have been falling back to the schema name when serialized name is not defined. However falling back to the name doesn't add value and isn't needed(except for xml scenarios), it is only required for properties for properly mapping and sending data over the wire. Hence I'm removing unnecessary serializedName properties from mappers

src/utils/nameUtils.ts
```diff
@@ -4,22 +4,117 @@ import { Operation, OperationGroup } from "@azure-tools/codemodel";
...
-  Property
+  Property,
+  Parameter,
+  Operation,
+  OperationGroup
+}
+
+const Newable = [NameType.Class, NameType.Interface, NameType.OperationGroup];
+
+const ReservedModelNames: ReservedName[] = [
```
daviwil, line 24: I like this!  Seems like all of them need `NameType.Parameter` so it feels like that should be checked by default in `guardReservedNames`, but I suppose it makes things less clear to not have it explicitly in the list.

joheredi: Yeah, I thought about that. However by having it there we don't need any NameType specific checks in guardReservedNames and as you said makes it makes it clear and gives us flexibility in case we decide we don't want to suffix a specific reserved word. For instance string would be a valid parameter name

PR #630, opened 2020-04-28
[v6] Add Smoke Test + Fix parameter name

joheredi:
* Add script for Smoke Testing supported sdks
   * Smoke test consists on Generating the sdk from a Readme.md and building without errors
   * Allow passing slice and size to smoke-test to be able to split smoke testing in chunks when the list of readmes grow.

* Fix for parameter names, we were using serializedNames instead of the name already processed by prenamer


src/transforms/parameterTransforms.ts
```diff
@@ -185,16 +185,22 @@ export function populateOperationParameters(
   targetMediaType?: KnownMediaType
 ): void {
   const parameterName = getParameterName(parameter);
-  let description =
-    getLanguageMetadata(parameter.language).description ||
-    getLanguageMetadata(parameter.schema.language).description;
+  const patameterMetadata = getLanguageMetadata(parameter.language);
```
daviwil, line 188:
```suggestion
  const parameterMetadata = getLanguageMetadata(parameter.language);
```

joheredi: Thanks for catching that one 😄 fixed it

PR #632, opened 2020-04-30
[v6] Fix Regex pattern constraint generation

joheredi:
**Problem**
When a string parameter includes a regex pattern constraint we are generating the mapper incorrectly causing a build error. This is because we were trying to stringify the Regex object.
image

**Fix**
Instead of stringify the Regex object, we need to special handle the pattern property to output the new RegExp creation
image

**A couple other small fixes included:**
- Fix missing Operations import when SDK has both client level and Operation classes
- Fix modelProperties mapper generation
   - We were simply stringifying the the mappers under modelProperties, with this we missed any custom generation needed. To fix this I'm called generateMapper instead


src/generators/clientFileGenerator.ts
```diff
@@ -54,16 +59,18 @@ export function generateClient(clientDetails: ClientDetails, project: Project) {
       moduleSpecifier: "@azure/core-http"
     });
 
+  if (hasImportedOperations) {
```
joheredi, line 62: Included this fix as I found an error where the SDK had both Client level operations and operation group classes, we were not importing the Operations

src/generators/mappersGenerator.ts
```diff
@@ -171,7 +195,14 @@ function writeModelProperties(
     //  color: { type: { name: "String" }, serializedName: "color" }
     // }
     writeParentMappers(parents, writer);
-    writeObjectProps(modelProperties, writer);
+    // Write all sub-mappers
+    if (modelProperties) {
```
joheredi, line 199: Added this to make sure we correctly generate mappers for each modelProperty, otherwise we just stringify and miss any custom handling

src/generators/clientFileGenerator.ts
```diff
@@ -29,11 +29,16 @@ export function generateClient(clientDetails: ClientDetails, project: Project) {
   const clientContextClassName = `${clientDetails.className}Context`;
   const hasMappers = !!clientDetails.mappers.length;
 
-  // A client has inline operations when it has a toplevel operation group
+  // Check if there ara any client level operations
```
daviwil, line 32:
```suggestion
  // Check if there are any client level operations
```

Issue #546, opened 2020-01-13
[v6] Correctly handle multiple Success Response codes


joheredi:
[Current version]

When a response defines more than one success response, for example:

chradek:
Do we have any cases where we'd expect to see a different response shape for different successful status codes? The 2 issues linked had multiple error status codes but only a single successful status code.

The reason I ask is because it seems weird to me that I might get a completely different shape based on the status code (except in the case of errors). What I could imagine is that in one status code, I get a shape that includes only a subset of the fields in the other (for example, one status code indicates an operation completed, while the other indicates that the service has accepted the input and will complete).

That said I don't know the current state of our models so I could be wrong!

PR #633, opened 2020-05-01
[v6] Avoid generating duplicate response types

joheredi:
**Problem**
We are generating duplicate response types. This is happening in swaggers where there are 2 success responses defined. This is is an example([link](https://github.com/Azure/azure-rest-api-specs/blob/49fc16354df7211f8392c56884a3437138317d1f/specification/storage/resource-manager/Microsoft.Storage/stable/2019-06-01/blob.json#L235)) where 2 responses referencing the same type are defined

```json
"responses": {
          "201": {
            "description": "Created -- Create Container operation completed successfully.",
            "schema": {
              "$ref": "#/definitions/BlobContainer"
            }
          },
          "200": {
            "description": "OK -- The Container is already created.",
            "schema": {
              "$ref": "#/definitions/BlobContainer"
            }
          }
        }
      }
```

**Fix**
I have added a `hash` build from the response object. this hash is used during generation to compare equality.

When an operation defines different success return types I'm throwing an exception. I decided to do it this way to unblock generation of the following swaggers: `Keyvault`, `Storage` and `Msi` 

However OpenAPI doesn't seem to restrict the definition of different response schemas for the same operation, so in theory we could get into that scenario. I'd need to find a real world example to understand scenario and requirements. I think we'll tackle the problem when we face a real world swagger, we have an issue tracking this #546

**Other changes in this PR**
* Added `Keyvault`, `Storage` and `Msi` resource-manager swaggers to the Smoke Tests
* Added missing `@azure/core-lro` dependency to generated package.json


src/generators/static/packageFileGenerator.ts
```diff
@@ -18,6 +21,7 @@ export function generatePackageJson(
       `A generated SDK for ${clientDetails.name}.`,
     version: packageDetails.version,
     dependencies: {
+      ...(hasLRO && { "@azure/core-lro": "^1.0.1" }),
```
chradek, line 24: Good catch!

src/transforms/operationTransforms.ts
```diff
@@ -291,13 +292,22 @@ export function transformOperationResponse(
 
   const isDefault = httpInfo.statusCodes.indexOf("default") > -1;
 
-  return {
-    statusCodes: httpInfo.statusCodes,
```
chradek, line 295: We weren't relying on the statusCodes for determining which error to use, were we? (E.g. if an operation supports multiple error types based on the status code)

src/transforms/operationTransforms.ts
```diff
@@ -291,13 +292,22 @@ export function transformOperationResponse(
...
+  const resultResponse = {
     mediaType: httpInfo.knownMediaType,
     mappers,
     types,
     isError: isDefault || isError
   };
+
+  const hash = createHash("md5")
+    .update(JSON.stringify(resultResponse))
```
chradek, line 303: I _think_ this should be fine...the only thing that has me slightly worried is if you had two JavaScript objects that had fields added to them in a different order, then the hash wouldn't match even if the contents were the same. I think a deep equality check (I think lodash has one) would cover that scenario and doesn't seem like it would be more expensive since we're only checking the hash in one place, but I'll leave it up to you (presumably our generated test clients should catch this scenario).

src/transforms/operationTransforms.ts
joheredi: Nope, currently we only handle Errors marked by `x-ms-error-response` or `default` status code. When a response is marked as `x-ms-error-response` M4 groups it under exceptions instead of Error. If a swagger defines an error response without that extension it will be treated as a success response and we'll very likely throw. However that would be a buggy swagger.

src/transforms/operationTransforms.ts
joheredi: Yeah, you're right. I didn't think about the order 😄. It is not impacting any of our current tests, however it would be a timebomb. I'll use lodash's isEqual instead, as you suggested. Thanks for catching this

PR #637, opened 2020-05-05
[v6] Check git tree on CI

joheredi:
Add a check for a clean git tree to make sure no un-intended changes in generated clients slip through

### DIFF
--- a/test/integration/generated/mediaTypesV3Lro/src/mediaTypesV3LROClient.ts
+++ b/test/integration/generated/mediaTypesV3Lro/src/mediaTypesV3LROClient.ts
@@ -224,7 +224,7 @@ const send$binaryOperationSpec: coreHttp.OperationSpec = {
   httpMethod: "POST",
   responses: { 204: {}, 400: {} },
   requestBody: Parameters.data,
-  queryParameters: [Parameters.excluded1],
+  queryParameters: [Parameters.excluded],
   urlParameters: [Parameters.$host, Parameters.thing],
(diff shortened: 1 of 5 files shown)


test/integration/generated/mediaTypesV3Lro/src/mediaTypesV3LROClient.ts
```diff
@@ -234,7 +234,7 @@ const send$textOperationSpec: coreHttp.OperationSpec = {
   httpMethod: "POST",
   responses: { 204: {}, 400: {} },
   requestBody: Parameters.data1,
-  queryParameters: [Parameters.excluded1],
+  queryParameters: [Parameters.excluded],
```
joheredi, line 237: This thing caught its first missing set of changes 😄

daviwil:
You should consider using `@autorest/compare` for this :)  Didn't get a chance to mention it before the PR got merged, but I can help with it at another time.

chradek:
Please tell us more about `@autorest/compare` @daviwil!

daviwil:
The short version is that it semantically diffs the generated output of AutoRest between two sets of input parameters.  I wrote it because using `git diff` is unreliable because it catches reordering of code output that has no real behavioral change.  Comparing the code on an AST level leads to less false positives.

joheredi:
Thanks @daviwil, @autorest/compare sounds great. I will try to move to use it soon, I have filed Issue #639 to track that work

PR #638, opened 2020-05-05
[v6] Clone Specs repo

joheredi:
I've noticed that sometimes our smoke-test script fails to read a file directly from github as we are doing right now.

We can get better reliability by cloning the whole specs repo and use swaggers from there. Also adding a way to target specific branch/commit for each swagger


test/utils/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const SPECS_PATH = "./.tmp/specs";
+export const DEFAULT_SPEC_BRANCH = "49fc16354df7211f8392c56884a3437138317d1f";
```
chradek, line 2: What made you decide to use this as the branch? Any reason not to choose `master` as the default?

package.json
```diff
@@ -17,6 +17,7 @@
         "stop-test-server": "stop-autorest-testserver",
         "debug": "node --inspect-brk ./dist/src/main.js",
         "generate-swaggers": "ts-node ./test/utils/test-swagger-gen.ts",
+        "clone:specs": "git clone https://github.com/Azure/azure-rest-api-specs.git ./.tmp/specs",
```
chradek, line 20: Does it make sense to do a shallow clone instead of a full clone to limit disk space usage/time to clone? You can use `--depth` to limit how many changes to pull down, and `--no-single-branch` to apply that depth to each branch. I don't know what happens though if you attempt to checkout a commit that's outside this depth though, so depends on what the behavior is with that, and if we only need to reference branches or arbitrary commits.

test/utils/constants.ts
joheredi: The problem with using master is that every run may potentially produce a different output, and we'll have a hard time maintaining CI and keeping it healthy. By locking to a point in time we won't be exposed to CI breaks for random changes, for example if we target master and a swagger updates its description for an Operation we'll end up with a broken CI.

I took this commit arbitrarily, I think it was the latest commit to master when I started adding smoke tests.

package.json
joheredi: Good point, let me play a bit with this and see how it works out

test/utils/constants.ts
chradek: Ah I see. That makes sense since the goal here is to make sure autorest.typescript doesn't regress, not that it supports the latest version of a model. Maybe someday we can have a bot that creates a PR with a new branch commit id for a spec when one of the models changes (so we can review/test it).

package.json
joheredi: I can do `depth=1` and `--single-branch` . And since we know the specific branch/commit we want we can fetch each of them. That would cut our disk usage from ~60MB to ~11MB

PR #640, opened 2020-05-06
[v6] Fix duplicate mappers and Response type name collisions

joheredi:
While trying to generate Table Storage swagger, I found that the generator has 2 issues:

1. When there are multiple success responses defined, we'd get duplicate mappers. This is similar to the bug fixed on PR #633. The difference is that in this scenario, the response types are defined inline instead of being references ([see here](https://github.com/Azure/azure-rest-api-specs/blob/f5cb6fb416ae0a06329599db9dc17c8fdd7f95c7/specification/cosmos-db/data-plane/Microsoft.TablesStorage/preview/2019-02-02/table.json#L137))

As in the previous fix, if we find an operation with multiple success responses but different mappers we throw an error as we don't  support that scenario yet.

2. There can be collisions between Response type names and properties inside the response. For example this operation is named `Table_Query` which would end up with an autogenerated response type `TableQueryResponse` however it's response's schema ([link](https://github.com/Azure/azure-rest-api-specs/blob/f5cb6fb416ae0a06329599db9dc17c8fdd7f95c7/specification/cosmos-db/data-plane/Microsoft.TablesStorage/preview/2019-02-02/table.json#L103)) is also named `TableQueryResponse`. In this case we end up with a name collision when generation the Response Body type and the Response Type

### DIFF
--- a/src/generators/utils/responseTypeUtils.ts
+++ b/src/generators/utils/responseTypeUtils.ts
@@ -1,0 +1,34 @@
+import { ClientDetails } from "../../models/clientDetails";
+import { normalizeName, NameType } from "../../utils/nameUtils";
+
+/**
+ * Helper function that gets a set of object model names,
+ * this can be used to check for possible duplicate names
+ * while generating other models
+ */
+export function getAllModelsNames(clientDetails: ClientDetails) {
+  return clientDetails.objects.reduce<Set<string>>((acc, { name }) => {
+    acc.add(name);
+    return acc;
+  }, new Set<string>());
+}
+
+/**
+ * Determines the correct name to use for a response type
+ * taking in consideration the already existing object models
+ * to avoid name collisions
+ * @param responseName The raw response name
+ * @param modelNames Set of existing model names
+ */
+export function getResponseTypeName(
+  responseName: string,
+  modelNames: Set<string>
+) {
+  const operationResponseName = normalizeName(responseName, NameType.Interface);
+
+  let typeName = `${operationResponseName}Response`;
+
+  return modelNames.has(typeName)
+    ? `${operationResponseName}OperationResponse`
+    : typeName;
+}
--- a/test/smoke/generated/msi-resource-manager/src/models/index.ts
+++ b/test/smoke/generated/msi-resource-manager/src/models/index.ts
@@ -24,14 +24,9 @@ export interface Resource {
 }
 
 /**
- * The resource model definition for a ARM proxy resource. It will have everything other than required location and tags
- */
-export type ProxyResource = Resource & {};
-
-/**
  * Describes a system assigned identity resource.
  */
-export type SystemAssignedIdentity = ProxyResource & {
+export type SystemAssignedIdentity = Resource & {
   /**
    * The geo-location where the resource lives
    */
@@ -211,6 +206,11 @@ export type IdentityUpdate = Resource & {
    */
   readonly clientId?: string;
 };
+
+/**
+ * The resource model definition for a ARM proxy resource. It will have everything other than required location and tags
+ */
+export type ProxyResource = Resource & {};
 
 /**
  * Contains response data for the getByScope operation.
--- a/test/smoke/generated/msi-resource-manager/src/models/mappers.ts
+++ b/test/smoke/generated/msi-resource-manager/src/models/mappers.ts
@@ -38,22 +38,12 @@ export const Resource: coreHttp.CompositeMapper = {
   }
 };
 
-export const ProxyResource: coreHttp.CompositeMapper = {
-  type: {
-    name: "Composite",
-    className: "ProxyResource",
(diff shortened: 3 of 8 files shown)


test/smoke/generated/msi-resource-manager/src/models/mappers.ts
```diff
@@ -38,22 +38,12 @@ export const Resource: coreHttp.CompositeMapper = {
...
-  }
-};
-
 export const SystemAssignedIdentity: coreHttp.CompositeMapper = {
   type: {
     name: "Composite",
     className: "SystemAssignedIdentity",
     modelProperties: {
-      ...ProxyResource.type.modelProperties,
+      ...Resource.type.modelProperties,
```
joheredi, line 46: @daviwil this change seems to be caused by the last version of M4. I thought the step to remove these empty parents was skipped for now. Do I need to update something in my autorest command?

This is the definition of ProxyResource ([link](https://github.com/Azure/azure-rest-api-specs/blob/e994a9b58f099f073dd60bba3fd53d7bd6e969a6/specification/common-types/resource-management/v1/types.json#L78))

```json
 "ProxyResource": {
      "description": "The resource model definition for a ARM proxy resource. It will have everything other than required location and tags",
      "allOf": [
        {
          "$ref": "#/definitions/Resource"
        }
      ]
    }
```

test/smoke/generated/msi-resource-manager/src/models/index.ts
```diff
@@ -23,15 +23,10 @@ export interface Resource {
...
-/**
- * The resource model definition for a ARM proxy resource. It will have everything other than required location and tags
- */
-export type ProxyResource = Resource & {};
-
 /**
  * Describes a system assigned identity resource.
  */
-export type SystemAssignedIdentity = ProxyResource & {
+export type SystemAssignedIdentity = Resource & {
```
joheredi, line 29: This change comes from taking the latest M4 build

test/smoke/generated/msi-resource-manager/src/models/mappers.ts
chradek: @joheredi 
Do you mean that the `ProxyResource` mapper shouldn't be generated, but it is because it's in the code model?

src/generators/utils/responseTypeUtils.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export function getResponseTypeName(
+  responseName: string,
+  modelNames: Set<string>
+) {
+  const operationResponseName = normalizeName(responseName, NameType.Interface);
+
+  let typeName = `${operationResponseName}Response`;
+
+  return modelNames.has(typeName)
```
chradek, line 31: Darn, it's kind of a shame that some operations won't have a consistently-named response type. 😢

joheredi: Yeah, I know. Autorest v5 used to suffix these with '2'. I thought that OperationResponse would make more sense

joheredi:
> Looks good @joheredi
> If I understand correctly, the ProxyResource still showing up isn't a blocker is it? It looks like it's extra stuff we don't need, but also doesn't break anything?

Yes, not blocking and not breaking anything. Just wanted to bring this up to make sure this is expected from latest M4 changes

test/smoke/generated/msi-resource-manager/src/models/mappers.ts
joheredi: Originally I thought that ProxyResource shouldn't have been removed. But it looks like I was wrong 😄. I think M4 removing it is expected with the last version.

Some background on why I thought ProxyResource shouldn't have been removed, feel free to ignore:

So there was some logic added in M4 recently that removes parents if the only thing they do is extend from another object.

In a recent PR, David mentioned that this logic was going to be disabled ([link](https://github.com/Azure/autorest.modelerfour/pull/274#issuecomment-624300730)), so I thought the removal of ProxyResource might've been unintended.

However I clearly didn't reach the bottom of the PR,  where David said this logic would be partially enabled instead of completely disabled ([link](https://github.com/Azure/autorest.modelerfour/pull/274#issuecomment-624329254)).

leonklingele:
With this patch applied I get the following error: `An error was encountered while handling a request: Error: There are different mapper definitions with the same name. This may be caused by a Swagger defining multiple Success responses with different return types`

Swagger file to reproduce:

```
swagger: "2.0"
info:
  title: demo
  version: 0.1.0
paths:
  /api:
    head:
      operationId: Demo
      responses:
        "200":
          description: OK
          headers:
            Cache-Control:
              type: string
              default: "no-store"
        "403":
          description: Forbidden
          headers:
            Cache-Control:
              type: string
```

It works fine when I add the `default: "no-store"` to the 403 response as well.

PR #642, opened 2020-05-08
[v6] Refactor parent and additionalProperties processing

joheredi:
**Problem**

Since M4 adds a Dictionary Schema to the parents collection when additionalProperties is defined, we were failing to identify uber parents correctly.

This bug surfaced while trying to generate Monitor arm swagger, which has a top level (uberParent) object with additionalProperties.

In this scenario, the our processing logic didn't identify such object as an uberParent since it had one parent, the Dictionary for additional properties.

**Fix**

Refactor how we figure out uberParents and additionalProperties by only taking into consideration parents that are defined as ObjectSchema

Similarly, when figuring out additional properties, we need to only get look at parents that are DictionaySchemas.

In the process I removed a duplicate uberParent tracking collection and reused the one we build during ObjectTransforms.

This change will unblock generation of the following swaggers:

https://github.com/Azure/azure-rest-api-specs/blob/master/specification/cosmos-db/resource-manager/readme.md

https://github.com/Azure/azure-rest-api-specs/blob/master/specification/graphrbac/data-plane/readme.md


src/transforms/mapperTransforms.ts
```diff
@@ -281,21 +291,25 @@ function transformObjectMapper(pipelineValue: PipelineValue) {
...
-  const immediateParents = objectSchema.parents
-    ? objectSchema.parents.immediate
-    : [];
+  const parents = getSchemaParents(objectSchema);
+  const immediateParents = getSchemaParents(
+    objectSchema,
+    true /** immediateOnly */
+  );
   const parentsRefs = immediateParents
+    .filter(p => p.type === SchemaType.Object)
```
chradek, line 300: I don't think this filter is needed because it's already being done by `getSchemaParents`.

src/transforms/mapperTransforms.ts
```diff
@@ -264,10 +276,8 @@ function getAdditionalProperties(allParents: Schema[]): Mapper | undefined {
 }
 
 function isUberParent(objectSchema: ObjectSchema) {
```
chradek, line 278: I've always been a little unclear on the uber parent. Is this just any ObjectSchema that has no parents but is extended?

src/transforms/mapperTransforms.ts
```diff
@@ -716,6 +730,7 @@ export function getMapperTypeFromSchema(type: SchemaType, format?: string) {
     case SchemaType.Number:
       return MapperType.Number;
     case SchemaType.Object:
+    case SchemaType.Any:
```
chradek, line 733: Hmm, should this be the "any" mapper type instead of "object"? They get serialized differently:
https://github.com/Azure/azure-sdk-for-js/blob/1e02861f3abec278cd5b3629b5737760727878ad/sdk/core/core-http/src/serializer.ts#L127-L130

That would mean adding "any" to the `MapperType` enum in core-http:
https://github.com/Azure/azure-sdk-for-js/blob/1e02861f3abec278cd5b3629b5737760727878ad/sdk/core/core-http/src/serializer.ts#L1036
but we could test the impact of this by returning the string "any" in the meantime.

src/transforms/mapperTransforms.ts
joheredi: Yes you're right I'll get rid of this one

src/transforms/mapperTransforms.ts
joheredi: Yes, that's an uber parent

joheredi: I'm adding that to the description when we extract Uber Parents so that it is clear

src/transforms/mapperTransforms.ts
joheredi: Good catch, try that out. 

It like core-http is exporting MapperType twice: https://github.com/Azure/azure-sdk-for-js/blob/1e02861f3abec278cd5b3629b5737760727878ad/sdk/core/core-http/src/serializer.ts#L901

joheredi: Thanks for the idea Chris, I have moved to use "any" string instead of Object

chradek: Yeah, not sure why that's the case, but they have very different shapes. I got really confused when I first saw that.

src/transforms/mapperTransforms.ts
joheredi: Oh sorry forgot to commit this change, pushing now

joheredi:
Thanks for the review @chradek! Before merging, I'll be pushing to this PR the generated clients under smoke-tests for the following management libraries:
- CosmosDB
- GraphRBAC
- Monitor

PR #643, opened 2020-05-13
[v6] Fix missing client imports

joheredi:
We're missing some named imports for client parameters in Client and ClientContext which causes build errors.

This problem surfaces particularly when the Client has Enum or Object parameters that require referencing the Models.

We were also missing a case for handling ShcemaType.Choice

This issue surfaced while generating cosmosdb data-plane swagger

### DIFF
--- a/src/generators/clientContextFileGenerator.ts
+++ b/src/generators/clientContextFileGenerator.ts
@@ -22,6 +22,7 @@ export function generateClientContext(
   packageDetails: PackageDetails,
   project: Project
 ) {
+  const importedModels = new Set<string>();
   const clientParams = clientDetails.parameters.filter(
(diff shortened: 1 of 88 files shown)


src/generators/clientContextFileGenerator.ts
```diff
@@ -64,23 +65,31 @@ export function generateClientContext(
...
     moduleSpecifier: "@azure/core-http"
   });
 
-  sourceFile.addImportDeclaration({
-    namespaceImport: "Models",
-    moduleSpecifier: "./models"
-  });
+  if (importedModels.size) {
+    sourceFile.addImportDeclaration({
+      namedImports: [...importedModels],
```
chradek, line 89: Oh that's cool, I didn't realize you could use spread notation with a `Set`.

PR #644, opened 2020-05-14
[v6] Handle credential schema

joheredi:
**Problem**
We don't have support for strings with `format: "password"`. Represented in M4 as CredentialSchema

According to OpenAPI 3. A string with `format: "password"` is used to

> A hint to UIs to obscure input.
[source](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#data-types)

**Fix**
- Handle SchemaCredential mapper transform as we do for regular strings
- When generating a parameter or property with this schema, add a hint to obscure as suggested in OpenAPI spec. As a comment in the generated code.


src/utils/schemaHelpers.ts
```diff
@@ -179,3 +180,18 @@ export function isSchemaResponse(response: any): response is SchemaResponse {
   // check fields that should exist to determine if this is a SchemaResponse
   return typeof response.schema !== "undefined";
 }
+
+/**
+ * Gets special information to include as documentation when generating certain
+ * schema types.
+ */
+export function getSchemaTypeDocumentation(schema: Schema) {
```
chradek, line 188: Good idea pulling this into a separate function.

joheredi:
> Looks good! Is this going to be covered in one of the smoke clients?

Yes I'll be adding[ Web management swagger]( https://github.com/Azure/azure-rest-api-specs/blob/master/specification/web/resource-manager/readme.md ) to the smoke tests which uses everything that this PR fixes

Thanks for the review!

PR #646, opened 2020-05-16
[v6] Add SQL and ARM to smoke test

joheredi:
- Adding SQL and ARM swaggers to smoke tests
- Allow to select which smoke-tests to run using "tags" to group swaggers together
- Split CI Smoke-Tests in 3 groups to run in parallel and keep CI run time around 10 minutes


joheredi:
@chradek, @daviwil yeah I was arbitrary and grouped them by trial and error to get similar run times. It may get impractical if we want to have many more swaggers as part of Smoke-Tests, we'll get to a point when I think the best bet would be to have a nightly run with bigger buckets, and keeping the CI groups to a core set of Swaggers

PR #647, opened 2020-05-19
 [v6] add mediaType to OperationSpecs and make param ordering deterministic

chradek:
Depends on https://github.com/Azure/azure-sdk-for-js/pull/8977

## Description of changes
This update makes 2 changes.
1. Generated `OperationSpec`s that have a request body will now also have a `mediaType` field that provides a hint for how the request body should be serialized in core-http.
2. Parameter ordering now matches the ordering present in the `CodeModel`.

## Why is `mediaType` needed?
mediaType describes the type of the request body in a more generic way than contentType.
For example, some valid values for mediaTypes would be "binary", "xml", "json", "text".

This change adds a check to see if the mediaType is text, and serializes the request body the same as if it had the contentType text/plain.

## Why do we care about parameter ordering?
One of our goals is to maintain parameter ordering across multiple runs and minor versions of `@autorest/typescript`. This is important so we can ensure that regenerating a client with either a new version of autorest.typescript or an updated, backwards-compatible model doesn't break existing operations.

## What was the problem with the current method for parameter ordering?
Currently when autorest.typescript gets the parameters for an operation, it gets them from a list of all parameters and filters out the ones that don't apply to the current operation.
This causes a problem when parameters are deduplicated.

For example, if 2 operations use the same parameter, that parameter only exists in the global list of parameters once. When getting the list of parameters for an operation overload, the ordering of parameters depends on their position in the global list. This means if a new operation is added that uses the same parameter as an existing operation, the existing operation may now have a different parameter ordering than it did previously.

## Ok...so how do we fix it?
This change determines what the parameter ordering for an operation request should be, then sorts the list of parameter definitions we generate to match. The order should now always match the following from the CodeModel:
1. operation.parameters: this is an ordered list of parameters that are shared across operation requests (e.g. they are not part of the request body).
2. operation.request[#].parameters: this is an ordered list of parameters specific to an operation request (method overload).

## What about grouped parameters?
~~Grouped parameters will currently be placed after required non-grouped parameters, but before the options bag.~~
Turns out required group parameters do have an entry in the operation.request,parameters list, so no special action is needed to get them properly sorted as well. I just had to move the sort operation to be after the grouped parameters are included, rather than before,


src/generators/operationGenerator.ts
```diff
@@ -482,11 +492,42 @@ function getOperationParameterSignatures(
       return acc;
     }, []);
 
+    // sort parameters to match the signature ordering
```
joheredi, line 495: Awesome! Thanks for doing this 😄!

Maybe it would help readability if this is extracted to its own `orderParameters` function

chradek: Good suggestion, I'll update this.

chradek: Updated! I also moved the sort call to be after the required parameter groups are included. Turns out they do appear in the operation.request[x].parameters list, so the same logic works for the groups as well.

PR #649, opened 2020-05-21
[v6] dedupe parameter mappers

chradek:
## Summary
This update changes the logic around how parameters are deduped to no longer take into account parameter descriptions (docs).

## What was happening before?
Previously, we treated parameters as unique based on a subset of their fields, including description. The description was included because a parameter may have different documentation depending on where it exists in an API model.

With this change, we no longer use the description field to determine if a parameter is 'unique' compared to another parameter.

## Well wait, won't that impact doc generation?
The `ParameterDetails.operationsIn` field has been updated to account for this. Previously, `operationsIn` was a list of strings describing which operations a parameter could be found in.

With this change, `operationsIn` has been changed to a map, where they key is the operation name, and the value is an object that contains the `description`. Now when a parameter description is needed, we first check if the description exists in the `operationsIn` map, and if not, fall back to the parameter's top-level `description` field (which is still used for non-method parameters).

## Testing
I've generated the integration test clients and ensured that documentation has stayed the same, the integration tests still pass, and duplicated parameters have been removed.

I've also regenerated the smoke test clients and observed the same from a sampling of the files.


src/models/parameterDetails.ts
```diff
@@ -23,7 +23,7 @@ export interface ParameterDetails {
   mapper: string | Mapper;
   isGlobal: boolean;
   parameter: Parameter;
-  operationsIn?: string[];
+  operationsIn?: { [operationName: string]: { description: string } };
```
joheredi, line 26: This is cool :)

PR #650, opened 2020-05-21
[v6] Add httpInfrastructure test

joheredi:
* Generate HttpInfrastructure test client and add tests
* Re-enable some skipped tests
* Add more Paging tests


test/integration/bodyArray.spec.ts
```diff
@@ -601,9 +601,8 @@ describe("Integration tests for BodyArrayClient", () => {
     });
   });
 
-  describe.skip("#getBooleanInvalidString", () => {
+  describe("#getBooleanInvalidString", () => {
     it("returns [true, 'boolean', false]", async () => {
-      // TODO: Should throw deserialization error
```
chradek, line 606: Are these skipped tests supposed to be throwing errors? (I think I might have left those TODOs originally). If not then nevermind!

test/integration/paging.spec.ts
```diff
@@ -66,6 +181,46 @@ describe("Integration tests for Paging", () => {
...
+        await client.paging.getMultiplePagesFailureNext(result.nextLink!);
+      } catch (error) {
+        assert.equal(error.statusCode, 400);
+      }
+    });
+
+    it("should fail on invalid next link URL in multiple pages", async () => {
+      try {
+        const result = await client.paging.getMultiplePagesFailureUri();
+        await client.paging.getMultiplePagesFailureUriNext(result.nextLink!);
```
chradek, line 197: For the tests that expect a failure, we should either do an assert.fail (I think that's a real method) after the operation is invoked but before the catch, or throw an error after the operation is called (e.g. `throw new Error('Test failure')`) and check that the error.message doesn't match 'Test failure' in the catch block.

Otherwise, if the operation doesn't throw an error the assert in the catch never gets ran and the test passes (false positive).

Edit: I see you did that for the httpInfrastructure tests, I think you just need to do the same for the paging tests.

test/integration/httpInfrastructure.spec.ts
```diff
@@ -0,0 +1,736 @@
...
+        assert.equal(error.statusCode, 400);
+      }
+    });
+  });
+
+  describe("httpClientFailure scenarios", () => {
+    it("delete400 should throw error", async () => {
+      try {
+        await client.httpClientFailure.delete400();
+        assert.fail("Expected error");
```
chradek, line 152: Ah, assert.fail does exist :)

test/integration/bodyArray.spec.ts
joheredi: Discussed offline. If we want to guard against arrays with elements that don't match what is defined in the type, for example here, `getBooleanInvalidString's` return type is an `Array<boolean>` but TestServer sends an array with booleans and a string.

This would need to be verified at runtime, since these arrays are valid in Javascript.

I've created an issue in `core-http` to discuss implementing this

test/integration/paging.spec.ts
joheredi: Good catch! 😄

joheredi: Found a few other places like this and fixed as part of this PR

PR #653, opened 2020-06-03
[v6] Add test coverage for Multiple inheritance

joheredi:
Adding test coverage for Multiple Inheritance test server scenarios


test/integration/multipleInheritance.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+
+describe("MultipleInheritance", () => {
+  let client: MultipleInheritanceClient;
+
+  beforeEach(() => {
+    client = new MultipleInheritanceClient();
+  });
+
+  it("should get a pet", async () => {
+    const { _response, ...result } = await client.getPet();
```
chradek, line 13: Ah, that's a good trick. I hadn't thought of doing that when I needed to write similar tests and went with a more complicated approach.

joheredi: It just occurred to me today, I was also doing other complicated stuff. I think I'll stick with this in future tests

PR #657, opened 2020-06-08
[v4] Make sure that the interface also extends the base model when generat…

joheredi:
Port of #616 

Make sure that the interface also extends the base model when generating PageCompositeTypes.

For example in [this swagger](https://github.com/Azure/azure-rest-api-specs/blob/0c0880833da9d695140d140071aea579fdc3da38/specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/stable/2019-06-15/bms.json#L3610-L3627) `ResourceList` is dropped when generating `JobResourceList`

```typescript
export interface JobResourceList extends Array<JobResource> {
}
```

Where it should generate the following in order to honor the swagger definition
```typescript
export interface JobResourceList extends Array<JobResource>, JobResourceList {
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft.azure/autorest.typescript",
-  "version": "4.4.0",
+  "version": "4.3.1",
   "description": "The typescript extension for classic generators in AutoRest.",
   "scripts": {
     "autorest": "autorest",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft.azure/autorest.typescript",
-  "version": "4.3.0",
+  "version": "4.4.0",
   "description": "The typescript extension for classic generators in AutoRest.",
   "scripts": {
     "autorest": "autorest",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft.azure/autorest.typescript",
-  "version": "4.2.4",
+  "version": "4.3.0",
   "description": "The typescript extension for classic generators in AutoRest.",
   "scripts": {
     "autorest": "autorest",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft.azure/autorest.typescript",
-  "version": "4.2.4",
+  "version": "4.3.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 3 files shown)


chradek:
I'm curious, why is the interface
```ts
export interface JobResourceList extends Array<JobResource>, ResourceList {
}
```
instead of
```ts
export interface JobResourceList extends ResourceList {
  value:  Array<JobResource>
}
```
Since `JobResourceList` is modeled as an object I would have thought it would look like the latter.
Note: I assumed you meant `JobResourceList` to extend `ResourceList`, not itself :smile:

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft.azure/autorest.typescript",
-  "version": "4.2.4",
+  "version": "4.4.0",
```
joheredi: Had to make it 4.3.1 since 4.3.0 failed with 
image

joheredi:
> I'm curious, why is the interface
> 
> ```ts
> export interface JobResourceList extends Array<JobResource>, ResourceList {
> }
> ```
> 
> instead of
> 
> ```ts
> export interface JobResourceList extends ResourceList {
>   value:  Array<JobResource>
> }
> ```
> 
> Since `JobResourceList` is modeled as an object I would have thought it would look like the latter.
> Note: I assumed you meant `JobResourceList` to extend `ResourceList`, not itself 😄

Right now the pattern that Autorest.Typescript follows for pageable responses is that it flattens the response object so that the Response type is actually the array of values with an extra property `nextLink`.

In the scenario where `nextLink` is a property on the object itself i.e. `JobResourceList` `nextLink` would've been included as

```typescript
export interface JobResourceList extends Array<JobResource> {
   nextLink?: string;
}
```

Allowing users to consume it as an array like this:
```typescript
const x: JobResourceList = await getJobResourceList();

do {
   x.forEach(jobResource => console.log(jobResource ));
} while(x.nextLink)
```

However in this scenario, since `nextLink` is a property in a base class it was being ignored, because Autorest was not considering it's parent properties. So this change is trying to honor the inheritance defined in the Swagger.

I agree that in general it would look cleaner to represent the interface in the way you described above, however in the specific scenario of pageable operations, I think it makes sense to have the values property flattened. This change only impacts Pageable operations.

@chradek does it make sense?

chradek:
@joheredi 
Yes, that makes. I think the part that tripped me up was that the x-ms-pageable extension is on the operation and changes the shape of the referenced type when that type is the response.

PR #659, opened 2020-06-09
[v4x] Polymorphic uber parent

joheredi:
In short the problem seems to be in the way we currently figure out the `uberParent` when building the `discriminators` object, which is used by `core-http` to select the mapper to use for serialization.

This `discriminators` object is a dictionary with key `{uberParent}.{discriminatorValue}` and a Mapper  as a value. When dealing with an operation `core-http` knows the `uberParent` or `base` object and needs to figure out which descendant it is dealing with at run time based on the value of a property defined as `discriminator` in the `uberParent`.

So the problem seems to be that in Autorest we're building the dictionary and assume that always the **topLevel parent** is the one which defines the polymorphic discriminator. However this is not always true. We could have this hierarchy

image

In which the following dictionary would be built

```typescript
// Current, incorrect
const discriminators = {
  "A.Ba": BaMapper,
  "A.Bb": BbMapper,
  "A.Ca": CaMapper,
  "A.Cb": CbMapper
}
// Fixed, correct
const discriminators = {
  "B.Ba": BaMapper,
  "B.Bb": BbMapper,
  "C.Ca": CaMapper,
  "C.Cb": CbMapper
}
```

The problem becomes more evident when the extension `x-ms-discriminator-value`

image

```typescript
// Current, incorrect
const discriminators = {
  "A.Foo": BaMapper,
  "A.Bar": BbMapper, // Duplicate Key makes it an invalid object :(
  "A.Foo": CaMapper,
  "A.Bar": CbMapper
}

// Fixed, correct
const discriminators = {
  "B.Foo": BaMapper,
  "B.Bar": BbMapper,
  "C.Foo": CaMapper,
  "C.Bar": CbMapper
}
```
**Fix**
So my fix is basically to stop crawling the parents when we find one that defines a polymorphic discriminator.

### DIFF
--- a/src/vanilla/ClientModelExtensions.cs
+++ b/src/vanilla/ClientModelExtensions.cs
@@ -88,7 +88,8 @@ namespace AutoRest.TypeScript
                 return "string";
             else if (primary.KnownPrimaryType == KnownPrimaryType.Credentials)
                 return "msRest.ServiceClientCredentials"; //TODO: test this, add include for it
-            else {
+            else
+            {
                 throw new NotImplementedException($"Type '{primary}' not implemented");
             }
         }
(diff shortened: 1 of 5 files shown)


src/vanilla/ClientModelExtensions.cs
```diff
@@ -637,6 +635,24 @@ void applyConstraints(TSObject obj)
             }
         }
 
+
+        /// <summary>
+        /// Finds the UberParent for a given Composite type.
```
chradek, line 640: Thanks for adding a description around this!

PR #664, opened 2020-06-10
New pipeline

joheredi:
Changes to run in the new Pipeline:

* Explicitly indicate PR triggers
   * Used to be a GUI override

* Publish package.json to artifacts, in order to simplify publishing script.
* Disable package version check
* Add releaseTag

### DIFF
--- a/.scripts/checkPackageJsonVersion.ts
+++ b/.scripts/checkPackageJsonVersion.ts
@@ -1,3 +1,0 @@
-import { checkPackageJsonVersion } from "@ts-common/azure-js-dev-tools";
-
-checkPackageJsonVersion();
--- a/.scripts/checkPackageJsonVersion.ts
+++ b/.scripts/checkPackageJsonVersion.ts
@@ -1,3 +1,3 @@
 import { checkPackageJsonVersion } from "@ts-common/azure-js-dev-tools";
 
-checkPackageJsonVersion();
+// checkPackageJsonVersion();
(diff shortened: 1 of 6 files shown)


azure-pipelines.yml
```diff
@@ -1,5 +1,14 @@
 trigger:
-- master
+  batch: true
+  branches:
+    include:
+    - master
+
+pr:
```
joheredi, line 7: Added this to avoid overriding settings in the AzureDev Ops GUI. This will make the pipeline trigger when there is a PR on master or v4x

.scripts/checkPackageJsonVersion.ts
```diff
@@ -1,3 +1,3 @@
 import { checkPackageJsonVersion } from "@ts-common/azure-js-dev-tools";
```
daviwil, line 1: Maybe we can get rid of this now that you have logic baked into the script?

azure-pipelines.yml
daviwil: If we include this pipeline for PRs, don't we run the risk of accidentally shipping builds from PRs (like I did recently on `ms-rest-nodeauth`)?

joheredi: If we set up auto publish yes. I've disabled auto publish for this package. But this is a very good point, I think we may want to have separate pipelines for CI (merge) and PR and just have CI plugged in to the auto release pipeline

PR #678, opened 2020-07-07
[v6] Split Mappers into their own files

joheredi:
**The change**
Before this PR all mappers have been generated into a single `mappers.ts` file. This change breaks mappers down into several files each one containing a single mapper.  These mappers are placed under the mappers which contains an `index.ts` file exporting all the mappers and a `discriminators.ts` file, which defines the polymorphic discriminator map to be used by `core-http` (this was also part of the `mappers.ts`)

Assuming a swagger that defines 3 Objects `Pet`, `Cat` and `Dog`, this is how the folder structure looks like

**Before this change**
```
src/
   models/
      mappers.ts
      index.ts
      parameters.ts
      etc...
```

**After this change**
```
src/
   mappers/
      index.ts
      discriminators.ts
      pet.ts
      cat.ts
      dog.ts
   models/
      index.ts
      parameters.ts
      etc...
```

**Motivation**
The motivation for this change comes from an issue that surfaced when upgrading to the latest ModelerFour build. In which the order of some mappers was slightly modified, causing Typescript errors if those changes resulted in using a type before it was defined.

@bterlson, @xirzec - I discussed this change with @daviwil and @chradek offline, and we think it is a good approach since it will result in a more robust generation, since we no longer make assumption on how the Swagger orders its definitions. On the other hand, this will result in the generation of additional files which can get large depending on the amount of objects defined in the Swagger. After bundling this is probably not going to be an issue, however we wanted to loop you guys in to get your thoughts


joheredi:
Talked to @bterlson offline, his main concern is around the potentially large number of files generated, it could easily get around a few hundreds just for mappers, and the impact it would have on build time for the generated sdk. I'm going prototype a the approach of generating the mappers in the same file in topographical order as he suggested, depending on its complexity we can make a call on which approach to take

PR #689, opened 2020-07-08
[v6] Generate mappers in hierarchical order

### LINKED ISSUE #687: [v6] Fix mapper generation ordering issue
The latest version of ModelerFour changed the order of some of the objects, this causes Typescript Generator to sometimes generate a child type before a parent, causing a Typescript error because of referencing a type not yet defined. There are 2 approaches to fix this

1. Break down the `mappers.ts` file so that each mapper has its own file
   - Good because it doesn't run into problems with cyclic references and generation is straightforward
   - No-so-good because it may create a high number of Files, bounded by the number of Objects defined in the Swagger, which can lead to a few hundreds of files generated just for mappers. The main concern with too many files is the impact it would have when compiling the generated SDK
2. Generate mappers in Topographical order
   - Good because we get a constant number of files generated
   - Not-so-good because generation is not as straightforward and need to handle cyclic references

joheredi:
Make sure mappers are generated in hierarchical order to avoid referencing parents before they have been defined.

This change unblocks consuming the latest M4 build.

This PR also updates the following dependencies:

```
@azure-tools/autorest-extension-base
@azure-tools/codegen
@azure-tools/codemodel
@microsoft.azure/autorest.testserver
```

Fixes #687 

/cc @bterlson - This is a different approach to generating a file per mapper, no extra files generated. The logic isn't too complex. This doesn't handle cyclic references, since I didn't find any swaggers with that in TestServer or our Smoke Tests, I guess we can tackle that when/if we run into a Swagger like that. (If you know one, I'll be happy to look at it)


src/utils/sortObjectSchemasHierarchically.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    const { name } = getLanguageMetadata(current.language);
+
+    if (processed.has(name)) {
+      continue;
+    }
+
+    result.push(current);
+    processed.add(name);
```
daviwil, line 46: I believe this strategy will work fine for handling recursive schemas, it's similar to what we're doing in m4

src/utils/sortObjectSchemasHierarchically.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    const currentChildren = !current.children
+      ? []
+      : (current.children.immediate.filter(
+          c => c.type === SchemaType.Object
+        ) as ObjectSchema[]);
+
+    if (currentChildren.length) {
+      stack.push(...currentChildren);
```
daviwil, line 55: This should work fine without the `currentChildren.length` check I think?

PR #690, opened 2020-07-09
[v6] Use normalized header name

joheredi:
Use the name provided by M4 to leverage naming functionality such as `x-ms-client-name` processing.

This PR depends on Azure/autorest.modelerfour/pull/318


src/utils/headersToSchema.ts
```diff
@@ -14,14 +20,23 @@ export function headersToSchema(
...
   );
 
-  headers.forEach(({ header, schema }) => {
+  headers.forEach(({ header, language, schema }) => {
     if (!headersSchema.properties) {
       headersSchema.properties = [];
     }
 
-    const { description } = getLanguageMetadata(schema.language);
+    if (!language) {
```
daviwil, line 28: Now `language` is mandatory so you won't need to check for it anymore :)

joheredi: Nice! Thanks!

PR #712, opened 2020-08-11
[v6] Extract headers for Exceptions

joheredi:
A couple of errors surfaced when generating service bus. There are 2 problems:

* Failing to correctly detect xmlSerializer
* Not generating header mappers for Exception/Error responses

### DIFF
--- a/src/generators/operationGenerator.ts
+++ b/src/generators/operationGenerator.ts
@@ -958,7 +958,7 @@ function hasMediaType(
   operationDetails: OperationDetails,
   mediaType: KnownMediaType
 ) {
-  if (!operationDetails.requests) {
+  if (!operationDetails.requests.some(r => !!r.mediaType)) {
     return operationDetails.mediaTypes.has(mediaType);
   }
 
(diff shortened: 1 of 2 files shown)


src/generators/operationGenerator.ts
```diff
@@ -958,7 +958,7 @@ function hasMediaType(
   operationDetails: OperationDetails,
   mediaType: KnownMediaType
 ) {
-  if (!operationDetails.requests) {
+  if (!operationDetails.requests.some(r => !!r.mediaType)) {
```
joheredi, line 961: requests is never undefined so this check was never true. Checking for the mediaType field inside request instead.

src/utils/extractHeaders.ts
```diff
@@ -9,8 +9,12 @@ export function extractHeaders(
   let responseHeaders: ObjectSchema[] = [];
 
   operationGroups.forEach(operationGroup =>
-    operationGroup.operations.forEach(operation =>
-      operation.responses?.forEach(response => {
+    operationGroup.operations.forEach(operation => {
+      const responsesAndExceptions = [
+        ...(operation.responses || []),
```
joheredi, line 14: We used to just process success responses. However Error responses can also contain headers

Issue #688, opened 2020-07-07
[v6] Integrate ApiExtractor


joheredi:
Add a flag to the generator to integrate APIExtractor.

This will probably require #684 to be fixed

joheredi:
This has been completed

PR #713, opened 2020-08-11
[v6] Integrate API Extractor

joheredi:
Integrating Api Extractor required the following changes

* Generate `api-extractor` config file
* Add api-extractor as a dependency
* Change entry point and exports from `libraryNameClient.ts` to `index.ts`
* Change to named imports on Operations

**Bonus!**
Switch to use sequential generation of our Integration Test swagger. Before we have been spawning several processes for each swagger which sometimes causes some locking issues. Also measured the time CI `npm test` takes before and after and the time is cut by half


joheredi:
Closes #688

Issue #721, opened 2020-09-04
[v4] Update readme with link to ms-rest-nodeauth and a note on other login options


ramya-rao-a:
The readme of the packages generated using the code generator have the below information around authentication
- Installation instruction for the ms-rest-nodeauth package: `npm install @azure/ms-rest-nodauth`
- Code snippet that shows how to use the `interactiveLogin()` method in the `ms-rest-nodeauth` package

What is missing here is information on the other options for login leading one to believe that interactive login is the only option.

We should add a note with a link to the ms-rest-nodeauth readme which has details on other options. Something like the below just before the node code snippet

> While the below sample shows how to authenticate using interactive login, other options are described in the [readme for @azure/ms-rest-nodeauth](https://www.npmjs.com/package/@azure/ms-rest-nodeauth) package.

Related to https://github.com/Azure/azure-sdk-for-js/issues/10993

ramya-rao-a:
cc @joheredi, @sarangan12, @deyaaeldeen

deyaaeldeen:
I did some investigation and found that the old version of the code generator creates this section in the README.md file [here](https://github.com/deyaaeldeen/autorest.typescript/blob/master/src/vanilla/Model/CodeModelTS.cs#L1068). However, the new version of the code generator (v6) does something different, does not inline instructions about authentication nor shows a sample. Instead it links to a broken link [here](Refer the sample code in the [azure-sdk-for-js-samples](https://github.com/Azure/azure-sdk-for-js-samples) repository.). I think we should be consistent here between the two versions and perhaps regenerate that sdk using v6.

ramya-rao-a:
@deyaaeldeen The entire set of management libraries (the ones whose name start with `@azure/arm-`) are on v4 of the code generator. There are no current plans, atleast for this year, to upgrade them to use v6 which introduces new dependencies for auth.

Therefore, for the time being, it is still beneficial for the v4 code generator to update the readme with the above content.

sarangan12:
I will update the v6 code generator today, @deyaaeldeen If you prefer you can take the v4 generator. Else, I will be happy to change the v4 gen also.

deyaaeldeen:
@ramya-rao-a Thanks for the clarification!

@sarangan12 Sounds good!

PR #722, opened 2020-09-09
Modify Readme to include Authentication Details

sarangan12:
This PR is to address Readme changes listed in Issue https://github.com/Azure/autorest.typescript/issues/721

@ramya-rao-a Please review and approve

### DIFF
--- a/src/generators/static/readmeFileGenerator.ts
+++ b/src/generators/static/readmeFileGenerator.ts
@@ -27,10 +27,12 @@ npm install ${packageDetails.name}
 \`\`\`
 
 ### How to use
+#### Authentication
+One of the first steps in accessing the services using SDK is to authenticate the client. There are several methods to authenticate such as interactive login, etc. Refer [readme for @azure/ms-rest-nodeauth](https://www.npmjs.com/package/@azure/ms-rest-nodeauth) package for all options to authenticate the client.
 
 #### Sample code
 
-Refer the sample code in the [azure-sdk-for-js-samples](https://github.com/Azure/azure-sdk-for-js-samples) repository.
+Refer the sample code in the [azure-sdk-for-js](https://github.com/Azure/azure-sdk-for-js/tree/master/samples) repository.
 
 ## Related projects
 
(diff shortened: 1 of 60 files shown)


deyaaeldeen (approved):
Looks good!

joheredi (approved):

src/generators/static/readmeFileGenerator.ts
```diff
@@ -27,10 +27,12 @@ npm install ${packageDetails.name}
 \`\`\`
 
 ### How to use
+#### Authentication
+One of the first steps in accessing the services using SDK is to authenticate the client. There are several methods to authenticate such as interactive login, etc. Refer [readme for @azure/ms-rest-nodeauth](https://www.npmjs.com/package/@azure/ms-rest-nodeauth) package for all options to authenticate the client.
```
ramya-rao-a, line 31: This PR is against the v6 branch.

From my understanding the packages generated using v6 of the code generator use `TokenCredential` and therefore a user will have to bring in `@azure/identity`. Here, we are talking about `@azure/ms-rest-nodeauth` which is for the packages generated using v4 of the code generator.

ramya-rao-a: The issue https://github.com/Azure/autorest.typescript/issues/721 is mainly for packages generated using the v4 code generator. The packages generated using the v6 generator are getting the convenience layer and hand written readmes anyway and so does not have this immediate need.

sarangan12: @ramya-rao-a  Does the below line makes sense? 

```
One of the first steps in accessing the services using SDK is to authenticate the client. There are several methods to authenticate such as DefaultAzureCredential, etc. Refer [readme for @azure/identity](https://www.npmjs.com/package/@azure/identity) package for all options to authenticate the client.
```

sarangan12:
Based on the above conversation, I dont think this change is required anymore. Closing this for now

deyaaeldeen:
@sarangan12 I think we still should at least fix the link to the samples.

PR #725, opened 2020-09-09
Adding Auth information to README

### LINKED ISSUE #721: [v4] Update readme with link to ms-rest-nodeauth and a note on other login options

deyaaeldeen:
Fixes #721.

### DIFF
--- a/src/vanilla/Model/CodeModelTS.cs
+++ b/src/vanilla/Model/CodeModelTS.cs
@@ -1065,7 +1065,12 @@ namespace AutoRest.TypeScript.Model
 
         private void GenerateHowToUseInNodeJs(MarkdownBuilder builder)
         {
-            builder.Section($"nodejs - Authentication, client creation and {GetSampleMethod()?.Name} {GetSampleMethodGroupName()} as an example written in TypeScript.", () =>
+            builder.Section($"Authentication", () => 
+            {
+                builder.Section("One of the first steps in accessing the services using SDK is to authenticate the client. There are several methods to authenticate such as interactive login, etc. Refer [readme for @azure/ms-rest-nodeauth](https://www.npmjs.com/package/@azure/ms-rest-nodeauth) package for all options to authenticate the client.");
+                
+            });
+            builder.Section($"nodejs - client creation and {GetSampleMethod()?.Name} {GetSampleMethodGroupName()} as an example written in TypeScript.", () =>
             {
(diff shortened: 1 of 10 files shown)


joheredi (approved):

src/vanilla/Model/CodeModelTS.cs
```diff
@@ -1065,7 +1065,13 @@ private void GenerateAvailableAPIVersions(MarkdownBuilder builder)
 
         private void GenerateHowToUseInNodeJs(MarkdownBuilder builder)
         {
-            builder.Section($"nodejs - Authentication, client creation and {GetSampleMethod()?.Name} {GetSampleMethodGroupName()} as an example written in TypeScript.", () =>
+            builder.Section($"Authentication", () => 
+            {
+                builder.Line("One of the first steps in accessing the services using SDK is to authenticate the client. There are several methods to authenticate such as interactive login, etc. Refer [readme for @azure/ms-rest-nodeauth](https://www.npmjs.com/package/@azure/ms-rest-nodeauth) package for all options to authenticate the client.");
```
ramya-rao-a, line 1070: Nit: To use `etc`, we should present more than 1 choice :)

ramya-rao-a: Also, this statement applies only to the node sample. From what I understand with this change, this line is being added before both the node and browser samples. Will it lead one to think that this applies to both node and browser?

deyaaeldeen: sorry I copied this from the other PR for v6. I updated the wording here https://github.com/Azure/autorest.typescript/pull/726. Please take a look.

PR #727, opened 2020-09-10
[v6] Add comment to readonly properties

joheredi:


deyaaeldeen (approved):
There is a lot of formatting noise, is that intentional? otherwise, it looks good!

Issue #742, opened 2020-10-08
Mappers for Non-string enum parameters are generated as String mappers


joheredi:
[This definition ](https://github.com/Azure/autorest.testserver/blob/c412c2e5dbee0404558e1a7d1cd7bf7a4a3475d6/swagger/non-string-enum.json#L98)in this swagger generates an incorrect mapper. The type should be `Number` not `String`

**Acrual**
```typescript

export const input1: OperationParameter = {
  parameterPath: ["options", "input"],
  mapper: {
    serializedName: "input",
    type: {
      name: "String"
    }
  }
};
```

**Expected**
```typescript

export const input1: OperationParameter = {
  parameterPath: ["options", "input"],
  mapper: {
    serializedName: "input",
    type: {
      name: "Number"
    }
  }
};
```

There is another issue related where the model is generated as

```typescript
/**
 * Defines values for IntEnum.
 */
export type IntEnum = 200 | 403 | 405 | 406 | 429 | string;
/**
 * Defines values for FloatEnum.
 */
export type FloatEnum = 200.4 | 403.4 | 405.3 | 406.2 | 429.1 | string;
```

where it should be at the very least

```typescript
/**
 * Defines values for IntEnum.
 */
export type IntEnum = 200 | 403 | 405 | 406 | 429 | number;
/**
 * Defines values for FloatEnum.
 */
export type FloatEnum = 200.4 | 403.4 | 405.3 | 406.2 | 429.1 | number;
```

sarangan12:
code complete with PR https://github.com/Azure/autorest.typescript/pull/763

Issue #741, opened 2020-10-08
Add test for MediaTypes.contentTypeWithEncoding


joheredi:
MediaTypes.contentTypeWithEncoding, there seems to be a problem where the [Swagger defines the content-type](https://github.com/Azure/autorest.testserver/blob/c412c2e5dbee0404558e1a7d1cd7bf7a4a3475d6/swagger/media_types.json#L47) as `text/plain` however the [test server](https://github.com/Azure/autorest.testserver/blob/c412c2e5dbee0404558e1a7d1cd7bf7a4a3475d6/legacy/routes/mediatypes.js#L32) looks for `text/plain; encoding=UTF-8`

Our implementation treats the value defined in consumes as a constant and automatically adds the header to the request without exposing it as a parameter. We may need to either update the swagger to define `text/plain; encoding=UTF-8` or the test in test server to expect `text/plain`

PR #740, opened 2020-10-08
adding the validation test

deyaaeldeen:
Fixes #521. I followed [Python's](https://github.com/Azure/autorest.python/blob/autorestv3/test/vanilla/AcceptanceTests/test_validation.py).

PR #758, opened 2020-10-09
[v6] Add tests for RequiredOptional and Non-String Enum + bonus

joheredi:
Changes in this PR:

**Test Coverage**
- Adding full coverage for RequiredOptional test server scenarios
- Initial coverage for Non-String enum scenarios, skipping 2 tests pending #742
- Missing scenario for Header test server
- Adding skipped test for mediaType missing scenario. Will be enabled with #741

**Extras:**
- Update test server dependency version
- Update autorest core dependency version
- Update autorest modelerfour dependency version
- Regenerated all test libraries to pickup changes from updated dependencies


README.md
```diff
@@ -26,9 +26,9 @@ More information about these can be found [here](https://github.com/Azure/autore
 3. Hook up plugins into the AutoRest pipeline DAG, e.g.
 
 ```yaml
-version: 3.0.6267
+version: 3.0.6320
```
deyaaeldeen, line 29: For my own edification, should I bump this number every time I add a new feature?

test/integration/header.spec.ts
```diff
@@ -1,5 +1,5 @@
 import * as coreHttp from "@azure/core-http";
-import { should } from "chai";
+import { should, assert } from "chai";
```
deyaaeldeen, line 2: is `assert` needed here?

test/integration/requiredOptional.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    it("should handle postRequiredArrayHeader", async () => {
+      try {
+        await client.explicit.postRequiredArrayHeader(null as any);
+        assert.fail("Expected client to throw");
+      } catch (error) {
+        assert(
+          error.message.indexOf("cannot be null or undefined") !== -1,
+          "Expected error to contain 'cannot be null or undefined'"
+        );
```
deyaaeldeen, line 96: Perhaps the following is more idiomatic:
```suggestion
       assert.include(error.message, "cannot be null or undefined");
```
If you agree, please apply it to the rest.

test/integration/requiredOptional.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+      }
+    });
+  });
+});
```
deyaaeldeen, line 160: There are a few extra ones in Python's, e.g.: https://github.com/Azure/autorest.python/blob/4cfd3e8e15f902e76c9ae9145ced542a459e78ef/test/vanilla/AcceptanceTests/test_required_optional.py#L92. Can we add them as well?

deyaaeldeen (commented):
Thanks for filling the gaps! I left a few comments

README.md
joheredi: We have been updating this on a need-to basis, so usually when there are new features or fixes in Autores.Core or ModelerFour or new tests when bumping test-server. In this particular case I needed this bump for a fix in Autorest.Core for supporting `x-ms-text` since TestServer started using that extension generating swaggers with it was failing.

I also try to get modeler four to the latest version once a month to be sure we keep up. In the near future we'd probably point to the latest builds within the same major build.

test/integration/header.spec.ts
joheredi: Nope, good catch 😄

README.md
deyaaeldeen: I see, thanks!

test/integration/requiredOptional.spec.ts
joheredi: I had to revert back as it looks like `assert.include` only works for arrays and not for strings 😞

sarangan12 (approved):

test/integration/requiredOptional.spec.ts
deyaaeldeen: weird, it works in my validation [PR](https://github.com/Azure/autorest.typescript/pull/740/files#r503975225).

deyaaeldeen (approved):
Looks great!

Issue #510, opened 2019-12-04
[Test] ExtensibleEnums


joheredi:
Full support for extensible enums swagger

[Swagger](https://github.com/Azure/autorest.testserver/blob/master/swagger/extensible-enums-swagger.json)
**Missing scenarios:** 
- [x] [`unexpectedEnum`](https://github.com/Azure/autorest.testserver/search?q=unexpectedEnum)
- [x] [`expectedEnum`](https://github.com/Azure/autorest.testserver/search?q=expectedEnum)
- [x] [`roundTripEnum`](https://github.com/Azure/autorest.testserver/search?q=roundTripEnum)
- [x] [`allowedValueEnum`](https://github.com/Azure/autorest.testserver/search?q=allowedValueEnum)

Bug #742 should be addressed when adding this coverage

joheredi:
We should look into @willmtemple's suggestion for representing extensible enums, here's a sample he shared:

```typescript
// Make an unmatchable symbol
declare const __ : unique symbol;
type Primitive =
	| null
	| undefined
	| string
	| number
	| boolean
	| symbol
	| bigint;
// Opaque<T> is a slightly more specific type than T that will not collapse
// subtypes of T when it is put into a union with them
type Opaque<T extends Primitive> = T & { [__]?: never};
type ExtensibleUnion<T extends Base, Base extends Primitive = string> = T | Opaque<Base>;

function foo(input: ExtensibleUnion<"a" | "b">) {
    console.log(input);
}

// If you make a new call to `foo` and open intellisense, it will suggest
// "a" and "b", but any other string is also valid.
foo("a");
foo("b");
foo("c");

```

sarangan12:
code complete with PR https://github.com/Azure/autorest.typescript/pull/763

PR #763, opened 2020-10-13
Mappers for Non-string enum parameters are generated as String mappers & Test Scenarios

sarangan12:
This PR is to address 2 Issues:

1. https://github.com/Azure/autorest.typescript/issues/742
2. https://github.com/Azure/autorest.typescript/issues/510

@joheredi Please review and approve


test/integration/generated/extensibleEnums/src/models/mappers.ts
```diff
@@ -0,0 +1,38 @@
...
+        serializedName: "DaysOfWeek",
+        type: {
+          name: "String"
+        }
+      },
+      intEnum: {
+        serializedName: "IntEnum",
+        required: true,
+        type: {
+          name: "String"
```
joheredi, line 33: I was expecting this to be name: "Number", you may be missing a commit updating the mapper generator?

test/integration/generated/nonStringEnum/src/models/index.ts
```diff
@@ -11,11 +11,11 @@ import * as coreHttp from "@azure/core-http";
 /**
  * Defines values for IntEnum.
  */
-export type IntEnum = 200 | 403 | 405 | 406 | 429 | string;
+export type IntEnum = 200 | 403 | 405 | 406 | 429 | number;
```
joheredi, line 14: Are we updating the mappers for these as well?

sarangan12: In the mapper the type will be set to number

joheredi: I would expect this not to be a string:

https://github.com/Azure/autorest.typescript/blob/c57733992fb0ba8917dd03b735dc004459b6e1aa/test/integration/generated/nonStringEnum/src/models/parameters.ts#L70-L75

sarangan12: Agreed. Give me few minutes. I will do it

joheredi:
It would be great if we can explore @willmtemple's suggestion on how to better type extensible enums. Quoting his suggested sample:

```typescript
// Make an unmatchable symbol
declare const __ : unique symbol;
type Primitive =
	| null
	| undefined
	| string
	| number
	| boolean
	| symbol
	| bigint;
// Opaque<T> is a slightly more specific type than T that will not collapse
// subtypes of T when it is put into a union with them
type Opaque<T extends Primitive> = T & { [__]?: never};
type ExtensibleUnion<T extends Base, Base extends Primitive = string> = T | Opaque<Base>;

function foo(input: ExtensibleUnion<"a" | "b">) {
    console.log(input);
}

// If you make a new call to `foo` and open intellisense, it will suggest
// "a" and "b", but any other string is also valid.
foo("a");
foo("b");
foo("c");

```

test/integration/generated/nonStringEnum/src/models/index.ts
sarangan12: Fixed it

test/integration/extensibleEnums.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+describe("Integration tests for extensible enums", () => {
+  let client: ExtensibleEnumsClient;
+
+  beforeEach(() => {
+    client = new ExtensibleEnumsClient();
+  });
+
+  it("sends an unexpected enum value successfully", async () => {
+    const response = await client.pet.getByPetId("casper");
+    expect(response).to.deep.equal({
```
joheredi, line 13: Would you mind switching to assert? This is in order to be consistent with the azure-sdks-for-js repo

src/transforms/transforms.ts
```diff
@@ -56,10 +56,18 @@ export function transformChoice(
   const metadata = getLanguageMetadata(choice.language);
   let name = normalizeName(metadata.name, NameType.Interface);
   let schemaType = choice.type;
+  let itemType =
+    //@ts-ignore
```
joheredi, line 60: What is this @ts-ignore for?

src/transforms/mapperTransforms.ts
```diff
@@ -462,20 +462,31 @@ function transformByteArrayMapper(pipelineValue: PipelineValue) {
...
 
   const choiceSchema = schema as ChoiceSchema;
-
-  const mapper = buildMapper(
-    schema,
-    {
+  let type;
+
+  if (isSchemaType([SchemaType.Choice], schema)) {
+    if (choiceSchema.choiceType && choiceSchema.choiceType.type) {
```
joheredi, line 473: According to the type definition of `ChoiceSchema` and `SealedChoiceSchema` it looks like `choiceType` and `choiceType.type` should never be undefined. Did you encounter any undefined during your testing?

joheredi: If not, we could probably skip this validation and simplify to something like this

```typescript
  if (isSchemaType([SchemaType.Choice], schema)) {
    type = {
      name: getMapperTypeFromSchema(choiceSchema.choiceType.type)
    };
  } else {
    type = {
      name: MapperType.Enum,
      allowedValues: choiceSchema.choices.map(choice => choice.value)
    };
  }

  const mapper = buildMapper(schema, type, options);
```

sarangan12: Yes. In one of the unit tests. In choiceSchema, choiceType is defined directly as String. It will not have any property called type.

src/transforms/transforms.ts
sarangan12: By definition `choice.choiceType.type` is defined as StringSchema. So, TS will complain that the condition will never be true. So, added this line to override that

src/transforms/mapperTransforms.ts
sarangan12: > If not, we could probably skip this validation and simplify to something like this
> 
> ```ts
>   if (isSchemaType([SchemaType.Choice], schema)) {
>     type = {
>       name: getMapperTypeFromSchema(choiceSchema.choiceType.type)
>     };
>   } else {
>     type = {
>       name: MapperType.Enum,
>       allowedValues: choiceSchema.choices.map(choice => choice.value)
>     };
>   }
> 
>   const mapper = buildMapper(schema, type, options);
> ```

:-) This was my original code. Unfortunately, because of the test failure, I added the if check and modified the code

joheredi: Talked offline with @sarangan12 , we should file a bug in M4 to either throw when ChoiceSchema is constructed without specifying the choiceType or make choiceType an optional property of ChoiceSchema

joheredi (approved):
Looks good, thanks!

Issue #752, opened 2020-10-09
Add coverage for missing body-dictionary test server scenarios


joheredi:
[Swagger](https://github.com/Azure/autorest.testserver/blob/ee473a350ebd680a728acda7d3bbd4334e9826a0/swagger/body-dictionary.json)
- [ ] [`getDictionaryArrayEmpty`](https://github.com/Azure/autorest.testserver/search?q=getDictionaryArrayEmpty)
- [ ] [`getDictionaryArrayItemNull`](https://github.com/Azure/autorest.testserver/search?q=getDictionaryArrayItemNull)
- [ ] [`getDictionaryArrayValid`](https://github.com/Azure/autorest.testserver/search?q=getDictionaryArrayValid)
- [ ] [`getDictionaryComplexNull`](https://github.com/Azure/autorest.testserver/search?q=getDictionaryComplexNull)
- [ ] [`getDictionaryDictionaryEmpty`](https://github.com/Azure/autorest.testserver/search?q=getDictionaryDictionaryEmpty)
- [ ] [`getDictionaryDictionaryItemEmpty`](https://github.com/Azure/autorest.testserver/search?q=getDictionaryDictionaryItemEmpty)
- [ ] [`getDictionaryDictionaryItemNull`](https://github.com/Azure/autorest.testserver/search?q=getDictionaryDictionaryItemNull)
- [ ] [`getDictionaryDictionaryValid`](https://github.com/Azure/autorest.testserver/search?q=getDictionaryDictionaryValid)
- [ ] [`putDictionaryComplexValid`](https://github.com/Azure/autorest.testserver/search?q=putDictionaryComplexValid)

PR #768, opened 2020-10-16
adding support for nullable

### LINKED ISSUE #680: [v6] Support x-ms-nullable
Nullable schemas type should be generated as as union with null i.e. `nullableNumber: number | null`

deyaaeldeen:
Fixes #680.
Fixes https://github.com/Azure/autorest.typescript/issues/752.

### DIFF
--- a/src/generators/modelsGenerator.ts
+++ b/src/generators/modelsGenerator.ts
@@ -589,7 +589,7 @@ function getProperties(
       property.name === "siblings"
         ? `${(objectDetails as PolymorphicObjectDetails).unionName}[]`
         : property.type;
-    return property.nullable ? typeName + " | null" : typeName;
+    return property.nullable ? `${typeName} | null` : typeName;
   };
 
   return properties
--- a/src/generators/modelsGenerator.ts
+++ b/src/generators/modelsGenerator.ts
@@ -585,9 +585,11 @@ function getProperties(
       )}"`;
     }
 
-    return property.name === "siblings"
-      ? `${(objectDetails as PolymorphicObjectDetails).unionName}[]`
-      : property.type;
+    const typeName =
+      property.name === "siblings"
+        ? `${(objectDetails as PolymorphicObjectDetails).unionName}[]`
+        : property.type;
+    return property.nullable ? typeName + " | null" : typeName;
   };
 
   return properties
--- a/src/utils/nameUtils.ts
+++ b/src/utils/nameUtils.ts
@@ -143,7 +143,9 @@ export function normalizeName(
   const [firstPart, ...otherParts] = parts;
   const normalizedFirstPart = toCasing(firstPart, casingConvention);
   const normalizedParts = (otherParts || [])
-    .map(part => toCasing(part, CasingConvention.Pascal))
+    .map(part =>
+      part === "null" ? part : toCasing(part, CasingConvention.Pascal)
+    )
     .join("");
 
   const normalized = `${normalizedFirstPart}${normalizedParts}`;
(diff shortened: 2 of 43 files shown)


src/generators/modelsGenerator.ts
```diff
@@ -585,9 +585,11 @@ function getProperties(
...
     }
 
-    return property.name === "siblings"
-      ? `${(objectDetails as PolymorphicObjectDetails).unionName}[]`
-      : property.type;
+    const typeName =
+      property.name === "siblings"
+        ? `${(objectDetails as PolymorphicObjectDetails).unionName}[]`
+        : property.type;
+    return property.nullable ? typeName + " | null" : typeName;
```
joheredi, line 592: nit: you can also use string interpolation, I personally like it better for readability. However it is just personal taste

```typescript
return property.nullable ? `${typeName} | null` : typeName;
```

src/utils/nameUtils.ts
```diff
@@ -143,7 +143,9 @@ export function normalizeName(
   const [firstPart, ...otherParts] = parts;
   const normalizedFirstPart = toCasing(firstPart, casingConvention);
   const normalizedParts = (otherParts || [])
-    .map(part => toCasing(part, CasingConvention.Pascal))
+    .map(part =>
+      part === "null" ? part : toCasing(part, CasingConvention.Pascal)
```
joheredi, line 147: Do you think there may be other scenarios where we'd like to keep the name? Would it be possible to have strings as `boolean`, `string`, `number` that would be impacted the same way `null` was?

If so, it may be worth using a collection of values that don't need to get a casing change and check that the current part is not present in that collection, instead of just checking for null

src/utils/schemaHelpers.ts
```diff
@@ -61,22 +61,30 @@ export function getAdditionalProperties(
...
       const arraySchema = schema as ArraySchema;
-      const itemsType = getTypeForSchema(arraySchema.elementType);
+      const itemsType = getTypeForSchema(
+        arraySchema.elementType,
+        arraySchema.nullableItems
+      );
       const itemsName = normalizeTypeName(itemsType);
       kind = itemsType.kind;
-      typeName = `${itemsName}[]`;
+      const wrappedItemsName = itemsType.nullable
```
joheredi, line 84: Would you mind adding a comment about how wrappedItemsName is used here?

test/integration/bodyDictionary.spec.ts
```diff
@@ -33,16 +33,55 @@ describe("BodyDictionary", () => {
       "1": [],
       "2": ["7", "8", "9"]
     });
+    assert.deepEqual(_response.status, 200);
   });
 
+  // FIXME we should de-serialize the incoming value as a null
```
joheredi, line 39: Is this comment still valid? If so, can you please add an Issue number so that we can keep track?

joheredi (approved):
Looks great! Thanks for taking on this work!

src/utils/nameUtils.ts
deyaaeldeen: This is a great feedback! I experimented with different ways to approach this. It turns out, `normalizeName` is a function that processes a single name and should not be called on a type expression. I updated `getTypeForSchema` to not call `normalizeTypeName` excessively and reverted `normalizeName` back.

test/integration/bodyDictionary.spec.ts
deyaaeldeen: Yes, I put a link to this issue: https://github.com/Azure/azure-sdk-for-js/issues/12009.

PR #769, opened 2020-10-20
Enable MediaType test scenario

joheredi:
Enabling skipped missing type for MediaType tests. This scenario is supposed to test that Content-Type can be overriden, we can do that by passing customHeaders in the options parameter. See https://github.com/Azure/autorest.testserver/pull/228 for a conversation about the expectations of the test.

### DIFF
--- a/test/integration/mediaType.spec.ts
+++ b/test/integration/mediaType.spec.ts
@@ -6,7 +6,7 @@ describe("Integration tests for MediaTypes", () => {
 
   beforeEach(() => {
     client = new MediaTypesClient({
-      requestPolicyFactories: defaultPolicies => {
+      requestPolicyFactories: (defaultPolicies) => {
         defaultPolicies.push({
           create(nextPolicy) {
             return {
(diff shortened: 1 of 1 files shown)


joheredi:
Fixes #741

test/integration/mediaType.spec.ts
```diff
@@ -59,12 +59,14 @@ describe("Integration tests for MediaTypes", () => {
...
     it("works with text/plain", async () => {
-      const response = await client.contentTypeWithEncoding(
-        "text/plain; encoding=UTF-8"
-      );
+      client = new MediaTypesClient();
+      const response = await client.contentTypeWithEncoding("test", {
+        requestOptions: {
+          customHeaders: { "content-type": "text/plain; encoding=UTF-8" }
+        }
+      });
```
deyaaeldeen, line 70: I think we should also check for the response value.
```suggestion

           expect(response.body).to.equal("Nice job sending content type with encoding");
```

deyaaeldeen: I might have messed up the spacing in my suggested code.

deyaaeldeen (approved):

PR #773, opened 2020-10-22
Only add polymorphic parent to union if the parent is a valid value

joheredi:
Only include the base class in the union type if the polymorphic parent's name is a valid discriminator value.

### DIFF
--- a/src/generators/modelsGenerator.ts
+++ b/src/generators/modelsGenerator.ts
@@ -444,13 +444,17 @@ function writeUniontypes({ objects }: ClientDetails, modelsFile: SourceFile) {
     .forEach(obj => {
       const polymorphicObject = obj as PolymorphicObjectDetails;
       const childrenNames = [
-        obj.name,
         ...polymorphicObject.children.map(c => {
           return c.schema.children && c.schema.children.immediate.length
             ? `${c.name}Union`
             : c.name;
         })
       ];
+
+      if (isPolymorphicParentInUnion(polymorphicObject)) {
+        childrenNames.unshift(polymorphicObject.name);
+      }
+
       modelsFile.addTypeAlias({
         name: `${obj.name}Union`,
         isExported: true,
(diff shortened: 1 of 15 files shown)


src/generators/modelsGenerator.ts
```diff
@@ -444,13 +444,17 @@ function writeUniontypes({ objects }: ClientDetails, modelsFile: SourceFile) {
     .forEach(obj => {
       const polymorphicObject = obj as PolymorphicObjectDetails;
       const childrenNames = [
-        obj.name,
         ...polymorphicObject.children.map(c => {
           return c.schema.children && c.schema.children.immediate.length
             ? `${c.name}Union`
             : c.name;
         })
       ];
```
deyaaeldeen, line 452: I think this could be better rewritten as:

```ts
const childrenNames = polymorphicObject.children.map(c =>
          c.schema.children && c.schema.children.immediate.length
            ? `${c.name}Union`
            : c.name
        );
```

src/generators/modelsGenerator.ts
```diff
@@ -444,13 +444,17 @@ function writeUniontypes({ objects }: ClientDetails, modelsFile: SourceFile) {
...
         ...polymorphicObject.children.map(c => {
           return c.schema.children && c.schema.children.immediate.length
             ? `${c.name}Union`
             : c.name;
         })
       ];
+
+      if (isPolymorphicParentInUnion(polymorphicObject)) {
+        childrenNames.unshift(polymorphicObject.name);
+      }
```
deyaaeldeen, line 456: Nit: I am not sure about the complexity of `unshift`. Perhaps we can do this check first to initialize the array and then add  the children names after.

deyaaeldeen (approved):
Thanks for doing it! I left a couple minor comments.

src/generators/modelsGenerator.ts
joheredi: Good catch, by removing obj.name the spread is no longer needed.

src/generators/modelsGenerator.ts
joheredi: Yeah, I'll do that

Issue #738, opened 2020-10-08
[v6] Should avoid name collisions


joheredi:
For example [this](https://github.com/Azure/azure-sdk-for-js/blob/a3f90de4dcc0bf9a8b7f4869914410552820fb0f/sdk/api-learn/design-activity/swagger/computation.json#L109) swagger would produce a name collision in operation group and top-level operation. Both called `computation`.

We should generate non colliding names, at the bare minimum `computation` and `computation1` or throw an error indicating that the swagger definition is ambiguous 

//cc: @xirzec

daviwil:
Is this a TypeScript generator issue or is something that modelerfour should handle?  I don't think that generators should have to deal with this unless there are language-specific naming conventions that cause further collisions.

joheredi:
That's a good question, let me debug and see if this is language specific. There are other examples of collisions for example, an operation named client would collide with the private client property in the operation group class, however this instance is TS generator specific

joheredi:
This issue seems related: https://github.com/Azure/autorest.typescript/issues/708

sarangan12:
Code complete with PR https://github.com/Azure/autorest.typescript/pull/779

PR #779, opened 2020-10-27
Added code to avoid name collisions

sarangan12:
This PR is to fix the issue with Issue: https://github.com/Azure/autorest.typescript/issues/738

@joheredi Please review and approve

### DIFF
--- a/test/unit/generators/clientFileGenerator.spec.ts
+++ b/test/unit/generators/clientFileGenerator.spec.ts
@@ -1,0 +1,84 @@
+import { KnownMediaType } from "@azure-tools/codegen";
+import * as assert from "assert";
+import { checkForNameCollisions } from "../../../src/generators/clientFileGenerator";
+import { PropertyKind } from "../../../src/models/modelDetails";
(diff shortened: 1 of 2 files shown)


joheredi (approved):
Looks good! Thanks for working on this, now we have more actionable error messages when running into name collisions 😄 

As discussed offline, it would be great to remove the validations from `test-swagger-gen.ts` and instead test the collision checking logic in a UnitTest.

test/unit/generators/clientFileGenerator.spec.ts
```diff
@@ -0,0 +1,84 @@
+import { KnownMediaType } from "@azure-tools/codegen";
+import * as assert from "assert";
+import { checkForNameCollisions } from "../../../src/generators/clientFileGenerator";
+import { PropertyKind } from "../../../src/models/modelDetails";
+import { OperationGroupDetails } from "../../../src/models/operationDetails";
+
+describe("clientFileGenerator", () => {
```
joheredi, line 7: Thanks for doing this 😄

joheredi (approved):

PR #780, opened 2020-10-27
Handle composite and sequence mapper separately

joheredi:
We got a report of element contraints not being generated correctly and caused a compilation error.

Specifically, we were having issues generating the Pattern constraint. The reason is that the mapper generator was not treating the Sequence element as a mapper, but instead just writing its JSON stringified content.

To fix this we'll handle Composite and Sequence mappers separately and generating the element content as an inline mapper


src/generators/mappersGenerator.ts
```diff
@@ -152,30 +153,84 @@ export function writeMapper(writer: CodeBlockWriter, mapper: Mapper) {
...
+  return writeCompositeMapperType(writer, mapperType, parents);
+}
+
+/**
+ * Check if a MapperType is a SequenceMapperType
+ */
+function isSequenceMapperType(
+  mapperType: MapperType
+): mapperType is SequenceMapperType {
+  return Boolean((<SequenceMapperType>mapperType).element);
```
deyaaeldeen, line 195: nit: I think this would be cleaner:
```suggestion
  return (mapperType as SequenceMapperType).element !== undefined;
```

deyaaeldeen (approved):
Looks great! Thanks!

PR #787, opened 2020-11-04
Make LROData a symbol in the response

joheredi:
* Adding missing coverage for LRO scenarios
* Converting _lroData into a Symbol


src/lro/lroPolicy.ts
```diff
@@ -24,11 +25,9 @@ class LROPolicy extends BaseRequestPolicy {
     webResource: WebResource
   ): Promise<HttpOperationResponse> {
     let result = await this._nextPolicy.sendRequest(webResource);
+    const _lroData = getLROData(result);
 
-    if (webResource.shouldDeserialize !== undefined) {
-      const _lroData = getLROData(result);
-      result.parsedBody = { ...result.parsedBody, _lroData };
-    }
+    Object.defineProperty(result, LROSYM, { value: _lroData });
```
xirzec, line 30: I don't think using `defineProperty` is necessary here, since symbols aren't enumerable in the same way normal properties are.

This should be equivalent:
```suggestion
    result[LROSYM] = _lroData;
```

src/lro/models.ts
```diff
@@ -2,9 +2,11 @@ import {
   OperationArguments,
   OperationSpec,
   RestResponse,
-  HttpMethods
+  HttpMethods,
+  HttpOperationResponse
 } from "@azure/core-http";
 import { PollOperationState, PollOperation } from "@azure/core-lro";
+export const LROSYM = Symbol.for("LROData");
```
xirzec, line 9: Can we have a longer name, since this is a global symbol cache?

Proposal: `packagename.identifier` e.g. `@azure/core-http.LROData`

xirzec (commented):
Overall looks good, just a couple small comments

deyaaeldeen (approved):

PR #789, opened 2020-11-10
Support for Parameter flattening

joheredi:
Our current support of model flattening didn't support parameter flattening. This PR adds:

* Support for Parameter Flattening and Parameter Grouping
* Fixes generator to avoid generating un-needed parameter definitions

**Note:** there are many files in the PR, most of them are under `test/integration/generated/*` in the form of removing duplicate parameter definitions for `accept` and `content-type`. The main interesting file to review are: 

* everything under `src/*`
* `test/**/*.spec.ts`
* `test/integration/generated/modelFlattening` - For an example of how the generated code will change with this PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,11 @@ modelerfour:
   prenamer: true
   # this will flatten modelers marked with 'x-ms-client-flatten'
   flatten-models: true
+  # this will flatten parameters marked with 'x-ms-client-flatten'
+  flatten-payloads: true
   # this will make the content-type parameter always specified
   always-create-content-type-parameter: true
+
   # enables parameter grouping via x-ms-parameter-grouping
   group-parameters: true
 
--- a/README.md
+++ b/README.md
@@ -39,7 +39,6 @@ modelerfour:
   flatten-payloads: true
   # this will make the content-type parameter always specified
   always-create-content-type-parameter: true
-
   # enables parameter grouping via x-ms-parameter-grouping
   group-parameters: true
 
--- a/src/generators/operationGenerator.ts
+++ b/src/generators/operationGenerator.ts
@@ -143,7 +143,7 @@ export function writeGetOperationOptions(
  * Generates a string representation of an Operation spec
  * the output is to be inserted in the Operation group file
  */
-function writeSpec(spec: OperationSpecDetails, writer: CodeBlockWriter) {
+function writeSpec(spec: OperationSpecDetails, writer: CodeBlockWriter): void {
   const responses = buildResponses(spec);
   const requestBody = buildRequestBody(spec);
   const queryParams = buildParameters(spec, "queryParameters");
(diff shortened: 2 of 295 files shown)


src/generators/operationGenerator.ts
```diff
@@ -151,45 +153,187 @@ function buildSpec(spec: OperationSpecDetails): string {
...
     ? "serializer: xmlSerializer"
     : "serializer";
 
-  return `{ path: "${spec.path}", httpMethod: "${
-    spec.httpMethod
-  }", responses: {${responses.join(
-    ", "
-  )}},${requestBody}${formDataParams}${queryParams}${urlParams}${headerParams}${isXML}${contentType}${mediaType}${serializerName}
-    }`;
+  writer.block(() => {
```
joheredi, line 160: Moved to a more explicit generation strategy which allows more flexibility when arranging stuff. Also although it is more verbose, it should improve readability as it is easier to follow what is happening

src/generators/operationGenerator.ts
```diff
@@ -151,45 +153,187 @@ function buildSpec(spec: OperationSpecDetails): string {
...
+      writer.write(
+        `requestBody: { parameterPath: ${JSON.stringify(
+          requestBody.parameterPath
+        )}, mapper: ${mapper}}, `
+      );
+    }
+
+    if (formDataParams) {
+      writer.write(formDataParams);
+      writer.write(", ");
```
joheredi, line 182: Also decided to explicitly add the commas here rather than within the helper `build*` functions to make it easier to manage

src/generators/parametersGenerator.ts
```diff
@@ -51,6 +51,8 @@ export function generateParameters(
 
   clientDetails.parameters
     .filter(p => !p.isSynthetic)
+    // No need to generate parameters for flattened ones, as the would never get used
```
deyaaeldeen, line 54:
```suggestion
    // No need to generate parameters for flattened ones, as they will never get used
```

src/generators/operationGenerator.ts
```diff
@@ -141,7 +143,7 @@ export function writeGetOperationOptions(
  * Generates a string representation of an Operation spec
  * the output is to be inserted in the Operation group file
  */
-function buildSpec(spec: OperationSpecDetails): string {
+function writeSpec(spec: OperationSpecDetails, writer: CodeBlockWriter) {
```
deyaaeldeen, line 146: can you add a return type to this function?

README.md
```diff
@@ -35,8 +35,11 @@ modelerfour:
   prenamer: true
   # this will flatten modelers marked with 'x-ms-client-flatten'
   flatten-models: true
+  # this will flatten parameters marked with 'x-ms-client-flatten'
+  flatten-payloads: true
   # this will make the content-type parameter always specified
   always-create-content-type-parameter: true
+
```
deyaaeldeen, line 42: is this added by mistake?

deyaaeldeen (approved):
Looks really good! I left a few minor comments.

src/generators/operationGenerator.ts
joheredi: :D good catch, this should be void. The return was a left over

PR #790, opened 2020-11-12
Generate Tracing

joheredi:
Create span and mark success or failure failure. Also refactored the generator code in hopes to improve readability and reusability.

Adding this behind a flag `--enable-tracing` as there may be a few things we'll want to collect feedback on and improve, for example, we may want to support overriding the operation name, this is for cases where there is a convenience layer on top and the public method has a different name.

Also, tracing for LRO operations will come in a separate PR.

**Note:** The most interesting files to look at outside of the changes under `src/` are `test/integration/generated/bodyComplexWithTracing/src/operations/*` and `test/integration/generated/mediaTypesWithTracing/mediaTypesWithTracingClient.ts` which contain test clients generated with the `--enable-tracing` flag

Here are a couple of  sample with and without tracing extracted from the test clients mentioned above.

**Without Tracing**
```typescript
getValid(
    options?: coreHttp.OperationOptions
  ): Promise<InheritanceGetValidResponse> {
    const operationArguments: coreHttp.OperationArguments = {
      options: coreHttp.operationOptionsToRequestOptionsBase(options || {})
    };
    return this.client.sendOperationRequest(
      operationArguments,
      getValidOperationSpec
    ) as Promise<InheritanceGetValidResponse>;
  }
```

**With Tracing**
```typescript
  async getValid(
    options?: coreHttp.OperationOptions
  ): Promise<InheritanceGetValidResponse> {
    const { span, updatedOptions } = createSpan(
      "BodyComplexWithTracing-getValid",
      coreHttp.operationOptionsToRequestOptionsBase(options || {})
    );
    const operationArguments: coreHttp.OperationArguments = {
      options: updatedOptions
    };
    try {
      const result = await this.client.sendOperationRequest(
        operationArguments,
        getValidOperationSpec
      );
      return result as InheritanceGetValidResponse;
    } catch (error) {
      span.setStatus({
        code: CanonicalCode.UNKNOWN,
        message: error.message
      });
      throw error;
    } finally {
      span.end();
    }
  }
```

Another sample with operations that have multiple media type overrides

**Without Tracing**
```typescript 
analyzeBody(
    contentType: ContentType,
    input: coreHttp.HttpRequestBody,
  /**
   */
  analyzeBody(
  /**
   * @param args Includes all the parameters for this operation.
   */
  analyzeBody(
    ...args:
      | [
          ContentType,
          coreHttp.HttpRequestBody,
          MediaTypesClientAnalyzeBody$binaryOptionalParams?
        ]
      | ["application/json", MediaTypesClientAnalyzeBody$jsonOptionalParams?]
  ): Promise<MediaTypesClientAnalyzeBodyResponse> {
    let operationSpec: coreHttp.OperationSpec;
    let operationArguments: coreHttp.OperationArguments;
    if (
      args[0] === "application/pdf" ||
      args[0] === "image/jpeg" ||
      args[0] === "image/png" ||
      args[0] === "image/tiff"
    ) {
      operationSpec = analyzeBody$binaryOperationSpec;
      operationArguments = {
        contentType: args[0],
        input: args[1],
        options: args[2]
      };
    } else if (args[0] === "application/json") {
      operationSpec = analyzeBody$jsonOperationSpec;
      operationArguments = { contentType: args[0], options: args[1] };
    } else {
      throw new TypeError(
        `"contentType" must be a valid value but instead was "${args[0]}".`
      );
    }
    operationArguments.options = coreHttp.operationOptionsToRequestOptionsBase(
      operationArguments.options || {}
    );
    return this.sendOperationRequest(
      operationArguments,
      operationSpec
    ) as Promise<MediaTypesClientAnalyzeBodyResponse>;
  }

```

**With Tracing**

```typescript
analyzeBody(
    contentType: ContentType,
    input: coreHttp.HttpRequestBody,
    options?: MediaTypesWithTracingClientAnalyzeBody$binaryOptionalParams
  ): Promise<MediaTypesWithTracingClientAnalyzeBodyResponse>;
  /**
   */
  analyzeBody(
    options?: MediaTypesWithTracingClientAnalyzeBody$jsonOptionalParams
  ): Promise<MediaTypesWithTracingClientAnalyzeBodyResponse>;
  /**
   * @param args Includes all the parameters for this operation.
   */
  async analyzeBody(
    ...args:
      | [
          ContentType,
          coreHttp.HttpRequestBody,
          MediaTypesWithTracingClientAnalyzeBody$binaryOptionalParams?
        ]
      | [
          "application/json",
          MediaTypesWithTracingClientAnalyzeBody$jsonOptionalParams?
        ]
  ): Promise<MediaTypesWithTracingClientAnalyzeBodyResponse> {
    let operationSpec: coreHttp.OperationSpec;
    let operationArguments: coreHttp.OperationArguments;
    if (
      args[0] === "application/pdf" ||
      args[0] === "image/jpeg" ||
      args[0] === "image/png" ||
      args[0] === "image/tiff"
    ) {
      operationSpec = analyzeBody$binaryOperationSpec;
      operationArguments = {
        contentType: args[0],
        input: args[1],
        options: args[2]
      };
    } else if (args[0] === "application/json") {
      operationSpec = analyzeBody$jsonOperationSpec;
      operationArguments = { contentType: args[0], options: args[1] };
    } else {
      throw new TypeError(
        `"contentType" must be a valid value but instead was "${args[0]}".`
      );
    }
    const { span, updatedOptions } = createSpan(
      "MediaTypesWithTracingClient-analyzeBody",
      coreHttp.operationOptionsToRequestOptionsBase(
        operationArguments.options || {}
      )
    );
    operationArguments.options = updatedOptions;
    try {
      const result = await this.sendOperationRequest(
        operationArguments,
        operationSpec
      );
      return result as MediaTypesWithTracingClientAnalyzeBodyResponse;
    } catch (error) {
      span.setStatus({
        code: CanonicalCode.UNKNOWN,
        message: error.message
      });
      throw error;
    } finally {
      span.end();
    }
  }
```


src/generators/tracingFileGenerator.ts
```diff
@@ -0,0 +1,140 @@
...
+    ],
+    parameters,
+    returnType: "{ span: Span; updatedOptions: T }"
+  });
+
+  const telemetryNamespace = getTelemetryNamespace(packageDetails);
+  const telemetryPackageName = `Microsoft.${getTelemetryPackageName(
+    packageDetails
+  )}`;
+  createSpanFunction.addStatements([
```
xirzec, line 90: I've been thinking for a while that we could move most of this helper into `core-tracing` itself and then do something like

```ts
import { createSpanFunction } from "@azure/core-tracing";
const createSpan = createSpanFunction({packagePrefix, namespace});
```

thoughts?

src/generators/tracingFileGenerator.ts
```diff
@@ -0,0 +1,140 @@
...
+      tracingOptions: newTracingOptions
+    };`,
+    `return {
+      span,
+      updatedOptions: newOperationOptions
+    };`
+  ]);
+}
+
+function getTelemetryPackageName({ nameWithoutScope }: PackageDetails) {
```
xirzec, line 125: curious how this spits out the right name, looks like we want to mangle `@azure/storage-blob` into `Azure.Storage.Blob`, but somehow `@azure/ai-form-recognizer` becomes `Azure.CognitiveServices.FormRecognizer`

Not sure how you map this properly.

src/generators/tracingFileGenerator.ts
```diff
@@ -0,0 +1,140 @@
...
+  return nameWithoutScope
+    .split(/[-._ ]+/)
+    .map(part => normalizeName(part, NameType.Class));
+}
+
+function getTelemetryNamespace(packageDetails: PackageDetails) {
```
xirzec, line 131: I'm not sure how this function is supposed to correctly map the package name to the resource provider list:

https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-services-resource-providers

Most of them are pretty 1:1, but some of them aren't, like Form Recognizer being `Microsoft.CognitiveServices`

Also they all begin with `Microsoft` but the `scopeName` is `Azure`, isn't it?

src/generators/utils/tracingUtils.ts
```diff
@@ -0,0 +1,26 @@
+import { SourceFile } from "ts-morph";
+import { ClientDetails } from "../../models/clientDetails";
+
+/**
+ * Adds the required imports to have operations tracing
+ * @param param0 clientDetails
```
xirzec, line 6: I never thought about what `@param` would call destructured parameters before...

test/integration/generated/bodyComplexWithTracing/src/tracing.ts
```diff
@@ -0,0 +1,59 @@
...
+  const tracingOptions = operationOptions.tracingOptions || {};
+  const spanOptions: SpanOptions = {
+    ...tracingOptions.spanOptions,
+    kind: SpanKind.INTERNAL
+  };
+  const span = tracer.startSpan(
+    `Body.Complex.Tracing.${operationName}`,
+    spanOptions
+  );
+  span.setAttribute("az.namespace", "Microsoft.Body,Complex,Tracing");
```
xirzec, line 34: why are there commas in this namespace?

xirzec (commented):
Nice work on this! The span wrapping of operations looks great!

My only quibble is the shape of generated `tracing.ts` -- I don't think we can programmatically generate the operation prefix or the az.namespace with enough reliability. 

Rather than risk someone generating garbage, I'd prefer us add required arguments to the generator to specify them when tracing is enabled. Does that sound reasonable?

src/generators/tracingFileGenerator.ts
joheredi: Yeah totally makes sense added it to `core-http` and `core-client` [#12525](https://github.com/Azure/azure-sdk-for-js/pull/12525)

src/generators/tracingFileGenerator.ts
joheredi: It makes more sense to have the user explicitly specify the namespace and prefix. I have changed the flag to contain those two required properties

xirzec (approved):
Nice! The update looks good to me

PR #792, opened 2020-11-17
Generate Async iterators

joheredi:
This PR adds the ability to generate pageable operations using AsyncIterators.

This will be the default value, however this functionality can be disabled by `---disable-async-iterators=true`.

The new paging operations look like this:

```typescript
public listNoItemNamePages(
    options?: coreHttp.OperationOptions
  ): PagedAsyncIterableIterator<Product, Product[]> {
    const iter = this.getNoItemNamePagesPagingAll(options);
    return {
      next() {
        return iter.next();
      },
      [Symbol.asyncIterator]() {
        return this;
      },
      byPage: () => {
        return this.getNoItemNamePagesPagingPage(options);
      }
    };
  }

  private async *getNoItemNamePagesPagingPage(
    options?: coreHttp.OperationOptions
  ): AsyncIterableIterator<Product[]> {
    let result = await this._getNoItemNamePages(options);
    yield result.value || [];
    let continuationToken = result.nextLink;
    while (continuationToken) {
      result = await this._getNoItemNamePagesNext(continuationToken, options);
      continuationToken = result.nextLink;
      yield result.value || [];
    }
  }

  private async *getNoItemNamePagesPagingAll(
    options?: coreHttp.OperationOptions
  ): AsyncIterableIterator<Product> {
    for await (const page of this.getNoItemNamePagesPagingPage(options)) {
      yield* page;
    }
  }


private async _getNoItemNamePages(
    options?: coreHttp.OperationOptions
  ): Promise<PagingGetNoItemNamePagesResponse> {
    const { span, updatedOptions } = createSpan(
      "PagingClient-_getNoItemNamePages",
      coreHttp.operationOptionsToRequestOptionsBase(options || {})
    );
    const operationArguments: coreHttp.OperationArguments = {
      options: updatedOptions
    };
    try {
      const result = await this.client.sendOperationRequest(
        operationArguments,
        getNoItemNamePagesOperationSpec
      );
      return result as PagingGetNoItemNamePagesResponse;
    } catch (error) {
      span.setStatus({
        code: CanonicalCode.UNKNOWN,
        message: error.message
      });
      throw error;
    } finally {
      span.end();
    }
  }

/**
   * GetNoItemNamePagesNext
   * @param nextLink The nextLink from the previous successful call to the GetNoItemNamePages method.
   */
  private async _getNoItemNamePagesNext(
    nextLink: string,
    options?: coreHttp.OperationOptions
  ): Promise<PagingGetNoItemNamePagesNextResponse> {
    const { span, updatedOptions } = createSpan(
      "PagingClient-_getNoItemNamePagesNext",
      coreHttp.operationOptionsToRequestOptionsBase(options || {})
    );
    const operationArguments: coreHttp.OperationArguments = {
      nextLink,
      options: updatedOptions
    };
    try {
      const result = await this.client.sendOperationRequest(
        operationArguments,
        getNoItemNamePagesNextOperationSpec
      );
      return result as PagingGetNoItemNamePagesNextResponse;
    } catch (error) {
      span.setStatus({
        code: CanonicalCode.UNKNOWN,
        message: error.message
      });
      throw error;
    } finally {
      span.end();
    }
  }
```

In the snipper above you may notice that there are two methods used `_getNoItemNamePages` and `_getNoItemNamePagesNext`. These two used to be the 2 public methods before implementing this feature, the first method gets the first page and the `next` method will get the next page taking a continuation token. With this the user needs to link the two operations and loop until there is no continuationToken.

When enabling AsyncIterators, those 2 methods become private and we expose a single method to the user, which takes care of looping and linking the initial and next operations.

### LRO Sample

```typescript
 /**
   * A long-running paging operation that includes a nextLink that has 10 pages
   */
  public listMultiplePagesLRO(
    options?: PagingGetMultiplePagesLROOptionalParams
  ): PagedAsyncIterableIterator<Product, Product[]> {
    const iter = this.getMultiplePagesLROPagingAll(options);
    return {
      next() {
        return iter.next();
      },
      [Symbol.asyncIterator]() {
        return this;
      },
      byPage: () => {
        return this.getMultiplePagesLROPagingPage(options);
      }
    };
  }

  private async *getMultiplePagesLROPagingPage(
    options?: PagingGetMultiplePagesLROOptionalParams
  ): AsyncIterableIterator<Product[]> {
    const poller = await this._getMultiplePagesLRO(options);
    let result: any = await poller.pollUntilDone();
    yield result.values || [];
    let continuationToken = result.nextLink;
    while (continuationToken) {
      result = await this._getMultiplePagesLRONext(continuationToken, options);
      continuationToken = result.nextLink;
      yield result.values || [];
    }
  }

  private async *getMultiplePagesLROPagingAll(
    options?: PagingGetMultiplePagesLROOptionalParams
  ): AsyncIterableIterator<Product> {
    for await (const page of this.getMultiplePagesLROPagingPage(options)) {
      yield* page;
    }
  }

/**
   * A long-running paging operation that includes a nextLink that has 10 pages
   */
  private async _getMultiplePagesLRO(
    options?: PagingGetMultiplePagesLROOptionalParams
  ): Promise<LROPoller<PagingGetMultiplePagesLROResponse>> {
    const { span, updatedOptions } = createSpan(
      "PagingClient-_getMultiplePagesLRO",
      this.getOperationOptions(options, "undefined")
    );
    const operationArguments: coreHttp.OperationArguments = {
      options: updatedOptions
    };
    const sendOperation = async (
      args: coreHttp.OperationArguments,
      spec: coreHttp.OperationSpec
    ) => {
      try {
        const result = await this.client.sendOperationRequest(args, spec);
        return result as PagingGetMultiplePagesLROResponse;
      } catch (error) {
        span.setStatus({
          code: CanonicalCode.UNKNOWN,
          message: error.message
        });
        throw error;
      } finally {
        span.end();
      }
    };

    const initialOperationResult = await sendOperation(
      operationArguments,
      getMultiplePagesLROOperationSpec
    );
    return new LROPoller({
      initialOperationArguments: operationArguments,
      initialOperationSpec: getMultiplePagesLROOperationSpec,
      initialOperationResult,
      sendOperation
    });
  }


 /**
   * GetMultiplePagesLRONext
   * @param nextLink The nextLink from the previous successful call to the GetMultiplePagesLRO method.
   */
  private async _getMultiplePagesLRONext(
    nextLink: string,
    options?: PagingGetMultiplePagesLRONextOptionalParams
  ): Promise<PagingGetMultiplePagesLRONextResponse> {
    const { span, updatedOptions } = createSpan(
      "PagingClient-_getMultiplePagesLRONext",
      coreHttp.operationOptionsToRequestOptionsBase(options || {})
    );
    const operationArguments: coreHttp.OperationArguments = {
      nextLink,
      options: updatedOptions
    };
    try {
      const result = await this.client.sendOperationRequest(
        operationArguments,
        getMultiplePagesLRONextOperationSpec
      );
      return result as PagingGetMultiplePagesLRONextResponse;
    } catch (error) {
      span.setStatus({
        code: CanonicalCode.UNKNOWN,
        message: error.message
      });
      throw error;
    } finally {
      span.end();
    }
  }
```


jonathandturner:
I see we're using `getNoItemNamePages` in the example, and in KeyVault we use `list*` (with `list` instead of `get` and no `Pages` at the end).

Also, are you thinking of adding the spans in a later PR?

Can you say a bit more about `_getNoItemNamePages`? This doesn't seem to be referenced in the example.

joheredi:
> I see we're using `getNoItemNamePages` in the example, and in KeyVault we use `list*` (with `list` instead of `get` and no `Pages` at the end).
> 
> Also, are you thinking of adding the spans in a later PR?
> 
> Can you say a bit more about `_getNoItemNamePages`? This doesn't seem to be referenced in the example.

Thanks for reminding me about the spans! I'll add them in this PR.

These test clients are being generated from swaggers in the TestServer ([link](https://github.com/Azure/autorest.testserver)). The name `getNoItemNamePages` comes directly from the name defined in the [paging swagger](https://github.com/Azure/autorest.testserver/blob/79e6a195ce625a3c26f18dc7c9f4d0aabd4219a6/swagger/paging.json#L22), it was not created by the generator. In a real-life client, to make this name conform to our guidelines we'd set transforms in the configuration readme file adding `x-ms-client-name = "listNoItemName"` or even better, add that in the Swagger itself (if it makes sense across languages)

We'll usually have 2 private methods prefixed with `_` per paging operation. One for the initial call and one for getting the `next` page which takes the continuationToken. Before this change, these would just be public methods and the consumer would need to create the loop and link them together.

When enabling AsyncIterators, those 2 methods become private and we expose a single method to the user, which takes care of looping and linking the initial and next operations.

I'm also adding his info to the description

xirzec:
I think the codegen looks pretty good here.

I'm not sure how to put spans in the iterator itself, see https://github.com/Azure/azure-sdk-for-js/issues/5607

Since the guidelines say paged methods should begin with `list` I think we should find a way to enforce that: https://azure.github.io/azure-sdk/typescript_design.html#ts-pagination-provide-list

joheredi:
Spoke with @xirzec offline about tracing and agreed we'll set tracing in the initial and next operations. 

@jonathandturner I also added some basic enforcement of the `list` naming. Basically we'll check if it already starts with `list` we keep the name, if it starts with `get` we replace it with `list` otherwise we prefix the name with `list`.

deyaaeldeen:
I did not look at the changes yet but from the the description I see that you are not allowing custom page sizes in the `byPage` method. I wonder if there was any discussion around having a standardized way to do this in our swagger definitions. For instance, in cognitive libraries, our continuation token is basically a `top` and `skip` parameters where `top` controls the page size.

src/generators/utils/pagingOperations.ts
```diff
@@ -0,0 +1,378 @@
...
+  operationGroupClass: ClassDeclaration,
+  importedModels: Set<string>
+) {
+  if (clientDetails.options.disablePagingAsyncIterators) {
+    return;
+  }
+
+  operationGroupDetails.operations
+    // We can skip "next" operations since covering the original is enough. Otherwise we'll end up with duplicate methods.
+    .filter(o => o.pagination && !o.pagination?.isNextLinkMethod)
```
sarangan12, line 111: is that ```?``` after pagination required?

sarangan12 (approved):
Left a minor comment. Otherwise LGTM

joheredi:
> I did not look at the changes yet but from the the description I see that you are not allowing custom page sizes in the `byPage` method. I wonder if there was any discussion around having a standardized way to do this in our swagger definitions. For instance, in cognitive libraries, our continuation token is basically a `top` and `skip` parameters where `top` controls the page size.

This is a good point! Unfortunately, we are not currently able to define these options in Swagger. However this is a good conversation to have to either expand the `x-ms-pageable` extension or allow some customization. I have filed #793 to track this

src/generators/utils/pagingOperations.ts
```diff
@@ -0,0 +1,378 @@
...
+ */
+export function hasAsyncIteratorOperations(
+  operationGroupDetails: OperationGroupDetails
+) {
+  return operationGroupDetails.operations.some(o => o.pagination);
+}
+
+/**
+ * This function prepares the initial and next operations to be generated using
+ * AsyncIterators, adding some extra metadata sunc as scope and prefix
```
deyaaeldeen, line 74:
```suggestion
 * AsyncIterators, adding some extra metadata such as scope and prefix
```

src/generators/utils/pagingOperations.ts
```diff
@@ -0,0 +1,378 @@
...
+          operation.pagination?.nextLinkOperationName?.toLocaleLowerCase()
+      );
+
+      // We need to figure out the body response type to be able to set the return types properly
+      const bodyResponseType = getPagingResponseBodyType(operation);
+
+      // Since we'll be using this type name for importing, we need to make sure that it is not in
+      // the form of array [], just need the name.
+      let bodyResponseTypeName =
+        bodyResponseType?.typeName.replace("[]", "") || "";
```
deyaaeldeen, line 135: I think the assumption here is that `bodyResponseType` is of the form `T[]` where `T` is a type name. I think we need to make sure this assumption is checked just in case. Otherwise, I would rather work with the type as a structure rather than as a string.

src/generators/utils/pagingOperations.ts
```diff
@@ -0,0 +1,378 @@
...
+      } = getOperationParameterSignatures(
+        operation,
+        clientDetails.parameters,
+        importedModels,
+        operationGroupClass
+      );
+
+      // Build an object with all the information about the paging methods
+      // while generating each of the paging methods, this will help up access
+      // information about the other methdos.
```
deyaaeldeen, line 179:
```suggestion
      // information about the other methods.
```

src/generators/utils/pagingOperations.ts
```diff
@@ -0,0 +1,378 @@
...
+        publicMethod: {
+          name: getPublicMethodName(initialOperationName),
+          parameters: initialMethodParameters
+        },
+        allMethod: {
+          name: `${initialOperationName}PagingAll`,
+          parameters: initialMethodParameters
+        },
+        pageMethod: {
+          name: `${initialOperationName}PagingPage`,
```
deyaaeldeen, line 200: interesting name...

src/generators/utils/pagingOperations.ts
```diff
@@ -0,0 +1,378 @@
...
+
+/**
+ * Generates the content of the public method, here we reference the other 2 methods All and Page
+ */
+function writePublicMethod(
+  operation: OperationDetails,
+  operationGroupClass: ClassDeclaration,
+  pagingMethodSettings: PagingMethodSettings
+) {
+  const returnType = `PagedAsyncIterableIterator<${pagingMethodSettings.bodyResponseType}, ${pagingMethodSettings.bodyResponseType}[]>`;
```
deyaaeldeen, line 239: I think the default type for the second type argument is already the array version of the first, so perhaps we can be less verbose here:
```suggestion
  const returnType = `PagedAsyncIterableIterator<${pagingMethodSettings.bodyResponseType}>`;
```

src/generators/utils/pagingOperations.ts
```diff
@@ -0,0 +1,378 @@
...
+
+  // Extract the parameter names for the All method to call it
+  let allMethodParameters = pagingMethodSettings.allMethod.parameters
+    .map(p => p.name)
+    .join(",");
+
+  // Extract the parameter names for the page method to call it
+  const pageMethodNameParams = pagingMethodSettings.pageMethod.parameters
+    .map(p => p.name)
+    .join(",");
```
deyaaeldeen, line 262: commas is the default behavior so perhaps we can be less verbose here as well unless you prefer to keep it for readability.
```suggestion
    .join();
```

src/generators/utils/pagingOperations.ts
```diff
@@ -0,0 +1,378 @@
...
+        },
+        byPage: () => {
+          return this.${pagingMethodSettings.pageMethod.name}(${pageMethodNameParams});
+        }
+      };`
+  ]);
+}
+
+/**
+ * Generates the All method which loops through all the pages and returns a single array with all the results
```
deyaaeldeen, line 281:
```suggestion
 * Generates the All method which loops through all the pages and returns an iterator for all the results
```

src/generators/utils/pagingOperations.ts
```diff
@@ -0,0 +1,378 @@
...
+    returnType,
+    isAsync: true
+  });
+
+  let firstRequestStatements = [
+    `let result = await this.${pagingMethodSettings.initialMethod.name}(${initialMethodParameters});`
+  ];
+
+  if (operation.isLRO) {
+    // Since this is also an LRO operation, we need to poll until done to get the result
```
deyaaeldeen, line 343: can you put an example for generated methods for a paginated LRO API in the PR description?

src/generators/utils/responseTypeUtils.ts
```diff
@@ -32,3 +36,65 @@ export function getResponseTypeName(
...
+
+  return "coreHttp.RestResponse";
+}
+
+/**
+ * This function extracts the body type for pageable operations. This is used to later on
+ * be able to return an array of items, instead of the Response objects. This will get the type
+ * of the "value" property from the response on a pageable operation.
+ */
+export function getPagingResponseBodyType(operation: OperationDetails) {
```
deyaaeldeen, line 70: Please add a return type.

deyaaeldeen (approved):
Looks great! Thank you so much for taking on such a complex feature! I left a few comments mostly minor.

src/generators/utils/pagingOperations.ts
joheredi: I was giving it only Page but it was running into some naming conflicts, so I wanted a more unique name

src/generators/utils/pagingOperations.ts
joheredi: Yeah, good suggestion

src/generators/utils/pagingOperations.ts
joheredi: Sure, added it to the description

src/generators/utils/pagingOperations.ts
joheredi: This is used to import the required Model, removing [] is a safeguard to make sure we don't try to do `import {Model[]} from "./"`

If the typeName is not an array, it should be fine the replace will have no effect

PR #794, opened 2020-11-19
Fix unused imports and adding esnext to tsconfig

joheredi:
There are 2 small issues when generating Pageable operations:

1. Some OperationGroup files are importing `PagedAsyncIterableIterator` even though they have no pageable operations. This was caused because when we were checking for the existence of a pageable operation, we were checking across operation groups instead of just checking the current operation group.

2. Missing `esnext` in `tsconfig.json` causes build error: `Property 'asyncIterator' does not exist on type 'SymbolConstructor'`

### DIFF
--- a/src/generators/indexGenerator.ts
+++ b/src/generators/indexGenerator.ts
@@ -13,9 +13,12 @@ export function generateIndexFile(
     }
   );
 
-  const hasOperationsFile = clientDetails.operationGroups.some(
-    og => !og.isTopLevel
-  );
+  if (
+    clientDetails.options.hasPaging &&
+    !clientDetails.options.disablePagingAsyncIterators
+  ) {
+    indexFile.addStatements([`/// <reference lib="esnext.asynciterable" />`]);
+  }
 
   indexFile.addExportDeclarations([
     {
--- a/src/generators/static/tsConfigFileGenerator.ts
+++ b/src/generators/static/tsConfigFileGenerator.ts
@@ -2,8 +2,18 @@
 // Licensed under the MIT License.
 
 import { Project } from "ts-morph";
+import { ClientDetails } from "../../models/clientDetails";
 
-export function generateTsConfig(project: Project) {
+export function generateTsConfig(
+  project: Project,
+  clientDetails: ClientDetails
+) {
+  const esNext =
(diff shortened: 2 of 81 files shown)


src/generators/static/tsConfigFileGenerator.ts
```diff
@@ -15,7 +25,7 @@ export function generateTsConfig(project: Project) {
       esModuleInterop: true,
       allowSyntheticDefaultImports: true,
       forceConsistentCasingInFileNames: true,
-      lib: ["es6", "dom"],
+      lib: ["es6", "dom", ...esNext],
```
deyaaeldeen, line 28: This could cause problems to the users because this way `esnext` is in the compiled sources which brings all their definitions into scope at runtime even if the user does not want to use it. I think a better approach is to include the `asyncIterable` package only in compiled sources by having a triple slash directive in `index.ts` file (e.g. [here](https://github.com/Azure/azure-sdk-for-js/blob/2c2e35f6f986afbf7f1a6568678e0a89a4b59a4c/sdk/textanalytics/ai-text-analytics/src/index.ts#L4)).

deyaaeldeen (changes requested):
I agree we need to fix these issues but I disagree with approach to use the entire `esnext`. This was something I have bumped into recently and learned how to do it from @willmtemple and @sadasant.

src/generators/static/tsConfigFileGenerator.ts
joheredi: Thanks for bringing this up! I'm fixing it

src/generators/indexGenerator.ts
```diff
@@ -13,9 +13,12 @@ export function generateIndexFile(
     }
   );
 
-  const hasOperationsFile = clientDetails.operationGroups.some(
-    og => !og.isTopLevel
-  );
```
deyaaeldeen, line 18: why is this var gone? is it not used anywhere?

deyaaeldeen (approved):
Looks great! Thanks!

src/generators/indexGenerator.ts
joheredi: Yeah it was not used at all, just cleaned it up

sadasant (approved):
Thank you for doing this!

PR #795, opened 2020-11-20
Improving support for extensible enums

joheredi:
**Problem:**
Currently Autorest generates extensible enums as an union type of the options plus the type i.e

```typescript
type Status = "success" | "failure" | "canceled" | string
```

However Typescript collapses it and the type ends up being `string` and IDEs such as VSCode don't provide any auto complete help.

**Soulution:**
The solution involves 2 parts:

1. Document all possible options with JSDoc while typing the "enum" as string
2. Generate a helper enum with the know options supported by the service.

**Note:** The helper enum is only generated for string and numbers. Other types only get (1). Autorest only supports enums of types `string`, `number` and `boolean`.

### Before

```typescript
/**
 * Defines values for EnforcementMode.
 */
export type EnforcementMode = "Default" | "DoNotEnforce" | string;
```

**VSCode docs:**
image

### After
**Generated Code**
```typescript
/**
 * Known values of {@link EnforcementMode} that the service accepts.
 */
export const enum KnownEnforcementMode {
  /**
   * The policy effect is enforced during resource creation or update.
   */
  Default = "Default",
  /**
   * The policy effect is not enforced during resource creation or update.
   */
  DoNotEnforce = "DoNotEnforce"
}

/**
 * Defines values for EnforcementMode. \
 * {@link KnownEnforcementMode} can be used interchangeably with EnforcementMode,
 *  this enum contains the known values that the service supports.
 * ### Know values supported by the service
 * **Default**: The policy effect is enforced during resource creation or update. \
 * **DoNotEnforce**: The policy effect is not enforced during resource creation or update.
 */
export type EnforcementMode = string;

```

**VSCode docs**
image

### DIFF
--- a/src/generators/modelsGenerator.ts
+++ b/src/generators/modelsGenerator.ts
@@ -41,6 +41,7 @@ import {
   getAllModelsNames
 } from "./utils/responseTypeUtils";
 import { getParameterDescription } from "../utils/getParameterDescription";
+import { UnionDetails } from "../models/unionDetails";
 
 export function generateModels(clientDetails: ClientDetails, project: Project) {
   const modelsIndexFile = project.createSourceFile(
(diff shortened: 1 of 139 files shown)


bterlson (approved):

xirzec (approved):
LGTM

ffMathy:
In my opinion, this made everything worse. Now all my enums are essentially just like `string`. I know I can refer the "Known" version, but it's not enforced in my code in requests and responses, which I think is much more critical than being able to "extend" the enums.

xirzec:
> In my opinion, this made everything worse. Now all my enums are essentially just like `string`. I know I can refer the "Known" version, but it's not enforced in my code in requests and responses, which I think is much more critical than being able to "extend" the enums.

@ffMathy I feel your pain here. The major issue is the service deciding that they don't want to commit to non-breaking API changes for these fields. 

Since the server can return any string without it being considered breaking, modeling it as a strongly typed enum seems misleading at best and might cause issues in your otherwise typesafe code, for example:

```ts
enum Example {
  a = "a",
  b = "b",
  c = "c"
}

function doSometing(value: Example) {
  if (value === Example.a) {
    console.log("a")
  } else if (value === Example.b) {
    console.log("b");
  } else {
    console.log("TypeScript assumes value is Example.c");
  }
}
```

If you were switching off the enum value and a response came back with a new value, your else would be invalid and your production code could break.

If you want to opt-out, perhaps you could declare your own interfaces that used the Known version instead? 

At any rate, happy to continue this discussion in an issue if you'd like to file one.

PR #799, opened 2020-12-02
Reliably add generated pipeline policies to default

joheredi:
They way we add LRO policy and honor policies passed in `options.requestPolicyFactories` is not reliable  as we effectively bypass the default policyFactories added by `core-http`, the reason for this is that passing requestPolicyFactories as an array of factories tells `core-http` to not try adding any of the default policies see [here](https://github.com/Azure/azure-sdk-for-js/blob/98d87615a511484ed797cb1b91102432e1b16e70/sdk/core/core-http/src/serviceClient.ts#L201).

The fix in this PR is to handle the 3 scenarios we can face when working with LRO and policies

1) We are given an array of policy factories by the user. In this scenario the user decides to provide a list of all the pipelines they want to have and `core-http` won't add any default pipelines. We append the LRO policy to the array provided

2) We are given a function in the form `(defaultPolicies) =>  {}` which is for the scenario where the user wants to provide a set of policies to add to the default policies created by `core-http` in this scenario we append the LRO policy to the policies provided by the user which will eventually get added to the default policies by `core-http`

3) The user doesn't provide any pipelinePolicies, in this case we just add the LRO policy to the default policies that core-http will eventually add.

This fix addresses a related issue where the Authorization header is not being set, this was caused by us passing the LRO policy in an array by itself in `options.requestPolicyPipelines` which bypasses all default pipelines, including the auth pipeline

### DIFF
--- a/src/generators/clientContextFileGenerator.ts
+++ b/src/generators/clientContextFileGenerator.ts
@@ -174,16 +174,33 @@ const writeStatements = (lines: string[], shouldAddBlankLine = false) => (
 };
 
 function writeDefaultOptions(hasCredentials: boolean, hasLRO: boolean) {
-  const policies = ` 
-  const defaultPipelines = Array.isArray(options.requestPolicyFactories)
-  ? options.requestPolicyFactories
-  : coreHttp.createPipelineFromOptions(options)
-      .requestPolicyFactories as coreHttp.RequestPolicyFactory[];
-
-  options = {
-      ...options,
-      requestPolicyFactories: [lroPolicy(), ...defaultPipelines]
-    };`;
+  const addLROPolicy = hasLRO
+    ? `
+    // Building the rquest policy fatories based on the passed factories and the
+    // any required factories needed by the client.
+    if (Array.isArray(options.requestPolicyFactories)) {
+      // When an array of factories is passed in, we'll just add the required factories,
+      // in this case lroPolicy(). It is important to note that passing an array of factories
+      // to a new client, bypasses core-http default factories. Just the pipelines provided will be run.
+      options.requestPolicyFactories = [lroPolicy(), ...options.requestPolicyFactories]
+    } else if (options.requestPolicyFactories) {
+      // When we were passed a requestPolicyFactories as a function, we'll create a new one that adds the factories provided
+      // in the options plus the required policies. When using this path, the pipelines passed to the client will be added to the
+      // default policies added by core-http
+      const optionsPolicies = options.requestPolicyFactories([lroPolicy()]) || [
+        lroPolicy()
+      ];
+      options.requestPolicyFactories = defaultFactories => [
+        ...optionsPolicies,
+        ...defaultFactories
+      ];
+     
+    } else {
+      // In case no request policy factories were provided, we'll just need to create a function that will add 
+      // the lroPolicy to the default pipelines added by core-http
+      options.requestPolicyFactories = (defaultFactories) => ([lroPolicy(), ...defaultFactories])
+    }`
+    : "";
 
   return `// Initializing default values for options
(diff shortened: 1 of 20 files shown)


src/generators/clientContextFileGenerator.ts
```diff
@@ -174,16 +174,33 @@ const writeStatements = (lines: string[], shouldAddBlankLine = false) => (
...
-  : coreHttp.createPipelineFromOptions(options)
-      .requestPolicyFactories as coreHttp.RequestPolicyFactory[];
-
-  options = {
-      ...options,
-      requestPolicyFactories: [lroPolicy(), ...defaultPipelines]
-    };`;
+  const addLROPolicy = hasLRO
+    ? `
+    // Building the rquest policy fatories based on the passed factories and the
```
sarangan12, line 179: Minor typo

src/generators/clientContextFileGenerator.ts
```diff
@@ -174,16 +174,33 @@ const writeStatements = (lines: string[], shouldAddBlankLine = false) => (
...
+    // any required factories needed by the client.
+    if (Array.isArray(options.requestPolicyFactories)) {
+      // When an array of factories is passed in, we'll just add the required factories,
+      // in this case lroPolicy(). It is important to note that passing an array of factories
+      // to a new client, bypasses core-http default factories. Just the pipelines provided will be run.
```
sarangan12, line 184: What is the passed in values already have an lroPolicy?

test/integration/azureSpecialProperties.spec.ts
```diff
@@ -12,6 +13,35 @@ import {
   HttpHeaders
 } from "@azure/core-http";
 
+describe.only("auth validation", () => {
+  it("should add authorization header", async () => {
```
sarangan12, line 17: This test case looks fine. But, is it really related to the lroPolicy changes?

test/utils/smoke-test-list.ts
```diff
@@ -14,20 +14,20 @@ export enum AutorestParams {
...
 const getArmReadmes = (): SpecDefinition[] => {
   const armTags = [
-    "package-features-2015-12",
-    "package-locks-2016-09",
-    "package-policy-2019-09",
+    // "package-features-2015-12",
+    // "package-locks-2016-09",
+    // "package-policy-2019-09",
     "package-resources-2019-08",
-    "package-subscriptions-2019-06",
```
sarangan12, line 21: Any specific reason these were removed?

sarangan12 (changes requested):

test/utils/smoke-test-list.ts
joheredi: this was a mistake thanks for catching it!

test/integration/azureSpecialProperties.spec.ts
joheredi: One of the side effects of having a faulty requestPolicyFactories option was that the default authentication policy was being inadvertently bypassed. This test is here to make sure we don't regress again in building the expected default auth pipeline

src/generators/clientContextFileGenerator.ts
joheredi: Yeah, I think this is one of the drawbacks of the current pipelines design, that we cannot identify what pipelines are already in place. We'll be able to handle this once we move to core-client and core-https.

Meanwhile we don't have a way to make sure we don't duplicate the lroPolicy. In the bright side, the lroPolicy is not publicly exposed so having someone passing it in when creating a new client wouldn't be too common. I think we should be safe in the meantime while we get migrated to corev2

sarangan12 (approved):

Issue #808, opened 2020-12-05
Export Operations in the v6 generator


sarangan12:
In the new v6 generated SDK, the index.ts is not exporting anything from operations. Eg: https://github.com/Azure/azure-sdk-for-js/pull/12787/files#diff-8224a291789d8d70439c3737d0f6a9728e77653acb53baf04c0ec58eacb5dc91R9

sarangan12:
Duplicate of Issue https://github.com/Azure/autorest.typescript/issues/856

Issue #807, opened 2020-12-05
Remove reference to "@azure/core-arm"


sarangan12:
In the new v6 generator, the rollup.config.js file still mentions "@azure/core-arm". We need to remove it.

sarangan12:
Code complete with https://github.com/Azure/autorest.typescript/pull/811

PR #810, opened 2020-12-07
Fix operations export

sarangan12:
This PR fixes the following issues:

1. **[Remove reference to "@azure/core-arm" ](https://github.com/Azure/autorest.typescript/issues/807)**
We have already deprecated @azure/core-arm package and deleted the source code from our repositories. But, the new generator references this package in the rollup config file. So, this PR removes that reference. 

2. **[Export Operations in the v6 generator](https://github.com/Azure/autorest.typescript/issues/808)**
In the generated SDK's src/index.ts file, we are not exporting the operations which is not correct. We need to export it. This PR adds code changes to add an export line for the operations in the src/index.ts file. 

But, while fixing this code, I encountered different issue. In some of the swagger files, the operation group name conflicts with the model name. That has to be fixed. There are 2 options:

1. I can change it in the autorest.testserver. But, I think that is not required. The reason is, to the best of my knowledge, this affects only TS SDK. So, we don't want to change it in the common place. 

2. I can copy the file locally (to this repo) and modify it. The only problem is we are losing the advantage of the common place for all test files. But, I think it is ok since we have only 3 such files. 

(Note: This problem did not occur in the old generated as the models is exported with '<XYZ>models'. Also, the operation group have suffix 'operations' added. So, there will not be any conflict.)

@ramya-rao-a FYI.....

### DIFF
--- a/test/utils/test-swagger-gen.ts
+++ b/test/utils/test-swagger-gen.ts
@@ -74,13 +74,13 @@ const testSwaggers: { [name: string]: SwaggerConfig } = {
     licenseHeader: true
   },
   bodyComplex: {
-    swagger: "body-complex.json",
+    swagger: "test/integration/swaggers/body-complex.json",
     clientName: "BodyComplexClient",
     packageName: "body-complex",
     licenseHeader: true
   },
   bodyComplexWithTracing: {
-    swagger: "body-complex.json",
+    swagger: "test/integration/swaggers/body-complex.json",
     clientName: "bodyComplexWithTracing",
     packageName: "body-complex-tracing",
     licenseHeader: true,
@@ -344,13 +344,13 @@ const testSwaggers: { [name: string]: SwaggerConfig } = {
     licenseHeader: true
   },
   extensibleEnums: {
-    swagger: "extensible-enums-swagger.json",
+    swagger: "test/integration/swaggers/extensible-enums-swagger.json",
     clientName: "ExtensibleEnumsClient",
     packageName: "extensible-enums",
     licenseHeader: true
   },
   xmsErrorResponses: {
-    swagger: "xms-error-responses.json",
+    swagger: "test/integration/swaggers/xms-error-responses.json",
     clientName: "XmsErrorResponsesClient",
     packageName: "xms-error-responses",
     licenseHeader: true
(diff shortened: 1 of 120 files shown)


test/utils/test-swagger-gen.ts
```diff
@@ -74,13 +74,13 @@ const testSwaggers: { [name: string]: SwaggerConfig } = {
     licenseHeader: true
   },
   bodyComplex: {
-    swagger: "body-complex.json",
+    swagger: "test/integration/swaggers/body-complex.json",
```
joheredi, line 77: I don't love this approach. I think the right thing to do is to detect collisions and suffix either the conflicting model with "Model" or the operation group class with "Operations" rather than changing the swaggers. When this is a problem across languages it makes sense to fix the swaggers, however in this case the collisions are unique to our way to do things in the Typescript generator.

On the other hand, we need these tests to be updated and in sync with the testserver, once we move them here we'll stop updating them.

Also the same way these 3 clients broke, we'll potentially break other existing generated libraries. To make sure we don't break existing libraries with the change in this PR is to make sure we detect collisions and give non-colliding names to those operationGroups and/or Models

joheredi (changes requested):
Thanks Sarangan, for the goal (1) of this PR I agree, thanks for removing core-arm from the dependencies!

For (2) I'm not sure I understand the reason why the operation classes need to be exported. Would you mind sharing an example of a scenario where someone needs to access the operation group class directly and how the interaction with that class works?

IMO the operation groups should only be accessible through the client itself, otherwise it can become a bit confusing for the users when they se a class for example `TableOperations` they would be tempted to try using it to `CreateTable` however when trying to instantiate `TableOperations` to find out that a client is needed to instantiate it, and the client contains table operations.

This is important to consider as we'd like to be able to have libraries published directly from the generated code, so we need to be careful with the public API we provide.

I can see the scenario where a Library has a convenience layer on top of the generated code and wants to have its own Client, however in that scenario, one would have direct access to each of the operation group files to import them without the need to expose them to the public API.

sarangan12:
I am closing this PR in favor of https://github.com/Azure/autorest.typescript/pull/811. For the exporting operations, after discussion with Ramya, I have decided to take that issue later in January 2021.

Issue #817, opened 2020-12-10
Multiple Headers causing SDK generation failure


sarangan12:
Look at the following swagger:

```
"delete": {
        "tags": [
          "KeyValues"
        ],
        "summary": "Deletes a key-value.",
        "operationId": "DeleteKeyValue",
        "consumes": [],
        "produces": [
          "application/vnd.microsoft.appconfig.kv+json",
          "application/json",
          "application/problem+json"
        ],
        "parameters": [
          {
            "name": "key",
            "in": "path",
            "description": "The key of the key-value to delete.",
            "required": true,
            "type": "string"
          },
          {
            "name": "label",
            "in": "query",
            "description": "The label of the key-value to delete.",
            "type": "string"
          },
          {
            "$ref": "#/parameters/SyncTokens"
          },
          {
            "$ref": "#/parameters/ApiVersion"
          },
          {
            "name": "If-Match",
            "in": "header",
            "description": "Used to perform an operation only if the targeted resource's etag matches the value provided.",
            "type": "string"
          }
        ],
        "responses": {
          "200": {
            "description": "Success",
            "schema": {
              "$ref": "#/definitions/KeyValue"
            },
            "headers": {
              "Sync-Token": {
                "description": "Enables real-time consistency between requests by providing the returned value in the next request made to the server.",
                "type": "string"
              },
              "ETag": {
                "description": "An identifier representing the returned state of the resource.",
                "type": "string"
              }
            }
          },
          "204": {
            "description": "Success",
            "headers": {
              "Sync-Token": {
                "description": "Enables real-time consistency between requests by providing the returned value in the next request made to the server.",
                "type": "string"
              }
            }
          },
          "default": {
            "description": "Error response describing why the operation failed",
            "schema": {
              "$ref": "#/definitions/Error"
            }
          }
        },
        "x-ms-examples": {
          "DeleteKeyValue": {
            "$ref": "./examples/DeleteKeyValue.json"
          },
          "DeleteKeyValue_IfMatch": {
            "$ref": "./examples/DeleteKeyValue_IfMatch.json"
          }
        }
      }
```

The SDK generation will fail will fail

sarangan12:
Code complete with https://github.com/Azure/autorest.typescript/pull/818

Issue #816, opened 2020-12-10
@nextLink is causing SDK generation failure


sarangan12:
```
"KeyListResult": {
      "description": "The result of a list request.",
      "type": "object",
      "properties": {
        "items": {
          "description": "The collection value.",
          "type": "array",
          "items": {
            "$ref": "#/definitions/Key"
          }
        },
        "@nextLink": {
          "description": "The URI that can be used to request the next set of paged results.",
          "type": "string"
        }
      }
    }
```

In the latest generator, for the above Swagger, the code will be generated as:

```
export const KeyListResult: coreHttp.CompositeMapper = {
  type: {
    name: "Composite",
    className: "KeyListResult",
    modelProperties: {
      items: {
        serializedName: "items",
        type: {
          name: "Sequence",
          element: {
            type: {
              name: "Composite",
              className: "Key"
            }
          }
        }
      },
      @nextLink: { <<<<<< Error
        serializedName: "@nextLink",
        type: {
          name: "String"
        }
      }
    }
  }
};
```

This error causes issues in multiple locations within the SDK

sarangan12:
Code complete with https://github.com/Azure/autorest.typescript/pull/818

PR #818, opened 2020-12-10
Fix app configuration

sarangan12:
This PR fixes 2 issues:

1. **[@nextLink is causing SDK generation failure](https://github.com/Azure/autorest.typescript/issues/816)**
The problem is simple and straightforward. When the nextLinkName has a ```@``` at the beginning it must be cleaned up before usage. This PR does the cleaning.

2. **[Multiple Headers causing SDK generation failure](https://github.com/Azure/autorest.typescript/issues/817)**
If you look at the swagger (in the issue), in the responses section, you can see that both responses have headers. In such a case, the Code gen is assuming they are two different objects (one for each response). This PR fixes it by checking the schema names. If they are the same, it merges the properties.

@ramya-rao-a FYI....


src/utils/extractHeaders.ts
```diff
@@ -54,7 +55,28 @@ function processHeaders(
     const headers = response.protocol.http?.headers;
     if (headers) {
       const headerSchema = headersToSchema(headers, operationName, isException);
-      headerSchema && responseHeaders.push(headerSchema);
+      if (headerSchema) {
+        let pushHeaderSchema: boolean = true;
+        for (let responseHeader of responseHeaders) {
+          if (
+            getLanguageMetadata(responseHeader.language).name ==
```
joheredi, line 62: I think this looks good. For readability, would you mind extracting this section into a separate function and add some comments please?

joheredi (approved):
Looks good, just left a small comment to help readability

test/integration/swaggers/appconfiguration.json
```diff
@@ -0,0 +1,1347 @@
+{
```
joheredi, line 1: Would you mind either

1) Add a simple appconfig.spec.ts test that instantiates this client or
2) Add this to the smoke tests instead of integration

This is so that we also get build validation. We currently don't build the integration test unless they are being actually tested, while smoke-test generates and the builds the generated project

src/utils/extractHeaders.ts
sarangan12: Done

test/integration/swaggers/appconfiguration.json
sarangan12: Done

PR #827, opened 2021-01-05
Low level client prototype

joheredi:


ramya-rao-a:
@joheredi Can we close this if it is no longer needed?

PR #831, opened 2021-01-11
Add hidden to generated class context

sarangan12:
This PR is to address the issue reported at https://github.com/Azure/autorest.typescript/issues/832. A detailed description has been provided in the issue. 

In this PR, I have done the following changes:

1. Updated ```ts-morph``` to version ```9.0.0``` from ```5.0.0```. One Year back, when this project was created, the latest version of ```ts-morph``` was ```5.0.0```. It did not have the ability to add tags which is what we are doing in this PR. That feature was introduced later. So, I upgraded it to latest version. 

2. With the latest upgrade to ```ts-morph``` the documentation comments (with only one line) have changed to one line format from multi-line format. 

3. I have introduced the option ```has-custom-layer```. In this PR, this option is used to determine if we need to hide the Client and Client Context classes. In the future, we could extend its usage.

@joheredi @deyaaeldeen Please review and approve.


src/generators/clientContextFileGenerator.ts
```diff
@@ -20,7 +20,8 @@ import { formatJsDocParam } from "./utils/parameterUtils";
 export function generateClientContext(
   clientDetails: ClientDetails,
   packageDetails: PackageDetails,
-  project: Project
+  project: Project,
+  hasCustomLayer: boolean
```
deyaaeldeen, line 24: Perhaps the name is a bit unclear. One alternative could be `exportedByPackage`. Either way, please document it.

deyaaeldeen (commented):
Looks reasonable to me but I am not sure why comments changed from:
```
/**
 *
 */
```
to:
`/** */`

Is that because of the upgrade to ts-morph 9?

src/generators/clientContextFileGenerator.ts
sarangan12: The ```hasCustomLayer``` could be used for multiple purposes in the future. Not just hiding classes. Added some documentation for the property.

joheredi (approved):
Looks good! Thanks for fixing this, as we discussed offline the only comment is to use the flag name `hideClients` instead.

Issue #833, opened 2021-01-14
Property 'required' missing in the Spec Mapper


sarangan12:
I found this issue while working with ```keyvault-certificates```. In the KeyVault Swagger, look [here](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/keyvault/data-plane/Microsoft.KeyVault/preview/7.2-preview/certificates.json#L350): 

```
.......
.......
{
    "name": "parameter",
    "in": "body",
    "required": true,
    "x-ms-client-flatten": true,
    "schema": {
      "$ref": "#/definitions/CertificateIssuerUpdateParameters"
    },
    "description": "Certificate issuer update parameter."
}
.........
.........
```

The above code is translated, in the old generator, as mentioned [here](https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/keyvault/keyvault-certificates/src/generated/keyVaultClient.ts#L1182):

```
........
........
requestBody: {
    parameterPath: {
      provider: [
        "options",
        "provider"
      ],
      credentials: [
        "options",
        "credentials"
      ],
      organizationDetails: [
        "options",
        "organizationDetails"
      ],
      attributes: [
        "options",
        "attributes"
      ]
    },
    mapper: {
      ...Mappers.CertificateIssuerUpdateParameters,
      required: true
    }
  }
...........
...........
```

But, with the new V6 generator, the code is different as mentioned [here](https://github.com/Azure/azure-sdk-for-js/blob/3cf07ab15f43415eff3514abb5303d1cb1c3a78d/sdk/keyvault/keyvault-certificates/src/generated/keyVaultClient.ts#L957): 

```
............
............
  requestBody: {
    parameterPath: {
      provider: ["options", "provider"],
      credentials: ["options", "credentials"],
      organizationDetails: ["options", "organizationDetails"],
      attributes: ["options", "attributes"]
    },
    mapper: Mappers.CertificateIssuerUpdateParameters
  },
..............
..............
```

As you can see the required option is missing in the new V6 generator. This causes the [test ](https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/keyvault/keyvault-certificates/test/public/CRUD.spec.ts#L515)in the keyvault-certificates fail. This must be fixed ASAP for the successful migration of keyvault-certificates.

@joheredi @ramya-rao-a @sadasant FYI....

sarangan12:
Code Completed with PR: https://github.com/Azure/autorest.typescript/pull/834

PR #834, opened 2021-01-14
Property 'required' missing in the Spec Mapper

sarangan12:
This PR is to address the issue https://github.com/Azure/autorest.typescript/issues/833. The issue has a detailed description of the problem and the solution.


src/generators/operationGenerator.ts
```diff
@@ -277,6 +286,19 @@ function buildRequestBody({
 
   // First get the mapper, this will be the mapper for the parameter before flattening
   const mapper = requestBody[0].mapper;
+  let required: boolean = requestBody[0].required ?? false;
+
+  if (!required) {
+    for (const rb of requestBody) {
```
joheredi, line 292: Can you please add a comment about what this block is doing?

joheredi (approved):
Looks good! Thanks for the fix!

src/generators/operationGenerator.ts
sarangan12: Done

PR #839, opened 2021-01-19
Fix Names specifically overridden in Configuration Files

sarangan12:
This is to fix the naming issue related to search SDK - https://github.com/Azure/autorest.typescript/issues/840.

### DIFF
--- a/src/utils/nameUtils.ts
+++ b/src/utils/nameUtils.ts
@@ -137,6 +137,9 @@ export function normalizeName(
   nameType: NameType,
   shouldGuard?: boolean
 ): string {
+  if (name.startsWith("$DO_NOT_NORMALIZE$")) {
+    return name.replace("$DO_NOT_NORMALIZE$", "");
+  }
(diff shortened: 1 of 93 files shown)


src/utils/nameUtils.ts
```diff
@@ -137,6 +137,9 @@ export function normalizeName(
   nameType: NameType,
   shouldGuard?: boolean
 ): string {
+  if (name.startsWith("$DO_NOT_NORMALIZE$")) {
+    return name.replace("$DO_NOT_NORMALIZE$", "");
```
joheredi, line 141: Can you please file an issue in M4 to see if we can get an overridenName property in the language.default object

sarangan12: I do not have access to create issue in M4. But, I have sent an email detailing the issue.

joheredi (approved):
Looks good, hopefully we can get M4 handle this in a way we can get rid of the $ workaround

Issue #800, opened 2020-12-02
Api-Extractor fails to create review folder


joheredi:
This happens some times, we need to make sure that the script is able to create the review folder

sarangan12:
Code complete with PR https://github.com/Azure/autorest.typescript/pull/844

PR #844, opened 2021-01-21
Api-Extractor fails to create review folder

sarangan12:
This PR is to address the issue reported in the issue https://github.com/Azure/autorest.typescript/issues/800

@ramya-rao-a FYI.....


joheredi (approved):
Thanks for fixing this issue!

PR #846, opened 2021-01-21
adding TokenCredential as part of the credential type accepted

deyaaeldeen:
This PR adds support to v4 to accept `@azure/identity`'s `TokenCredential`. Most of the work for that support is already in `ms-rest-js` v2.2.0.

Things to note:
- we need typescript v3.6 to be able to work with abort-controller type definitions, a dependency of core-auth, which in turn is the home of `TokenCredential`.
- versions of ms-rest-js and ms-rest-azure-js are bumped to bring in the support for `TokenCredential` in them.

Todo:
- ~generate an updated README.md that promotes using `@azure/identity` based on https://github.com/Azure/azure-sdk-for-js/pull/13403.~


test/azure/generated/AzureReport/autoRestReportServiceForAzure.ts
```diff
@@ -17,10 +18,15 @@ import { AutoRestReportServiceForAzureContext } from "./autoRestReportServiceFor
 class AutoRestReportServiceForAzure extends AutoRestReportServiceForAzureContext {
   /**
    * Initializes a new instance of the AutoRestReportServiceForAzure class.
-   * @param credentials Credentials needed for the client to connect to Azure.
+   * @param credentials Credentials needed for the client to connect to Azure. The simplest
+   * TokenCredential credential can be obtained as follows:
```
joheredi, line 22: Can you please add a link to Identity's documentation so that they can learn about the ENV variables that need to be setup and more info?

deyaaeldeen: Fixed.

src/vanilla/Model/CodeModelTS.cs
```diff
@@ -588,7 +591,8 @@ public virtual void ConstructRuntimeImportForModelIndex(TSBuilder builder)
 
         public virtual void PackageDependencies(JSONObject dependencies)
         {
-            dependencies.StringProperty("@azure/ms-rest-js", "^2.0.4");
+            dependencies.StringProperty("@azure/ms-rest-js", "^2.2.0");
+            dependencies.StringProperty("@azure/core-auth", "^1.1.4");
```
deyaaeldeen, line 595: I think this dependency should not be added unless there is a credential, right? Is there a simple way to check for this condition here? cc @joheredi

joheredi: I think you could check either Settings.AddCredentials or Settings.AzureArm

test/azure/generated/AzureCompositeModelClient/azureCompositeModel.ts
```diff
@@ -29,10 +30,17 @@ class AzureCompositeModel extends AzureCompositeModelContext {
...
    * Initializes a new instance of the AzureCompositeModel class.
-   * @param credentials Credentials needed for the client to connect to Azure.
+   * @param credentials Credentials needed for the client to connect to Azure. The simplest
+   * TokenCredential credential can be obtained as follows:
+   * ```js
+   * const { DefaultAzureCredential } = require("@azure/identity");
+   * const credential = new DefaultAzureCredential();
+   * ```
+   * For more information about these credentials, see {@link
+   * https://github.com/Azure/azure-sdk-for-js/tree/master/sdk/identity/identity#authenticating-with-the-defaultazurecredential}
```
joheredi, line 40: Does this comment render correctly in VSCode tooltip>? I wonder if the docs tools would have any issues because `{@link` and link are not in the same line

joheredi (approved):
Looks great! Thanks for taking on this work, this is going to make things smoother for ARM users

test/azure/generated/AzureCompositeModelClient/azureCompositeModel.ts
ramya-rao-a: It would be odd that the constructor supports two credential types, but our docs for the parameter talks only about the second one. Moreover, we don't want to alarm users into thinking that the package they are currently using is not supported anymore. Because it is not easy to determine just looking at the type `msRest.ServiceClientCredentials` that their old code would still work.

I would recommend going with something like this:

> Credentials needed to authenticate the client using Azure Active Directory. 
Credentials implementing the `TokenCredential` interface from the @azure/identity package are recommended. 
Credentials implementing the `ServiceClientCredentials` interface from the older packages @azure/ms-rest-nodeauth and @azure/ms-rest-browserauth are also supported.

ramya-rao-a:
Are the changes from https://github.com/Azure/azure-sdk-for-js/pull/13403 going to be part of this PR or a separate one?
Where can I see a diff of the resulting changes?

deyaaeldeen:
@ramya-rao-a I will address feedback and generate the updated README.md in this PR. I will mention you in the diffs after I finish.

test/skip-subtypes/generated/BodyComplex/README.md
```diff
@@ -1,88 +1,83 @@
 ## An isomorphic javascript sdk for - AutoRestComplexTestService
 
-This package contains an isomorphic SDK for AutoRestComplexTestService.
+This package contains an isomorphic SDK (runs both in node.js and in browsers) for AutoRestComplexTestService.
```
deyaaeldeen, line 3: @ramya-rao-a , here is an updated readme file. @joheredi how can I get a display name for a package? I currently just use a normalized package name.

test/skip-subtypes/generated/BodyComplex/README.md
```diff
@@ -1,88 +1,83 @@
...
+Alternatively, you can add these to the dependencies section in your package.json and then run `npm install`.
 ```bash
 npm install
+npm install @azure/identity
 ```
+Please note that while the credentials from the older `@azure/ms-rest-nodeauth` and `@azure/ms-rest-browserauth` packages are still supported, these packages are in maintenance mode receiving critical bug fixes, but no new features.
+We strongly encourage you to use the credentials from `@azure/identity` where the latest versions of Azure Active Directory and MSAL APIs are used and more authentication options are provided.
 
-### How to use
+### How to Use
```
ramya-rao-a, line 24: This can stay lower case
```suggestion
### How to use

```

test/skip-subtypes/generated/BodyComplex/README.md
```diff
@@ -1,88 +1,83 @@
...
 ### Currently supported environments
 
-- Node.js version 6.x.x or higher
+- Node.js version 8.x.x or higher
 - Browser JavaScript
 
 ### How to Install
 
+To use this SDK in your project, you will need to install two packages.
+- `` that contains the client.
```
ramya-rao-a, line 13: Should we have some place holder?

test/skip-subtypes/generated/BodyComplex/README.md
```diff
@@ -1,88 +1,83 @@
...
-##### Install @azure/ms-rest-nodeauth
-
-- Please install minimum version of `"@azure/ms-rest-nodeauth": "^3.0.0"`.
-```bash
-npm install @azure/ms-rest-nodeauth@"^3.0.0"
-```
+There are multiple credentials available in the `@azure/identity` package to suit your different needs.
+Read about them in detail in [readme for @azure/identity package](https://www.npmjs.com/package/@azure/identity).
+To get started you can use the [DefaultAzureCredential](https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/identity/identity/README.md#defaultazurecredential) which tries different credentials internally until one of them succeeds.
+Most of the credentials would require you to register to [create an Azure App Registration](https://docs.microsoft.com/en-us/azure/active-directory/develop/app-objects-and-service-principals#application-registration) first.
```
ramya-rao-a, line 29:
```suggestion
Most of the credentials would require you to [create an Azure App Registration](https://docs.microsoft.com/en-us/azure/active-directory/develop/app-objects-and-service-principals#application-registration) first.

```

test/skip-subtypes/generated/BodyComplex/README.md
```diff
@@ -1,88 +1,83 @@
...
+#### nodejs - client creation and getValid basic as an example written in JavaScript.
 
 ##### Sample code
 
-While the below sample uses the interactive login, other authentication options can be found in the [README.md file of @azure/ms-rest-nodeauth](https://www.npmjs.com/package/@azure/ms-rest-nodeauth) package
 ```typescript
```
ramya-rao-a, line 34:
```suggestion
```javascript

```

test/skip-subtypes/generated/BodyComplex/README.md
```diff
@@ -1,88 +1,83 @@
...
+client.basic.getValid().then((result) => {
+  console.log("The result is:");
+  console.log(result);
 }).catch((err) => {
+  console.log("An error occurred:");
   console.error(err);
 });
 ```
 
 #### browser - Authentication, client creation and getValid basic as an example written in JavaScript.
```
ramya-rao-a, line 50: In the header for node, we don't have `Authentication`. We should add it there

test/skip-subtypes/generated/BodyComplex/README.md
```diff
@@ -1,88 +1,83 @@
...
-See https://github.com/Azure/ms-rest-browserauth to learn how to authenticate to Azure in the browser.
-
 - index.html
+
 ```html
 <!DOCTYPE html>
 <html lang="en">
   <head>
     <title> sample</title>
     <script src="node_modules/@azure/ms-rest-js/dist/msRest.browser.js"></script>
```
ramya-rao-a, line 64: We dont need msrest here anymore right?

test/skip-subtypes/generated/BodyComplex/README.md
deyaaeldeen: I addressed this by only capitalizing the first word in all subtitles in the file.

test/skip-subtypes/generated/BodyComplex/README.md
deyaaeldeen: I used the following as a place holder: "this package that contains the client".

test/skip-subtypes/generated/BodyComplex/README.md
deyaaeldeen: Fixed.

test/skip-subtypes/generated/BodyComplex/README.md
deyaaeldeen: Done.

test/skip-subtypes/generated/BodyComplex/README.md
deyaaeldeen: Done.

test/skip-subtypes/generated/BodyComplex/README.md
deyaaeldeen: Done.

PR #851, opened 2021-01-25
Use TypeScript v3.6.0 to be able to compile abort-controller

deyaaeldeen:
Otherwise, we will get the error in this log when adding a dependency for core-auth: https://dev.azure.com/azure-public/azsdk/_build/results?buildId=10244&view=logs&j=499e71a5-ab92-5cb7-8e77-9c19c2280f63&t=331a722e-d89d-5d12-1500-1d0f556ecaa0

```
node_modules/@azure/abort-controller/types/src/AbortController.d.ts(73,9): error TS1086: An accessor cannot be declared in an ambient context.
node_modules/@azure/abort-controller/types/src/AbortSignal.d.ts(40,9): error TS1086: An accessor cannot be declared in an ambient context.
node_modules/@azure/abort-controller/types/src/AbortSignal.d.ts(46,16): error TS1086: An accessor cannot be declared in an ambient context.
```

core-auth depends on abort-controller and it is needed for the work on supporting TokenCredential in v4. For more information, see https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#class-field-mitigations.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9054,9 +9054,9 @@
       "dev": true
     },
     "typescript": {
-      "version": "3.5.1",
-      "resolved": "https://registry.npmjs.org/typescript/-/typescript-3.5.1.tgz",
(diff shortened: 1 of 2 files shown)


joheredi (approved):
Looks good!

PR #855, opened 2021-01-28
Check for operation groups

joheredi:
There are cases where we can have swaggers without any operation groups.

**Problem**
When extracting the endpoint we were assuming the existence of operationGroups however we should allow the non-existence of these and have an empty endpoint

**Fix**
Add a null check for operation groups


deyaaeldeen (approved):
Looks good! Can you please add a test case? I also wonder what examples we have for such endpoints?

PR #857, opened 2021-02-01
Rename @azure-tools/autorest-extension-base dependency to new @autorest/extension-base pkg

timotheeguerin:
This package was renamed when we moved it to the autorest monorepo

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,48 @@
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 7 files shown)


deyaaeldeen (approved):
Thanks! Do we need to do something similar for v4x branch?

timotheeguerin:
@deyaaeldeen I don't see this as a dependency in the v4 branch so I guess we are good. Also the old package is not going to stop working, it will never get any new updates and I'll mark it as deprecated soon.

PR #861, opened 2021-02-11
[V4] Add a note to README files that older credentials are supported

deyaaeldeen:
Implements @ramya-rao-a's suggestions here: https://github.com/Azure/azure-sdk-for-js/pull/13512.

### DIFF
--- a/src/vanilla/Model/CodeModelTS.cs
+++ b/src/vanilla/Model/CodeModelTS.cs
@@ -914,6 +914,8 @@ namespace AutoRest.TypeScript.Model
         public void GenerateReadmeMdBrowserSampleCode(JSBuilder builder)
         {
             builder.ConstQuotedStringVariable("subscriptionId", "<Subscription_Id>");
+            builder.Line("// Create credentials using the `@azure/identity` package.");
+            builder.Line("// Please note that you can also use credentials from the `@azure/ms-rest-browserauth` package instead.");
             builder.Line("const credential = new InteractiveBrowserCredential(");
             builder.Object(tsObject =>
             {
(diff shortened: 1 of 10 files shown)


src/vanilla/Model/CodeModelTS.cs
```diff
@@ -919,6 +921,8 @@ public void GenerateReadmeMdBrowserSampleCode(JSBuilder builder)
                 tsObject.QuotedStringProperty("tenant", "<optional tenant for your organization>");
             });
             builder.Line(");");
+            builder.Line("// Pass the credentials from `@azure/identity` to the client constructor.");
+            builder.Line("// Please note that the credentials from `@azure/ms-rest-nodeauth` are supported here as well.");
```
ramya-rao-a, line 925: For browsers, this should be ms-rest-browserauth

test/azuremetadata/generated/Lro/README.md
```diff
@@ -37,6 +37,8 @@ const { AutoRestLongRunningOperationTestService } = require("");
 const subscriptionId = process.env["AZURE_SUBSCRIPTION_ID"];
 
 const creds = new DefaultAzureCredential();
+// Pass the credentials from `@azure/identity` to the client constructor.
+// Please note that the credentials from `@azure/ms-rest-nodeauth` are supported here as well.
```
ramya-rao-a, line 41: Just a thought...
Do you think it is better to have these comments above the line where the creds are created rather than above the client constructor? In that case, it would be

> Create credentials using the `@azure/identity` package.
Please note that you can also use credentials from the `@azure/ms-rest-nodeauth` package instead.

deyaaeldeen: @ramya-rao-a Yeah I like it

src/vanilla/Model/CodeModelTS.cs
```diff
@@ -919,6 +921,8 @@ public void GenerateReadmeMdBrowserSampleCode(JSBuilder builder)
                 tsObject.QuotedStringProperty("tenant", "<optional tenant for your organization>");
             });
             builder.Line(");");
+            builder.Line("// Pass the credentials from `@azure/identity` to the client constructor.");
+            builder.Line("// Please note that the credentials from `@azure/ms-rest-browserauth` are supported here as well.");
```
ramya-rao-a, line 925: Can you repeat the changes in https://github.com/Azure/autorest.typescript/pull/861/commits/9568c9c00d78b0e91244746a870ade1f4e4edacd over here for the browser?

ramya-rao-a (approved):

Issue #867, opened 2021-02-17
[V6] - Use ClientContext Class in Operations Class


sarangan12:
In V4 Generators, when there are classes within the operations folder, such classes use(in constructor)  from ClientContext Class. Please refer [here](https://github.com/Azure/azure-sdk-for-js/blob/0e28a4544886ebf9aa4053b80784ea261da5fd26/sdk/batch/batch/src/operations/file.ts#L25)

But, in V6 generators, such classes use(in constructor) from Client Class. Please refer [here](https://github.com/Azure/azure-sdk-for-js/blob/0e28a4544886ebf9aa4053b80784ea261da5fd26/sdk/search/search-documents/src/generated/service/operations/indexers.ts#L38)

sarangan12:
Code Complete.

PR #868, opened 2021-02-17
[V6] - Use ClientContext Class in Operations Class

sarangan12:
In V4 Generators, when there are classes within the operations folder, such classes use(in constructor)  from ClientContext Class. Please refer [here](https://github.com/Azure/azure-sdk-for-js/blob/0e28a4544886ebf9aa4053b80784ea261da5fd26/sdk/batch/batch/src/operations/file.ts#L25)

But, in V6 generators, such classes use(in constructor) from Client Class. Please refer [here](https://github.com/Azure/azure-sdk-for-js/blob/0e28a4544886ebf9aa4053b80784ea261da5fd26/sdk/search/search-documents/src/generated/service/operations/indexers.ts#L38)

This PR fixes this issue reported in https://github.com/Azure/autorest.typescript/issues/867.

@joheredi Please review and approve this PR. 

@ramya-rao-a FYI...


deyaaeldeen (approved):
LGTM

Issue #869, opened 2021-02-18
[v6] SDK Generation failed for bodyFormData


sarangan12:
This issue is happening as of February 17, 2021. 

**Repro Steps**
1. ```git clone https://github.com/Azure/autorest.typescript.git```
2. ```cd autorest.typescript```
3. ```git checkout v6```
4. ```npm install```
5. ```npm run build```
6. ```npm run generate-swaggers -- -i bodyFormData```

You will get the following error:

```
> @autorest/typescript@6.0.0 generate-swaggers C:\Users\sarajama\Projects\autorest.typescript
> ts-node ./test/utils/test-swagger-gen.ts "-i" "bodyFormData"

Not building Autorest.Typescript, use --build or -b to build
Failed to generate bodyFormData with error:
 Error: Command failed: autorest --clear-output-folder=true    --license-header=true --add-credentials=false --typescript --output-folder=./test/integration/generated/bodyFormData --use=. --title=BodyFormDataClient --package-name=body-formdata --package-version=1.0.0-preview1 --hide-clients=false --input-file=node_modules/@microsoft.azure/autorest.testserver/swagger/body-formdata.json

WARNING: AutoRest has not been tested with Node versions greater than v13.

ERROR: Schema violation: Data does not match any schemas from 'oneOf' (paths > /formdata/stream/uploadfiles > post > parameters > 0)
    - file:///C:/Users/sarajama/Projects/autorest.typescript/node_modules/@microsoft.azure/autorest.testserver/swagger/body-formdata.json:1:1
ERROR: Schema violation: Data does not match any schemas from 'oneOf' (paths > /formdata/stream/uploadfiles > post > parameters > 0)
    - file:///C:/Users/sarajama/Projects/autorest.typescript/node_modules/@microsoft.azure/autorest.testserver/swagger/body-formdata.json:1:1
Process() cancelled due to failure

Failed to generate bodyFormData with error:
 Error: Command failed: autorest --clear-output-folder=true    --license-header=true --add-credentials=false --typescript --output-folder=./test/integration/generated/bodyFormData --use=. --title=BodyFormDataClient --package-name=body-formdata --package-version=1.0.0-preview1 --hide-clients=false --input-file=node_modules/@microsoft.azure/autorest.testserver/swagger/body-formdata.json

WARNING: AutoRest has not been tested with Node versions greater than v13.

[...]

PR #870, opened 2021-02-18
Upgrade Autorest dependency in package

sarangan12:
This PR is to fix the issue https://github.com/Azure/autorest.typescript/issues/869. 

The issue happens only with a previous version of autorest. Once you upgrade the version, this issue will be resolved. 

@ramya-rao-a FYI......

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2500,9 +2500,9 @@
             "dev": true
         },
         "autorest": {
-            "version": "3.0.6187",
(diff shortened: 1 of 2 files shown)


deyaaeldeen (approved):
The change looks good but how can we verify the issue has been resolved? do we need to write a test case?

joheredi (approved):

PR #871, opened 2021-02-18
Adding a flag to model enum mappers as string mappers

joheredi:
This change is intended to provide an alternative to our current approach of modeling extensible enums. There are two main uses of extensible enums that we have identified

1) Extensible enums are used to hint users about values they may want. But any value is accepted by the service
2) Extensible enums are used to provide the Service flexibility to add new accepted values. In this scenario the list of values in the extensible enum are mostly what the service expects, values outside of this list are not expected most of the cases for this scenario.

Our current approach handles (1) nicely as we provide a convenience Enum with the known values plus documentation on what values are known. However for (2) most of the cases users would be using the KnownValues and it will be very rare  when they need to use an "unknown" value.

We have got feedback that for the scenario (2) it would be preferable to keep the string literal union with the options, and have users cast to any in case they need to use a different string.

```typescript
type EnumValues = "foo" | "bar"
declare function foo(value: EnumValues): void;

foo("baz" as any)
```

I spoke to @bterlson about this, and we think we can provide a short term solution for (2) if we allow skipping core-http's validation for `allowedValues`. We can do this by modeling `SealedChoice` mappers as string mappers.

This PR adds a flag `skip-enum-validation` that will model Enums as string mappers while keeping the model definition as a sting literal union.

Note that skipping this validation also aligns with the [Azure SDK Typescript Design guidelines](https://azure.github.io/azure-sdk/typescript_implementation.html#general-parameter-validation)

image


DominikMe (approved):
Thanks a lot for adding this flag @joheredi!

PR #873, opened 2021-02-18
Fix polymorphic skip normalization

joheredi:
We currently have the ability to tell the generator to skip property name normalization by setting an M4 name override like the following 

```yaml
modelerfour:
  naming:
    override:
      '@type': $DO_NOT_NORMALIZE$@type
```

However in some cases, such as the one in the example, when generating the object key is enclosed by quotes to make it a valid javascript property name. This was causing 2 issues

1. normalizeName wouldn't recognize the $DO_NOT_NORMALIZE$ token, as the overriden name would not start with $DO_NOT_NORMALIZE$ but "$DO_NOT_NORMALIZE$. So it normalized it anyway
2. When generating non-discriminated properties we are note wrapping the property name in quotes which while doesn't make any difference in the actual generated code it would make generation fail if the example above was used with a non-discriminated property since the following object is not valid in JS
```javascript
const foo = {
   @type: "someType"
}
```

So the fixes are:

1. Make sure to strip quotes at the beginning and end when normalizing discriminated property name. This is not an issue with normal properties because we use a "dictionary" to store discriminated props info which requires the property name to be enclosed with "". While in regular properties the name is stored in a string property so they are not enclosed by quotes.
2. Explicitly set quotes in the property name when generating both regular and discriminated-properties. Our codegen runs the generated code through prettier which removes un-needed quotes so this won't have any adverse side-effects

### DIFF
--- a/test/integration/generated/polymorphicSkipNormalize/src/generated/models/index.ts
+++ b/test/integration/generated/polymorphicSkipNormalize/src/generated/models/index.ts
@@ -1,0 +1,768 @@
+import * as coreHttp from "@azure/core-http";
+
+export type MethodRequestUnion =
+  | MethodRequest
+  | MediaGraphTopologySetRequest
+  | MediaGraphTopologySetRequestBody
+  | MediaGraphInstanceSetRequest
+  | MediaGraphInstanceSetRequestBody
+  | ItemNonSetRequestBaseUnion
+  | MediaGraphTopologyListRequest
+  | MediaGraphInstanceListRequest;
+export type MediaGraphSourceUnion =
+  | MediaGraphSource
+  | MediaGraphRtspSource
+  | MediaGraphIoTHubMessageSource;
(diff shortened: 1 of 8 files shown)


test/integration/generated/polymorphicSkipNormalize/src/generated/models/index.ts
```diff
@@ -0,0 +1,768 @@
...
+  /** Description of the parameter. */
+  description?: string;
+  /** The default value for the parameter to be used if the media graph instance does not specify a value. */
+  default?: string;
+}
+
+/** A source node in a media graph. */
+export interface MediaGraphSource {
+  /** Polymorphic discriminator, which specifies the different types this object can be */
+  "@type":
```
joheredi, line 125: This is the result of the `$DO_NOT_NORMALIZE$` token

sarangan12 (approved):

PR #877, opened 2021-02-25
Update Pipeline

joheredi:
Updating Pipelines, use a single pipeline definition for PR, CI and Release.

This configuration will serve 2 devops pipelines

1. Public pipeline for PR and CI
   * Triggered on PR or Merge 
2. Private pipeline for release
    * Manually triggered for release and requires approval to publish

Additionally this pipeline runs tests against Node 8, 10, 12 and 14.

This PR will enable us to start publishing dev (`alpha`) versions of @autorest/typescript to npmjs.org

As part of testing on Node 8 a blocking issue was found, as Node8 doesn't support AsyncIterators. The fix is to use the polyfill provided in @azure/core-auth and `/// <reference lib="esnext.asynciterable" />`. Fixing this is what caused the bulk of the changes in this PR.


daviwil (approved):

sarangan12 (approved):

.scripts/ci.yml
```diff
@@ -1,35 +1,117 @@
...
+
+          - script: npm run test:browser
+            displayName: "Run Browser Tests"
+            condition: eq(variables['TestType'], 'browser')
+
+          - script: npm run check:tree
+            displayName: "Check git Tree"
+
+          - script: |
+              npm run coverage-push -- $(Build.Repository.Name) $(Build.SourceBranch) $(github-token) $(storage-coverage-user) $(storage-coverage-pass)
```
praveenkuttappan, line 93: Is this step intended for PR run? Where are these variables stored?

.scripts/ci.yml
```diff
@@ -1,35 +1,117 @@
...
+            workingDirectory: node_modules/@microsoft.azure/autorest.testserver
+            displayName: "Upload Coverage Report"
+
+      - template: smoke-test.yml
```
praveenkuttappan, line 97: Can you please add a comment what's the difference among these 3 different smoke tests?

.scripts/ci.yml
```diff
@@ -1,35 +1,117 @@
...
+            displayName: "Copy packages"
+
+          - task: PublishPipelineArtifact@1
+            condition: succeededOrFailed()
+            displayName: "Publish artifacts"
+            inputs:
+              artifactName: packages
+              path: $(Build.ArtifactStagingDirectory)
+
+      - job: Test
```
praveenkuttappan, line 39: This step should also depend on build. right? There is no point in running them if build fails.

.scripts/release.yml
```diff
@@ -0,0 +1,68 @@
...
+stages:
+  - ${{if eq(parameters.PublishAlpha, 'true')}}:
+      - stage: Integration
+        dependsOn: ${{parameters.DependsOn}}
+        jobs:
+          - deployment: PublishToGithub
+            environment: github
+            displayName: Publish dev version to GitHub
+            pool:
+              vmImage: ubuntu-18.04
```
praveenkuttappan, line 19: This should also use pool name similar to one in Build job

.scripts/release.yml
```diff
@@ -0,0 +1,68 @@
...
+                        npm install
+                        npm run build
+                      displayName: "npm install and build"
+                    - script: |
+                        export DEV_VERSION=$(node -p -e "require('./package.json').version")-alpha.$BUILD_BUILDNUMBER
+                        npm version --no-git-tag-version $DEV_VERSION
+                        npm pack
+                        npx publish-release --token $(azuresdk-github-pat) --repo autorest.typescript --owner azure --name v$DEV_VERSION --tag v$DEV_VERSION --notes='prerelease build' --prerelease --editRelease false --assets autorest-typescript-$DEV_VERSION.tgz --target_commitish $(Build.SourceBranchName)
+                      displayName: "Publish to GitHub"
+          - deployment: PublishToNpm
```
praveenkuttappan, line 39: Can we simplify these two steps to combine common steps and just have two sub  steps to publish to github and npm?

.scripts/ci.yml
joheredi: sure! I'll add the comment

They contain different test clients that are generated and built. We decided to split them to run each set in parallel and have a faster CI. Each bucket takes about 8 minutes

.scripts/ci.yml
joheredi: Good catch I'll switch it off when BuildReason is PullRequest. The variables are defined in the pipeline config in DevOps

Issue #875, opened 2021-02-22
Fix createSpan options


joheredi:
createSpan takes `OperationOptions` as a second parameter, however our generated clients are passing `RequestOptionsBase`. Richard fixed it manually in Synapse. We should take this change in the generator

https://github.com/Azure/azure-sdk-for-js/pull/13887/files/00cbabac1dc13c5a6a1b118477ac3b95fd0e1472#r580649030
![image](https://user-images.githubusercontent.com/21109913/108783262-02b9c900-7522-11eb-8f28-3bb73a4f310c.png)

We also need to consume createSpanFunction from core-tracing
![image](https://user-images.githubusercontent.com/21109913/108783162-d30ac100-7521-11eb-9181-860f7a28b9e5.png)

and add core-tracing as a dependency
![image](https://user-images.githubusercontent.com/21109913/108783194-e28a0a00-7521-11eb-8f15-a6ccb6e6c6d3.png)

deyaaeldeen:
Code complete https://github.com/Azure/autorest.typescript/pull/882

PR #882, opened 2021-03-08
Upgrade core-tracing to v1.0.0-preview.10

deyaaeldeen:
This version upgrade brings a change to the `createSpan` function. This PR updates the call to `createSpan` accordingly and updates `getTracingSpanStatement` to return the output options bags variable name. Fixes https://github.com/Azure/autorest.typescript/issues/875.

Multimedia is an interesting case so please review that file: `test/integration/generated/mediaTypesWithTracing/src/mediaTypesWithTracingClient.ts`.


joheredi (approved):
Looks great! Thanks :)

Issue #856, opened 2021-02-01
Include Operations in the index.ts file


sarangan12:
Refer the index.ts file [here](https://github.com/Azure/autorest.typescript/blob/5a4bbaad4e1322b2189d437ae748ac7104747352/test/integration/generated/additionalProperties/src/index.ts). 

```js
export * from "./models";
export { AdditionalPropertiesClient } from "./additionalPropertiesClient";
export { AdditionalPropertiesClientContext } from "./additionalPropertiesClientContext";
```

There is no export from ```operations``` folder. 

But, if you look at the code in the ```additionalPropertiesClient``` - [here](https://github.com/Azure/autorest.typescript/blob/5a4bbaad4e1322b2189d437ae748ac7104747352/test/integration/generated/additionalProperties/src/additionalPropertiesClient.ts#L23) there are references to classes from the Operations folder. 

If the operations folder is not exported, then the SDK users will not have any method to access them directly. Also, you can see a ```forgotten-export``` error in the API Extractor. 

So, we need to fix this error by adding ```operations``` export in the index.ts file.

Note: This ```operations``` export is not required for packages (such as [noLicenseHeader](https://github.com/Azure/autorest.typescript/tree/5a4bbaad4e1322b2189d437ae748ac7104747352/test/integration/generated/noLicenseHeader/src)), where there is no operations folder. 

@joheredi @deyaaeldeen @ramya-rao-a FYI......

qiaozha:
I had similar problem in mgmt plane SDK. 
I figured out a way to include the operation in track1 mgmt SDK, which is to change the src/index.ts like this 
```
 /// <reference lib="esnext.asynciterable" />
export * from "./models";
export * from "./operations";
export { ComputeManagementClient } from "./computeManagementClient";
export { ComputeManagementClientContext } from "./computeManagementClientContext";
```
In the api extractor result, you will see the these operations are included. 
![image](https://user-images.githubusercontent.com/55688292/106831292-a0ac3780-66ca-11eb-8087-1ed99a8a37f9.png)

But when I tried to add this line `export * from "./operations";` in track2 mgmt SDK src/index.ts, it reports error about duplicate export. 
For example, when I use compute RP, 
![image](https://user-images.githubusercontent.com/55688292/106831191-72c6f300-66ca-11eb-9696-e423fd87755b.png)

qiaozha:
cc: @ramya-rao-a

joheredi:
[...]

PR #890, opened 2021-03-19
Create operation interfaces in V6 Generator

sarangan12:
This PR is to address the issue reported at https://github.com/Azure/autorest.typescript/issues/856.  (A detailed explanation of the issue and proposed resolution is provided in the issue)

@deyaaeldeen @ramya-rao-a FYI....


src/generators/indexGenerator.ts
```diff
@@ -33,4 +34,32 @@ export function generateIndexFile(
...
+        moduleSpecifier: "./operationsInterfaces"
+      }
+    ]);
+  }
+
+  operationGroups.forEach(og => {
+    if (hasLROOperation(og)) {
+      includeLRO = true;
+    }
+  });
```
deyaaeldeen, line 55: NIT: `includeLRO` could be defined in one line as follows:
```suggestion
  const operationGroups = clientDetails.operationGroups.filter(
    og => !og.isTopLevel
  );

  if (operationGroups.length) {
    indexFile.addExportDeclarations([
      {
        moduleSpecifier: "./operationsInterfaces"
      }
    ]);
  }

  const includeLRO = operationGroups.find(hasLROOperation) !== undefined;
```

src/generators/operationGenerator.ts
```diff
@@ -435,9 +435,10 @@ function addClass(
     true /** shouldGuard */
   );
   const operationGroupClass = operationGroupFile.addClass({
-    name: className,
+    name: `${className}Impl`,
     docs: [`Class representing a ${className}.`],
-    isExported: true
+    isExported: true,
+    implements: [`${className}`]
```
deyaaeldeen, line 441: `className` should be renamed to `interfaceName`.

src/generators/operationGenerator.ts
```diff
@@ -1098,6 +1099,17 @@ function addImports(
     operationGroupFile
   );
 
+  const interfaceName = normalizeName(
+    operationGroupDetails.name,
+    NameType.OperationGroup,
+    true /** shouldGuard */
+  );
```
deyaaeldeen, line 1106: this is a redefinition of the `className` variable above. I think it is better to rename `className` to `interfaceName` and remove this redefinition here.

src/generators/clientFileGenerator.ts
```diff
@@ -248,7 +265,7 @@ function writeConstructor(
 
   clientConstructor.addStatements([
     ...operationDeclarationDetails.map(
-      ({ name, typeName }) => `this.${name} = new ${typeName}(this)`
+      ({ name, typeName }) => `this.${name} = new ${typeName}Impl(this)`
```
deyaaeldeen, line 268: Perhaps there could be reasons in the future to have custom names for those classes so I think it is a good idea to include the name of the class in the `operationDeclarationDetails` instead of hardcoding it as one that has the postfix `Impl`.

src/generators/operationInterfaceGenerator.ts
```diff
@@ -0,0 +1,240 @@
...
+    modelNames
+  );
+
+  return operation.isLRO
+    ? `Promise<LROPoller<${responseName}>>`
+    : `Promise<${responseName}>`;
+}
+
+/**
+ * Adds an Operation group class to the generated file
```
deyaaeldeen, line 113:
```suggestion
 * Adds an Operation group interface to the generated file
```

src/generators/operationInterfaceGenerator.ts
```diff
@@ -0,0 +1,240 @@
...
+function addInterface(
+  operationGroupFile: SourceFile,
+  operationGroupDetails: OperationGroupDetails,
+  clientDetails: ClientDetails
+) {
+  let importedModels = new Set<string>();
+
+  let allModelsNames = getAllModelsNames(clientDetails);
+
+  const className = normalizeName(
```
deyaaeldeen, line 124: `className` needs to be renamed to `interfaceName`.

src/generators/operationInterfaceGenerator.ts
```diff
@@ -0,0 +1,240 @@
...
+
+    operationGroupFile.addImportDeclaration({
+      namedImports,
+      moduleSpecifier: "../models"
+    });
+  }
+}
+
+/**
+ * Write operations implementation, extracted from OperationGroupDetails, to the generated file
```
deyaaeldeen, line 162:
```suggestion
 * Write operations signatures, extracted from OperationGroupDetails, to the generated file
```

src/generators/operationInterfaceGenerator.ts
```diff
@@ -0,0 +1,240 @@
...
+) {
+  preparePageableOperations(operationGroupDetails, clientDetails);
+  writeAsyncIterators(
+    operationGroupDetails,
+    clientDetails,
+    operationGroupClass,
+    importedModels
+  );
+  operationGroupDetails.operations.forEach(operation => {
+    if (operation.scope != Scope.Private) {
```
deyaaeldeen, line 179: should we be more strict?
```suggestion
    if (operation.scope !== Scope.Private) {
```

src/generators/operationInterfaceGenerator.ts
```diff
@@ -0,0 +1,240 @@
...
+    });
+  }
+}
+
+/**
+ * Write operations implementation, extracted from OperationGroupDetails, to the generated file
+ */
+export function writeOperations(
+  operationGroupDetails: OperationGroupDetails,
+  operationGroupClass: InterfaceDeclaration,
```
deyaaeldeen, line 166:
```suggestion
  operationGroupInterface: InterfaceDeclaration,
```

src/generators/operationInterfaceGenerator.ts
```diff
@@ -0,0 +1,240 @@
...
+
+/**
+ * Adds required imports at the top of the file
+ */
+function addImports(
+  operationGroupDetails: OperationGroupDetails,
+  operationGroupFile: SourceFile,
+  clientDetails: ClientDetails
+) {
+  const { className, mappers } = clientDetails;
```
deyaaeldeen, line 212: `className` => `interfaceName`

src/generators/clientFileGenerator.ts
sarangan12: You can have custom names for classes even now with swagger transforms. The extent of having different names for interface and class name seems to be somewhat far fetched idea. I believe we can settle for Impl.

src/generators/operationInterfaceGenerator.ts
sarangan12: Done (in the next commit)

src/generators/operationInterfaceGenerator.ts
sarangan12: Removed that line.

src/generators/operationGenerator.ts
sarangan12: Removed the interfaceName and used the className directly

src/generators/operationGenerator.ts
sarangan12: In this place, it is easier for understanding to keep it a class name. Not changing here.

joheredi (approved):
Looks good, thanks for taking on this work!

Issue #594, opened 2020-03-23
[v6] Generate useful README files


joheredi:
Auto generate useful Readme files with code samples in Typescript that demonstrates the use of the generated SDK.

We need to make sure generated samples are valid and can be run. Previous Autorest.Typescript had some issues generating invalid samples, see #576 

- [ ] Samples should be generated in Javascript instead of Typescript to help portability

joheredi:
We should consider looking into https://github.com/Azure/autorest.typescript/issues/455 when working on the samples for v6

joheredi:
None of the languages are currently generating heavy readmes here's what the other languages do right now:

**Python**: Readme with general information about Azure Management, nothing specific about the generated library.
**Java**: no README
**C#**: no README

I have started a conversation with the Autorest crew to see if we can reach agreement on consistent Readmes across languages.

I would propose we generate a README with the following information:

**Service Description:** From the swagger
**Supported platforms:** Which node and browser versions are supported
**How to install the package**: npm install @azure/arm-resource-manager

I'm reluctant about generating samples, usually samples are hand-picked and carefully crafted to describe a set of steps that are meaningful for each service. For example a service may be interested to show the following hero scenario 1. authenticate 2. create and entity 3. poll the creation operation until completed 4. list all the entities where another service may have a very different flow.

If we want to provide samples, I think Python's approach is reasonable, where they point users to https://github.com/Azure-Samples/azure-samples-python-management

Python sample README: https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/containerservice/azure-mgmt-containerservice/README.md

/cc: @lmazuel, @sarangan12

qiaozha:
Maybe we could at least provide one simple example for customers just like track1 mgmt. SDK?  Otherwise it may cause user's confusion on how to use the track2 SDK once we publish it.  @nickzhums How do you think of it? 
Actually, we also have sample repo for JS mgmt. SDK https://github.com/Azure-Samples/azure-samples-js-management. we can consider to point to this sample repo, but we will only add some samples for ring0 RP due to lack of resources. 
cc @lirenhe

nickzhums:
it will be great if we can show how to authenticate and make a simple request in README for Track 2 package (this is what track 1 package currently has, so would be a bummer if we don't continue it in track 2)

sarangan12:
[...]

PR #895, opened 2021-04-01
Improve the generated README.md to be close to Track 2's

deyaaeldeen:
This PR improves the generated README.md to be close to Track 2's ones. In particular:
- Generates required sections: `Getting Started`, `Key Concepts`, `Troubleshooting`, `Next Steps`, `Contributing`, and `Related Projects`.
- If the generation is not for an Azure package, Azure-specific information is omitted.

Example: https://github.com/Azure/azure-sdk-for-js/pull/14655
Related issue: https://github.com/Azure/azure-sdk-for-js/issues/14449
[Swagger](https://raw.githubusercontent.com/Azure/azure-rest-api-specs/d2982a8962885e8506b5ebc0b33cb8caf1dc7551/specification/cognitiveservices/data-plane/TextAnalytics/preview/v3.1-preview.4/TextAnalytics.json)

Fixes https://github.com/Azure/autorest.typescript/issues/594

### DIFF
--- a/src/generators/static/readmeFileGenerator.ts
+++ b/src/generators/static/readmeFileGenerator.ts
@@ -89,6 +89,8 @@ function createMetadata(
      * "Client".
      */
     title.match(/(.*) Client/)?.[1] ??
+    /** I noticed management-plane swaggers do not use spaces in their titles */
+    title.match(/(.*)Client/)?.[1] ??
     clientClassName.match(/(.*)Client/)?.[1] ??
     title.match(/(.*) Service/)?.[1] ??
     "Service";
--- a/src/generators/static/readmeFileGenerator.ts
+++ b/src/generators/static/readmeFileGenerator.ts
@@ -204,7 +204,7 @@ function writeKeyConcepts(metadata: Metadata): string {
 
 ### ${metadata.clientClassName}
 
-\`${metadata.clientClassName}\` is the primary interface for developers using the ${metadata.clientDescriptiveName} library. It provides asynchronous methods to access a specific use of ${metadata.serviceName}.`;
+\`${metadata.clientClassName}\` is the primary interface for developers using the ${metadata.clientDescriptiveName} library. Explore the methods on this client object to understand the different features of the ${metadata.serviceName} service that you can access.`;
 }
 
 function writeTroubleshooting(metadata: Metadata): string {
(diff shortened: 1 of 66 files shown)


test/integration/generated/appconfiguration/README.md
```diff
@@ -1,27 +1,29 @@
...
-### How to use
-
-#### Sample code
-
-Refer the sample code in the [azure-sdk-for-js/samples](https://github.com/Azure/azure-sdk-for-js/tree/master/samples) folder.
-
-## Related projects
 
-- [Microsoft Azure SDK for Javascript](https://github.com/Azure/azure-sdk-for-js)
+## Key concepts
```
joheredi, line 25: Maybe here we can add a reference to Identity so users can refer to to learn how to authenticate

joheredi (approved):
Looks great! Thanks for taking on this work 😄 just left a small comment about including a reference to Identity docs

test/integration/generated/appconfiguration/README.md
deyaaeldeen: There is a reference to the docs for the `DefaultAzureCredential` [here](https://github.com/Azure/azure-sdk-for-js/blob/f16cc05b4061b5e0e5a48b9e31d112bfd9cc9b86/sdk/textanalytics/ai-text-analytics/README.md#using-an-azure-active-directory-credential). However, I am not sure if @ramya-rao-a wants to keep it. We can continue the discussion in https://github.com/Azure/azure-sdk-for-js/pull/14655.

src/generators/static/readmeFileGenerator.ts
```diff
@@ -5,39 +5,296 @@ import { Project } from "ts-morph";
...
+}
 
-This package contains an isomorphic SDK for ${clientDetails.name}.
+/**
+ * Creates the string for the "Getting Started" section in the README.md file.
+ * @param metadata metadata about the service, package, and the client.
+ * @returns the required getting started section in the readme file
+ */
+function writeGettingStarted(metadata: Metadata): string {
+  const text = `## Getting started
```
maorleger, line 128: I have a bit of a general question (for anyone, not specific to this PR) - is it common that we put all the text inline like this when generating static files? Have we explored simple templating engines (remarkable, handlebars, etc)

I'm not proposing we go all-in on super complex templates I'm just wondering if that's been explored at all just to separate content/layout from data and (for one example) avoid having to escape backticks everywhere

joheredi: We used to have template, if I recall correctly we moved to ts-morph as it allows more flexibility by crafting TS programmatically (@daviwil can correct me here). It may be interesting to do templates for static files though as they are usually not complex.

deyaaeldeen:
Closing in favor of https://github.com/Azure/autorest.typescript/pull/971.

Issue #785, opened 2020-10-29
Revisit LRO generated operations


joheredi:
Revisit LRO generated operations and evaluate if we can align better with the Typescpt SDK guidelines and the hand-written LROs in Azure/azure-sdk-for-js repo. Some examples:

https://github.com/Azure/azure-sdk-for-js/blob/6bcfe0d943662784f196561a5bb48c05460b0b8f/sdk/keyvault/keyvault-certificates/src/index.ts#L693

https://github.com/Azure/azure-sdk-for-js/blob/6bcfe0d943662784f196561a5bb48c05460b0b8f/sdk/formrecognizer/ai-form-recognizer/src/formRecognizerClient.ts#L306

- [x] Support pollingInterval
- [ ] prefix operation with `begin` #785
- [x] Support rehidrating the poller by passing poller state
- [ ] Log polling status at `info` level
- [x] Progress reporting

Guidelines: https://azure.github.io/azure-sdk/typescript_design.html#ts-lro

joheredi:
Received the following feedback from @qiaozha that we should consider.

In addition to providing an operation that returns the poller, it'd be helpful for customers to also have a function that performs the polling internally and just returns the response.

Another thing is that to align with the guidelines we need to rename the operations that return a poller to be prefixed with **begin**

joheredi:
I talked to Brian about adding a function that performs the polling internally and just returns the response.

We would not like to expose a function that returns the Poller result. The reason is that these operations may take days to complete and wrapping them in a function that returns the final result will be misleading and potentially painful for the users. 

Additionally, none of the other languages are currently exposing this method, if we were to implement it, we’d need to do it consistently across generators

qiaozha:
The current track1 SDK is to return the result directly and customers can still use beginXXX to get a poller if they like. https://github.com/Azure/azure-sdk-for-js/blob/b679003645bceac0685826ead6e2df5a02b86322/sdk/compute/arm-compute/src/operations/disks.ts#L40 

Not sure about other languages' SDKs but the Python track2 SDK will return a ARMPoller by default, and azure-mgmt-core will handle that poller. 
https://github.com/Azure/azure-sdk-for-python/blob/0bab49318faf4ef10193c9d99fe7766568430cdd/sdk/compute/azure-mgmt-compute/azure/mgmt/compute/v2020_09_30/operations/_disks_operations.py#L136. 
[...]

PR #899, opened 2021-04-08
Use PollerLike instead of LROPoller as return type for LROs

deyaaeldeen:
This change brings auto-generated LROs closer to the hand-written ones in our Track 2 packages.

Related to https://github.com/Azure/autorest.typescript/issues/785

The CI failure is due to a bug in @azure/core-lro@1.0.4. v1.0.5 will have the fix and will be released on 4/12.


joheredi (approved):
Looks good! Thanks for doing this work

Issue #907, opened 2021-04-13
[V6][Storage Blob]Missing serializer property in the operation spec


sarangan12:
In the Storage Blob Swagger, there is an operation: `BlockBlob_Upload`. In the generated SDK, the operation spec refers to serializer. But, there is no serializer property defined in the file

sarangan12:
Code Complete with PR https://github.com/Azure/autorest.typescript/pull/917

PR #917, opened 2021-04-22
Code changes to generate serializer

sarangan12:
This PR is to address the issue reported in the issue https://github.com/Azure/autorest.typescript/issues/907. Basically, per swagger we have a mediatype set to binary. But, Our SDK code checks only for xml & json which results in the error. This PR will fix it.

@joheredi Please review and approve it. 

@ramya-rao-a FYI.....


joheredi (approved):
LGTM!

Issue #909, opened 2021-04-13
[V6] Remove null on an optional property


sarangan12:
In the new V6 SDK generator, if a property is optional and also is nullable, then the generated SDK defines that property as:

```js
sample?: Number | null;
```

This does not make much sense. It would be nice to have it as:

```js
sample?: Number
```

sarangan12:
@joheredi What is your opinion on this issue?

sarangan12:
Created a PR for this issue at https://github.com/Azure/autorest.typescript/pull/919

sarangan12:
Code complete with PR https://github.com/Azure/autorest.typescript/pull/919

PR #919, opened 2021-04-26
Code to fix the Optional Null

sarangan12:
This is to fix the issue reported in https://github.com/Azure/autorest.typescript/issues/909.

@ramya-rao-a FYI....


test/integration/generated/bodyComplex/src/models/index.ts
```diff
@@ -16,7 +16,7 @@ export type SharkUnion = Shark | Sawshark | Goblinshark | Cookiecuttershark;
 
 export interface BasicDef {
   /** Basic Id */
-  id?: number | null;
```
sarangan12, line 19: Per Autorest, this is a arm swagger. So this change is expected.

test/integration/generated/bodyComplexWithTracing/src/models/index.ts
```diff
@@ -16,7 +16,7 @@ export type SharkUnion = Shark | Sawshark | Goblinshark | Cookiecuttershark;
```
sarangan12, line 16: Per Autorest, this is a arm swagger. So this change is expected.

src/typescriptGenerator.ts
```diff
@@ -76,22 +77,34 @@ export async function generateTypeScriptLibrary(
 
   const shouldGenerateLicense: boolean =
     (await host.GetValue("license-header")) || false;
-
   const hideClients: boolean = (await host.GetValue("hide-clients")) || false;
+  const armLibrary: boolean =
+    (await host.GetValue("azure-arm")) ||
+    (await host.GetValue("openapi-type")) === "arm" ||
+    false;
+  const ignoreNullableOnOptional: boolean =
```
joheredi, line 85: I think you can do the following and later on you just need to check for `ignoreNullableOnOptional` without also checking for armLibrary

```typescript
const ignoreNullableOnOptional: boolean = (await host.GetValue("ignore-nullable-on-optional")) ??  (await host.GetValue("azure-arm")) 
```

src/generators/modelsGenerator.ts
```diff
@@ -701,7 +712,14 @@ function getProperties(
       property.name === "siblings"
         ? `${(objectDetails as PolymorphicObjectDetails).unionName}[]`
         : property.type;
-    return property.nullable ? `${typeName} | null` : typeName;
+
+    if (optionsBag.armLibrary || optionsBag.ignoreNullableOnOptional) {
```
joheredi, line 716: If you do something like I suggested below, you should be able to just check for `optionsBag.ignoreNullableOnOptional`. Which would be safer as you don't need to remember to check `armLibrary` every time you check for `ignoreNullableOnOptional`

joheredi (approved):
Looks good! Thanks 😄 
Left one suggestion to avoid checking both `ignoreNullableOnOptional` and `armLibrary`

PR #922, opened 2021-04-30
Simplify the LRO logic

deyaaeldeen:
The Azure SDK design guidelines specifies some rules for the long running operations to follow, you can find them here: https://azure.github.io/azure-sdk/typescript_design.html

In addition to the simplification, this work applies many of those rules to auto-generated long running operations, notably:
- allowing performing custom actions at each polling occurrence
- allowing specifying the polling interval
- almost allowing resuming from a serialized state (this is still work in progress that will be finished up in a subsequent PR)

In addition, this PR does the following: 
- makes it so that a custom interface is always generated for the operation options bag even it it only have the standard `OperationOptions`. This is something we have been doing in Track 2 packages to decrease changes to the public surface if a new option is added in the future
- deletes lro unit tests that are no longer needed


.gitattributes
```diff
@@ -0,0 +1,6 @@
+# Don't allow people to merge changes to these generated files, because the result
+# may be invalid.  You need to run "npm install" again.
+package-lock.json               merge=binary
+
```
joheredi, line 4: What's this for?

deyaaeldeen: Context: I copied what is useful for us here from https://github.com/Azure/azure-sdk-for-js/blob/master/.gitattributes. I needed the `* text=auto eol=lf` line so we do not get conflicts regarding different new lines.

This line is to disable auto-merging of `package-lock.json`

joheredi (approved):
Looks good! Thanks for the improvements!

Issue #924, opened 2021-05-05
[V6][Storage Blob]isXML is not set correctly in the operation spec


sarangan12:
In Storage Blob Swagger, there is an operation id `PageBlob_UploadPages`. If you look at the operation spec in the V4 generator, it is:

```js
const serializer = new coreHttp.Serializer(Mappers, true);

const uploadPagesOperationSpec: coreHttp.OperationSpec = {
  httpMethod: "PUT",
  path: "{containerName}/{blob}",
  urlParameters: [
    Parameters.url
  ],
  queryParameters: [
    Parameters.timeoutInSeconds,
    Parameters.comp17
  ],
  headerParameters: [
    Parameters.contentLength,
    Parameters.transactionalContentMD5,
    Parameters.transactionalContentCrc64,
    Parameters.range0,
    Parameters.encryptionScope,
    Parameters.version,
    Parameters.requestId,
    Parameters.pageWrite0,
    Parameters.leaseId0,
    Parameters.encryptionKey,
    Parameters.encryptionKeySha256,
    Parameters.encryptionAlgorithm,
    Parameters.ifSequenceNumberLessThanOrEqualTo,
    Parameters.ifSequenceNumberLessThan,
    Parameters.ifSequenceNumberEqualTo,
    Parameters.ifModifiedSince,
    Parameters.ifUnmodifiedSince,
    Parameters.ifMatch,
    Parameters.ifNoneMatch,
    Parameters.ifTags
  ],
  requestBody: {
    parameterPath: "body",
    mapper: {
      required: true,
      serializedName: "body",
      type: {
        name: "Stream"
      }
    }
  },
  contentType: "application/octet-stream",
  responses: {
    201: {
      headersMapper: Mappers.PageBlobUploadPagesHeaders
    },
    default: {
      bodyMapper: Mappers.StorageError,
      headersMapper: Mappers.PageBlobUploadPagesHeaders
    }
  },
  isXML: true,
  serializer
};
```

There is no problem with the above code. Specifically look at the last two lines, the `isXML` is set to true and `serializer` is also set to xml serializer. 

Now, Let us take a look at the operation spec from the V6 SDK generator. It is:

```js
const serializer = new coreHttp.Serializer(Mappers, /* isXml */ false);

const uploadPagesOperationSpec: coreHttp.OperationSpec = {
  path: "/{containerName}/{blob}",
  httpMethod: "PUT",
  responses: {
    201: {
[...]

PR #925, opened 2021-05-05
Fix the isXml & Serializer values correctly

sarangan12:
This PR is to address the issue reported at  https://github.com/Azure/autorest.typescript/issues/924. The details of the problem is provided in the issue itself. 

@ramya-rao-a FYI...


src/generators/operationGenerator.ts
```diff
@@ -1026,7 +1038,6 @@ export function addOperationSpecs(
     operation =>
       hasMediaType(operation, KnownMediaType.Json) ||
       hasMediaType(operation, KnownMediaType.Form) ||
-      hasMediaType(operation, KnownMediaType.Binary) ||
```
sarangan12, line 1029: Normally, in our swaggers anything set to binary, the mediaType is set to XML. So, we need not have this condition here.

joheredi (approved):
LGTM

PR #926, opened 2021-05-05
Update CODEOWNERS

daviwil:
I haven't worked on this repo in a while so I'm removing myself from CODEOWNERS so I can reduce the amount of GitHub notifications I'm getting.  Please feel free to add me as a reviewer on any PRs where you feel my feedback would be useful!


deyaaeldeen (approved):
yeah gh notifications can be annoying, you're welcome here anytime though :)

@chradek I think you are in the same boat

PR #934, opened 2021-05-11
lro names start with begin

deyaaeldeen:
Fixes https://github.com/Azure/autorest.typescript/issues/785.


src/generators/utils/operations.ts
```diff
@@ -0,0 +1,14 @@
+import { OperationDetails } from "../../models/operationDetails";
+import { NameType, normalizeName } from "../../utils/nameUtils";
+
+export function calculateMethodName(operation: OperationDetails): string {
+  const name = `${operation.namePrefix || ""}${normalizeName(
+    operation.name,
+    NameType.Property
+  )}`;
+  const firstLetter = name.substr(0, 1)[0].toUpperCase();
+  const lroName = operation.isLRO
```
joheredi, line 10: maybe you can name this const `operationName` instead, as in the else case it won't be a lroName right?

Issue #936, opened 2021-05-11
[LRO] pollUntilDone to throw on "failed" or "canceled" status.


joheredi:
For consistency, we need to throw an error when the LRO returns a "failed" or "canceled" status. We should preserve the response body in the error message.

For reference, this is what Python is doing:
https://github.com/Azure/azure-sdk-for-python/blob/99ba7fa4baf6ee4be4082097315a46f2090e438d/sdk/core/azure-core/azure/core/polling/base_polling.py#L62

deyaaeldeen:
code complete

PR #942, opened 2021-05-12
LRO throws when fails

deyaaeldeen:
According to the code gen crew, LROs in other languages throw when fail. This PR implements the same behavior in JS. Fixes https://github.com/Azure/autorest.typescript/issues/936


joheredi (approved):
Looks great!

Issue #938, opened 2021-05-12
Provide an LRO overload that returns the result directly


deyaaeldeen:
As per the plan discussed here: https://github.com/Azure/azure-sdk/issues/2638#issuecomment-816936864. The name should have the `begin` prefix and `AndWait` suffix.

deyaaeldeen:
Code complete

PR #943, opened 2021-05-13
Generate additional method for LRO operations that return the results directly

deyaaeldeen:
Fixes https://github.com/Azure/autorest.typescript/issues/938

Notice that when an LRO returns a paged result, code gen generates a method that waits for the result. So in this PR I renamed those methods so that they have the suffix `AndWait` and opened https://github.com/Azure/autorest.typescript/issues/944 to track the addition of an overload that returns a poller. This should not be a blocker for the first beta release.


joheredi (approved):
LGTM! :shipit:

PR #945, opened 2021-05-18
Update readme

deyaaeldeen:
To add a list of supported options and a simple getting started guide

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,18 +48,7 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
 For more information see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/) or
 contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.
 
-# Simple AutoRest Extension
-
-A simple AutoRest extension that can be used as a template for or documentation about writing such extensions.
-Since this a TypeScript project, we distinguish between language agnostic and language dependent requirements.
-
-## Language Agnostic Requirements
-
-More information about these can be found [here](https://github.com/Azure/autorest/tree/master/docs/developer), summary:
-
-1. Specify the start command of the extension as the `start` script in the `package.json`
-2. Implement the AutoRest extension protocol (here: handled by library, see below)
-3. Hook up plugins into the AutoRest pipeline DAG, e.g.
+## Autorest Typescript Plugin Configuration
 
 ```yaml
 version: 3.1.2
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # TypeScript Plugin for Autorest
 
-[Autorest](https://github.com/Azure/autorest/blob/master/docs/readme.md) is a suite of tools to automatically generate SDKs for cloud services. This project provides a backend that generates SDKs in TypeScript.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,93 +1,48 @@
-# Contributing
+# TypeScript Plugin for Autorest
 
-This project welcomes contributions and suggestions. Most contributions require you to agree to a
-Contributor License Agreement (CLA) declaring that you have the right to, and actually do, grant us
-the rights to use your contribution. For details, visit https://cla.microsoft.com.
+[Autorest](https://github.com/Azure/autorest/blob/master/docs/readme.md) is a suite of tools to automatically generate SDKs for cloud services. This project provides a backend that generates SDKs in TypeScript.
```
ramya-rao-a, line 3: Why the term "backend" when there is no "frontend" in this context?

README.md
```diff
@@ -1,93 +1,48 @@
...
 
-## Language Agnostic Requirements
-
-More information about these can be found [here](https://github.com/Azure/autorest/tree/master/docs/developer), summary:
-
-1. Specify the start command of the extension as the `start` script in the `package.json`
-2. Implement the AutoRest extension protocol (here: handled by library, see below)
-3. Hook up plugins into the AutoRest pipeline DAG, e.g.
+```bash
+npm install -g autorest
```
ramya-rao-a, line 12: Shouldnt this be `@autorest/typescript`?

README.md
```diff
@@ -1,93 +1,48 @@
...
-    scope: typescript-scope/emitter
+## Options
 
-typescript-scope/emitter:
-  input-artifact: typescript-files
+In addition to the [list of Autorest flags](https://github.com/Azure/autorest/blob/master/docs/generate/flags.md), you can further control the behavior of the typescript generator with the following flags:
 
-output-artifact: typescript-files
-```
+| Flag                            | Description                                                                                                                             |
```
ramya-rao-a, line 27: Can we have the description include a note on when the user would probably want to use it?

For example, the flag for core v2 can point to https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/core/README.md so that reader gets more information on what they are opting into

--hide-client is probably useful if one is planning to write a convenience layer over the generated code rather than use the generated code directly

etc

README.md
deyaaeldeen: No, we use autorest itself and when you pass `--typescript` argument to it, it will use the TS extension.

README.md
deyaaeldeen: the frontend is m4 but I understand there is no need to be overly academic here :D I will replace `backend` with `extension`.

deyaaeldeen:
This is how the readme would look like: https://github.com/Azure/autorest.typescript/blob/addf504dd6019c96242b4ba1c319f3969a799f48/README.md

README.md
ramya-rao-a: So, the typescript extension is installed automatically when we install `autorest`?

joheredi (approved):
I think you can remove this block, doesn't say much
image

You can just add a title before the config like # Autorest Typescript Plugin Configuration

README.md
deyaaeldeen: it gets downloaded to a cache directory when you request it using the TS argument. You can also specify which version to use with the use arg: `--use=@microsoft.azure/autorest.typescript@4.4.4`

PR #946, opened 2021-05-18
Core V2 Migration Work 

sarangan12:
This PR is for the Core V2 Migration Work. Among the test files, the following has not been migrated yet:

1. lro.spec.ts
2. lroParametrizedEndPoints.spec.ts
3. mediaTypesV3LroClient.spec.ts
4. paging.spec.ts
5. pagingNoIterators.spec.ts

Other than the above, the package is able to build and test successfully. This PR can be merged only after the release and rebase of `core-client` and `core-rest-pipeline` packages. 

@deyaaeldeen @joheredi @ramya-rao-a FYI....


package.json
```diff
@@ -1,91 +1,98 @@
...
+        "README.md",
+        "LICENSE"
+    ],
+    "dependencies": {
+        "@autorest/codemodel": "^4.14.5",
+        "@autorest/extension-base": "^3.2.0",
+        "@azure-tools/codegen": "^2.4.267",
+        "@azure/core-client": "^1.1.2",
+        "@azure/core-http": "^1.2.1-alpha.20201203.2",
+        "@azure/core-https": "^1.0.0-beta.1",
```
joheredi, line 39: Can we get rid of these 2 dependencies?

package.json
```diff
@@ -1,91 +1,98 @@
...
+        "@azure/core-lro": "^1.0.5",
+        "@azure/core-paging": "^1.1.3",
+        "@azure/core-rest-pipeline": "^1.1.0-beta.2",
```
joheredi, line 42: We'll need to use ^1.0.3

src/generators/clientContextFileGenerator.ts
```diff
@@ -326,3 +379,12 @@ function getRequiredParamAssignments(requiredParameters: ParameterDetails[]) {
     .filter(({ name }) => !disallowedClientParameters.includes(name))
     .map(({ name }) => `this.${name} = ${name};`);
 }
+
+function getEndpointV2({ endpoint }: EndpointDetails, optionsBag: OptionsBag) {
+  const updatedEndpoint = optionsBag.allowInsecureConnection
+    ? endpoint?.replace("https://", "http://")
+    : endpoint?.replace("http://", "https://");
+  return `options.endpoint ${
```
joheredi, line 387: Do we need to replace the protocol? It is not guaranteed that if http is available https would be available for the same endpoint.

I would prefer to keep the original protocol and fix it in swagger or transform if needed. Users would get a very clear error saying that http are not allowed by default so they would have 2 options

1) use their https endpoints
2) set the allowInsecureConnection flag to autorest

joheredi: otherwise it may be hard to track down what's going on. If I set my server to http://localhost and no allowInsecureConnection flag, my server url would be mutated under the hoods to https://localhost and the error I'm going to get is probably a resource not found or connection error. In my experience noticing in the error message that there is an extra (s) in http could take hours.

src/generators/indexGenerator.ts
```diff
@@ -46,4 +46,6 @@ export function generateIndexFile(
       }
     ]);
   }
+
+  indexFile.fixUnusedIdentifiers();
```
joheredi, line 50: I didn't know about this function! Pretty cool :)

src/generators/modelsGenerator.ts
```diff
@@ -332,26 +356,43 @@ function buildResponseType(
...
-    ]
-  });
 
-  let intersectionTypes: WriterFunctionOrValue[] = [innerTypeWriter];
+  let intersectionTypes: WriterFunctionOrValue[] = [];
+  let innerTypeWriter: WriterFunctionOrValue = Writers.objectType({});
+
+  if (optionsBag.useCoreV2) {
+    if (
+      bodyProperties?.mainProperties &&
```
joheredi, line 365: nit: you may be able to make this in one line. `bodyProperties?.mainProperties?.length`

src/generators/operationGenerator.ts
```diff
@@ -909,12 +1009,20 @@ function writeMultiMediaTypeOperationBody(
...
   clientDetails: ClientDetails,
-  isInline = false
+  isInline = false,
+  optionsBag: OptionsBag
 ): void {
-  operationMethod.addStatements([
-    "let operationSpec: coreHttp.OperationSpec;",
-    "let operationArguments: coreHttp.OperationArguments;"
-  ]);
+  if (!optionsBag.useCoreV2) {
```
joheredi, line 1015: nit: you could also do something like this

```typescript
const coreImport = !optionsBag.useCoreV2 ? "coreHttp" :  "coreClient"
operationMethod.addStatements([      
  `let operationSpec: ${coreImport}.OperationSpec;`,
  `let operationArguments: ${coreImport}.OperationArguments;`
 ]);


```

src/generators/operationGenerator.ts
```diff
@@ -1119,7 +1233,9 @@ export function addOperationSpecs(
         declarations: [
           {
             name: operationSpec.name,
-            type: "coreHttp.OperationSpec",
+            type: !optionsBag.useCoreV2
+              ? "coreHttp.OperationSpec"
```
joheredi, line 1237: I wonder if we can have named import for OperationSpec? it may reduce the number of checks for useCoreV2 considerably

src/generators/static/packageFileGenerator.ts
```diff
@@ -27,7 +29,11 @@ export function generatePackageJson(
     dependencies: {
       ...(hasLRO && { "@azure/core-lro": "^1.0.5" }),
       ...(hasAsyncIterators && { "@azure/core-paging": "^1.1.1" }),
-      "@azure/core-http": "^1.1.4",
+      ...(!optionsBag.useCoreV2 && { "@azure/core-http": "^1.1.4" }),
+      ...(optionsBag.useCoreV2 && { "@azure/core-client": "1.0.0-beta.1" }),
+      ...(optionsBag.useCoreV2 && {
+        "@azure/core-rest-pipeline": "1.0.0-beta.1"
```
joheredi, line 35: We need to update these versions

src/transforms/groupTransforms.ts
```diff
@@ -7,14 +7,29 @@ import {
...
-export function transformGroups(codeModel: CodeModel) {
-  return (codeModel.schemas.groups || []).map(transformGroup);
+export function transformGroups(codeModel: CodeModel, optionsBag: OptionsBag) {
+  codeModel.schemas.groups = codeModel.schemas.groups || [];
+  const objectDetails = [];
+  for (const grp of codeModel.schemas.groups) {
+    objectDetails.push(transformGroup(grp, optionsBag));
+  }
+  return objectDetails;
+  // return (codeModel.schemas.groups || []).map(transformGroup);
```
joheredi, line 19: remove this commented line?

src/transforms/groupTransforms.ts
```diff
@@ -7,14 +7,29 @@ import {
...
+  optionsBag: OptionsBag
+): ObjectDetails {
   const { name, description } = getLanguageMetadata(language);
-  const groupProperties = (properties || []).map(transformProperty);
+  properties = properties || [];
+  const groupProperties = [];
+  for (const prop of properties) {
+    groupProperties.push(transformProperty(prop, optionsBag));
+  }
+  // const groupProperties = (properties || []).map(transformProperty);
```
joheredi, line 32: this one as well

src/transforms/groupTransforms.ts
joheredi: I'm curious about this change, would it be the same as 

```typescript
return (codeModel.schemas.groups ?? []).map(group => transformGroup(group, optionsBag));
```

I like more the map approach as it is easier to read and has less surface for bugs

src/transforms/objectTransforms.ts
```diff
@@ -61,23 +64,26 @@ export function transformObject(
...
+    serializedName,
+    required,
+    readOnly,
+    nullable
+  }: Property | GroupProperty,
+  optionsBag: OptionsBag
+): PropertyDetails {
   const metadata = getLanguageMetadata(language);
-  const typeDetails = getTypeForSchema(schema);
+  const typeDetails = getTypeForSchema(schema, false, optionsBag.useCoreV2);
```
joheredi, line 86: Can you add a comment about what `false` mean?

src/transforms/urlTransforms.ts
```diff
@@ -29,7 +29,7 @@ export async function transformBaseUrl(
   }
 
   return {
-    endpoint: endpoint,
+    endpoint: endpoint?.replace("http://", "https://"),
```
joheredi, line 32: I think we shouldn't replace urls, let users fix them in swagger or transforms. Or set the allowInsecureConnection flag

src/typescriptGenerator.ts
```diff
@@ -87,12 +71,33 @@ export async function generateTypeScriptLibrary(
   ))
     ? true
     : armLibrary;
+  const useCoreV2: boolean = (await host.GetValue("use-core-v2")) || false;
```
joheredi, line 74: Can we make this true by default?

My worry about having this false by default is that even when we get these changes in, people will keep creating new SDKs using core v1, and we'll need to do the migrations at one point.

I think it is better to force core v2 as soon as this is merged and set `use-core-v2: false` for spceial cases

test/integration/azureSpecialProperties.spec.ts
```diff
@@ -32,36 +32,38 @@ describe("auth validation", () => {
...
-    // Validate operation
-    assert.equal(result._response.status, 200);
+    await client.apiVersionDefault.getMethodGlobalValid(responseStatusChecker);
 
     // Validate auth header
-    assert.equal(
-      result._response.request.headers.get("authorization"),
-      "Bearer test-token"
-    );
+    // assert.equal(
```
joheredi, line 42: Is there a reason why we cannot check this with onResponse?

test/integration/azureSpecialProperties.spec.ts
```diff
@@ -232,26 +241,30 @@ describe("AzureSpecialProperties", () => {
...
             "x-ms-client-request-id": validClientId
           }
-        }
+        },
+        ...responseStatusChecker
       };
-      let result = await client.xMsClientRequestId.get(options);
-      assert.equal(result._response.status, 200);
-      assert.equal(result._response.headers.get("x-ms-request-id"), "123");
+      await client.xMsClientRequestId.get(options);
```
joheredi, line 247: Here we are missing the test for "x-ms-request-id" equality to "123"

test/integration/bodyDictionary.spec.ts
```diff
@@ -220,13 +209,12 @@ describe("BodyDictionary", () => {
...
-    const result = await client.dictionary.putEmpty({});
-    assert.equal(result._response.status, 200);
+    await client.dictionary.putEmpty({}, responseStatusChecker);
   });
 
   it("should get null dictionary", async () => {
-    const result = await client.dictionary.getNull();
-    assert.deepEqual(result._response.parsedBody, undefined);
+    await client.dictionary.getNull();
+    // assert.deepEqual(result._response.parsedBody, undefined);
```
joheredi, line 217: We are not testing for anything here

test/integration/bodyFormData.spec.ts
```diff
@@ -19,48 +20,47 @@ describe("Integration tests for BodyFormData", () => {
...
+      const buff = await readStreamToBuffer(result.readableStreamBody);
+      assert.deepEqual(buff, fileContent);
     } else {
       assert.fail("ReadableStreamBody must not be null!!!");
     }
   });
 
-  it("should correctly accept file via body", async function() {
-    client = new BodyFormDataClient();
+  it.skip("should correctly accept file via body", async function() {
```
joheredi, line 30: Why are we skipping this?

test/integration/bodyFormData.spec.ts
```diff
@@ -19,48 +20,47 @@ describe("Integration tests for BodyFormData", () => {
...
     let uploadNotified = false;
     let downloadNotified = false;
     const response = await client.formdata.uploadFileViaBody(content, {
       requestOptions: {
         onUploadProgress: ev => {
           uploadNotified = true;
-          ev.loadedBytes.should.be.a("Number");
+          console.log(ev);
+          assert.ok(typeof ev.loadedBytes === "number");
+          //ev.loadedBytes.should.be.a("Number");
```
joheredi, line 57: I think we can remove this commented line and the one on L63

sarangan12: Done

joheredi (commented):
The changes look great! I added a couple of comments, most are small things. There are a couple of things that worry me:

* The number one thing is that we are removing tests for corev1 and I'm pretty sure we are going to break generated code not using use-core-v2 flag pretty quickly. I think, if we are going to support generating corev1 we should keep testing it, this means we'd need to duplicate the whole suite. Keep a copy without any changes which will target corev1 and the tests you have updated to test corev2.
* I think we should make --use-core-v2 **true** by default. This will save us and our partners time migrating new libraries that miss setting the flag

@sarangan12, @ramya-rao-a, @xirzec, @deyaaeldeen - Any thoughts?

Code-wise I think the changes look pretty good!

package.json
joheredi: nevermind, we still support generating corev1

test/integration/bodyDictionary.spec.ts
sarangan12: Fixed. Thanks

test/integration/azureSpecialProperties.spec.ts
sarangan12: Fixed.

src/transforms/urlTransforms.ts
sarangan12: I have fixed it by using the options.allowInsecureConnection flag in this logic.

deyaaeldeen:
> - The number one thing is that we are removing tests for corev1 and I'm pretty sure we are going to break generated code not using use-core-v2 flag pretty quickly. I think, if we are going to support generating corev1 we should keep testing it, this means we'd need to duplicate the whole suite. Keep a copy without any changes which will target corev1 and the tests you have updated to test corev2.
> - I think we should make --use-core-v2 true by default. This will save us and our partners time migrating new libraries that miss setting the flag

These are exactly my thoughts as well and I talked to @sarangan12 and @ramya-rao-a privately about this. For now the flag should be set to true and we will keep the flag until all packages migrate to core v2 just in case there is a need to do a patch release for a version that uses core v1.

src/transforms/objectTransforms.ts
sarangan12: Done

src/transforms/groupTransforms.ts
sarangan12: Done

src/typescriptGenerator.ts
sarangan12: As mentioned in one of the comments by Deya, below, we are going to turn it on by default once all packages have moved to core V2. This is to support the patch version releases.

sarangan12: I am changing the code so the default mode will be use-core-v2:true. SDKs who want core V1 can set it to false.

src/transforms/groupTransforms.ts
sarangan12: Done

src/generators/operationGenerator.ts
sarangan12: Done

src/generators/modelsGenerator.ts
sarangan12: Done

src/generators/clientContextFileGenerator.ts
sarangan12: Agreed. This replace is not necessary. Removed it in both clientContextFileGenerator and transforms

src/generators/operationGenerator.ts
sarangan12: I think this one we can keep it as such. The main reason is we have dependencies such as OperationSpec, OperationOptions, RequestOptionsBase, operationOptionsToRequestOptionsBase, OperationArguments, etc.

src/generators/static/packageFileGenerator.ts
sarangan12: Done

package.json
sarangan12: This is intentional. I am expecting some more changes in core-packages. Before releasing the preview version of SDK generator, there will be a core-rest-pipeline GA release (1.0.4) and this will be dependent on that. For now, this version is intentional as this version has a specific bug fix that I did yesterday.

test/integration/bodyFormData.spec.ts
sarangan12: This will be dealt after the merge as part of this issue: https://github.com/Azure/autorest.typescript/issues/947

test/integration/azureSpecialProperties.spec.ts
sarangan12: This will be dealt after the merge as part of this issue: https://github.com/Azure/autorest.typescript/issues/947

package.json
sarangan12: I have removed the https package. We still need the core-http package since our generator is using several definitions from it.

joheredi (approved):

PR #948, opened 2021-05-22
LROs use core v2

deyaaeldeen:
# Rewrites the generated code for LROs to use core v2. 

## Overview

One of the main aspects of the core v2 is that it ditched the `_response` property and instead accepts a callback that gets applied to the raw response. However, LROs send one or more requests until they are done so when should the customer-provided callback get called? There are mainly three times when this could happen:
- the initial response, but then it does not have interesting information to the customer because if it has anything, it would be a location to where to poll or where the provisioned resource will be allocated.
- the response of each subsequent poll request, but then again this does not necessarily have everything the customer is interested in yet.
- the very last response which is probably what the customer care about because it has the final results.

This PR implements this third behavior where customer-provided callbacks get called on the last response. This behavior necessitated quite a large refactoring so that LROs are implemented explicitly as a state transition machine and makes sure the customer-provided callback is called only once in the last transition. This is easy for most cases except for the case when we know we are done but still need to send a GET request somewhere to retrieve the results. Using the state machine, we can queue that request to be done in the next transition. Note that we can not send GET requests inside the onResponse callback directly because they are not async. One little advantage of this design is that now we maintain the guarantee that exactly one request is sent in a polling cycle which is what our customers expect.

All LRO tests are passing.

## Issues

core v2 dropped support for cookies but the test server relies heavily on them in LRO/retry scenarios. @jeremymeng created a client that support cookies for test purposes and I used that to get around this issue.


joheredi (approved):
Changes look great!

What happens if we generate with `--use-core-v2=false`? It looks like the new polling strategies use corev2 exclusively (which I think is fine). Do we get a mix of some pieces using corev2 and other parts using core v1?

If that's the case, I think we should pull the band-aid and remove corev1 altogether rather soon. If we need to service something in corev1 we could spawn off a service branch off the last commit before merging corev2. Otherwise, I think we'll end up in an awkward situation. Of course, that would be in a PR of itself, just wanted to gauge what you think

deyaaeldeen:
Closing this in favor of https://github.com/Azure/autorest.typescript/pull/955 and https://github.com/Azure/autorest.typescript/pull/956

Issue #949, opened 2021-05-24
Core V2 Migration Pending Test Cases - Non LRO Related - Part I


sarangan12:
In the `test/integration/azureSpecialProperties.spec.ts`, we have the following code:

```js
it("should add authorization header", async () => {
    const expectedScopes = [
      "https://microsoft.com/.default",
      "http://microsoft.com/.default"
    ];

    const mockCredential: TokenCredential = {
      getToken: async scopes => {
        assert.deepEqual(scopes, expectedScopes);
        return {
          token: "test-token",
          expiresOnTimestamp: 111111
        };
      }
    };

    const client = new AzureSpecialPropertiesClient(
      mockCredential,
      "1234-5678-9012-3456",
      { allowInsecureConnection: true }
    );

    await client.apiVersionDefault.getMethodGlobalValid(responseStatusChecker);

    // Validate auth header
    // assert.equal(
    //   result._response.request.headers.get("authorization"),
    //   "Bearer test-token"
    // );
  });
```

In the above code we have to remove the comments from the last assert. We should find why the `authorization` header is not returned and then remove the comments.

sarangan12:
Code Complete with https://github.com/Azure/autorest.typescript/pull/950

PR #950, opened 2021-05-24
Code Changes to fix credential scopes

sarangan12:
During the previous merge of the Core V2 Migration work, 1 of the test case was marked as skipped. The following issues were created to address them: https://github.com/Azure/autorest.typescript/issues/949

This PR is to address that issue 

@deyaaeldeen @ramya-rao-a FYI....


test/integration/azureSpecialProperties.spec.ts
```diff
@@ -33,16 +33,26 @@ describe("auth validation", () => {
     const client = new AzureSpecialPropertiesClient(
       mockCredential,
       "1234-5678-9012-3456",
-      { allowInsecureConnection: true }
+      {
+        allowInsecureConnection: true,
+        credential: mockCredential,
```
joheredi, line 38: Why do we have to pass the credential as part of the ClientOptions? We are already passing it as the constructor's first parameter

test/integration/azureSpecialProperties.spec.ts
```diff
@@ -33,16 +33,26 @@ describe("auth validation", () => {
...
 
     // Validate auth header
-    // assert.equal(
-    //   result._response.request.headers.get("authorization"),
-    //   "Bearer test-token"
-    // );
+    assert.equal(
+      _response!.request.headers.get("authorization"),
+      "Bearer test-token"
+    );
```
deyaaeldeen, line 55: can we move the check to inside the callback?

sarangan12: 1. We are using this pattern across all test cases. 
2. I think it is more cleaner approach to have the asserts outside.

test/integration/azureSpecialProperties.spec.ts
sarangan12: Modified the code now. Also created https://github.com/Azure/autorest.typescript/issues/952 to fix the user agent. (Because that needs changes to the core)

deyaaeldeen (approved):

test/integration/azureSpecialProperties.spec.ts
joheredi: TBH I usually worry when I see code extracting stuff from the callback, in general you don't have guarantees on when the callback is going to be called, so consuming a value set inside the callback is risky and can lead to race conditions. 

In this case it works because we know that by the time getMethodGlobalValid has resolved onResponse would've been called.

I'm not going to block on this but would also prefer to see the checks inside

joheredi (approved):
Looks good! Consider moving checks inside onResponse callback

PR #951, opened 2021-05-24
Refactor options

joheredi:
Refactoring how we access Autorest Options (flags). Currently, we need to plumb the options object from the Autorest Typescript plugin entry point, often to a very deep function where parents don't need the options.

This PR makes a global function to access options information anywhere in the code without having to plumb it through. It reads the flags once at the beginning of the plugin and uses cached options afterward.

Additional change:
- Updated xmsErrorResponses test client to use core-v2 and set 0 as retry timeout
- Removed `console.log` in test
- Update dependencies


src/autorestSession.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export interface AutorestOptions {
+  restLevelClient?: boolean;
+  azureArm?: boolean;
+  addCredentials?: boolean;
+  credentialKeyHeaderName?: string;
+  credentialScopes?: string[];
+  srcPath: string;
+  packageDetails: PackageDetails;
+  licenseHeader: boolean;
```
sarangan12, line 15: In the existing code base, it is `shouldGenerateLicense`. The `licenseHeader` makes more sense. Please make sure corresponding change is done in the places where `shouldGenerateLicense` is used.

src/autorestSession.ts
```diff
@@ -0,0 +1,71 @@
...
+  allowInsecureConnection?: boolean;
+  disablePagingAsyncIterators?: boolean;
+}
+
+let host: Host;
+let session: Session<CodeModel>;
+let options: AutorestOptions;
+
+export async function initializeSession(autorestHost: Host) {
+  if (!host) {
```
sarangan12, line 30: nit: 

```js
host = host ?? autorestHost;
session = session ?? await startSession<CodeModel>(host, undefined, codeModelSchema);
options = options ?? await extractAutorestOptions();
```

sarangan12 (approved):
Looks terrific. Just left one nit comment

deyaaeldeen (approved):
This is great!, thanks!

PR #953, opened 2021-05-24
Generate static files for REST Level Clients

joheredi:
Generating static files for REST Level Client

### DIFF
--- a/src/generators/static/packageFileGenerator.ts
+++ b/src/generators/static/packageFileGenerator.ts
@@ -53,8 +53,6 @@ function restLevelPackage(packageDetails: PackageDetails) {
     main: "esm/index.js",
     types: "esm/index.d.ts",
     scripts: {
-      regenerate:
-        "autorest --reset && autorest --typescript --use=https://aka.ms/azsdk/typescript/rlc --rest-level-client=true https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/specification/iothub/resource-manager/readme.md --credential-scopes=https://management.azure.com/.default --add-credentials=true --output-folder=. --generate-metadata=false",
       test: 'echo "Error: no test specified" && exit 1',
       build: "tsc --build && npm run extract-api",
       "extract-api": "mkdirp ./review && api-extractor run --local"
@@ -64,11 +62,8 @@ function restLevelPackage(packageDetails: PackageDetails) {
     license: "ISC",
     dependencies: {
       "@azure-rest/core-client": "1.0.0-beta.2",
-      "@azure/ai-text-analytics": "^5.0.1",
       "@azure/core-auth": "^1.1.4",
-      "@azure/core-rest-pipeline": "^1.0.3",
-      "@azure/identity": "^1.2.2",
-      "@azure/storage-blob": "^12.5.0"
+      "@azure/core-rest-pipeline": "^1.0.3"
     },
     devDependencies: {
       autorest: "latest",
--- a/src/generators/static/packageFileGenerator.ts
+++ b/src/generators/static/packageFileGenerator.ts
@@ -61,9 +61,9 @@ function restLevelPackage(packageDetails: PackageDetails) {
     author: "",
     license: "ISC",
     dependencies: {
-      "@azure-rest/core-client": "1.0.0-beta.2",
+      "@azure-rest/core-client": "1.0.0-beta.4",
       "@azure/core-auth": "^1.1.4",
-      "@azure/core-rest-pipeline": "^1.0.3"
+      "@azure/core-rest-pipeline": "^1.0.4"
     },
     devDependencies: {
       autorest: "latest",
@@ -72,7 +72,7 @@ function restLevelPackage(packageDetails: PackageDetails) {
       dotenv: "^8.2.0",
       prettier: "^2.2.1",
       "ts-node": "^9.1.1",
-      typescript: "^4.1.5",
+      typescript: "~4.2.4",
       mkdirp: "^1.0.4"
     }
   };
--- a/src/generators/static/packageFileGenerator.ts
+++ b/src/generators/static/packageFileGenerator.ts
@@ -29,7 +29,7 @@ export function generatePackageJson(
     }
     packageJsonContents = regularAutorestPackage(clientDetails, packageDetails);
   } else {
-    packageJsonContents = restLevelPackage();
+    packageJsonContents = restLevelPackage(packageDetails);
   }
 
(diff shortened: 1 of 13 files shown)


src/generators/static/packageFileGenerator.ts
```diff
@@ -46,7 +46,41 @@ export function generatePackageJson(
...
-  throw Error("Rest Level Client - Not yet implemented");
+  return {
+    name: `${packageDetails.name}`,
+    version: `${packageDetails.version}`,
+    description: `${packageDetails.description}`,
+    main: "esm/index.js",
+    types: "esm/index.d.ts",
+    scripts: {
+      regenerate:
+        "autorest --reset && autorest --typescript --use=https://aka.ms/azsdk/typescript/rlc --rest-level-client=true https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/specification/iothub/resource-manager/readme.md --credential-scopes=https://management.azure.com/.default --add-credentials=true --output-folder=. --generate-metadata=false",
```
deyaaeldeen, line 57: it is intended to be hard coded for one package?

src/generators/static/packageFileGenerator.ts
```diff
@@ -46,7 +46,41 @@ export function generatePackageJson(
...
+      test: 'echo "Error: no test specified" && exit 1',
+      build: "tsc --build && npm run extract-api",
+      "extract-api": "mkdirp ./review && api-extractor run --local"
+    },
+    keywords: [],
+    author: "",
+    license: "ISC",
+    dependencies: {
+      "@azure-rest/core-client": "1.0.0-beta.2",
+      "@azure/ai-text-analytics": "^5.0.1",
```
deyaaeldeen, line 67: I think this should not be here.

src/generators/static/packageFileGenerator.ts
```diff
@@ -46,7 +46,41 @@ export function generatePackageJson(
...
+      "@azure/core-auth": "^1.1.4",
+      "@azure/core-rest-pipeline": "^1.0.3",
+      "@azure/identity": "^1.2.2",
+      "@azure/storage-blob": "^12.5.0"
```
deyaaeldeen, line 71: and this one?

src/generators/static/tsConfigFileGenerator.ts
```diff
@@ -4,32 +4,56 @@
...
+    outDir: "./esm",
+    importHelpers: true
+  },
+  include: ["./src/**/*.ts"],
+  exclude: ["node_modules"]
+};
+
+const restLevelTsConfig = {
+  compilerOptions: {
+    module: "commonjs",
```
deyaaeldeen, line 30: why cjs?

src/generators/static/tsConfigFileGenerator.ts
```diff
@@ -4,32 +4,56 @@
...
+    "browser",
+    "dist",
+    "dist-esm",
+    "./samples/**/*.ts"
+  ]
+};
+
 export function generateTsConfig(project: Project) {
-  const { generateMetadata } = getAutorestOptions();
+  const { generateMetadata, restLevelClient } = getAutorestOptions();
```
deyaaeldeen, line 48: NIT: I am wonder if we should generalize `restLevelClient` to be more like an enum instead of a boolean

deyaaeldeen (commented):
Looks good for the most part! I left a few minor comments

src/generators/static/tsConfigFileGenerator.ts
joheredi: Updated to es6

src/generators/static/tsConfigFileGenerator.ts
joheredi: What values do you think we should get? Something like `--client-type=<"rest" | "highlevel">`?

deyaaeldeen: I think that would be great

joheredi: It looks like all other languages are using the boolean flag. It is also less prone to typos. I think I'll keep it this way and if we need to generate other kind of clients in the future we can turn it to an enum like

src/generators/static/packageFileGenerator.ts
```diff
@@ -45,8 +45,37 @@ export function generatePackageJson(
...
+    scripts: {
+      test: 'echo "Error: no test specified" && exit 1',
+      build: "tsc --build && npm run extract-api",
+      "extract-api": "mkdirp ./review && api-extractor run --local"
+    },
+    keywords: [],
+    author: "",
+    license: "ISC",
+    dependencies: {
+      "@azure-rest/core-client": "1.0.0-beta.2",
```
deyaaeldeen, line 64: do you want to switch this to 1.0.4 instead?

deyaaeldeen: ops sorry, I thought it is the non-rest one.

src/generators/static/packageFileGenerator.ts
```diff
@@ -45,8 +45,37 @@ export function generatePackageJson(
...
+      "@azure/core-auth": "^1.1.4",
+      "@azure/core-rest-pipeline": "^1.0.3"
```
deyaaeldeen, line 66:
```suggestion
      "@azure/core-rest-pipeline": "^1.0.4"
```

src/generators/static/packageFileGenerator.ts
```diff
@@ -45,8 +45,37 @@ export function generatePackageJson(
...
+    },
+    devDependencies: {
+      autorest: "latest",
+      "@microsoft/api-extractor": "^7.13.2",
+      "@types/node": "^14.14.22",
+      dotenv: "^8.2.0",
+      prettier: "^2.2.1",
+      "ts-node": "^9.1.1",
+      typescript: "^4.1.5",
```
deyaaeldeen, line 75: would not you want to play it safe and use `~` rather than `^` here?

src/generators/static/tsConfigFileGenerator.ts
deyaaeldeen: Fair!

deyaaeldeen (approved):
Looks good! I am looking forward to have the RLC up and running soon!

src/generators/static/packageFileGenerator.ts
joheredi: Yeah, I'll set it ot ~4.2.4 which is the min version we need for Template Types

PR #954, opened 2021-05-25
Update rollup config to mark all deps as external

deyaaeldeen:
This will simplify bundling because we do not have to deal with bundling external deps. It is worth noting that this config is based on @witemple-msft's dev-tool shared one. I wish we could have used his directly but it is in the unpublished dev-tool.


ramya-rao-a:
@chradek, @witemple-msft Can you help with the review here?

src/generators/static/packageFileGenerator.ts
```diff
@@ -131,7 +135,7 @@ function regularAutorestPackage(
     scripts: {
       build:
         "tsc && rollup -c rollup.config.js && npm run minify && mkdirp ./review &&  npm run extract-api",
-      minify: `uglifyjs -c -m --comments --source-map "content='./dist/${packageDetails.nameWithoutScope}.js.map'" -o ./dist/${packageDetails.nameWithoutScope}.min.js ./dist/${packageDetails.nameWithoutScope}.js`,
+      minify: `uglifyjs -c -m --comments --source-map "content='./dist/index.js.map'" -o ./dist/${packageDetails.nameWithoutScope}.min.js ./dist/index.js`,
```
sarangan12, line 138: Could you confirm this line of change with Will?

sarangan12 (approved):

src/generators/static/packageFileGenerator.ts
deyaaeldeen: This change accounts for the new name for the bundle (`index.js`) and here it uses that name for minifying. I forgot to update the value of the `main` entry in `package.json` to use the new name as well. I will push that change shortly after I regenerate smoke tests.

As for the reasoning to change the bundle name, I am just using what we currently do in Track 2 dataplane, I do not think it affects anything. cc @witemple-msft

chradek (approved):

witemple-msft (approved):
LGTM. Should bring the default experience for codegen closer to the current convenience layer patterns.

PR #956, opened 2021-05-26
Refactor autorest run command

deyaaeldeen:
To be shared between integration tests and smoke tests. @sarangan12 helped with plumbing extra params to the command.

### DIFF
--- a/test/integration/generated/readmeFileChecker/src/keyVaultClient.ts
+++ b/test/integration/generated/readmeFileChecker/src/keyVaultClient.ts
@@ -36,6 +36,7 @@ import {
   KeyVaultClientGetDeletedSecretsNextResponse
 } from "./models";
 
+/** @internal */
 export class KeyVaultClient extends KeyVaultClientContext {
   /**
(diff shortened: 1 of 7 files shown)


test/integration/generated/readmeFileChecker/src/keyVaultClient.ts
```diff
@@ -36,6 +36,7 @@ import {
   KeyVaultClientGetDeletedSecretsNextResponse
 } from "./models";
 
+/** @internal */
```
joheredi, line 39: What triggered this change?

joheredi (approved):
Thanks for doing this!

test/integration/generated/readmeFileChecker/src/keyVaultClient.ts
deyaaeldeen: hide-client is set in the md file: https://github.com/Azure/autorest.typescript/blob/02eefa6ee665805141b9291e71a2f719d690fa4d/test/integration/swaggers/keyvaults-secrets.md so I am not sure why it was not triggered before.

PR #971, opened 2021-05-28
[V6] Update generated README.md using handlebar

deyaaeldeen:
To address feedback in https://github.com/Azure/autorest.typescript/pull/895.

~TODO: regenerate integration and smoke tests after finalizing the change.~


joheredi (approved):
This looks promising, lets see how it goes with the README and evaluate if it is worth moving some parts from ts-morph to handlebars templates

maorleger (commented):
This is very nice! Glad it worked out 😄 A few comments / questions if you don't mind

test/integration/generated/nameChecker/README.md
```diff
@@ -1,27 +1,87 @@
...
+# Azure SearchClient library for JavaScript
 
-This package contains an isomorphic SDK for SearchClient.
+This package contains an isomorphic SDK (runs both in node.js and in browsers) for Azure SearchClient.
+
+Client that can be used to query an index and upload, merge, or delete documents.
+
+[Source code](https://github.com/Azure/azure-sdk-for-js/tree/master/undefined) |
+[Package (NPM)](https://www.npmjs.com/package/@azure/search-documents) |
+[Samples](https://github.com/Azure/azure-sdk-for-js/tree/master/undefined/samples)
```
ramya-rao-a, line 9: Are we taking the source code path as input parameter? The samples folder can be inferred from that.
And why don't we have the link to API ref docs here? That can be inferred from the package name right?

src/generators/static/README.md.hbs
```diff
@@ -0,0 +1,101 @@
...
+- [{{ projectName }}]({{ repoURL }})
+
+[azure_sub]: https://azure.microsoft.com/free/
+[Impressions]({{ impressionURL }})
+[azure_cli]: https://docs.microsoft.com/cli/azure
+[azure_sub]: https://azure.microsoft.com/free/
+[azure_portal]: https://portal.azure.com
+[azure_identity]: {{ identityPackageURL }}
+[defaultazurecredential]: {{ identityPackageURL }}#defaultazurecredential
+{{/if}}
```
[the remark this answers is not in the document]
deyaaeldeen: these links are all azure-specific so they should be rendered only when this is an azure package.

test/integration/generated/nameChecker/README.md
```diff
@@ -1,27 +1,87 @@
-## Azure SearchClient SDK for JavaScript
+# Azure SearchClient library for JavaScript
```
ramya-rao-a, line 1: _Not blocking for this PR, but we should address this later, so please log an issue_

Our guideline for readme title is slightly different...

This should be `Azure Search client library for JavaScript` i.e `Azure <Name of the service> client library for JavaScript`. Same for the next line.

src/generators/static/readmeFileGenerator.ts
```diff
@@ -2,51 +2,135 @@
...
+ * @param clientDetails - the client details
+ * @param packageDetails - the package details
+ * @returns inferred metadata about the service, the package, and the client
+ */
+function createMetadata(
+  packageDetails: PackageDetails,
+  clientDetails: ClientDetails,
+  azureOutputDirectory?: string
+): Metadata {
+  const azureHuh = packageDetails.scopeName === "azure";
```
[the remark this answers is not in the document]
deyaaeldeen: meant to be a word for a question mark since I can not use the symbol itself in the var name :D

src/generators/static/README.md.hbs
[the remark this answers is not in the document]
maorleger: Oh I see, I missed the `if` statement on line 43. Thank you!

src/generators/static/README.md.hbs
```diff
@@ -0,0 +1,101 @@
...
+```bash
+npm install {{ clientPackageName }}
+```
+
+### Create and authenticate a `{{ clientClassName}}`
+
+To create a client object to access the {{ serviceName }} API, you will need the `endpoint` of your {{ serviceName }} resource and a `credential`.{{#if azure}} The {{ clientDescriptiveName }} can use Azure Active Directory credentials to authenticate.
+You can find the endpoint for your {{ serviceName }} resource in the [Azure Portal][azure_portal].
+
+#### Using an Azure Active Directory Credential
```
ramya-rao-a, line 46: Shouldnt this section be conditional based on whether or not the service supports AAD?

src/generators/static/README.md.hbs
```diff
@@ -0,0 +1,101 @@
...
+
+Client API key authentication is used in most of the examples, but you can also authenticate with Azure Active Directory using the [Azure Identity library][azure_identity]. To use the [DefaultAzureCredential][defaultazurecredential] provider shown below, or other credential providers provided with the Azure SDK, please install the `@azure/identity` package:
```
ramya-rao-a, line 48: > Client API key authentication is used in most of the examples,

The code generator should not be making this assumption.

src/generators/static/README.md.hbs
```diff
@@ -0,0 +1,101 @@
...
+
+Install the {{ clientDescriptiveName }} library for JavaScript with `npm`:
+
+```bash
+npm install {{ clientPackageName }}
+```
+
+### Create and authenticate a `{{ clientClassName}}`
+
+To create a client object to access the {{ serviceName }} API, you will need the `endpoint` of your {{ serviceName }} resource and a `credential`.{{#if azure}} The {{ clientDescriptiveName }} can use Azure Active Directory credentials to authenticate.
```
ramya-rao-a, line 43: Do all auto generated clients take "endpoint"?

src/generators/static/README.md.hbs
```diff
@@ -0,0 +1,101 @@
...
+{{/if}}
+{{#if samplesURL}}
+[Samples]({{samplesURL}})
+{{/if}}
+
+## Getting started
+
+### Currently supported environments
+
+- Node.js version 8.x.x or higher
```
ramya-rao-a, line 24: In v6 of the code generator, lets start with Node.js version 10

src/generators/static/README.md.hbs
```diff
@@ -0,0 +1,101 @@
+# {{ clientDescriptiveName }} library for JavaScript
+
+This package contains an isomorphic SDK (runs both in node.js and in browsers) for {{ clientDescriptiveName }}.
```
ramya-rao-a, line 3: node.js -> Node.js

src/utils/autorestOptions.ts
```diff
@@ -207,3 +209,19 @@ export async function getCredentialScopes(
...
   return undefined;
 }
+
+async function getAzureOutputDirectoryPath(
+  host: Host
+): Promise<string | undefined> {
+  const outputDirectoryPath: string | null = await host.GetValue(
+    "outputFolderUri"
+  );
+  const outputDirectoryRelativePath: string | undefined = outputDirectoryPath
```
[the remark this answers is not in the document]
deyaaeldeen: I do not think that would be equivalent. `outputFolderUri` is an undocumented and internal autorest variable so we do not have much guarantees about it. I think currently it is absolute. Anyway, that path could be anywhere so computing the `sdk/<pkg name>/<service name>` from it relatively to the home dir would not work generally. What I do here is to just get the last three parts of the path and make sure the first one is "sdk".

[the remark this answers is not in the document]
maorleger: Oh I see, but then why `join` it and then do a `substr`? it's already split so you can check the first element of the slice?

src/generators/static/README.md.hbs
deyaaeldeen: Done.

src/generators/static/README.md.hbs
deyaaeldeen: Made it behind the ``add-credentials` autorest argument

test/integration/generated/nameChecker/README.md
deyaaeldeen: Done.

test/integration/generated/nameChecker/README.md
deyaaeldeen: Fixed.

src/generators/static/README.md.hbs
deyaaeldeen: I believe so, @joheredi could we generate a client that does not take an endpoint?

ramya-rao-a: > can use Azure Active Directory credentials to authenticate

This should be behind the `addCredentials` check as well right?
And are we sure `addCredentials` is for indicating AAD support? Because we have "credentials" that are not AAD as well.

ramya-rao-a: Just looked at Track 1 mgmt plane package. The client constructor there does not require the endpoint.

deyaaeldeen: `addCredentials` means make the client constructor take the standard identit's credential type as input which implies that you do not have any fancy auth requirements. So I think it makes sense to make the AAD part of the readme behind that flag since it is part of the standard credential. If the package has custom auth requirements (e.g. API keys, conn strings, etc) then the author will most likely set that flag to false and implement these requirements manually in the convenience layer. At least this is what I saw so far.

deyaaeldeen: Ok I put that section behind `addCredentials` as well.

deyaaeldeen: I think we can merge this PR with the current behavior and we can revisit the endpoint issue later

Issue #1003, opened 2021-06-02
[Track2] Add the api.md generated by api extractor to published package


dw511214992:
api.md generated by api extractor is important for changelog detection tool. So it should be added in published package. However, it's ignored in packed tgz package. The "files" in package.json should includes api.md:

![image](https://user-images.githubusercontent.com/40835867/120438230-af260980-c3b3-11eb-9c96-126b9bf9db94.png)

sarangan12:
Code Complete with https://github.com/Azure/autorest.typescript/pull/1007

Issue #993, opened 2021-05-31
[v6] @azure/core-util is missing in the package.json


sarangan12:
In the arm-compute generated library (using V6 Code Generator), we have a file called `lroPoller.ts` which imports the `@azure/core-util` package.  But, this `@azure/core-util` package is not imported in the `package.json` file. So, this package must be included in the `package.json` using the `packageFileGenerator.ts` file.

**Why wasn't this error not caught in the integration test cases?**
This scenario is actually covered in several test cases including `paging`. But, it was running successfully. _Why?_ Because, the parent project (which is the SDK generator) has the `@azure/core-util` as one of the devDependencies. That was the reason this was passing in the integration testing.

**What is the Fix?**
In the `packageFileGenerator.ts` file, we need to add a line similar to the following line:

```js
.......
.......
...(hasLRO && { "@azure/core-util": "^1.0.0-beta.1" }),
.......
.......
```

**Next Steps**
It is actually simple to add the line above. But, the only issue is that We cannot add a beta package to Management Library. I will discuss with @jeremymeng to determine if this `@azure/core-util` package can have a GA release. Once it is released, I will update the version correctly in the above line and fix it. 

@ramya-rao-a @deyaaeldeen  FYI.......

sarangan12:
If `@azure/core-util` cannot have a GA release at this point, I can just clone the `delay` function. That will also resolve the issue

sarangan12:
Code Complete with https://github.com/Azure/autorest.typescript/pull/1007

Issue #902, opened 2021-04-12
[V6] headerCollectionPrefix is not generated in V6 SDK Generator


sarangan12:
While migrating `storage-queue` SDK to V6 SDK Generator, I found the following issue. Per the [swagger](https://github.com/Azure/azure-rest-api-specs/blob/storage-dataplane-preview/specification/storage/data-plane/Microsoft.QueueStorage/preview/2018-03-28/queue.json#L473), you can find the following JSON code:

```json
"x-ms-meta": {
    "type": "string",
    "x-ms-client-name": "Metadata",
    "x-ms-header-collection-prefix": "x-ms-meta-"
}
```

The above JSON Code is supposed to generate the following TS Code:

**parameters.ts**
```js
export const metadata: OperationParameter = {
  parameterPath: [ "options",  "metadata"],
  mapper: {
    serializedName: "x-ms-meta",
    type: {
      name: "Dictionary",
      value: {
        type: {
          name: "String"
        }
      }
    },
    headerCollectionPrefix: "x-ms-meta-"
  }
};
```

**mappers.ts**
```js
export const QueueGetPropertiesHeaders: coreHttp.CompositeMapper = {
  serializedName: "Queue_getPropertiesHeaders",
  type: {
    name: "Composite",
    className: "QueueGetPropertiesHeaders",
    modelProperties: {
      metadata: {
        serializedName: "x-ms-meta",
        type: {
          name: "Dictionary",
          value: { type: { name: "String" } }
        },
        headerCollectionPrefix: "x-ms-meta-"
      },
      ........
```

But, the following code is generated instead:

**parameters.ts**
```js
export const metadata: OperationParameter = {
  parameterPath: ["options", "metadata"],
  mapper: {
    serializedName: "x-ms-meta",
    xmlName: "x-ms-meta",
    type: {
      name: "Dictionary",
      value: { type: { name: "String" } }
    }
  }
};
```

**mappers.ts**
```js
export const QueueGetPropertiesHeaders: coreHttp.CompositeMapper = {
  serializedName: "Queue_getPropertiesHeaders",
  type: {
    name: "Composite",
    className: "QueueGetPropertiesHeaders",
    modelProperties: {
      metadata: {
        serializedName: "x-ms-meta",
        xmlName: "x-ms-meta",
        type: {
          name: "Dictionary",
          value: { type: { name: "String" } }
        }
         ............
```

[...]

PR #1007, opened 2021-06-02
Fix core-util package import, add review folder, and fix header prefix

sarangan12:
This PR is to address 3 issues:

1. [v6] @azure/core-util is missing in the package.json https://github.com/Azure/autorest.typescript/issues/993
I have explained the problem in the issue itself. After discussing with Ramya in https://github.com/Azure/azure-sdk-for-js/pull/15498, I have decided to clone the `delay` function in our SDK Generator. 

2. [Track2] Add the api.md generated by api extractor to published package https://github.com/Azure/autorest.typescript/issues/1003
This is a simple and straightforward issue. 

3. [V6] headerCollectionPrefix is not generated in V6 SDK Generator https://github.com/Azure/autorest.typescript/issues/902
I have explained the problem in the issue itself. The root cause is that we have been ignoring the extensions in the `parametersTransform.ts` and `mappersTransform.ts` files. This PR fixes it. 

@joheredi @deyaaeldeen Please review and approve the PR.

@ramya-rao-a FYI......


deyaaeldeen (commented):
It looks good for the most part. Not sure why smoke tests fail though. Perhaps because of the header collection prefix thing?

deyaaeldeen (approved):

PR #1010, opened 2021-06-04
Fix handling of constant parameter with default value

deyaaeldeen:
This PR makes sure to handle the special case for a constant parameter with a default value before the general case for all other default values. This fix is needed to generate text analytics v3.1 swagger.

Fixes https://github.com/Azure/autorest.typescript/issues/1012.


sarangan12:
1. Code Changes looks fine. 
2. Please provide an example in the description as what is the expected value ad what is the actual value we are getting before the code change. That will help the reviewers understand the problem that is resolved by this PR.
3. Thanks for adding the test case. Can you point to the location (in the generated code for the test case) , your code change is impacting?
4. CI is failing for all the 3 smoke tests. I am getting a feeling this has nothing to do with your code changes (To be on the safer side, you can run the smoke test once again locally and confirm). @joheredi The CI Smoke Test Failures in this PR looks similar to the issue that we were discussing on Wednesday evening (with the nightly release job). Doesn't it?

sarangan12 (changes requested):
1. Code Changes looks fine. 
2. Please provide an example in the description as what is the expected value ad what is the actual value we are getting before the code change. That will help the reviewers understand the problem that is resolved by this PR.
3. Thanks for adding the test case. Can you point to the location (in the generated code for the test case) , your code change is impacting?
4. CI is failing for all the 3 smoke tests. I am getting a feeling this has nothing to do with your code changes (To be on the safer side, you can run the smoke test once again locally and confirm). @joheredi The CI Smoke Test Failures in this PR looks similar to the issue that we were discussing on Wednesday evening (with the nightly release job). Doesn't it?

sarangan12:
Ok. After offline discussion with Deya, I found the example and description in the linked issue. Thanks for that. Approving the PR now. Will wait for input from Jose too regarding the CI failures.

sarangan12 (approved):

PR #1016, opened 2021-06-07
Remove unused express dependency

joheredi:
We have an un needed dependency on express that is causing issues when building after dependabot updating our dependencies.

Removing the dependency on express seems to fix the iussue


deyaaeldeen (approved):

sarangan12 (approved):
Great Job

Issue #1017, opened 2021-06-07
Add note to readme pointing people interested in v4 to the right place


ramya-rao-a:
This repo has been the house for the v4 code generator which has a different package name than our v6 code generator which is now in the master branch.

For those customers landing here looking for v4, we should add a pointer to the readme of the v4 branch.
A small description of the difference would be useful to add in the v6 readme.
Also a disclaimer that v6 is still in preview

And we should do vice versa i.e. add pointer to v6 from v4 for anyone interested in checking out our new work.

sarangan12:
Code complete with PR https://github.com/Azure/autorest.typescript/pull/1107

PR #1018, opened 2021-06-08
Update Version and Changelog for Beta 2 Release

sarangan12:
This PR is to update the version and changelog for the beta 2 release of the SDK Generator. 

@deyaaeldeen @joheredi Please review and approve the PR.

@ramya-rao-a FYI...


ramya-rao-a:
Can we also update the readme as per #1017 in this PR?

README.md
```diff
@@ -1,3 +1,11 @@
+**Note**:
+
+This branch (main) is for the latest V6 version of the TS/JS SDK generator - [@autorest/typescript](https://www.npmjs.com/package/@autorest/typescript).
```
deyaaeldeen, line 3: We probably should not call it V6 anymore. cc @joheredi

deyaaeldeen: I think we should refer to them now using their names on npm, `@autorest/typescript` and `@microsoft.azure/autorest.typescript`

deyaaeldeen: @ramya-rao-a what do you think?

ramya-rao-a: What is the approach used by the code generators in other languages? Usually in other feature crews, we start with the last major version + 1 even if the package name is different.

deyaaeldeen: @ramya-rao-a to give you more context, we as the crew, already committed to have v6-branded version released as v1.0.0-beta.1. I know Python used the approach of last major version + 1. @sarangan12 did some research on what the other languages did but I do not remember the findings tbh.

Now, since we already committed to v1 release, I feel like it is very confusing to still call it v6. I think the following naming proposal might work better than v6:
- `@autorest/typescript` vs `@microsoft.azure/autorest.typescript`
- New code gen vs legacy code gen
- v1 vs v4 (least favorite because it is still confusing IMO)

sarangan12: 1. Removed the reference to V6 and V4 Versions. 
2. As for the version number, we have discussed this earlier with autorest crew and decided to start with 1.0.0-beta.x version for the SDK Generator.

joheredi (approved):

Issue #883, opened 2021-03-09
Autorest typecript generator having issues ignoring type checking in node_modules 


andreskim:
**Describe the bug**
I'm in the process of utilizing autorest to generate a client for one of our REST microservices. Whille attempting to use autorest to generate a service client, (specifically for typescript)  I ran into the following issue. 

After generating using Autorest with typescript flag: 
```
  - npm run build fails  due to type checking in node_modules
  - In order to manually fix this, we have to add a "skipLibCheck": true to the tsconfig compiler options. 
```

The issue arises since our workflow depends on running a script to use autorest to generate a client on a given instance. The current workaround would involve a convoluted script that uses vim to edit the tsconfig after generation. 

**Potential Solutions**
```
1. Add the skipLibCheck in the tsconfig.json on default
2. Or, Allow users to pass in a config for the tsconfig

```

sarangan12:
@andreskim Thanks for considering autorest for generating SDK for your service. We could not add `skipLibCheck: true` by default to all the generated libraries. But, I realize that there might be cases (such as yours) where you know you would want to pass such a check. I think I could add a flag to achieve this. I will work on that and update this issue soon. Thanks.

sarangan12:
@andreskim After discussion with my team, I believe it is not required to add this flag in the generator. Because, each project might have unique requirements around tsconfig flags and it would be too cumbersome to add flags for each one in the generator. (You can find the discussion in this issue: https://github.com/Azure/autorest.typescript/pull/1022)

My suggestion would be to:

1. Generate your library using autorest.
2. Then, you could modify the tsconfig manually for the first time. 
3. For the followup cases, you could set  use `--generate-metadata=false` when you execute autorest command. This will generate only the src files and will not override your tsconfig files. 

Please let me know if this resolves your issue or if you have any more questions or issues.

sarangan12:
No more action items pending on this issue. Closing this issue. Feel free to reopen if you encounter the same issue again

PR #1022, opened 2021-06-09
Code to add skipLibCheck to tsconfig json

sarangan12:
Before  a few days, We encountered a CI failure in our SDK Generator repository that was caused by a break in one of our underlying dependency packages (express). We were able to remove that dependency (since it was not used anyway) and resolve the CI failures. If we could not have removed it, then we might be forced to add skipLibCheck to our tsconfig json files.

Accidentally, a similar request has come from a customer at https://github.com/Azure/autorest.typescript/issues/883. So, I have added a flag `skip-lib-check` to check and add the flag to the generated tsconfig.json file.

@joheredi @deyaaeldeen Please review and approve.

@ramya-rao-a FYI...

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ In addition to the [list of Autorest flags](https://github.com/Azure/autorest/bl
 | `--tracing-info`                | Controls specification of meta info attached to requests for tracing purposes                                                                                                                                                                                                                                                        |
 | `--disable-async-iterators`     | Does not generate async iterators needed for paging operations                                                                                                                                                                                                                                                                       |
 | `--allow-insecure-connection`   | Allow generated clients to make requests to HTTP endpoints                                                                                                                                                                                                                                                                           |
+| `--skip-lib-check`              | Sets `skipLibCheck` to `true` in the `tsconfig.json` file of the generated SDK.                                                                                                                                                                                                                                                      |
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -35,7 +35,8 @@ In addition to the [list of Autorest flags](https://github.com/Azure/autorest/bl
 | `--tracing-info`                | Controls specification of meta info attached to requests for tracing purposes                                                                                                                                                                                                                                                        |
 | `--disable-async-iterators`     | Does not generate async iterators needed for paging operations                                                                                                                                                                                                                                                                       |
 | `--allow-insecure-connection`   | Allow generated clients to make requests to HTTP endpoints                                                                                                                                                                                                                                                                           |
-| `--skip-lib-check`              | Sets `skipLibCheck` to `true` in the `tsconfig.json` file of the generated SDK.                                                                                                                                                                                                                                                      |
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -35,6 +35,7 @@ In addition to the [list of Autorest flags](https://github.com/Azure/autorest/bl
...
 | `--disable-async-iterators`     | Does not generate async iterators needed for paging operations                                                                                                                                                                                                                                                                       |
 | `--allow-insecure-connection`   | Allow generated clients to make requests to HTTP endpoints                                                                                                                                                                                                                                                                           |
+| `--skip-lib-check`              | Sets `skipLibCheck` to `true` in the `tsconfig.json` file of the generated SDK.                                                                                                                                                                                                                                                      |
```
deyaaeldeen, line 38: NIT: all other sentences do not end with a period
```suggestion
| `--skip-lib-check`              | Sets `skipLibCheck` to `true` in the `tsconfig.json` file of the generated SDK                                                                                                                                                                                                                                                      |

```

src/generators/static/tsConfigFileGenerator.ts
```diff
@@ -46,16 +46,23 @@ const restLevelTsConfig = {
...
+    restLevelClient,
+    skipLibCheck
+  } = getAutorestOptions();
 
   if (!generateMetadata) {
     return;
   }
 
-  const tsConfigContents = restLevelClient
+  const tsConfigContents: any = restLevelClient
```
deyaaeldeen, line 59: I would rather not do this and instead define an interface for things we support in tsconfig.

deyaaeldeen (commented):
This PR adds a customer-requested TS config to tsconfig.json, behind a flag. I am not against this change but here is one thought: Are we going to add a new autorest flag every time a customer asks for an additional TS config? Instead, did you consider if we can make our generated tsconfig.json file extend a customer-provided one?

I left a few comments.

joheredi:
Hi Sarangan, do we know what specific error this user was hitting? If it was the same due to the express dependency it should be fixed.

There is currently a flag `--generate-metadata=false` that would skip generating tsconfig.json. I'm a little worried about adding this flag as it seems to start getting too specific. I think there must be a line in which users are better off setting `--generate-metadata=false` and build their own tsconfig, package, etc.

For example if we add this flag, we may get a request in the future asking for a flag to change the target in tsconfig, and so on.

My proposal is that we generate a set of static files that work out of the box, and any additional stuff users may want they need to opt-out of the generated metadata and keep building on top of the generated ones.

There are 2 scenarios:

1. Users want to create their metadata files from scratch, so they set `--generate-metadata=false` from the first time they generate
2. Users start using the generated metadata files but realize they need special settings, at that point they opt-out of metadata generation and keep building on top of what we generated. Future codegen runs won't override these files since the flag `--generate-metadata=false` is set.

What do you think?

sarangan12:
@joheredi @deyaaeldeen I agree that this flag is more specific. I was Ok even with that point only because the error that happens is on external library and does not have anything to do with the code itself. My point was that the customer need not suffer because of some external dependency. 

But, I think it is a valid point to say we cannot keep on adding flags. I will suggest to modify the metadata setting and generate it manually. 

Will close this PR for now. Will reopen if it is really required after discussion with the customer.

Issue #1021, opened 2021-06-09
Dropping support for Node.js versions that have reached end of life


ramya-rao-a:
Our [support policy](https://github.com/Azure/azure-sdk-for-js/blob/master/SUPPORT.md#microsoft-support-policy) does not guarantee support for Node.js versions that have reached end of life.

We have the below generated in the readme when using the v4 code generator

```
### Currently supported environments

- Node.js version 6.x.x or higher
- Browser JavaScript
```

This should be replaced with

```
### Currently supported environments

- [LTS versions of Node.js](https://nodejs.org/about/releases/)
- Latest versions of Safari, Chrome, Edge and Firefox.
```

In the v6 code generator, in addition to the above, the `engines` field in the generated package.json file should use `>=12.0.0` instead of `>=8.0.0`

ramya-rao-a:
@deyaaeldeen Can you update https://github.com/deyaaeldeen/azure-sdk-for-js/blob/master/common/tools/add-identity-mgmt-lib.js as well?

deyaaeldeen:
@ramya-rao-a done: https://github.com/Azure/azure-sdk-for-js/pull/15648.

deyaaeldeen:
code complete

PR #1024, opened 2021-06-09
Update supported environments in readme

deyaaeldeen:
Updates the list of supported environments as per https://github.com/Azure/autorest.typescript/issues/1021 for the legacy code generator.


sarangan12 (approved):
I think CI is failing because of the `regenerate` command. Could you please check?

Issue #1005, opened 2021-06-02
[Track1] User Agent Can be Override


dw511214992:
I do a test to try to override the Use-Agent in Track1 JS SDK, and it succeed. The Following is the details:
Codes:
```
const msRestNodeAuth = require("@azure/ms-rest-nodeauth");
const { ComputeManagementClient } = require("@azure/arm-compute");

msRestNodeAuth.interactiveLogin().then((creds: any) => {
    const client = new ComputeManagementClient(creds, "eece696d-f03b-47a5-9e4c-157d7ad85b20", {
        userAgent: "aaaaaaaaaaaaaaaaaaaaaaaa"
    });
    client.operations.list().then((result: any) => {
        console.log("The result is:");
        console.log(result);
    });
}).catch((err: any) => {
    console.error(err);
});
```
Requests Header captured by [fiddler](https://www.telerik.com/fiddler):

![image](https://user-images.githubusercontent.com/40835867/120439204-d6c9a180-c3b4-11eb-97c3-690301c1c0cb.png)

I think the user-agent should only be extended, instead of overrode. Please check it.

ramya-rao-a:
Thanks for finding this @dw511214992 

@sarangan12, Can you please check with the auto rest crew as to what the behavior is in other languages (both Track 1 & Track 2)?  

Our new SDK guidelines does say that we should only be allowing adding to the default user agent string and not replacing it. 
I guess, we have been using the convenience layer on top of the generated code to ensure we append the user provided value to the default user agent string and then pass it to the generated layer.

@dw511214992, 
Have you found data to suggest that there is a significant of usage data where the user agent string has been replaced?

sarangan12:
@dw511214992 

Thanks a lot for reporting this issue. I will work on fixing the issue and update you.

@ramya-rao-a 

Yes. I will sync up with the autorest crew and get their behavior also and document it here.

dw511214992:
Hi @ramya-rao-a The [telemetry](https://msit.powerbi.com/groups/de995b1a-16c1-4cc7-a49b-bc9f90bc6acd/reports/8a8582d5-8fdf-4014-8cf9-4d183c5db2c8/ReportSection) count JS SDK's request based on User Agent, and logic is:
```
// azure-sdk-for-node
userAgent.Contains("azure-sdk-for-node") or userAgent.Contains("azure-sdk-for-node.js")
// azure-sdk-for-js
userAgent.Contains("ms-rest-js")
```
[...]

PR #1033, opened 2021-06-11
Modify User Agent Strings in Core V1 & V2

sarangan12:
This PR is related to Issue https://github.com/Azure/autorest.typescript/issues/1005. Although, the issue is reported only for Track 1,  it happens both in Track 1 & Track 2. So, this PR fixes the issue in Track 2. (I will send out a separate one for Track 1)

This PR brings the user agent strings in sync with the guidelines.

In Track 2 + Core V1 (core-http), if the user specifies **no** custom user agent string, the automatically generated user agent string will be similar to `core-http/1.2.4 Node/v14.15.4 OS/(x64-Windows_NT-10.0.19043)`. With this PR, it has been modified to `useragent-corev1/1.0.0-preview1 core-http/1.2.4 Node/v14.15.4 OS/(x64-Windows_NT-10.0.19043)` (The `useragent-corev1` is the name of the package)

In Track 2 + Core V1 (core-http), if the user specifies a custom user agent string, the automatically generated user agent string will be similar to `CustomUserAgentPrefix core-http/1.2.4 Node/v14.15.4 OS/(x64-Windows_NT-10.0.19043)`. With this PR, it has been modified to `CustomUserAgentPrefix useragent-corev1/1.0.0-preview1 core-http/1.2.4 Node/v14.15.4 OS/(x64-Windows_NT-10.0.19043)`. 

In Track 2 + Core V2 (core-client+core-rest-pipeline), if the user specifies **no** custom user agent string, the automatically generated user agent string will be similar to `core-rest-pipeline/1.0.3 Node/v14.15.4 OS/(x64-Windows_NT-10.0.19043)`. With this PR, it has been modified to `useragent-corev2/1.0.0-preview1 core-rest-pipeline/1.0.3 Node/v14.15.4 OS/(x64-Windows_NT-10.0.19043)`.

In Track 2 + Core V2 (core-client+core-rest-pipeline), if the user specifies a custom user agent string, the automatically generated user agent string will be similar to `CustomUserAgentPrefix`. With this PR, it has been modified to `CustomUserAgentPrefix useragent-corev2/1.0.0-preview1 core-rest-pipeline/1.0.3 Node/v14.15.4 OS/(x64-Windows_NT-10.0.19043)`. 

In have included integration test case (`userAgents.spec.ts`) that covers all the above 4 cases. 

@joheredi @deyaaeldeen Please review and approve the PR.

@ramya-rao-a FYI....


sarangan12:
I am working on getting the smoke tests work correctly in my local machine. Once, I resolve the issue, I will update the PR with smoke test cases. Thanks

src/generators/clientContextFileGenerator.ts
```diff
@@ -250,14 +250,14 @@ function writeDefaultOptions(
...
+  }
 
-  if (!options.userAgent) {
-     const defaultUserAgent = coreHttp.getDefaultUserAgentValue();
-     options.userAgent = \`\${packageName}/\${packageVersion} \${defaultUserAgent}\`;
-   }
+  const defaultUserAgent = coreHttp.getDefaultUserAgentValue();
+  options.userAgent = options.userAgent
+  ? \`\${options.userAgent} \${packageName}/\${packageVersion} \${defaultUserAgent}\`
+  : \`\${packageName}/\${packageVersion} \${defaultUserAgent}\`;
```
deyaaeldeen, line 259: NIT: You can build this bit once: `\${packageName}/\${packageVersion} \${defaultUserAgent}\` before the check.

src/generators/clientContextFileGenerator.ts
```diff
@@ -269,6 +269,17 @@ function writeDefaultOptions(
     options = {};
   }
   ${defaults}
+
+  const userAgentPrefix =
+      options.userAgentOptions && options.userAgentOptions.userAgentPrefix
+        ? \`\${options.userAgentOptions.userAgentPrefix} ${
+          packageDetails.name
+        }/${packageDetails.version}\`
+        : \`${packageDetails.name}/${packageDetails.version}\`;
```
deyaaeldeen, line 278: same comment here about doing the concat work once.

test/integration/userAgents.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+      }
+    });
+    const result: GroupGetSampleResourceGroupResponseV2 = await client.group.getSampleResourceGroup(
+      "testgroup101",
+      {
+        onResponse: (response: any) => {
+          const array = response.request.headers
+            .get("user-agent")!
+            .match(
+              /CustomUserAgentPrefix useragent-corev2\/1\.0\.0-preview1 core-rest-pipeline\/.* Node\/v.* OS\/.*/g
```
deyaaeldeen, line 84: I see the word Node here, would this work for browser tests?

sarangan12: I have added it to the skip browser tests.

src/generators/clientContextFileGenerator.ts
sarangan12: Done

src/generators/clientContextFileGenerator.ts
sarangan12: Done

test/integration/generated/readmeFileChecker/src/keyVaultClientContext.ts
```diff
@@ -26,6 +26,15 @@ export class KeyVaultClientContext extends coreClient.ServiceClient {
       requestContentType: "application/json; charset=utf-8"
     };
 
+    const packageDetails = `@azure/keyvault-secrets/1.0.0-preview1`;
```
joheredi, line 29: This format doesn't seem to be consistent with existing Track2 packages. For example `@azure/search-documents` has as prefix `azsdk-js-search-documents/${SDK_VERSION}` ( [link](https://github.com/Azure/azure-sdk-for-js/blob/a760eadea01c1c11960ad9071f704340fd867ea5/sdk/search/search-documents/src/searchClient.ts#L119))

Other packages follow the same as search, we may need to adjust to that. Also, it would be a good idea to update the guidelines with the format we want for user-agent, I didn't find anything specific about that ([link](https://azure.github.io/azure-sdk/typescript_implementation.html#general-distributed-tracing))

ramya-rao-a:
So... to summarize, what we are saying is that in Track 2, user provided custom string has always been being prefixed to the user agent string properly. What was missing is adding the package name and version details as prefix before adding the user provided custom string as prefix. Is that right?

test/integration/generated/readmeFileChecker/src/keyVaultClientContext.ts
ramya-rao-a: Here is the guideline on user agent strings: https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy

We need to have `azsdk-js-` followed by the package name without the `@azure/`

sarangan12:
@ramya-rao-a 

1. For Track 2 + Core V1 - The answer is yes. The prepending of Cutsom User agent string has always been happening. Only change is to add `azsdk-js-<package-name>`.
2. For Track 2 + Core V2 - The answer is no. When the user provides a custom user agent string, it will override the entire string and that is a bug. So, this PR fixes that also.

test/integration/generated/readmeFileChecker/src/keyVaultClientContext.ts
sarangan12: Thanks @joheredi. I will modify the user agent string to `azsdk-js-<package_name>` in the next commit

ramya-rao-a:
> For Track 2 + Core V1 - The answer is yes. The prepending of Cutsom User agent string has always been happening. Only change is to add azsdk-js-<package-name>

After this change, current package owners will have to remove the part where they form the string `azsdk-js-package-name/SDK_VERSION` right? We should call that out in the changelog

sarangan12:
@ramya-rao-a Agreed. I have updated the changelog (in this PR)

joheredi (approved):

Issue #1034, opened 2021-06-11
LROs should respect the retry-after header


deyaaeldeen:
When the service's response contains a retry-after header, the poller should use its time value to be the new polling interval.

PR #1036, opened 2021-06-14
update LRO's intervalInMs from the Retry-After header

deyaaeldeen:
Fixes https://github.com/Azure/autorest.typescript/issues/1034.

The algorithm is simple, if the service's response has `Retry-After` header, use that instead of the customer-provided/default polling interval.

TODO: update core-lro if necessary to make sure it always use the updated intervalInMs.


sarangan12 (approved):

src/lro/stateMachine.ts
```diff
@@ -133,12 +139,36 @@ export function createGetLROState<TResult>(
...
+  return async (
+    path: string,
+    pollerConfig: PollerConfig
+  ): Promise<LROState<TResult>> => {
     await poll(path);
+    if (retryAfter !== undefined) {
+      const retryAfterInMs = parseInt(retryAfter);
+      pollerConfig.intervalInMs = isNaN(retryAfterInMs)
+        ? calculatePollingIntervalFromDate(
+            new Date(retryAfter),
```
joheredi, line 151: Just double checking if you are passing `retryAfter` instead of `retryAfterInMs` on purpose.

joheredi: If we don't get neither pollingInterval, nor Retry-After. Are we setting a default value?

deyaaeldeen: For your first question, yes this is intended because `retryAfterInMs` is unusable since it is `NaN` in this branch.

joheredi (approved):
Thanks!

PR #1051, opened 2021-06-22
Bump dependency versions of ModelerFour and AutorestCore

joheredi:
Bumping dependencies to the latest ModelerFour and Autorest Core.

In order to take the latest version of M4 I had to implement support for AnyObject schema. The resulting type is a `Record<string, unknown>`


src/transforms/mapperTransforms.ts
```diff
@@ -397,6 +399,14 @@ function transformDictionaryMapper(pipelineValue: PipelineValue) {
   };
 }
 
+function isDictionarySchema(schema: Schema): schema is DictionarySchema {
```
sarangan12, line 402: nit: I will not write a seperate function for such a small code. But, it is just personal choice.

sarangan12 (approved):

src/transforms/mapperTransforms.ts
joheredi: I like these small functions because they help narrowing down the type. The typescript server in the editor knows if the condition is met and gives you the type without you having to do a manual cast

Issue #1049, opened 2021-06-22
[Track 2] Add sdk-type in package.json


dw511214992:
This feature request comes from https://github.com/Azure/azure-sdk-for-js/commit/7026a578a61bc4d9332131c9ff47c2f84ed39b05#r52479640

As the `sdk-type` already exsits in dataplane sdk, can we also add it in mgmt plane sdk?

cc: @ramya-rao-a @sarangan12

sarangan12:
Code complete with PR https://github.com/Azure/autorest.typescript/pull/1074

PR #1074, opened 2021-06-24
Add sdk-type in package.json

sarangan12:
This PR is to address the issue reported in https://github.com/Azure/autorest.typescript/issues/1049. This PR is a simple PR to add `sdk-type` to generated `package.json` file. 

@deyaaeldeen @joheredi Please review and approve.

@ramya-rao-a FYI....


deyaaeldeen (approved):
I am not sure why do we want to distinguish between `client` and `management` but that is a question the EngSys team. Otherwise, this looks good for now.

src/generators/static/packageFileGenerator.ts
```diff
@@ -174,7 +201,29 @@ function regularAutorestPackage(
         "tsc && rollup -c rollup.config.js && npm run minify && mkdirp ./review &&  npm run extract-api",
       minify: `uglifyjs -c -m --comments --source-map "content='./dist/index.js.map'" -o ./dist/index.min.js ./dist/index.js`,
       prepack: "npm install && npm run build",
-      "extract-api": "api-extractor run --local"
+      pack: "npm pack 2>&1",
+      "extract-api": "mkdirp ./review && api-extractor run --local",
+      lint: "echo skipped",
```
dw511214992, line 206: please wrap it with double quotes.

sarangan12: @dw511214992 It will automatically be wrapped in double quotes at the time of generation. Ref: https://github.com/Azure/autorest.typescript/blob/main/test/integration/generated/additionalProperties/package.json#L3

Issue #1066, opened 2021-06-24
[Track2] Generated codes by latest codegen compile failed


dw511214992:
generated codes: https://github.com/Azure/azure-sdk-for-js/tree/5e3b6d37d570ba71a421c28de58b60a639f3c34e/sdk/iotspaces/arm-iotspaces
command:
```
rush update
node common/scripts/install-run-rush.js build --from @azure/arm-iotspaces --verbose -p max
```
Compile Error:
```
Invoking: tsc && rollup -c rollup.config.js && npm run minify && mkdirp ./review &&  npm run extract-api

src/ioTSpacesClient.ts(1,27): error TS2307: Cannot find module '@azure/core-auth' or its corresponding type declarations.
src/ioTSpacesClientContext.ts(2,27): error TS2307: Cannot find module '@azure/core-auth' or its corresponding type declarations.
src/lro/operation.ts(4,33): error TS2307: Cannot find module '@azure/abort-controller' or its corresponding type declarations.
```

dw511214992:
cc @ramya-rao-a @sarangan12

sarangan12:
@dw511214992 I am investigating this issue. I will get back to you shortly

ramya-rao-a:
@sarangan12 In case you are updating the dev dependencies as part of this issue, also include updating `@types/node` to `^12.0.0` as we are moving away from Node.js 8 support. I am doing this change in the azure-sdk-for-js repo as par tof https://github.com/Azure/azure-sdk-for-js/pull/15996

sarangan12:
@dw511214992 There is an issue with packagefilegenerator code that does not include the core-auth package in the package.json file. I will get the fix in and include it on Tuesday's release. 

One thing that was surprising for me in your code. I am checking the link here: https://github.com/Azure/azure-sdk-for-js/tree/5e3b6d37d570ba71a421c28de58b60a639f3c34e/sdk/iotspaces/arm-iotspaces Here, there is no `lro` folder under `src` folder. (Which is correct and as expected)

But, your error message (in the first comment above), I see the following line:
```
src/lro/operation.ts(4,33): error TS2307: Cannot find module '@azure/abort-controller' or its corresponding type declarations.
```

Where is this line coming from? 

For your reference, I generated the package using the following command:

```
[...]

PR #1082, opened 2021-06-28
Fix the core-auth dependency

sarangan12:
This is to fix the issue reported in https://github.com/Azure/autorest.typescript/issues/1066.

Minor fix in the logic to add core-auth in the package json file. 

@joheredi @deyaaeldeen @dw511214992 Please review and approve.

@ramya-rao-a FYI.....


dw511214992 (approved):

src/generators/static/packageFileGenerator.ts
```diff
@@ -88,7 +88,7 @@ function restLevelPackage(packageDetails: PackageDetails) {
     },
     dependencies: {
       "@azure-rest/core-client": "1.0.0-beta.4",
-      "@azure/core-auth": "^1.1.4",
+      "@azure/core-auth": "^1.3.0",
       "@azure/core-rest-pipeline": "^1.0.4"
```
maorleger, line 92: `@azure/core-rest-pipeline` should be updated to "^1.1.0" as well - can we do it as part of this PR or a separate PR?

sarangan12: I can combine it with this PR. Will update it shortly

test/integration/generated/iotspaces/package.json
```diff
@@ -0,0 +1,84 @@
...
+  "name": "iotspaces",
+  "sdk-type": "management",
+  "author": "Microsoft Corporation",
+  "description": "A generated SDK for IoTSpacesClient.",
+  "version": "1.0.0-preview1",
+  "engines": { "node": ">=12.0.0" },
+  "dependencies": {
+    "@azure/core-client": "^1.1.2",
+    "@azure/core-auth": "^1.3.0",
+    "@azure/core-rest-pipeline": "1.0.0-beta.2",
```
deyaaeldeen, line 11: why are generating a beta version here?

test/integration/generated/iotspaces/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "compilerOptions": {
+    "module": "es6",
+    "moduleResolution": "node",
+    "strict": true,
+    "target": "es6",
```
deyaaeldeen, line 6: No need to be fixed in this PR but I think we can target ES2017 everywhere now to be consistent with Track 2 packages in the mono repo.

test/utils/test-swagger-gen.ts
```diff
@@ -702,6 +702,15 @@ const testSwaggers: { [name: string]: SwaggerConfig } = {
     allowInsecureConnection: true,
     addCredentials: false
   },
+  iotspaces: {
+    swaggerOrConfig: "test/integration/swaggers/iotspaces.json",
```
deyaaeldeen, line 706: what is interesting about this swagger to be added in our tests?

src/generators/static/packageFileGenerator.ts
ramya-rao-a: 1.1.0 of core-rest-pipeline is not published yet...

sarangan12: Spoke offline with Maor. Will keep core-rest-pipeline to 1.0.4 all through for this week.

test/integration/generated/iotspaces/package.json
sarangan12: I think this is a left over. Changed it now

test/integration/generated/iotspaces/tsconfig.json
sarangan12: Will do it in a seperate PR

PR #1110, opened 2021-07-19
Simplify lroImpl

deyaaeldeen:
Simplify `lroImpl` class. Apparently, core v2 does the right thing and there is no need to do custom deserialization.


sarangan12 (approved):
I would highly recommend creating a detailed document explaining:

1. What is LRO? 
2. How was LRO handled in v4x?
3. What was the drawbacks?
4. How was LRO handled in V6 - Core V1?
5. How is LRO handled in V6 - Core V2?

with examples.

src/lroImpl.ts
```diff
@@ -1,136 +1,18 @@
...
-    operationLocation: result.headers.get("operation-location"),
-    location: result.headers.get("location"),
-    requestMethod: result.request.method
-  };
-}
-
 export class LroImpl<T> implements LongRunningOperation<T> {
   constructor(
-    private sendOperationFn: SendOperationFn<T>,
+    private sendOperationFn: (args: any, spec: any) => Promise<LroResponse<T>>,
```
joheredi, line 5: Is there a more specific type we can use instead of any?

joheredi (approved):
I love this change! Thanks for the hard work improving our LRO story!

src/lroImpl.ts
deyaaeldeen: Good point, I am addressing this in https://github.com/Azure/autorest.typescript/pull/1118.

deyaaeldeen:
I also tested locally against the old testsuite that was using core v1 and it passes all the tests.

PR #1119, opened 2021-07-23
[RLC] Add test coverage for body string

joheredi:
Adding test coverage for body string and making the following changes to the generator:

1. Generate index.ts
2. generate byteArray schemas as strings and add documentation to tell customers that the value may contain a base64 encoded string. We are not making it a Uint8Array because there is no good way to deserialize these strings without using some kind of mappers. We have decided to keep runtime thin and hand off that responsibility to the user, the SDK will return the value that the service returns, plus extra documentation to hint the user about the possible contents.

### DIFF
--- a/src/generators/indexGenerator.ts
+++ b/src/generators/indexGenerator.ts
@@ -1,21 +1,67 @@
-import { Project } from "ts-morph";
+import { Project, SourceFile } from "ts-morph";
 import { ClientDetails } from "../models/clientDetails";
-import { getAutorestOptions } from "../autorestSession";
+import { getAutorestOptions, getSession } from "../autorestSession";
+import { NameType, normalizeName } from "../utils/nameUtils";
 
 export function generateIndexFile(
-  clientDetails: ClientDetails,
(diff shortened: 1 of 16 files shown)


joheredi:
Fixes #996

deyaaeldeen (approved):
Looks good!

src/generators/indexGenerator.ts
```diff
@@ -1,21 +1,67 @@
...
+  project: Project,
+  clientDetails?: ClientDetails
 ) {
-  const { srcPath, disablePagingAsyncIterators } = getAutorestOptions();
+  const { restLevelClient, srcPath } = getAutorestOptions();
   const indexFile = project.createSourceFile(`${srcPath}/index.ts`, undefined, {
     overwrite: true
   });
 
+  if (!restLevelClient) {
```
sarangan12, line 15: nit:   if (!restLevelClient && !clientDetails) {

sarangan12 (approved):
Looks good

PR #1122, opened 2021-07-27
Issue with npm install scripts in test libraries

sarangan12:
In our SDK repository, we have several test libraries created for integration & smoke test scenarios. In these test libraries, we have the `README.md` in which we have installation instructions with `npm install .....` command. But, all these libraries are test libraries and will not be released to npm.  

Now, this causes a security vulnerability. An attacker can create an actual npm library with that name and if someone follows the instruction from our repository and installed that library, it could potentially contain malicious code and could be used for hacking. This scenario has already happened and security team wanted to remove that instruction from our test libraries. This PR is created to remove that instruction from test libraries. 

@deyaaeldeen @joheredi Please review and approve.

@ramya-rao-a FYI.....

### DIFF
--- a/src/generators/static/readmeFileGenerator.ts
+++ b/src/generators/static/readmeFileGenerator.ts
@@ -47,8 +47,8 @@ interface Metadata {
   addCredentials?: boolean;
   /** The link to the identity package in the repository */
   identityPackageURL?: string;
-  /** Indicates if the package is a test/Package package. */
-  isRealPackage?: boolean;
+  /** Indicates if the package is a test/releasable package. */
+  isReleasablePackage?: boolean;
 }
 
 /**
@@ -124,7 +124,7 @@ function createMetadata(
     projectName: azureHuh ? "Microsoft Azure SDK for JavaScript" : undefined,
     addCredentials,
     identityPackageURL,
-    isRealPackage: !isTestPackage
+    isReleasablePackage: !isTestPackage
   };
 }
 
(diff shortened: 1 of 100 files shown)


src/generators/static/readmeFileGenerator.ts
```diff
@@ -47,6 +47,8 @@ interface Metadata {
   addCredentials?: boolean;
   /** The link to the identity package in the repository */
   identityPackageURL?: string;
+  /** Indicates if the package is a test/Package package. */
+  isRealPackage?: boolean;
```
deyaaeldeen, line 51: I think this flag could use a better name, e.g. `isReleasedPackage`

src/utils/autorestOptions.ts
```diff
@@ -61,6 +63,11 @@ async function getHeadAsBoolean(host: Host): Promise<boolean> {
   return Boolean(headAsBoolean);
 }
 
+async function getIsTestPackage(host: Host): Promise<boolean> {
+  const isTestPackage: boolean = await host.GetValue("is-test-package");
```
deyaaeldeen, line 67: same comment about the name, `is-released-package`

deyaaeldeen (commented):
I agree this could be a nasty issue. I left a comment about the name for the flag.

src/generators/static/readmeFileGenerator.ts
sarangan12: I am going to change it to `isReleasablePackage`.

src/utils/autorestOptions.ts
sarangan12: Here the is-test-package looks fine

PR #1145, opened 2021-08-06
[RLC] Generate missing header interfaces

joheredi:
Generator builds the Header interfaces but never actually generates them


qiaozha (approved):
LGTM

Issue #1138, opened 2021-08-03
readme.md should add a link to how to use app registration docs.


qiaozha:
In the readme.md there's a section about how to use app registration in @azure/identity. 
Customer suggeests us to put a docs link on how to set up app registration https://github.com/Azure/azure-sdk-for-js/issues/16702#issuecomment-891474532

JonathanCrd:
Fixed in https://github.com/Azure/autorest.typescript/pull/1146

PR #1146, opened 2021-08-06
Linking guide on how to create an Azure AD application in the Readme

JonathanCrd:
Addressing issue https://github.com/Azure/autorest.typescript/issues/1138.

### DIFF
--- a/src/generators/static/README.md.hbs
+++ b/src/generators/static/README.md.hbs
@@ -58,6 +58,8 @@ npm install @azure/identity
 You will also need to register a new AAD application and grant access to {{ serviceName}} by assigning the suitable role to your service principal (note: roles such as `"Owner"` will not grant the necessary permissions).
 Set the values of the client ID, tenant ID, and client secret of the AAD application as environment variables: `AZURE_CLIENT_ID`, `AZURE_TENANT_ID`, `AZURE_CLIENT_SECRET`.
 
+For more information about how to create an Azure AD Application check out [this guide](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal).
+
 ```javascript
 const { {{ clientClassName }} } = require("{{ clientPackageName }}");
 const { DefaultAzureCredential } = require("@azure/identity");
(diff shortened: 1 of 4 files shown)


JonathanCrd:
@qiaozha I added a line in the readme referring to the guide, please let me know if you want me to add something else.

Also, I was thinking about putting the following sentence in bold like this: 

> You will also need to **register a new AAD application and grant access to {{ serviceName}}** by assigning the suitable role to your service principal (note: roles such as `"Owner"` will not grant the necessary permissions) ...

What do you think?

qiaozha:
@JonathanCrd Thanks. I think the link part looks good and I like the idea of putting that sentence in bold.

qiaozha:
@JonathanCrd You probably need to update the generated folder in the integration test and smoke test ?

sarangan12 (approved):
Looks Good. Please close the original issue with the comment as Issue Fixed

src/generators/static/README.md.hbs
```diff
@@ -55,9 +55,11 @@ You can authenticate with Azure Active Directory using the [Azure Identity libra
 npm install @azure/identity
 ```
 
-You will also need to register a new AAD application and grant access to {{ serviceName}} by assigning the suitable role to your service principal (note: roles such as `"Owner"` will not grant the necessary permissions).
+You will also need to **register a new AAD application and grant access to {{ serviceName}}** by assigning the suitable role to your service principal (note: roles such as `"Owner"` will not grant the necessary permissions).
 Set the values of the client ID, tenant ID, and client secret of the AAD application as environment variables: `AZURE_CLIENT_ID`, `AZURE_TENANT_ID`, `AZURE_CLIENT_SECRET`.
 
+For more information about how to create an Azure AD Application check out [this guide](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal).
```
ramya-rao-a, line 61: Remove the locale i.e. `en-us` from the url here to allow people at different locations to fall back to their own locale

JonathanCrd: Good catch, I opened PR https://github.com/Azure/autorest.typescript/pull/1155 for this.

Issue #1120, opened 2021-07-26
Bug in Boolean parameters with false default value


deyaaeldeen:
Swagger: https://github.com/Azure/azure-rest-api-specs/blob/e3dbb827d99eaadf132075707373f6b4bda08fd9/specification/cognitiveservices/data-plane/TextAnalytics/stable/v3.1/TextAnalytics.json#L839

Expected: generate a parameter with default value: false
```ts
export const EntitiesTaskParameters: coreClient.CompositeMapper = {
  type: {
    name: "Composite",
    className: "EntitiesTaskParameters",
    modelProperties: {
      modelVersion: {
        defaultValue: "latest",
        serializedName: "model-version",
        type: {
          name: "String"
        }
      },
      loggingOptOut: {
        defaultValue: false,
        serializedName: "loggingOptOut",
        type: {
          name: "Boolean"
        }
      },
      stringIndexType: {
        serializedName: "stringIndexType",
        type: {
          name: "String"
        }
      }
    }
  }
};
```
Actual: generate a parameter with no default values
```ts
export const EntitiesTaskParameters: coreClient.CompositeMapper = {
  type: {
    name: "Composite",
    className: "EntitiesTaskParameters",
    modelProperties: {
      modelVersion: {
        defaultValue: "latest",
        serializedName: "model-version",
        type: {
          name: "String"
        }
      },
      loggingOptOut: {
        serializedName: "loggingOptOut",
        type: {
          name: "Boolean"
        }
      },
      stringIndexType: {
        serializedName: "stringIndexType",
        type: {
          name: "String"
        }
      }
    }
  }
};
```

PR #1151, opened 2021-08-11
Fixing bug in Boolean parameters with false default value

JonathanCrd:
Fixing issue: https://github.com/Azure/autorest.typescript/issues/1120

### DIFF
--- a/src/transforms/mapperTransforms.ts
+++ b/src/transforms/mapperTransforms.ts
@@ -220,7 +220,7 @@ function buildMapper(
   return {
     ...{ type: mappeType },
     ...(serializedName && { serializedName }),
-    ...(!!schema.defaultValue && {
+    ...(schema.defaultValue !== undefined && {
       defaultValue: schema.defaultValue
     }),
     ...(required && { required }),
--- a/test/integration/generated/textanalytics/src/models/parameters.ts
+++ b/test/integration/generated/textanalytics/src/models/parameters.ts
@@ -97,6 +97,7 @@ export const top: OperationQueryParameter = {
 export const skip: OperationQueryParameter = {
   parameterPath: ["options", "skip"],
   mapper: {
+    defaultValue: 0,
     constraints: {
       InclusiveMinimum: 0
     },
--- a/test/smoke/generated/compute-resource-manager/src/models/mappers.ts
+++ b/test/smoke/generated/compute-resource-manager/src/models/mappers.ts
@@ -4287,6 +4287,7 @@ export const RunCommandParameterDefinition: coreClient.CompositeMapper = {
         }
       },
       required: {
+        defaultValue: false,
         serializedName: "required",
         type: {
           name: "Boolean"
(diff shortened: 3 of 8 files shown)


test/utils/test-swagger-gen.ts
```diff
@@ -819,6 +819,16 @@ const testSwaggers: { [name: string]: SwaggerConfig } = {
     restLevelClient: true,
     licenseHeader: true,
     isTestPackage: true
+  },
+  textAnalytics: {
```
deyaaeldeen, line 823: let's remove the entry at line 698 because it is a duplicate and gets overwritten by the new test you add here.

test/integration/generated/textanalytics/src/models/mappers.ts
```diff
@@ -124,6 +209,13 @@ export const EntitiesTaskParameters: coreClient.CompositeMapper = {
           name: "String"
         }
       },
+      loggingOptOut: {
+        defaultValue: false,
```
deyaaeldeen, line 213: nice! thanks a lot!

sarangan12 (dismissed):
Please rename the textanlytics test case as it overrides the existing test case

test/utils/test-swagger-gen.ts
JonathanCrd: I renamed it, thanks for pointing this out :)

deyaaeldeen: I think you should delete the old entry all together, there is no need to generate the same swagger twice in our tests. Technically the two swaggers are not exactly the same but the one you're adding have everything in the old one and more so let's stick with that instead.

JonathanCrd: That makes sense, @sarangan12 what are your thoughts on this?

joheredi (approved):
CI is failing, it looks like you need to re-generate the tests and commit the changes. `npm run generate-swaggers`.

**Up for discussion**: Since there are many smokeTests and Test-Swaggers that are changing with this fix. Do we still want to add the whole new swagger to the test suite? I think the swaggers we already generate would provide enough coverage to make it evident if we have a regression. Every new swagger we add to the test suite slows down the CI and our local development workflow, so we should be strategic when adding new load to the test suite.

What do you guys think? @sarangan12, @JonathanCrd , @deyaaeldeen 

I don't think this is a blocker, however, I wanted to bring up this conversation

deyaaeldeen:
@joheredi I agree! For this PR, I think the easiest thing @JonathanCrd can do is to add the new TA swagger and remove the old one if the boolean with the false default value does not show up in any of our existing swaggers (I did not check). If it does, then he does not need to add any new swaggers. Also, I think we should create an issue to trim down the existing integration-tests swaggers and rely on smoke tests for full ones.

test/utils/test-swagger-gen.ts
sarangan12: @JonathanCrd Please check if the old entry has this boolean parameter. If so, You can delete it.

sarangan12:
@joheredi Yes. If the old swagger has an entry for the boolean value, we do not want a new test case at all.

JonathanCrd:
> @joheredi Jose Manuel Heredia Hidalgo FTE I agree! For this PR, I think the easiest thing @JonathanCrd Jonathan Cardenas FTE can do is to add the new TA swagger and remove the old one if the boolean with the false default value does not show up in any of our existing swaggers (I did not check). If it does, then he does not need to add any new swaggers. Also, I think we should create an issue to trim down the existing integration-tests swaggers and rely on smoke tests for full ones.

@deyaaeldeen I just checked and there are other swaggers that had the boolean default value = false, so I'm removing the new test that I added

Issue #1159, opened 2021-08-17
Generated Code fails to build on Container Service management plane. 


qiaozha:
We found the current v6 code generator fails to generate container service management plane package.
To repo: 
1. use the latest v6 code generator 6.0.0-alpha.8.20210816.1 and checkout the rest api specs main branch. 
2. use this readme.md https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerservice/resource-manager/readme.md 

This is because there's a numeric enum in container service's swagger 
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerservice/resource-manager/Microsoft.ContainerService/stable/2021-07-01/managedClusters.json#L2272-L2286 

And the generated code looks like this 

`export type Count = One | Three | Five;`

But it should be 

`export type Count =1 | 3 | 5;`

I guess it has something to do with x-ms-enum modelAsString: false

sarangan12:
Thanks @qiaozha for reporting this issue. As you have mentioned that this might be a GA Blocker Issue (in the email), I will take this as high priority and started working on this issue. Will update my progress on this issue soon.

sarangan12:
The Code Changes are completed and merged. The changes are available in https://www.npmjs.com/package/@autorest/typescript/v/6.0.0-alpha.8.20210817.4

sarangan12:
A new version https://www.npmjs.com/package/@autorest/typescript/v/6.0.0-beta.9 has been released with this fix. Please check and feel free to reopen this issue if you still face any more issue. Thanks

PR #1160, opened 2021-08-17
Fix Non String Sealed Choices Generated Code

sarangan12:
This PR is to fix the issue reported in https://github.com/Azure/autorest.typescript/issues/1159. The Swagger has a sealed choice object such as:

```
"count": {
    "type": "integer",
    "format": "int32",
    "enum": [ 1, 3, 5 ],
    "x-ms-enum": {
        "name": "Count",
        "modelAsString": false
    },
    "description": "Number of masters (VMs) in the container service cluster. Allowed values are 1, 3, & 5.The default value is 1.",
    "default": 1
}
```

The above swagger generates the following SDK code:

`export type Count = One | Three | Five;`

But it should be 

`export type Count = 1 | 3 | 5;`

This is happening because for Non-String Enum Values, the generator uses the `name` instead of `value`. This PR fixes it.

@joheredi Please approve it.

### DIFF
--- a/src/generators/modelsGenerator.ts
+++ b/src/generators/modelsGenerator.ts
@@ -537,7 +537,9 @@ const writeSealedChoice = (
   modelsIndexFile: SourceFile
 ) => {
   const values = choice.properties
-    .map(p => (choice.itemType === SchemaType.String ? `"${p.value}"` : p.name))
+    .map(p =>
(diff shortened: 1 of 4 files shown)


JonathanCrd (approved):
LGTM

joheredi (approved):
LGTM!

PR #1164, opened 2021-08-18
Fixed typing bug

### LINKED ISSUE #1163: Critical bug within the last 24 hours
We had a build that succeeded around 5 PM GMT + 1 yesterday, but now that very same build fails, after regenerating our Autorest client.

It looks like the `package.json` of the Autorest package generated has a `types` property, which points to a file `./types/client.d.ts` which isn't there.

That's causing issues in VS Code and at compile-time:

> Could not find a declaration file for module '@anonymized/client'. 'c:/web/node_modules/@anonymized/client/dist/index.js' implicitly has an 'any' type.
>   Try `npm i --save-dev @types/anonymized__client` if it exists or add a new declaration (.d.ts) file containing `declare module '@anonymized/client';`

Also, it looks like it's trying to look for typings in the `dist` folder, and there are none there. However, there are typings in the `dist-esm` folder, but that doesn't seem to be the path it's using (from the error).

ffMathy:
Fixes #1163

### DIFF
--- a/src/generators/static/packageFileGenerator.ts
+++ b/src/generators/static/packageFileGenerator.ts
@@ -54,7 +54,7 @@ function restLevelPackage(packageDetails: PackageDetails) {
     version: `${packageDetails.version}`,
     description: `${packageDetails.description}`,
     main: "dist-esm/index.js",
-    types: `./types/${packageDetails.nameWithoutScope}.d.ts`,
+    types: "./dist-esm/index.d.ts",
     scripts: {
       test: 'echo "Error: no test specified" && exit 1',
       build: "tsc --build && npm run extract-api",


sarangan12:
Thanks @ffMathy for this PR. But, I do not think this is the right fix for this issue. A fix has been merged and a new version https://www.npmjs.com/package/@autorest/typescript/v/6.0.0-beta.10 has been released. Please try that and if you still face the issue, please update the issue https://github.com/Azure/autorest.typescript/issues/1163.

Also, It is great you are contributing to the repository. Please refer the steps in the [Contribution](https://github.com/Azure/autorest.typescript/blob/main/CONTRIBUTING.md) file for detailed steps.

PR #1165, opened 2021-08-19
fix-missing-types-folder-in-package

qiaozha:
fix issue https://github.com/Azure/azure-sdk-for-js/issues/16932 in code generator side.


sarangan12 (approved):

sarangan12:
This might be the root cause for the issue reported in https://github.com/Azure/autorest.typescript/issues/1163.  After the merge, I will release a new version with this fix tomorrow.

qiaozha:
> This might be the root cause for the issue reported in #1163. After the merge, I will release a new version with this fix tomorrow.

I have no idea you already have this issue tracked.

PR #1168, opened 2021-08-20
[RLC] Support header request generation and.add test for header.json

qiaozha:


src/restLevelClient/generateParameterTypes.ts
```diff
@@ -38,6 +41,16 @@ export function generateParameterInterfaces(
     );
     const parameterInterfaceName = `${operationName}Parameters`;
     const parameters = getOperationParameters(operation);
+    // const requestProperties: PropertySignatureStructure[] = [];
```
sarangan12, line 44: nit: remove the comment

src/restLevelClient/generateParameterTypes.ts
```diff
@@ -38,6 +41,16 @@ export function generateParameterInterfaces(
...
+
+    const headerParameterDefinition =  buildHeaderParameterDefinition(
```
sarangan12, line 46: nit: headerParameterDefinition`s`

src/restLevelClient/generateParameterTypes.ts
```diff
@@ -82,6 +108,67 @@ export function generateParameterInterfaces(
...
+  }
+  const headersInterfaceName = `${baseName}Headers`;
+  return {
+    kind: StructureKind.Interface,
+    isExported: true,
+    name: headersInterfaceName,
+    properties: headerParameters.map((h: Parameter) => {
+      const description = getLanguageMetadata(h.language).description;
+      return {
+        name: `"${getLanguageMetadata(h.language).name}"`,
```
joheredi, line 128: Do headers have a serializaedName? If they do, we'll need to use that one instead of `name` name can be changed by M4 if `x-ms-client-name` is set in the swagger. `serializedName` would be the value that the service understands so that'd be the one we want to use

src/restLevelClient/generateParameterTypes.ts
```diff
@@ -82,6 +108,67 @@ export function generateParameterInterfaces(
...
+        ...(description && { docs: [{ description }] }),
+        type: primitiveSchemaToType(h.schema),
+        hasQuestionToken: true
```
joheredi, line 131: Are all Headers optional? Is there any chance a header may be required?

joheredi, line 131: Is there any way to mark a header as required? If so we may want to calculate hasQuestionToken instead of hard coding it

src/restLevelClient/generateResponseTypes.ts
```diff
@@ -194,7 +197,7 @@ function getResponseHeaderInterfaceDefinition(
         return {
           name: `"${h.header.toLowerCase()}"`,
           ...(description && { docs: [{ description }] }),
-          type: "string",
+          type: primitiveSchemaToType(h.schema),
```
joheredi, line 200: After my current PR is merged this will need to take an additional param

`primitiveSchemaToType(h.schema, SchemaContext.Output)`

test/integration/azureSpecialProperties.spec.ts
```diff
@@ -275,6 +275,10 @@ describe("AzureSpecialProperties", () => {
           ...clientOptions
         }
       );
+      client.pipeline.removePolicy({
+        name: bearerTokenAuthenticationPolicyName
```
joheredi, line 279: These changes are already in main so you should be able to remove these changes in yout PR

joheredi (approved):
Looks good, just double check to make sure you are using `serializedName` or `language.default.serializedName` instead of `.name`

src/restLevelClient/generateParameterTypes.ts
qiaozha: I have changed it into serializedName now

PR #1169, opened 2021-08-20
[RLC] Generate pagination helper

joheredi:
Generate pagination helper and export it whenever a REST Client contains operations that are marked with `x-ms-pageable`


package.json
```diff
@@ -34,12 +34,12 @@
...
         "@autorest/codemodel": "^4.15.0",
         "@autorest/extension-base": "^3.2.1",
-        "@azure-rest/core-client": "^1.0.0-beta.4",
+        "@azure-rest/core-client": "^1.0.0-beta.6",
+        "@azure-rest/core-client-paging": "^1.0.0-alpha.20210820.2",
         "@azure-tools/codegen": "^2.5.294",
         "@azure/core-client": "^1.1.2",
         "@azure/core-http": "^1.2.4",
         "@azure/core-lro": "^2.1.0",
-        "@azure/core-paging": "^1.1.3",
```
qiaozha, line 42: is it for regular sdk paging ? Are we sure it's okay to remove it ?

joheredi: Good catch!

deyaaeldeen: do we import `@azure/core-paging` directly from code gen?

test/integration/generated/bodyComplexRest/src/models.ts
```diff
@@ -31,12 +31,12 @@ export interface FloatWrapper {
 
 export interface DoubleWrapper {
   field1?: number;
-  field56ZerosAfterTheDotAndNegativeZeroBeforeDotAndThisIsALongFieldNameOnPurpose?: number;
+  field_56_zeros_after_the_dot_and_negative_zero_before_dot_and_this_is_a_long_field_name_on_purpose?: number;
```
deyaaeldeen, line 34: what is up with the change in naming convention here?

test/integration/pagingRest.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+      let response = await client.path("/paging/multiple").get();
+
+      const iter = paginate(client, response);
+
+      let index = 1;
+      let items: Product[] = [];
+
+      for await (const item of iter) {
+        assert.equal(item.properties?.id, index);
+        index++;
```
deyaaeldeen, line 185: `index` is incremented but never used

package.json
joheredi: Yeah for the High-level clients we import it. We just consume the PageableAsyncITerator type I think

PR #1172, opened 2021-08-24
Remove Node Resolve Dependency

sarangan12:
Removes Node Resolve Dependency which is obsolete.

@jeremymeng Could you please approve it?


jeremymeng (approved):
LGTM. Thanks for the fix!

Issue #1166, opened 2021-08-20
unintentional clear output folder


qiaozha:
Autorest will clear output folder when I don't specify --clear-output-folder in command line or anywhere.
Even I try to use --clear-output-folder=false to force it not to clear the output folder, it won't work.

To repro:
1. run this command `autorest --typescript --version=3.1.3 --use=@autorest/typescript@6.0.0-beta.9 --modelerfour.lenient-model-deduplication --license-header=MICROSOFT_MIT_NO_VERSION --typescript-sdks-folder=../azure-sdk-for-js ../azure-rest-api-specs/specification/dns/resource-manager/readme.md `
2. run `rush update && rush build`, we should be able to see the review folder after it finishes.
3. run this command again `autorest --typescript --version=3.1.3 --use=@autorest/typescript@6.0.0-beta.9 --modelerfour.lenient-model-deduplication --license-header=MICROSOFT_MIT_NO_VERSION --typescript-sdks-folder=../azure-sdk-for-js ../azure-rest-api-specs/specification/dns/resource-manager/readme.md ` You will see review or dist or dist-esm folder is gone. 

I also tried to remove the --version=3.1.3 and run `autorest --reset` before I run those commands, still not working.

sarangan12:
Hi @qiaozha Actually the command does have clear-output-folder. It is available in the readme.typescript.md. If you remove it there, then it will not be overriding any values. 

Is there anything else you need from me?

qiaozha:
@sarangan12 Hi, Thanks for the info, we added this option a long time ago due to https://github.com/Azure/azure-rest-api-specs/pull/7676  this pr. If we simply remove it, the older issue will still be there, especially when we generate track2 SDK for existing track1 SDK for the first time as track2 code generator has removed the *Mappers.ts in src/models folder. 
![image](https://user-images.githubusercontent.com/55688292/130384388-8114ffc2-c267-40b9-b74f-0a7ab9b2f52d.png)
I am thinking whether we can clear the src folder in code generator side ? As the src folder is intend to be clear up when we generate the code with or without `--generate-metadata` option and those metadata outside of src folder such as review or package.json can still be reserved if we don't add `--generate-metadata`  option.

sarangan12:
So, I am assuming your statement is:

1. Clear the src folder 
2. Then generate the code (irrespective of clear-output-folder/generate-metadata option)

That looks like a reasonable thing to do. Let me send the PR soon for this. Let me know if my understanding is correct

[...]

PR #1173, opened 2021-08-25
Clear Source Folder before generation

sarangan12:
This is to address the issue reported at https://github.com/Azure/autorest.typescript/issues/1166. 

Basically, the src folder could have many files (especially while migrating from Track 1). It is a pain to check and remove the files. Normally, we do not want the metadata to be regenerated but we would like to clear all the previous files within the src folder. This PR accomplishes the task.

@joheredi @qiaozha  Please review and approve it.


qiaozha (approved):
LGTM

Issue #860, opened 2021-02-05
[v4/v6] Update readme with a note on how to use an existing access token to authenticate


ramya-rao-a:
As seen in https://github.com/Azure/azure-sdk-for-js/issues/13626 and https://github.com/Azure/azure-sdk-for-js/issues/11243, we are seeing cases where people want to use an access token they already have at hand with the client libraries generated by this code generator.

While we documented the solution for this in the readme for `@azure/ms-rest-nodeauth` package (See [Authenticating with existing token](https://www.npmjs.com/package/@azure/ms-rest-nodeauth#authenticating-with-an-existing-token)), this package may not be the first place people look at. It is much more likely for them to look at the readme for the package that contains the client library.

So, this feature request to add the relevant message both in v4 and v6 code generator.
We can copy over the code snippet from the readme of the ms-rest-nodeauth package to the readme of the generated package.
We can have a similar code snippet for the `TokenCredential` interface in case of the v6 code gen

cc @joheredi, @deyaaeldeen, @sarangan12

PR #1174, opened 2021-08-25
Reference how to use pre-fetched access tokens in Readme

JonathanCrd:
Issue https://github.com/Azure/autorest.typescript/issues/860.

After talking with Ramya, we concluded to put a reference to [this document](https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/identity/identity/samples/AzureIdentityExamples.md#authenticating-with-a-pre-fetched-access-token) in the main readme, so it's more visible and clearer for customers when they want to use an access token that they already have.

For track 2, the reference is inside the auth section of the readme.


ramya-rao-a (changes requested):
Here we are placing the section `Authenticating with a pre-fetched access token` as a sibling to `Using an Azure Active Directory Credential` where as it should be under the latter. Both talk about using AAD. Its just that one uses the credentials from `@azure/identity` that will make a service call to get the token and in another the customer already has an AAD token that they got through different means

I would recommend the below
- Remove the existing section header `Using an Azure Active Directory Credential`. The header above that already calls out that we are covering authentication
- You can update the line `You can authenticate with Azure Active Directory using the Azure Identity library.` to `You can authenticate with Azure Active Directory using a credential from the `@azure/identity` library or [use an existing AAD token](https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/identity/identity/samples/AzureIdentityExamples.md#authenticating-with-a-pre-fetched-access-token).

sarangan12 (approved):

ramya-rao-a (approved):

PR #1181, opened 2021-08-30
Removing `Operations` keyword from params name

JonathanCrd:
For issue https://github.com/Azure/autorest.typescript/issues/1143, the `Operations` keyword is being added to avoid conflicts between the operation groups and model names.

However, only the operation group name should be modified, and not the params and interfaces name, as it can be seen in this track1 generated [file](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/storage/storage-datalake/src/operations/pathOperations.ts).

The code where the params name were being modified has been deleted, and a new swagger test case was added.


sarangan12 (approved):
Excellent Work

PR #1184, opened 2021-09-07
Fix the value of clientName Property in the Discriminators

sarangan12:
Recently, a customer reported an issue with the `search-documents` SDK. The problem is that while they are using the types with Discriminators, only the parent type is getting accessed. So, the properties that are specific to the child type are ignored and not sent to the service side. 

On analysis, I have found that the issue is because of the problem with the property `clientName`. We have 2 properties:

1. clientName
2. serializedName

As of now, both the properties are set to same value. That was not the intention of the 2 properties and is incorrect. If you look at the code within `core-http` (where these values are getting used), it is clear that the property `clientName` should indicate the `name` that is used on the client/SDK side. 

This PR has been created to fix the issue and set the value of the `clientName` correctly. I have not added a seperate test case for this change, as we already have scenarios covering it. 

### DIFF
--- a/src/transforms/mapperTransforms.ts
+++ b/src/transforms/mapperTransforms.ts
@@ -362,7 +362,10 @@ function transformObjectMapper(pipelineValue: PipelineValue) {
         uberParent,
         polymorphicDiscriminator: {
           serializedName: discriminator!.property.serializedName,
-          clientName: discriminator!.property.serializedName
+          clientName: normalizeName(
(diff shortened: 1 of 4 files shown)


joheredi (approved):
LGTM!

joheredi (approved):
I wonder why the test server scenarios were not failing. We should probably add a scenario like this in Test Server, would you mind filing an issue in test server to track adding one?

PR #1186, opened 2021-09-10
Bump azure identity version to make rush update happy

qiaozha:


sarangan12 (approved):
I think this is the one that Jose has told about as a temporary measure. Please create an issue and assign it to yourself to remind you to remove the skips after the next release.

qiaozha:
> I think this is the one that Jose has told about as a temporary measure. Please create an issue and assign it to yourself to remind you to remove the skips after the next release.

I have this issue here #1187

Issue #1140, opened 2021-08-05
Enum Value with only one option is always present in the request


sarangan12:
In the Storage Blob Data Plane SDK - [file](https://github.com/Azure/azure-rest-api-specs/blob/storage-dataplane-preview/specification/storage/data-plane/Microsoft.BlobStorage/preview/2020-10-02/blob.json), we have the following swagger definition:

```
"EncryptionAlgorithm": {
      "name": "x-ms-encryption-algorithm",
      "x-ms-client-name": "encryptionAlgorithm",
      "type": "string",
      "in": "header",
      "required": false,
      "enum": [
        "AES256"
      ],
      "x-ms-enum": {
        "name": "EncryptionAlgorithmType",
        "modelAsString": false
      },
      "x-ms-parameter-location": "method",
      "x-ms-parameter-grouping": {
        "name": "cpk-info"
      },
      "description": "The algorithm used to produce the encryption key hash. Currently, the only accepted value is \"AES256\". Must be provided if the x-ms-encryption-key header is provided."
}
```

With the above swagger definition, the parameter definition looks as:

```
export const encryptionAlgorithm: OperationParameter = {
  parameterPath: ["options", "cpkInfo", "encryptionAlgorithm"],
  mapper: {
    defaultValue: "AES256",
    isConstant: true,
    serializedName: "x-ms-encryption-algorithm",
    type: {
      name: "String"
    }
  }
};
```

With the above code, the request always has this encryptionAlgorithm even when the user does not specify anything. Such behavior is not present in the old generator.

CC: @ramya-rao-a @EmmaZhu

PR #1197, opened 2021-09-16
Removing defaultValue for optional enum parameters with only one value.

JonathanCrd:
Issue: https://github.com/Azure/autorest.typescript/issues/1140

Removing the `defaultValue` property from optional parameters, so the value is not present in the request headers. Only constant schemas with serialize name equals to `"Content-Type"` is to have the default value.

For this, a new test swagger and integration tests were created (enumvalues), 4 parameters tests cases were assessed:

- An enum with 1 value.
- An enum with 2 values.
- A required param with an enum with 1 value.
- A required param with an enum with 2 values.

Only the required parameters with the enums were present in the request call, which is the intended behavior.


JonathanCrd:
Closing this old PR, if changes are needed it will be done in a new one

PR #1201, opened 2021-09-24
[RLC] support multiple mediatypes

joheredi:
Add support and test coverage for multiple media types


qiaozha:
Just wondering if it's possible that the response could also be media types? Should we support this kind of scenarios?

joheredi:
> Just wondering if it's possible that the response could also be media types? Should we support this kind of scenarios?

This should be handled already if the Swagger defines the options a "Content-Type" header can have. For example a swagger like this

```json
"responses": {
          "200": {
            "description": "Empty Response",
            "headers": {
              "Content-Type": {
                "description": "response with header value \"Content-Type\": \"text/html\"",
                "type": "string",
                "enum": [ "text/html" ]
              }
            }
          },
          "default": {
            "description": "Unexpected error",
            "schema": {
              "$ref": "#/definitions/Error"
            }
          }
        }
```

would end up generating a response type like this

```typescript
export interface ResponseProtectedKey200Headers {
  /** response with header value "Content-Type": "text/html" */
  "content-type"?: "text/html";
}
```

test/integration/generated/mediaTypesRest/src/parameters.ts
```diff
@@ -0,0 +1,54 @@
...
+   * Input parameter.
+   *
+   * Value may contain any sequence of octets
+   */
+  body?: string;
+}
+
+export interface AnalyzeBodyMediaTypesParam {
+  /** Request content type */
+  contentType?: "application/pdf" | "image/jpeg" | "image/png" | "image/tiff";
```
qiaozha, line 18: Will we pass the value for the customer by default if this contentType is a single value but not equals to application/json ?

qiaozha:
Are we suppose to make this before or after the stream support ?

test/integration/generated/mediaTypesRest/src/parameters.ts
joheredi: Nope, we can't pass values by default as we are not generating any runtime code, just type definitions. We could explore making contentType required and only accept one contentType. But I would like to wait until we get feedback on that. In the meantime we'll keep it optional and default to json

joheredi:
> Are we suppose to make this before or after the stream support ?

Before as there is still some design discussion and testing for stream support, meanwhile the response content will be loaded in memory.

test/integration/generated/mediaTypesRest/src/parameters.ts
qiaozha: I see. Thanks

qiaozha (approved):
LGTM

PR #1234, opened 2021-11-04
bump m4 version and autorest version

qiaozha:


test/smoke/generated/web-resource-manager/src/models/index.ts
```diff
@@ -106,6 +81,31 @@ export interface CertificateDetails {
...
+  location: string;
+  /**
+   * Resource type.
+   * NOTE: This property will not be serialized. It can only be populated by the server.
+   */
+  readonly type?: string;
+  /** Resource tags. */
+  tags?: { [propertyName: string]: string };
+}
+
```
qiaozha, line 108: Why this definition changed order ?

sarangan12 (approved):

Issue #1213, opened 2021-10-22
API reference docs links for preview packages should point to the preview docs


ramya-rao-a:
The API reference docs link generated in the readme for the beta version of a package should point to the preview version of the docs.

For example, see the [readme for @azure/arm-eventhub version 5.0.0-beta.1](https://www.npmjs.com/package/@azure/arm-eventhub/v/5.0.0-beta.1). The API reference docs link point to the docs corresponding to the stable version of this package. Adding the query parameter `?view=azure-node-preview` in the end should work

PR #1235, opened 2021-11-04
Reference docs links for preview packages

JonathanCrd:
Solves https://github.com/Azure/autorest.typescript/issues/1213

If the generated package is a preview version, the API reference docs URL contains `?view=azure-node-preview` at the end. To determine if a package is a preview SDK it looks for the keyword `preview` or `beta` on the package version.

Tested this change by manually modifying the `const package_version` and `packageName` to add `@azure` in _test-swagger-gen.ts_. These changes were not committed.

### DIFF
--- a/src/generators/static/readmeFileGenerator.ts
+++ b/src/generators/static/readmeFileGenerator.ts
@@ -50,7 +50,7 @@ interface Metadata {
   /** Indicates if the package is a test/releasable package. */
   isReleasablePackage?: boolean;
   /** indicate if the package is management plane SDK */
-  azureArm?: boolean
+  azureArm?: boolean;
 }
 
 /**
@@ -100,6 +100,12 @@ function createMetadata(
     : simpleServiceName;
   const identityPackageURL =
     repoURL && `${repoURL}/tree/main/sdk/identity/identity`;
+
+  var PackageNPMURLSufix: string = "";
+  if (clientPackageName.includes("-beta")) {
+    PackageNPMURLSufix = "?view=azure-node-preview";
+  }
+
   return {
     serviceName: serviceName,
     clientPackageName: clientPackageName,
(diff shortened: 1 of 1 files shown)


src/generators/static/readmeFileGenerator.ts
```diff
@@ -100,6 +100,15 @@ function createMetadata(
     : simpleServiceName;
   const identityPackageURL =
     repoURL && `${repoURL}/tree/main/sdk/identity/identity`;
+
+  var apiRefUrlPreviewSufix: string = "";
+  if (
+    packageDetails.version.includes("preview") ||
```
sarangan12, line 106: please change preview to -preview. and beta to -beta.

sarangan12 (changes requested):

src/generators/static/readmeFileGenerator.ts
sarangan12: On second thought, since you are validating the version, you need not change

sarangan12 (approved):

src/generators/static/readmeFileGenerator.ts
```diff
@@ -100,6 +100,15 @@ function createMetadata(
     : simpleServiceName;
   const identityPackageURL =
     repoURL && `${repoURL}/tree/main/sdk/identity/identity`;
+
+  var apiRefUrlPreviewSufix: string = "";
+  if (
```
ramya-rao-a, line 105: Why check for `preview`? All our packages now use `beta` in the package version to denote a preview package.

src/generators/static/readmeFileGenerator.ts
```diff
@@ -100,6 +100,15 @@ function createMetadata(
...
+    packageDetails.version.includes("preview") ||
+    packageDetails.version.includes("beta")
+  ) {
+    apiRefUrlPreviewSufix = "?view=azure-node-preview";
```
ramya-rao-a, line 109: Nit: I would recommend naming this variable in a way that denotes that what we are adding is a query parameter and not just any suffix

Maybe `apiRefUrlQueryParameter`?

src/generators/static/readmeFileGenerator.ts
JonathanCrd: Saw the `preview` keyword on the integration tests, if it's no longer being used on new SDKs I can remove it in a separate PR

src/generators/static/readmeFileGenerator.ts
JonathanCrd: I like `apiRefUrlQueryParameter`, making this change in a new PR

PR #1237, opened 2021-11-09
Set default RLC flags

joheredi:
Setting a few commonly used RLC flags to minimize the boilerplate needed when creating a new config file.


README.md
```diff
@@ -87,6 +87,14 @@ output-artifact: typescript-files
 ### REST Client Generator overrides
 
 ```yaml $(rest-level-client)
+generate-metadata: false
+license-header: MICROSOFT_MIT_NO_VERSION
+source-code-folder-path: ./src
```
qiaozha, line 92: I don't think that's applicable for multi-client ?

README.md
```diff
@@ -87,6 +87,14 @@ output-artifact: typescript-files
 ### REST Client Generator overrides
 
 ```yaml $(rest-level-client)
+generate-metadata: false
+license-header: MICROSOFT_MIT_NO_VERSION
+source-code-folder-path: ./src
+add-credentials: true
+
+typescript:
+  output-folder: ../
```
qiaozha, line 96: Same here

README.md
joheredi: When adding multi client support we would probably need to add another check for the multi-client flag and setup the right directives for this.  Such as `$(rest-level-client) && !$(multi-client)`

dw511214992: For multi-client, we can define the `source-code-folder-path` in the README.md which is stored in sdk repo. Then the `source-code-folder-path` will be overrided.

README.md
```diff
@@ -117,12 +116,23 @@ pipeline:
...
     scope: typescript-scope/emitter
-
 ```
 
 ### REST Client Generator overrides
 
 ```yaml $(rest-level-client)
+license-header: MICROSOFT_MIT_NO_VERSION
+
+$(add-credentials == undefined):
```
qiaozha, line 126: Are you sure this kind of yaml condition really work ?

Issue #1214, opened 2021-10-22
Provide link to bundling docs when talking about browser support


ramya-rao-a:
As being discussed in https://github.com/Azure/azure-sdk-for-js/issues/18326, we need to provide a link to the [bundling docs](http://aka.ms/AzureSDKBundling) in the generated readme.

This applies to both v4 and v6 of the code generator

For v4, it can be another bullet point in the below shown section

![image](https://user-images.githubusercontent.com/16890566/138530015-7b43cf1e-375b-4559-8f7d-e25f26cb3aa2.png)

For v6, it should follow the format we used in say https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/storage/storage-blob/README.md#javascript-bundle

JonathanCrd:
This has been solved for v6.
There are some issues with v4 that need a deeper investigation. Issue https://github.com/Azure/autorest.typescript/issues/1245 keeps track of this.

PR #1240, opened 2021-11-11
Adding link to bundling docs

JonathanCrd:
Solves https://github.com/Azure/autorest.typescript/issues/1214 for v4.

Bullet point referencing the Javascript Bundle added under the `browser - Authentication, client creation, and ...` section.

### DIFF
--- a/src/vanilla/Model/CodeModelTS.cs
+++ b/src/vanilla/Model/CodeModelTS.cs
@@ -1098,7 +1098,7 @@ namespace AutoRest.TypeScript.Model
                 builder.Line("In browser applications, we recommend using the `InteractiveBrowserCredential` that interactively authenticates using the default system browser.");
                 builder.Line("  - See [Single-page application: App registration guide](https://docs.microsoft.com/azure/active-directory/develop/scenario-spa-app-registration) to configure your app registration for the browser.");
                 builder.Line("  - Note down the client Id from the previous step and use it in the browser sample below.");
-                builder.Line("  - To use this client library in the browser, first you need to use a bundler. For details on how to do this, please refer to our bundling documentation.");
+                builder.Line("  - To use this client library in the browser, first you need to use a bundler. For details on how to do this, please refer to our [bundling documentation](https://github.com/Azure/azure-sdk-for-js/blob/master/documentation/Bundling.md).");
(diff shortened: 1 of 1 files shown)


src/vanilla/Model/CodeModelTS.cs
```diff
@@ -1098,6 +1098,7 @@ private void GenerateHowToUseInBrowser(MarkdownBuilder builder)
                 builder.Line("In browser applications, we recommend using the `InteractiveBrowserCredential` that interactively authenticates using the default system browser.");
                 builder.Line("  - See [Single-page application: App registration guide](https://docs.microsoft.com/azure/active-directory/develop/scenario-spa-app-registration) to configure your app registration for the browser.");
                 builder.Line("  - Note down the client Id from the previous step and use it in the browser sample below.");
+                builder.Line("  - To use this client library in the browser, first you need to use a bundler. For details on how to do this, please refer to our bundling documentation.");
```
joheredi, line 1101: Can you add a link to the bundling docs?

JonathanCrd: My bad, Thanks for spotting this!

joheredi (approved):

ramya-rao-a:
Don't we have to update the package version or something to ensure a new version is released with this change?

joheredi:
Yes, we need to bump the patch version before releasing it. I usually run `npm version patch` and push the changes. @JonathanCrd ping me when this is ready to release and I can help with that

sarangan12:
@joheredi The CI is not running for this PR. Any ideas??

JonathanCrd:
This is the output of the pipelines: 
```
##[warning]An image label with the label Ubuntu 16.04 does not exist.
,##[error]The remote provider was unable to process the request.
```

Could the service or VMs be down?

joheredi:
It looks like the Ubuntu 16 vm image has been discontinued in Azure DevOps. You should be able to fix this updating `azure-pipelines.yml`.

This change should fix it:
image

ramya-rao-a:
By the end of January 2022, almost all mgmt plane packages would have moved to using v6 code generator. Therefore, it is not worth our time to fix this for the v4 code generator

PR #1243, opened 2021-11-12
Update generated tests and remove `.only` in url tests

joheredi:
There were a couple of issues from the last 2 PRs

1. There was an unresolved merge conflict that caused a couple of RLC tests to be outdated. Re-generating fixes this issue.
2. Accidentally the urlRest tests added a `.only` in their describe, which skips all other tests in the suite. This hid a few issues that came up when upgrading to the latest @azure-rest/core-client. Updated the tests to align with the changes, which were
   2.1 core-client-rest now throws when the response has application/json content-type but it is an invalid JSON
   2.2 We are now encoding the URL, so one of the LRO tests needed to set skipUrlEncoding property.
   2.3 Media Types REST PDF test fails, I have filed an issue to investigate and fix in a separate PR

### DIFF
--- a/test/integration/urlRest.spec.ts
+++ b/test/integration/urlRest.spec.ts
@@ -1,16 +1,16 @@
 import * as assert from "assert";
-import UrlRestClient, { UrlRestClientRestClient } from "./generated/urlRest/src";
+import UrlRestClient, {
+  UrlRestClientRestClient
+} from "./generated/urlRest/src";
 import { UriColor } from "./generated/url/src";
 
-describe.only("Integration tests for UrlRest", () => {
+describe("Integration tests for UrlRest", () => {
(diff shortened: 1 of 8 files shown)


test/integration/urlRest.spec.ts
```diff
@@ -1,16 +1,16 @@
 import * as assert from "assert";
-import UrlRestClient, { UrlRestClientRestClient } from "./generated/urlRest/src";
+import UrlRestClient, {
+  UrlRestClientRestClient
+} from "./generated/urlRest/src";
 import { UriColor } from "./generated/url/src";
 
-describe.only("Integration tests for UrlRest", () => {
```
joheredi, line 5: Everything in this file, but this line are just formatting changes

sarangan12 (approved):

PR #1244, opened 2021-11-12
[v4x] Fix CI

joheredi:
It looks like the vminage was discontinued, updating it and moving the ci file to the .scripts folder to match the main branch


sarangan12:
The change looks fine. But, I think there is something else strange that is going on. The `npm install` fails, due to a possible timeout. I could not get this working even locally.

sarangan12 (approved):

PR #1249, opened 2021-11-18
samples generation for js mgmt sdk

qiaozha:
This PR is to generate management SDK samples from swagger examples.


joheredi:
> _No description provided._

Would you mind providing some context on the change in the description?

test/smoke/generated/sql-resource-manager/samples-dev/createOrUpdateAnExtendedDatabaseSAzureMonitorAuditingPolicyWithMinimalParameters.ts
```diff
@@ -0,0 +1,44 @@
...
+  const databaseName = "testdb";
+  const parameters: ExtendedDatabaseBlobAuditingPolicy = {
+    isAzureMonitorTargetEnabled: true,
+    state: "Enabled"
+  };
+  await client.extendedDatabaseBlobAuditingPolicies
+    .createOrUpdate(resourceGroupName, serverName, databaseName, parameters)
+    .then((res) => {
+      console.log(res);
+    });
```
joheredi, line 36:
```suggestion
  const result = await client.extendedDatabaseBlobAuditingPolicies
    .createOrUpdate(resourceGroupName, serverName, databaseName, parameters);
  console.log(result);
```

src/generators/static/samples.ts.hbs
```diff
@@ -0,0 +1,48 @@
...
+    resArray.push(item);
+  }
+  console.log(resArray);
+  {{else}}  
+  await client
+    {{#unless isTopLevel}}
+    .{{operationGroupName}}
+    {{/unless}}
+    .{{operationName}}({{methodParameterNames}})
+    .then((res) => {
```
sarangan12, line 35: Why do we use `await` and `then` together?

src/models/operationDetails.ts
```diff
@@ -103,6 +103,7 @@ export interface OperationSpecDetails {
 export interface OperationGroupDetails {
   key: string;
   name: string;
+  originalKey: string;
```
sarangan12, line 106: what is the purpose of this `originalKey`? This name does not convey the true meaning and might create confusion in the future.  Could you explain?

test/smoke/generated/arm-package-deploymentscripts-2019-10-preview/review/arm-package-deploymentscripts-2019-10-preview.api.md
```diff
@@ -118,7 +118,21 @@ export interface DeploymentScripts {
     getLogs(resourceGroupName: string, scriptName: string, options?: DeploymentScriptsGetLogsOptionalParams): Promise<DeploymentScriptsGetLogsResponse>;
     getLogsDefault(resourceGroupName: string, scriptName: string, options?: DeploymentScriptsGetLogsDefaultOptionalParams): Promise<DeploymentScriptsGetLogsDefaultResponse>;
     listByResourceGroup(resourceGroupName: string, options?: DeploymentScriptsListByResourceGroupOptionalParams): PagedAsyncIterableIterator<DeploymentScriptUnion>;
+    listByResourceGroupNext(resourceGroupName: string, nextLink: string, options?: DeploymentScriptsListByResourceGroupNextOptionalParams): PagedAsyncIterableIterator<DeploymentScriptUnion>;
```
sarangan12, line 121: How are these changes related to this PR?

src/models/operationDetails.ts
qiaozha: This is because you have changed the name in conflict resolver and I need the originalKey value to find out which operationGroup/operation matches to one specific example in test modeler.

qiaozha:
I will convert this PR into draft stage as there's still some unresolved issues.

qiaozha:
Hi @sarangan12 and @joheredi  I think this pr is ready for review now, Could you help take a look again ? Thank you

joheredi:
It's a little bit hard to add comments inline because the browser hangs due to the size of the PR. I'm leaving some comments here:

I see that we are generating a commented line, is this on purpose? https://github.com/qiaozha/autorest.typescript/blob/062707da9b19edadf48f6d081a366e722f64acf9/test/smoke/generated/sql-resource-manager/samples-dev/assignsMaintenanceConfigurationToAnElasticPool.ts#L23

I don't love the idea of having a global client in the sample file and then assigning it in the main function. This makes the sample functions non-pure without a real need for that. I would prefer to move the client initialization into each sample function instead of having a shared client. Ideally I would like to see the sample like this

```typescript
import {
  ElasticPoolUpdate,
  SqlManagementClient
} from "@msinternal/sql-resource-manager";
import { DefaultAzureCredential } from "@azure/identity";

async function assignsMaintenanceConfigurationToAnElasticPool() {
  const credential = new DefaultAzureCredential();
  const subscriptionId = "00000000-1111-2222-3333-444444444444";
  const client = new SqlManagementClient(credential, subscriptionId);

  const resourceGroupName = "sqlcrudtest-2369";
  const serverName = "sqlcrudtest-8069";
  const elasticPoolName = "sqlcrudtest-8102";
  const parameters: ElasticPoolUpdate = {
    maintenanceConfigurationId:
      "/subscriptions/00000000-1111-2222-3333-444444444444/providers/Microsoft.Maintenance/publicMaintenanceConfigurations/SQL_JapanEast_1"
  };
  const result = await client.elasticPools.beginUpdateAndWait(
    resourceGroupName,
    serverName,
    elasticPoolName,
    parameters
  );
  console.log(result);
}

async function main() {
  await assignsMaintenanceConfigurationToAnElasticPool();
}

// To make sure we log any errors
main().catch(console.error)
``` 

If we have a single sample function per file we might as well remove the main function altogether and call the sample directly

```typescript
assignsMaintenanceConfigurationToAnElasticPool().catch(console.error)
```

joheredi:
I think we could add this behind a flag `--generate-samples` which can be true if `--azure-arm` is set. This way data-plane libraries can choose to turn off and write their own samples.

Also, what happens if a swagger has no examples? Do we just skip sample generation or do we fail?

qiaozha:
> It's a little bit hard to add comments inline because the browser hangs due to the size of the PR. I'm leaving some comments here:
> 
> I see that we are generating a commented line, is this on purpose? https://github.com/qiaozha/autorest.typescript/blob/062707da9b19edadf48f6d081a366e722f64acf9/test/smoke/generated/sql-resource-manager/samples-dev/assignsMaintenanceConfigurationToAnElasticPool.ts#L23
> 
> I don't love the idea of having a global client in the sample file and then assigning it in the main function. This makes the sample functions non-pure without a real need for that. I would prefer to move the client initialization into each sample function instead of having a shared client. Ideally I would like to see the sample like this
> 
> ```ts
> import {
>   ElasticPoolUpdate,
>   SqlManagementClient
> } from "@msinternal/sql-resource-manager";
> import { DefaultAzureCredential } from "@azure/identity";
> 
> async function assignsMaintenanceConfigurationToAnElasticPool() {
>   const credential = new DefaultAzureCredential();
>   const subscriptionId = "00000000-1111-2222-3333-444444444444";
>   const client = new SqlManagementClient(credential, subscriptionId);
> 
>   const resourceGroupName = "sqlcrudtest-2369";
>   const serverName = "sqlcrudtest-8069";
>   const elasticPoolName = "sqlcrudtest-8102";
>   const parameters: ElasticPoolUpdate = {
>     maintenanceConfigurationId:
>       "/subscriptions/00000000-1111-2222-3333-444444444444/providers/Microsoft.Maintenance/publicMaintenanceConfigurations/SQL_JapanEast_1"
>   };
>   const result = await client.elasticPools.beginUpdateAndWait(
>     resourceGroupName,
>     serverName,
>     elasticPoolName,
>     parameters
>   );
>   console.log(result);
> }
> 
> async function main() {
>   await assignsMaintenanceConfigurationToAnElasticPool();
> }
> 
> // To make sure we log any errors
> main().catch(console.error)
> ```
> 
> If we have a single sample function per file we might as well remove the main function altogether and call the sample directly
> 
> ```ts
> assignsMaintenanceConfigurationToAnElasticPool().catch(console.error)
> ```

The sample template has been updated, @joheredi Could you take a look again ? Thanks

qiaozha:
> I think we could add this behind a flag `--generate-samples` which can be true if `--azure-arm` is set. This way data-plane libraries can choose to turn off and write their own samples.

According to our offline discussion, we already has a --generate-test flag for this sample generation, and it's fine keep as it is. 

> Also, what happens if a swagger has no examples? Do we just skip sample generation or do we fail?

We will skip sample generation in that case.

test/integration/generated/additionalProperties/package.json
```diff
@@ -84,5 +84,11 @@
     "docs": "echo skipped"
   },
   "sideEffects": false,
-  "autoPublish": true
+  "autoPublish": true,
+  "//sampleConfiguration": {
```
joheredi, line 88: I don't see any samples generated for this integration test. Why are we adding this sample metadata config?

test/integration/lro.spec.ts
```diff
@@ -540,7 +540,7 @@ describe("Custom Headers", () => {
     await client.lROsCustomHeader.beginPost202Retry200AndWait(options);
     check200(lastResponse);
   });
-});
+}).timeout(5000);
```
joheredi, line 543: Is there a way we can avoid doing this long timeout?

I think we can use Sinon fakeTimers: https://sinonjs.org/releases/v1.17.7/fake-timers/

joheredi (commented):
Changes look good to me. I left a question about one of the integration tests adding sample metadata. Please make sure to check why that is being generated.

Other than that I'm good with the changes, please follow up with @sarangan12 about his comment on `originalName`. Feel free to merge once he approves the PR

test/integration/generated/additionalProperties/package.json
qiaozha: changed it

test/integration/lro.spec.ts
qiaozha: The ci somehow fails and report some errors like timeout if I don't add this. I am also quite superised about this as this should not be related to my changes.

sarangan12 (approved):
Great work on this mega PR :-)

joheredi (approved):

test/integration/lro.spec.ts
qiaozha: removed.

package.json
```diff
@@ -32,9 +32,10 @@
         "LICENSE"
     ],
     "dependencies": {
-        "@autorest/codemodel": "^4.15.0",
+        "@autorest/codemodel": "^4.17.0",
         "@autorest/extension-base": "^3.4.1",
         "@azure-rest/core-client": "1.0.0-alpha.20211206.10",
+        "@autorest/testmodeler": "1.0.4",
```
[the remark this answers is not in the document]
qiaozha: updated

Issue #1131, opened 2021-08-02
Better heuristics for generated names


xirzec:
We often end up with bad names, like:

- names with suffixes with incrementally higher numbers
-  things like ApiVersion20200630Preview
- Acronyms looking bad: #430

We should come up with better heuristics or invent syntax to address these issues where we can't resolve them automatically

sarangan12:
@sarangan12 Discuss with Autorest crew about naming issues?

sarangan12:
[Example 1](https://github.com/Azure/autorest.typescript/blob/main/test/integration/generated/appconfiguration/src/models/index.ts)
KnownGet6ItemsItem, Get6ItemsItem, Head6ItemsItem, Get7ItemsItem, Head7ItemsItem 

[Example 2](https://github.com/Azure/autorest.typescript/blob/main/test/integration/generated/arrayConstraints/src/models/index.ts)
Enum0, ArrayConstraintsClientApiV1ValueGetOptionalParams, ArrayConstraintsClientApiV1ValueGetResponse 

[Example 3](https://github.com/Azure/autorest.typescript/blob/main/test/integration/generated/azureSpecialProperties/src/models/index.ts)
SubscriptionInCredentialsPostMethodGlobalNotProvidedValidOptionalParams, SubscriptionInCredentialsPostMethodGlobalNotProvidedValidOptionalParams, SubscriptionInCredentialsPostMethodGlobalNotProvidedValidOptionalParams

[Example 4](https://github.com/Azure/autorest.typescript/blob/main/test/integration/generated/httpInfrastructure/src/models/index.ts)
MultipleResponsesGet200Model201ModelDefaultError200ValidOptionalParams, MultipleResponsesGet200Model201ModelDefaultError200ValidResponse, MultipleResponsesGet200Model201ModelDefaultError201ValidOptionalParams, MultipleResponsesGet200Model201ModelDefaultError400ValidOptionalParams, MultipleResponsesGet200ModelA201ModelC404ModelDDefaultError200ValidOptionalParams, MultipleResponsesGet200ModelA201ModelC404ModelDDefaultError200ValidResponse

[Example 5](https://github.com/Azure/azure-sdk-for-js/blob/9ac7dd03347d1d52417016d3d52901e42de024ae/sdk/search/search-documents/src/generated/service/models/index.ts)
[...]

Issue #1134, opened 2021-08-02
Merge generated Client and ClientContext together


xirzec:
Today generated clients have two levels of inheritance to them, but we really don't need to be doing this. By flattening now we will make it easier for convenience clients to extend the generated client without it getting too deep.

As part of this work, we should make sure the generated client doesn't expose any properties that we wouldn't want convenience clients to inherit.

sarangan12:
Yes. This will be a breaking change (to a certain extent). But, we could implement it with deprecate flag/removing its functionality in steps and eventually remove it without much impact to the customer.

ramya-rao-a:
Any use cases where having these two separate is actually better?

xirzec:
> Any use cases where having these two separate is actually better?

I honestly don't understand why they are separate today. It doesn't appear to offer any tangible benefit to developers. What would someone use ClientContext for? Maybe I'm missing a use-case?

ramya-rao-a:
It is definitely something we inherited from track 1 code gen. If this was something that was followed in other languages, then maybe @sarangan12 would know from his ruby times?

joheredi:
Yeah, this is definitely inherited from track1 codegen. I can't think of a scenario where using the context class directly would be useful, it wouldn't even have any operations on it as they are added in the Client class which inherits the Context.

xirzec:
As we regenerate SDKs will this change now start showing up?

joheredi:
Yes! We already have some sdks without the ClientContext class: https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/resources/arm-resources/src/resourceManagementClient.ts

Issue #430, opened 2019-02-12
Why are Client Model Class Property names created with the first character lower case?


abhishekgoenka:
I am generating typescript clients using AutoRest. All my properties in model->index.ts are converted to lower case(only first character). Is there any switch to ignore this behavior?

Amareis:
+1, looks ugly
![image](https://user-images.githubusercontent.com/4806617/54084384-af103e00-4351-11e9-962a-5d36fc129145.png)

bricejar:
https://en.wikipedia.org/wiki/Camel_case 

camelCase is a javascript naming convention.

RikkiGibson:
related to #236

joheredi:
This is still an issue in v6, and I think there may be 2 parts:

1) Our rename logic blindly tries to camel case, however Acronyms should be left unchanged
2) There may be some times where users may want to disable re-naming

We'll take this fix in the new generator version

sarangan12:
@joheredi Do you think if a flag is good for this approach? I think we could have names in the same swagger which should be normalized (as is now) and names that should not be normalized (like JWT...) I think the right candidate should be an extension on AutoRest level like "x-ms-normalize-name". Any thoughts???

sarangan12:
I have verified that this issue has been resolved with PT #1250. Closing this issue

PR #1250, opened 2021-11-18
Fix track2 ga naming and client merge issues

qiaozha:


test/integration/generated/datalakestorage/src/models/index.ts
```diff
@@ -180,7 +180,7 @@ export interface PathUpdateHeaders {
   /** User-defined properties associated with the file or directory, in the format of a comma-separated list of name and value pairs "n1=v1, n2=v2, ...", where each value is a base64 encoded string. Note that the string may only contain ASCII characters in the ISO-8859-1 character set. */
   properties?: string;
   /** When performing setAccessControlRecursive on a directory, the number of paths that are processed with each invocation is limited.  If the number of paths to be processed exceeds this limit, a continuation token is returned in this response header.  When a continuation token is returned in the response, it must be specified in a subsequent invocation of the setAccessControlRecursive operation to continue the setAccessControlRecursive operation on the directory. */
-  xMsContinuation?: string;
```
qiaozha, line 183: This is because Style.camel("XMsContinuation", true, {}, 5} gives us XMsContinuation

test/integration/generated/datalakestorage/src/models/index.ts
```diff
@@ -82,7 +82,7 @@ export interface BlobItemInternal {
   deleted: boolean;
   snapshot: string;
   versionId?: string;
-  isCurrentVersion?: boolean;
```
qiaozha, line 85: Not sure why we have this change

test/integration/generated/datalakestorage/src/models/index.ts
```diff
@@ -301,7 +301,7 @@ export interface PathGetPropertiesHeaders {
   /** The POSIX access permissions for the file owner, the file owning group, and others. Included in the response if Hierarchical Namespace is enabled for the account. */
   permissions?: string;
   /** The POSIX access control list for the file or directory.  Included in the response only if the action is "getAccessControl" and Hierarchical Namespace is enabled for the account. */
-  acl?: string;
+  ACL?: string;
```
qiaozha, line 304: probably should use override for this case?

test/integration/generated/lro/src/index.ts
```diff
@@ -7,6 +7,5 @@
  */
 
 export * from "./models";
-export { LROClient } from "./lROClient";
-export { LROClientContext } from "./lROClientContext";
+export { LROClient } from "./LROClient";
```
qiaozha, line 10: Does the file name change look ok ? This should also because of Style.camel("LRO") = "LRO"

test/integration/generated/lro/src/models/mappers.ts
```diff
@@ -198,10 +198,10 @@ export const SubProduct: coreClient.CompositeMapper = {
   }
 };
 
-export const LROsPatch200SucceededIgnoreHeadersHeaders: coreClient.CompositeMapper = {
+export const LROsPatch200SucceededIgnoreHeaders: coreClient.CompositeMapper = {
```
qiaozha, line 201: Not sure why we change it from HeadersHeaders to Headers ?

test/integration/generated/lro/src/operations/index.ts
```diff
@@ -6,7 +6,7 @@
  * Changes may cause incorrect behavior and will be lost if the code is regenerated.
  */
 
-export * from "./lROs";
-export * from "./lRORetrys";
+export * from "./LROs";
+export * from "./LRORetrys";
 export * from "./lrosaDs";
-export * from "./lROsCustomHeader";
+export * from "./LROsCustomHeader";
```
qiaozha, line 12: The same as the previous one.

test/integration/generated/nameChecker/src/models/parameters.ts
```diff
@@ -340,8 +340,8 @@ export const selectedFields: OperationQueryParameter = {
   collectionFormat: "CSV"
 };
 
-export const search$DONotNormalize$Text: OperationQueryParameter = {
-  parameterPath: "search$DONotNormalize$Text",
+export const searchDONOTNormalizeText: OperationQueryParameter = {
+  parameterPath: "searchDONOTNormalizeText",
```
qiaozha, line 344: Maybe this is because I set the Style.camel(str, true, {}, 5) but if I don't set it as 5, I get a mismatch in bodyComplex of CMYKColor in models/index.ts.

test/smoke/generated/network-resource-manager/src/models/index.ts
```diff
@@ -775,7 +775,7 @@ export interface BastionShareableLinkListRequest {
 /** Bastion Shareable Link. */
 export interface BastionShareableLink {
   /** Reference of the virtual machine resource. */
-  vm: Vm;
+  vm: VM;
```
qiaozha, line 778: Guess this is looking correct?

test/integration/generated/datalakestorage/src/models/index.ts
qiaozha: look like this is also because Style.camel("IsCurrentVersion", true, {}, 3) gives us IsCurrentVersion. Really weird

ramya-rao-a:
@qiaozha The PR description is empty. Can you please update it with details of 
- links to issues you are resolving with this PR
- your approach to resolving those issues
- anything that the PR reviewer needs to know upfront before reviewing the PR

Both @sarangan12 and @joheredi are out this week. So, @deyaaeldeen will be reviewing this PR. Therefore, any context you can provide will be appreciated

ramya-rao-a:
@deyaaeldeen I believe the issues related to this PR are #1131 and #1134

qiaozha:
@deyaaeldeen This PR is trying to resolve issue #1131 #430 and #1134.
For issue #1131 and issue #430 @joheredi and I are trying to use Style.camel and Style.pascal  from @azure-tools/codegen library to fix issue like method name  JWTLogin will be convert to jWTLogin, but as I commentted above there's some weird behaviour and unexpected result of Style.camel. I am still trying to find out a solution, Could you provide any suggestions? 
For issue #1134, we are trying to merge the Client and ClientContext into one file. 

@ramya-rao-a @deyaaeldeen  I will split them into two PRs seperately, as the fix for issue #1134 are pretty straight forward. But the naming issue fix probably going to take some time. I think that fix can be done after the GA, because it only impact some corner cases and we don't want this impact the GA date. let me know if that works for you ?

ramya-rao-a:
Splitting the PR sounds like a good idea

test/integration/generated/nameChecker/src/operations/documents.ts
```diff
@@ -143,16 +143,16 @@ export class DocumentsImpl implements Documents {
    * Autocompletes incomplete query terms based on input text and matching terms in the index.
    * @param suggesterName The name of the suggester as specified in the suggesters collection that's part
    *                      of the index definition.
-   * @param search$DONotNormalize$Text The incomplete term which should be auto-completed.
+   * @param searchDONOTNormalizeText The incomplete term which should be auto-completed.
```
qiaozha, line 146: why there's a $ sign here before?

test/integration/generated/url2/src/operations/GET.ts
```diff
@@ -6,18 +6,18 @@
  * Changes may cause incorrect behavior and will be lost if the code is regenerated.
```
qiaozha, line 6: Does this file name look okay ?

test/integration/generated/xmlservice/src/models/index.ts
```diff
@@ -19,7 +19,7 @@ export interface RootWithRefAndNoMeta {
 /** I am a complex type with no XML node */
 export interface ComplexTypeNoMeta {
   /** The id of the res */
-  id?: string;
+  ID?: string;
```
qiaozha, line 22: Also this one.

test/integration/generated/lro/src/index.ts
deyaaeldeen: The new name feels better!

test/integration/generated/datalakestorage/src/models/index.ts
```diff
@@ -119,7 +119,7 @@ export interface BlobPropertiesInternal {
   accessTierChangeTime?: Date;
   tagCount?: number;
   expiresOn?: Date;
-  isSealed?: boolean;
+  IsSealed?: boolean;
```
qiaozha, line 122: Is this 'Is' the problem ?

test/smoke/generated/network-resource-manager/src/operations/virtualWans.ts
```diff
@@ -163,13 +163,13 @@ export class VirtualWansImpl implements VirtualWans {
    * Creates a VirtualWAN resource if it doesn't exist else updates the existing VirtualWAN.
    * @param resourceGroupName The resource group name of the VirtualWan.
    * @param virtualWANName The name of the VirtualWAN being created or updated.
-   * @param wANParameters Parameters supplied to create or update VirtualWAN.
+   * @param WANParameters Parameters supplied to create or update VirtualWAN.
```
qiaozha, line 166: This is also a good candidate

src/utils/nameUtils.ts
```diff
@@ -125,12 +127,17 @@ function getSuffix(nameType?: NameType) {
  */
 export function normalizeTypeName({ kind, typeName }: TypeDetails) {
   // Only Enum and Composite kinds need normalization
+  let tempTypeName = "";
+  if (typeName.endsWith("[]")) {
+    tempTypeName = "[]"
+    typeName = typeName.replace("[]", "");
+  }
```
deyaaeldeen, line 134: NIT:

`tempTypeName = "[]"` could be confusing. Could this be rewritten into a boolean variable instead that can be checked in line 136? e.g. `isArray = true;` and then 
```ts
return `${normalizeName(typeName, NameType.Interface)}${isArray ? "[]" : ""}`;
```

sarangan12:
@qiaozha I have discussed with Jose about this PR and here is the summary:

1. Currently, the naming changes are done by 2 systems - M4 & SDK Generator. 
2. Similarly, We could override the names using 2 methods - Swagger Overrides & M4 Overrides 
3. After the discussion, we have come to the agreement that there can always be cases which might not be handled only by the M4 Logic and M4 Overrides. So, it is not safe or in our advantage to remove the naming logic we have in our SDK Generator. 
4. The edge cases where the name is different now (from what we want) are rare and could be easily handled with a combination of Swagger Overrides & M4 Overrides. 
5. So, I would recommend not to proceed with this PR and close it. The only recommended change would be to remove `$DO_NOT_NORMALIZE$` and instead could use the modeler four overrides directly (as they could be read from the host). But, again it is not a mandatory or urgent change or a change that will produce any change in the generated code.

qiaozha:
close this PR based on @sarangan12 's comment. Thanks

PR #1251, opened 2021-11-22
[RLC] Support binary and form-data content

joheredi:
* Generates the correct types for multi-part body
* Generates correct type for binary request properties "string | Uint8Array"
* Generates correct type for binary response body "Uint8Array"

This PR depends on https://github.com/Azure/azure-sdk-for-js/pull/18753


test/integration/bodyFormDataRest.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+      87,
+      111,
+      114,
+      108,
+      100
+    ]);
+
+    const result = await client.path("/formdata/stream/uploadfile").post({
+      contentType: "multipart/form-data",
+      body: { fileContent: content, fileName },
```
qiaozha, line 29: `{ fileContent: content, fileName: fileName }` missed a fileName key ?

src/restLevelClient/schemaHelpers.ts
```diff
@@ -115,14 +115,17 @@ export function primitiveSchemaToType(
...
     case SchemaType.ByteArray:
-    case SchemaType.Binary:
     case SchemaType.Duration:
     case SchemaType.Credential:
     case SchemaType.UnixTime:
     case SchemaType.Uri:
     case SchemaType.Uuid:
     case SchemaType.String:
       return "string";
+    case SchemaType.Binary:
```
qiaozha, line 125: just wondering why SchemaType.ByteArray is not mapping to mediaType ?

qiaozha:
The CI is failing

test/integration/bodyFormDataRest.spec.ts
joheredi: If the key has the same name as a variable there is no need to add key value. For example

```typescript
let foo = "foo"

console.log({ foo: foo })
// output:
// {
//  "foo": "foo"
// } 


console.log({ foo})
// output:
// {
//  "foo": "foo"
// } 
```

package.json
```diff
@@ -45,6 +45,7 @@
         "@azure/logger": "^1.0.0",
         "@opentelemetry/api": "^0.10.2",
         "@types/lodash": "^4.14.149",
+        "cd": "^0.3.3",
```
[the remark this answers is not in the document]
joheredi: Yeah I'll remove this, not sure why it got in

src/restLevelClient/schemaHelpers.ts
```diff
@@ -115,14 +115,17 @@ export function primitiveSchemaToType(
     case SchemaType.Char:
       return "string";
     case SchemaType.ByteArray:
-    case SchemaType.Binary:
     case SchemaType.Duration:
     case SchemaType.Credential:
```
sarangan12, line 119: Should Credential be mapped to string?

test/integration/bodyFileRest.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+
+describe("BodyFile Client", () => {
+  let client: BodyFileRestClient;
+
+  beforeEach("create client", () => {
+    client = BodyFile({ allowInsecureConnection: true });
+  });
+
+  describe("node.js", () => {
+    it("should getFile", async () => {
```
sarangan12, line 12: Can you add a test case with empty file?

test/integration/bodyFormDataRest.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    const response = await client.path("/formdata/stream/uploadfile").put({
+      body: content,
+      binaryResponse: true,
+      contentType: "application/octet-stream"
+    });
+
+    if (response.status !== "200") {
+      const error = `Unexpected error ${response.status}`;
+      assert.fail(error);
+      throw error;
```
sarangan12, line 58: is this throw really required?

sarangan12 (approved):

src/restLevelClient/schemaHelpers.ts
joheredi: That's a goof point. I don't think we have yet any logic for special handling Credentials. We'd need to dig more into that

test/integration/bodyFormDataRest.spec.ts
joheredi: It is somehow for IntelliSense :(

PR #1252, opened 2021-11-22
Pin @autorest/extension package to patch versions

timotheeguerin:
That package doesn't follow semver as it need to have major version match autorest major version. this means current state pulls in the latest always with potential breaking changes


joheredi:
Discussed offline with @timotheeguerin and merged a change to update the generator to be compatible with the latest version. Closing this PR in favor of: https://github.com/Azure/autorest.typescript/pull/1253

PR #1253, opened 2021-11-22
Fix breaking changes from M4

joheredi:
Recently there was a breaking change introduced in @autorest/extension-base without a major version bump. This change updates the code to align with the latest API. Most of the changes are on method name casing and the shape of writeFile parameters

### DIFF
--- a/src/utils/autorestOptions.ts
+++ b/src/utils/autorestOptions.ts
@@ -1,4 +1,4 @@
-import { Channel, Host } from "@autorest/extension-base";
+import { Channel, AutorestExtensionHost } from "@autorest/extension-base";
 import { AutorestOptions, getHost, getSession } from "../autorestSession";
 import { TracingInfo } from "../models/clientDetails";
 import { PackageDetails } from "../models/packageDetails";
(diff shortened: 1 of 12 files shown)


src/utils/autorestOptions.ts
```diff
@@ -199,11 +221,11 @@ async function getPackageDetails(host: Host): Promise<PackageDetails> {
 }
 
 export async function getCredentialScopes(
-  host: Host
+  host: AutorestExtensionHost
 ): Promise<string[] | undefined> {
-  const addCredentials = await host.GetValue("add-credentials");
-  const credentialScopes = await host.GetValue("credential-scopes");
-  const azureArm = await host.GetValue("azure-arm");
+  const addCredentials = await host.getValue("add-credentials");
```
timotheeguerin, line 226: Maybe that's for a follow up PR if you care about it but the latest version just expose the configuration in the session object with `.configuration` so you can get it without await everywhere
```ts
session.configuration["add-credentials"]
```

timotheeguerin (approved):

deyaaeldeen (approved):
Looks great!

PR #1258, opened 2021-12-07
RLC shortcut methods

joheredi:
This PR introduces shortcut methods to RLC clients. These methods are meant to be in addition to the regular `path` and `pathUnchecked` methods already available.

These shortcut methods call internally `path(...).get` or the right combination.

For example the following calls would be equivalent

```typescript
await client.path("/foo/{fooId}", "1").post({ body: {name: "TestFoo" } });

// Shortcut method calls the path method above internally.
await client.createFoo("1", { body: {name: "TestFoo" } })
```

If Swagger groups methos in operation groups we'd produce the same structure

```typescript
// foo being the operation group name defined in the swagger
await client.foo.create("1", { body: {name: "TestFoo" } })

// In this example the swagger didn't define an operation group so the shortcut method lives in the client
await client.createFoo("1", { body: {name: "TestFoo" } })
```

Note: Generating shortcut methods is currently behind the flag `--rlc-shortcut` which will be removed upon discussion and approval of arch board.


src/restLevelClient/generateClient.ts
```diff
@@ -28,28 +24,16 @@ import { transformBaseUrl } from "../transforms/urlTransforms";
...
-  [key: string]: {
-    name: string;
-    pathParameters: PathParameter[];
-    methods: Methods;
-  };
-};
+import {
+  buildMethodDefinitions,
+  getOperationParameters,
+  getParhParamDefinitions
```
sarangan12, line 30: Is that a typo?

src/restLevelClient/generateMethodShortcuts.ts
```diff
@@ -0,0 +1,116 @@
...
+  MethodDeclarationStructure,
+  ObjectLiteralElement,
+  PropertyDeclaration,
+  PropertyDeclarationStructure
+} from "ts-morph";
+import { NameType, normalizeName, ReservedName } from "../utils/nameUtils";
+import { buildMethodDefinitions } from "./helpers/operationHelpers";
+import { Methods, PathParameter, Paths } from "./interfaces";
+
+const RESCT_CLIENT_RESERVED: ReservedName[] = [
```
sarangan12, line 14: Typo?

sarangan12 (approved):

src/restLevelClient/generateClient.ts
joheredi: Yup it definitely is! good catch

Issue #1057, opened 2021-06-23
[Rest Level Client] TestServer coverage for custom-baseurl


joheredi:
https://github.com/Azure/autorest.testserver/blob/main/swagger/custom-baseUrl.json

joheredi:
We need to define what's the UX we want to provide for this scenario where the BaseURL is parametrized:

It takes as a client parameter the `{host}` part and as a method parameter the `{accountName}` part.



In HCL we call it like this

```typescript
const client = new CustomURLClient("host:300");
await client.paths.getEmpty("local")
```


And ends up building http://localhost:3000 as base url.

In RLC I was thinking we can still take the `{host}` part in the client. And the {accountName} as a  `urlParameter` option in the get call


```typescript
const client = CustomUrl("host:3000");
await client.path("/customuri").get({urlParameters: { accountName: "local"} })
```

Maybe we could also add it as a required positional param, but it is not going to the path so it may not be evident where it goes and it may get a bit confusing when the path also has params

```typescript
const client = CustomUrl("host:3000");
await client.path("/customuri", "local").get()
```

joheredi:
@bterlson, @xirzec do you have any preference on the above. I think I like the urlParameters option.

xirzec:
I also like the `urlParameters` option, since it feels more explicit

bterlson:
I like the option approach, but I think `pathParameters` is a more recognizable name for this, if I'm not misunderstanding anything?

PR #1260, opened 2021-12-09
[RLC] add-custom-base-url-test-for-llc

qiaozha:
This PR is to add functionality test customBaseUrl from test server for RLC.


test/integration/customUrlRest.spec.ts
```diff
@@ -0,0 +1,16 @@
+import CustomUrlRestClient, { CustomUrlRestClientRestClient } from "./generated/customUrlRest/src";
+import { assert } from 'chai';
+
+describe("Custom Endpoint", () => {
+  let client: CustomUrlRestClientRestClient;
+  let clientOptions: any;
+  beforeEach(() => {
+    clientOptions = { host: "host:3000", allowInsecureConnection: true };
```
sarangan12, line 8: From what I see in the client options, there is no `host` property. We have only `baseUrl`. How are you setting this property? I think this  might be the reason the CI is failing with the error message:

```
TypeError [ERR_INVALID_URL]: Invalid URL: http://{accountName}[object Object]//customuri
```

Could you please check and fix it? Does this code run locally?

sarangan12 (changes requested):
nit: Please add a description on the issue detailing what and why the changes are done for this PR. Thanks

test/integration/customUrlRest.spec.ts
qiaozha: yeah, I am still trying to understand why

test/integration/customUrlRest.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+describe("CustomRest Endpoint", () => {
+  let client: CustomUrlRestClientRestClient;
+  let clientOptions: any;
+  beforeEach(() => {
+    clientOptions = { allowInsecureConnection: true };
+    client = CustomUrlRestClient("host:3000", clientOptions);
+  });
+
+  it("should return 200", async () => {
+    const result = await client.path("/customuri", 'local').get();
```
joheredi, line 13: According to the conversation in https://github.com/Azure/autorest.typescript/issues/1057

I'm expecting this call to look like this
```typescript
// Internally as defined in the swagger, the base url has parameters in this case the base url is: `{accountName}{suffix}`, 
// where suffix is a global parameter that we pass when building the client as above, and `accountName` is a method 
// parameter that we'll require as part of the operation options.
await client.path("/customuri").get({pathParameters: { accountName: "local"} })
```

For this we need to:

1. Add a new `pathParameters` property to `RequestParameters` in @azure-rest/core-client`
2. Update logic in `@azure-rest/core-client` to use `pathParameters` when building the url.
3. Generate the parameter types to include `accountName` as a pathParameter

For the changes in (2) you'd need to add logic before this loop https://github.com/Azure/azure-sdk-for-js/blob/a8b7f42993733e43a3dbaee89e66393df1b8bfa0/sdk/core/core-client-rest/src/urlHelpers.ts#L27

In which you loop the object options.pathParameters (added as part of (1)) to replace any occurrences of its keys in the URL with the value.

qiaozha: Got it. Thanks for info and instructions

qiaozha:
@joheredi @sarangan12 Could you help review this PR again, I have test it with https://github.com/Azure/azure-sdk-for-js/pull/19463 in my local and it works. Thanks

test/integration/customUrlRest.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+
+describe("CustomRest Endpoint", () => {
+  let client: CustomUrlRestClientRestClient;
+  let clientOptions: any;
+  beforeEach(() => {
+    clientOptions = { allowInsecureConnection: true };
+    client = CustomUrlRestClient("host:3000", clientOptions);
+  });
+
+  it("should return 200", async () => {
```
joheredi, line 12: Can you also add a test using the shortcut methods?

test/utils/test-swagger-gen.ts
```diff
@@ -265,6 +265,17 @@ const testSwaggers: { [name: string]: SwaggerConfig } = {
     addCredentials: false,
     isTestPackage: true
   },
+  customUrlRest: {
```
joheredi, line 268: Would you mind moving this to the bottom next to all the other REST Client tests?

joheredi (approved):
Looks good, left a couple of minor comments

test/integration/customUrlRest.spec.ts
qiaozha: added

test/utils/test-swagger-gen.ts
qiaozha: moved

qiaozha:
@sarangan12 @joheredi Could you take a look at this PR again ? Thanks

sarangan12 (approved):

PR #1264, opened 2021-12-15
How to resolve compliance violations

sarangan12:
This Change is to add instructions in the contributing document to resolve compliance violations reported in the JS SDK Repository.

@joheredi @ramya-rao-a  Please review and approve the PR

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -179,3 +179,10 @@ To debug code generation the flag &#x2013;typescript.debugger can be used
 - Make sure to re-build autorest.typescript after any changes you want to test
 
       npm run build
+
+### How to update dependencies in generated package.json files?
+
+Our generated SDKs take dependency on various packages which you can see in the generated package.json files. These will need to be upgraded from time to time to stay on the latest major version so that we get bug fixes automatically due to semver.
+
+- Update the version of the dependency you are looking for in the methods `restLevelPackage` and/or `regularAutorestPackage` in the [`packageFileGenerator.ts`](https://github.com/Azure/autorest.typescript/blob/main/src/generators/static/packageFileGenerator.ts) file.
+- Run `npm run build && npm run generate-swaggers && npm run smoke-tests` to update the generated files in the repo and run smoke tests before creating the PR.
--- a/test/integration/generated/additionalProperties/LICENSE
+++ b/test/integration/generated/additionalProperties/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2021 Microsoft
+Copyright (c) 2022 Microsoft
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/test/integration/generated/appconfiguration/LICENSE
+++ b/test/integration/generated/appconfiguration/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2021 Microsoft
+Copyright (c) 2022 Microsoft
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/test/integration/generated/appconfigurationexport/LICENSE
+++ b/test/integration/generated/appconfigurationexport/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2021 Microsoft
+Copyright (c) 2022 Microsoft
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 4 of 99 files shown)


docs/contributing.md
```diff
@@ -179,3 +179,12 @@ To debug code generation the flag &#x2013;typescript.debugger can be used
...
 
       npm run build
+
+### I've got compliance violations for the generated SDKs. How do I resolve it?
+
+There are CI Jobs which are setup to monitor the compliance violations in the SDK Generator repositories. For the JS SDK Repository, the CI Job could be found [here](https://dev.azure.com/azure-sdk/internal/_componentGovernance/101977/alert/5809940?typeId=7613604). Now, these violations should be resolved permanently. 
+
+Such dependencies are added to the `package.json` file in the [`packageFileGenerator.ts`](https://github.com/Azure/autorest.typescript/blob/main/src/generators/static/packageFileGenerator.ts). This file has 2 methods `restLevelPackage` & `regularAutorestPackage`. Determine which package is causing the violation and make the appropriate changes. For example, let us say `@opentelemetry/api` package of version `0.10.2` is causing a compliance violation. The solution is to upgrade the version to `1.0.3`. (Usually, such recommendations are found in the compliance issue itself) So, change the version and create a PR. Once the PR is merged & released, any and all SDKs generated with the new version will not have the compliance issue.
```
joheredi, line 187: Would it be worth mentioning that this change would require an `npm run build && npm run generate-swaggers && npm run smoke-tests` to regenerate all test swaggers?

joheredi (approved):

docs/contributing.md
```diff
@@ -179,3 +179,12 @@ To debug code generation the flag &#x2013;typescript.debugger can be used
 - Make sure to re-build autorest.typescript after any changes you want to test
 
       npm run build
+
+### I've got compliance violations for the generated SDKs. How do I resolve it?
```
ramya-rao-a, line 183: I would recommend having this section talk about dependency updates in general and not in the context of compliance violations. In the azure-sdk-for-js repo we now have automation looking for outdated dependencies and logging [issues with label dependency-upgrade-required](https://github.com/Azure/azure-sdk-for-js/issues?q=is%3Aopen+is%3Aissue+label%3Adependency-upgrade-required). If any of these are part of the generated code, then we will be opening issues in the autorest.typescript repo to update the dependency in the generated code. I would like to add help-wanted label on such issues in the code gen repo and point to the contributing guide as to how to go about making the change.

So, think of a section like `How to update dependencies in generated package.json files?".

docs/contributing.md
```diff
@@ -179,3 +179,11 @@ To debug code generation the flag &#x2013;typescript.debugger can be used
...
+
+Now, there is a scenario where there will be a compliance violation and there is no such dependency entry in the SDK generator. But, the SDKs in the JS Repository will be having that dependency. For example, there was a violation reported for the `nyc` package. The issue was resolved through this [PR](https://github.com/Azure/azure-sdk-for-js/pull/19248). But, there is no dependency entry for `nyc` in the [`packageFileGenerator.ts`](https://github.com/Azure/autorest.typescript/blob/main/src/generators/static/packageFileGenerator.ts). It means that entry has been made manually and not part of the SDK generator. In such cases, you do not have to do any code changes in the generator. (It is sufficient to do the changes manually in the JS Repo) But, Please raise the absence of dependency as an issue in the repository. This will help the Generator owners to consider adding that dependency to the [`packageFileGenerator.ts`](https://github.com/Azure/autorest.typescript/blob/main/src/generators/static/packageFileGenerator.ts).
```
ramya-rao-a, line 189: When we need to upgrade a dependency to use a new feature, there will be many other changes in the generated code as well. Anyone doing that will be knowing how to update the version in generated package.json file as well. So, we don't need to mention that particularly here. I would also stay away from all talks of compliance violations for the same reasons I mentioned in https://github.com/Azure/autorest.typescript/pull/1264#discussion_r773401058




```suggestion
Our generated SDKs take dependency on various packages which you can see in the generated package.json files. These will need to be upgraded from time to time to stay on the latest major version so that we get bug fixes automatically due to semver.

Such dependencies are added to the `package.json` file in the [`packageFileGenerator.ts`](https://github.com/Azure/autorest.typescript/blob/main/src/generators/static/packageFileGenerator.ts). Find the dependency you are trying to update in either of the 2 methods `restLevelPackage` & `regularAutorestPackage` here and make the change. Run `npm run build && npm run generate-swaggers && npm run smoke-tests` before creating the PR.
```

test/integration/generated/additionalProperties/LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2021 Microsoft
+Copyright (c) 2022 Microsoft
```
ramya-rao-a, line 3: Why are these changes in this PR? Can we move it to another PR instead?

JonathanCrd: I also have these changes on a PR I opened today, looks like the first PR of the year will have this when regenerating the tests swaggers.

sarangan12: This is the result of regeneration.

Issue #1089, opened 2021-06-30
[Track2]Cannot generate expected codes for Resources which includes batch task：the generated codes are the same


dw511214992:
RP resources contains many packages in [readme.typescript.md](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/resources/resource-manager/readme.typescript.md):
```
  - package-resources: true
  - package-features: true
  - package-locks: true
  - package-policy: true
  - package-links: true
  - package-managedapplications: true
```
When I generate codes for them, all the generated codes of diferent package are same.
To reproduce it, please follow the following steps (You also can checkout [my branch](https://github.com/Azure/azure-rest-api-specs/tree/dw/resource-test) which already contains the changes in step1 and 2):
1. Remove the `profile` in `readme.typescript.md` because of https://github.com/Azure/autorest/issues/4013:
![image](https://user-images.githubusercontent.com/40835867/123928601-a4fe2780-d9c0-11eb-8f20-edd5b62bed82.png)

2. Delete some batch tasks in `readme.typescript.md` because it's time consuming to generate codes for all packages:
![image](https://user-images.githubusercontent.com/40835867/123928772-c9f29a80-d9c0-11eb-85e3-b335b10e01e0.png)

3. Run autorest command:
```
autorest --version=3.1.3 --typescript --typescript-sdks-folder=D:/projects/gener
ated ../azure-rest-api-specs/specification/resources/resource-manager/readme.md --use=@autorest/typescript@
6.0.0-beta.5 --debug
```
Then you will find the generated codes in folder `locks` are the same as the ones in `features`.

This is not a blocking issue for preview, but we hope it can be fixed in GA because it blocks our swagger pipeline.
@ramya-rao-a @sarangan12

PR #1265, opened 2021-12-16
fix-typescript-batch-task-generation-issues

qiaozha:
to fix issue #1089 and prepare for multi-client generation for RLC


joheredi (approved):
Thanks for fixing this. I didn't realize host, session and options would be kept the same for every run in the batch.

PR #1267, opened 2021-12-17
Generate multi clients for rlc

qiaozha:


src/generators/indexGenerator.ts
```diff
@@ -24,21 +24,82 @@ export function generateIndexFile(
       );
     }
     generateHLCIndex(clientDetails, indexFile);
-  } else {
+  } else if (!batch || batch?.length === 1) {
```
joheredi, line 27: Can you add a comment that this is the case where a single client is being generated?

src/generators/indexGenerator.ts
```diff
@@ -24,21 +24,82 @@ export function generateIndexFile(
...
     generateRLCIndex(indexFile);
+  } else {
+    generateRLCIndexForMultiClient(indexFile);
   }
 }
 
-function generateRLCIndex(file: SourceFile) {
+function generateRLCIndexForMultiClient(file: SourceFile) {
```
joheredi, line 34: Can you add a comment about what this is doing? For future reference.

src/restLevelClient/generateRestLevel.ts
```diff
@@ -10,23 +10,34 @@ import { generateSchemaTypes } from "./generateSchemaTypes";
...
 import { hasPagingOperations } from "../utils/extractPaginationDetails";
 import { hasPollingOperations } from "./helpers/hasPollingOperations";
+import { NameType, normalizeName } from "../utils/nameUtils";
 import { generateKarmaConfigFile } from "../generators/static/karmaConfigFileGenerator";
 import { generateEnvFile } from "../generators/test/envFileGenerator";
 import { generateEnvBrowserFile } from "../generators/test/envBrowserFileGenerator";
 import { generateRecordedClientFile } from "../generators/test/recordedClientFileGenerator";
 import { generateSampleTestFile } from "../generators/test/sampleTestGenerator";
+
+const batchOutputFolder: [string, string, string][] = [];
```
joheredi, line 28: Can you move this inside `generateRestLevelClient`? Having it as global can bring us some hard to find issues in the future

src/restLevelClient/generateRestLevel.ts
```diff
@@ -10,23 +10,34 @@ import { generateSchemaTypes } from "./generateSchemaTypes";
...
 /**
  * Generates a Rest Level Client library
  */
 export async function generateRestLevelClient() {
   const host = getHost();
   const { model } = getSession();
+  const { batch, srcPath } = getAutorestOptions();
+  if (srcPath) {
```
joheredi, line 36: it feels like this should be inside your if block in line 76? 

Would it be possible to move this check and the check in line 76 inside the `generateTopLevelIndexFile` function? If the conditions are not met `generateTopLevelIndexFile` does nothing. This will keep this function cleaner

test/integration/generated/bodyComplexRest/src/index.ts
```diff
@@ -1,12 +1,10 @@
...
-
 export * from "./bodyComplexRestClient";
 export * from "./parameters";
 export * from "./responses";
+export * from "./clientDefinitions";
 export * from "./models";
 export * from "./outputModels";
-
-export default BodyComplexRestClient;
+export { BodyComplexRestClient } from "./bodyComplexRestClient";
```
joheredi, line 10: If we are generating a single client, we should keep the defatult export. Named exports should only happen when generating a multi client package

joheredi (commented):
It looks good overall. The only thing that I think is blocking this PR is the change from default to named exports when generating a single client package. We should keep default export for single client packages and have named exports for multi-client packages

src/restLevelClient/generateRestLevel.ts
qiaozha: Actually we can't this variable is suppose to be global level so that it won't get override while running the next batch task.

joheredi (approved):
Great work! :shipit:

PR #1268, opened 2021-12-18
[RLC] Introduce version tolerance tests

joheredi:
This PR adds tooling to support version tolerance validation for our Rest Clients. Version tolerance will provide coverage against unexpected Breaking changes when generating a newer version of the swagger. RLCs are not expected to produce breaking changes, unless the swagger itself has an API breaking change.

This change includes the following npm scripts:

* `npm run validate-version-tolerance`: generates SDKs and tests for breaking changes between swagger versions
* `generate-version-tolerance`: Generates version tolerance before and after sdks
* `generate-version-tolerance:tests`: generates `updated` spec.ts based on hand written tests against the initial SDK
* `test-version-tolerance`: executes all .spec.ts tests under `test/version-tolerance/`

In CI we'll be running `npm run validate-version-tolerance` which calls all the other npm scripts. The other individual npm scripts are useful for development.

**Context:**
 
 This tool helps us run Rest Client Version Tolerance validation. With this we have coverage
 against unexpected SDK breaking changes. Rest Clients are not expected to create any breaking changes
 if the new swagger doesn't contain API breaking changes.

 Version tolerance tests are configured in the variable versionToleranceTests. Where we define the pair of
 swaggers to generate as `initialSwagger` and `updatedSwagger`

 The tool will generate the 2 swaggers.

 We need to hand author tests against the initialSwagger. When adding a new versionTolerance swagger, a new
 test needs to be created in test/version-tolerance/initial-tests/{name}.spec.ts.

 NOTE: The test file name must be the same name defined in the `versionToleranceTests` object below.

 We only need to author test for the initial swagger, and the tool will auto-generate a test against the updated swagger
 taking the test we write and changing the import paths to point to the updated generated sdk.

 When this tool is invoked without cli parameters, it will generate all the client pairs defined below in `versionToleranceTests`.
 If --test-gen is passed, the tool will generate a spec.ts file for each file under `test/version-tolerance/initial-tests`
 change the imports to point to the updated version of the sdks and put the new tests under `test/version-tolerance/updated-tests`

 Running all tests under `test/version-tolerance` should pass. If a test failure is reported in any of the `updated-tests` it is
 likely because of an unexpected breaking change.


test/version-tolerance/initial-tests/rlcClient.spec.ts
```diff
@@ -0,0 +1,47 @@
+import RLCClient, { LLCClientRestClient } from "../generated/llc-initial/src";
+import { assert } from "chai";
+
+const phase = "initial";
+
+describe(`RLC Version Tolerance ${phase} phase`, async () => {
```
qiaozha, line 6: will the inital-tests/rlcClient.spec.ts and updated-test/rlcClient.specs.ts be mostly the same for non breaking change scenarios ?

qiaozha (approved):

test/version-tolerance/initial-tests/rlcClient.spec.ts
joheredi: Yes they are the same, the only thing that changes is the import. This way if there is a breaking change the updated test will fail

Issue #1156, opened 2021-08-13
The package.json file should have metadata required by the automation to increment package version after release


ramya-rao-a:
In the azure-sdk-for-js repo, we have automated PRs that increment the package version after an update for the package is released. This automation updates the version in package.json and changelog files by default. Any other places that hard codes the package version needs to be listed as metadata in the package.json. For example: https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/keyvault/keyvault-keys/package.json#L76-L89

https://github.com/Azure/azure-sdk-for-js/pull/16903 is one such PR for the newly generated mgmt plane package that used the v6 code generator. It missed updating the package version at https://github.com/Azure/azure-sdk-for-js/blob/308a805d8f58addfb06cfcf6e225e58fa6bd721a/sdk/sql/arm-sql/src/sqlManagementClientContext.ts#L44

This issue is a feature request to update the v6 code generator to add the metadata section in the pacakge.json file pointing to this file and the package version that needs to be updated here. For example:

```json
"//metadata": {
    "constantPaths": [
     {
        "path": "src/sqlManagementClientContext.ts",
        "prefix": "packageDetails"
      }
    ]
  }
```

JonathanCrd:
File `src/sqlManagementClientContext.ts` does not longer exists, now the `packageDetails` constant is inside `src/sqlManagementClient.ts`.

I'm wondering if there are other places where there's a version specified. From what I found after a little research, it looks like `sqlManagementClient.ts` is the only one.

ramya-rao-a:
The other place is the swagger/readme.md file, but that is not part of the generated code

PR #1278, opened 2022-01-05
Added metadata needed to automatically increment package version after release

JonathanCrd:
Solves https://github.com/Azure/autorest.typescript/issues/1156.

Added the following lines in package files generator:
```
"//metadata":{
      constantPaths: [{
        path: `src/${clientDetails.name}.ts`,
        prefix: "packageDetails",
      }],
    },
```
In order to have an output like this:
```
"//metadata": {
    "constantPaths": [
      {
        "path": "src/AdditionalPropertiesClient.ts",
        "prefix": "packageDetails"
      }
    ]
  },
```

Integration and smoke tests were regenerated.

This change only applies to regular autorest packages, not for REST packages.


test/integration/generated/appconfiguration/package.json
```diff
@@ -81,5 +81,10 @@
     "docs": "echo skipped"
   },
   "sideEffects": false,
+  "//metadata": {
+    "constantPaths": [
+      { "path": "src/AppConfigurationClient.ts", "prefix": "packageDetails" }
```
sarangan12, line 86: The code changes look good. But, I have a question. Why is it formatted in one file and not formatted to two lines in other files?

JonathanCrd: It might be caused by VScode. I'm actually not sure, all of these files are generated so I didn't touch them.
Going to take a look locally to see if this was caused by my dev environment.

JonathanCrd: After some tests and research, I found that this is done by Prettier, it will try to put all elements in a single line only if they fit with the default print width, as it written [here](https://prettier.io/docs/en/rationale.html#multi-line-objects). So, paths being larger than others is causing this discrepancy between the generated files.

sarangan12 (approved):
LGTM

PR #1289, opened 2022-01-25
Fix sample generation for several mgmt plane RPs

qiaozha:
This PR is trying to fix several issues for mgmt plane RPs.
1. We are not handle number/boolean type constant or single value number/boolean enum correctly. 
2. In sample generation, if we don't add type cast logic, it will complain about imcompatible issue, 
3. In sample generation, iwe don't add the logic for reserved words in the parameter,  For example type => typeParam, default => defaultParam


joheredi:
Can you please add a description about what's the problem and how are you fixing it? It is very difficult to make sense of a PR without proper description

qiaozha:
@joheredi I have updated the description for this PR. Could you take a look again ?

package.json
```diff
@@ -15,7 +15,7 @@
         "integration-test:new": "npm-run-all start-test-server generate-and-test integration-test:alone stop-test-server",
         "integration-test:browser": "npm run build:test:browser && karma start karma.conf.js",
         "generate-and-test": "npm-run-all -s build generate-swaggers integration-test:alone integration-test:browser",
-        "integration-test:alone": "mocha -r ts-node/register --timeout 2000 './test/integration/**/!(sampleTest).spec.ts'",
+        "integration-test:alone": "mocha -r ts-node/register --timeout 20000 './test/integration/**/!(sampleTest).spec.ts'",
```
joheredi, line 18: Why do we need to increase this timeout so dramatically?

test/smoke/generated/network-resource-manager/samples-dev/virtualWanCreate.ts
```diff
@@ -24,7 +24,7 @@ async function virtualWanCreate() {
   const subscriptionId = "subid";
   const resourceGroupName = "rg1";
   const virtualWANName = "wan1";
-  const WANParameters: VirtualWAN = {
+  const wANParameters: VirtualWAN = {
```
joheredi, line 27: What's up with this name, it doesn't look like something we'd want right?

test/smoke/generated/network-resource-manager/samples-dev/startPacketCaptureOnVpnGatewayWithFilter.ts
```diff
@@ -28,7 +28,9 @@ async function startPacketCaptureOnVpnGatewayWithFilter() {
     filterData:
       "{'TracingFlags': 11,'MaxPacketBufferSize': 120,'MaxFileSize': 200,'Filters': [{'SourceSubnets': ['20.1.1.0/24'],'DestinationSubnets': ['10.1.1.0/24'],'SourcePort': [500],'DestinationPort': [4500],'Protocol': 6,'TcpFlags': 16,'CaptureSingleDirectionTrafficOnly': true}]}"
   };
-  const options = { parameters: parameters };
+  const options = {
+    parameters: parameters as VpnGatewayPacketCaptureStartParameters
```
joheredi, line 32: Why is this cast needed? is there any way to make the types compatible without forcing a cast?

package.json
qiaozha: removed. The reason I added it is because the ci keep failing earlier.

test/smoke/generated/network-resource-manager/samples-dev/startPacketCaptureOnVpnGatewayWithFilter.ts
qiaozha: Actually this is to mainly fix issues like this 
```
enum A {
    "AAA",
    "BBB"
};
const aa: A = "AAA";
console.log(aa);
```
Then the typescript compiler will complain about Type '"AAA"' is not assignable to type 'A'. And since we get the value from swagger example, we can only get the string type. I also don't think adding cast logic everywhere look like a good idea. just wondering what's the better way of resolve it?

joheredi: I'm not sure I follow the relationship of this change with enums. The cast should no be needed in this case or in any other instances this cast appears in this PR.

This line already makes parameters be type `VpnGatewayPacketCaptureStartParameters` so the cast is not needed.
```typescript
const parameters: VpnGatewayPacketCaptureStartParameters = {
```

Also please generate this line as 

```diff
- const options = {
-    parameters: parameters as VpnGatewayPacketCaptureStartParameters
- }
+ const options = { parameters }
```

joheredi (commented):
I'll defer the approval to @sarangan12, from my side, I'd like the unnecessary cast to be removed.

test/smoke/generated/network-resource-manager/samples-dev/startPacketCaptureOnVpnGatewayWithFilter.ts
qiaozha: @joheredi I have updated the code to remove the casting usage in our samples and instead I add a type declaration for options parameter. which could also resolve the case I am facing with vmwarecloudsimple. 
You may use the following command with autorest.typescript main branch to repro the original issue if you are interested
```
autorest --version=3.7.3 --typescript --modelerfour.lenient-model-deduplication --head-as-boolean=true --license-header=MICROSOFT_MIT_NO_VERSION --generate-test --typescript-sdks-folder=. ../azure-rest-api-specs/specification/vmwarecloudsimple/resource-manager/readme.md --use=../autorest.typescript --generate-sample=true 
```

test/smoke/generated/storage-resource-manager/samples-dev/acquireALeaseOnAShare.ts
```diff
@@ -32,7 +33,7 @@ async function acquireALeaseOnAShare() {
     leaseId: undefined,
     proposedLeaseId: undefined
   };
-  const options = { parameters: parameters };
+  const options: FileSharesLeaseOptionalParams = { parameters: parameters };
```
joheredi, line 36: I would like this to be like this instead, `{parameters: parameters}` is unnecessary

```typescript
const options: FileSharesLeaseOptionalParams = { parameters: parameters };
```

qiaozha: changed

qiaozha:
I think all the comments from @joheredi should have been resolved. 
@sarangan12 Could you also take a look ? Thanks

sarangan12 (approved):

src/generators/static/samples.ts.hbs
```diff
@@ -8,8 +8,10 @@
  * x-ms-original-file: {{originalFileLocation}}
  */
 import {
-{{#if hasBody }}
-  {{bodySchemaName}},
```
MaryGao, line 12: If we don't use these two columns does that mean we could delete them in model and transform.ts?

src/transforms/samplesTransforms.ts
```diff
@@ -18,6 +18,8 @@ import { calculateMethodName } from "../generators/utils/operationsUtils";
 import { camelCase } from "@azure-tools/codegen";
 import { OperationGroupDetails } from "../models/operationDetails";
 import { getPublicMethodName } from '../generators/utils/pagingOperations';
+import { BodiedNode } from "ts-morph";
```
MaryGao, line 21: No usage delete this import?

src/transforms/samplesTransforms.ts
```diff
@@ -117,22 +122,31 @@ export async function getAllExamples(codeModel: TestCodeModel, clientDetails: Cl
...
             if (methodParameter.parameter.protocol?.http?.["in"] === "body") {
               sample.hasBody = true;
-              sample.bodySchemaName = getLanguageMetadata(
-                methodParameter.exampleValue.schema.language
-              ).name;
+              sample.bodySchemaName = parameterTypeName;
+              sample.importedTypes?.push(parameterTypeName);
+              if (methodParameter.exampleValue.schema.type === SchemaType.AnyObject || methodParameter.exampleValue.schema.type  === SchemaType.Any) {
+                sample.bodySchemaName = "Record<string, unknown>"
+                sample.isAnyTypeBody = true;
```
MaryGao, line 148: We re-correct the value of bodySchemaName & isAnyTypeBody but we didn't put the latest into importedTypes list or any usage directly, do we need to refine this?

PR #1290, opened 2022-01-25
decouple multiclient and batch

dw511214992:


joheredi:
Is there a readme.md file that runs into this issue? I would like to reproduce the issue locally. Would you mind sharing the swaggers and autorest commands to reproduce?

In general I won't like the idea of assuming from the path, I would prefer to set this up declaratively instead. Adding extra flags to link together the swaggers that are meant to be in the same client.

dw511214992:
@joheredi You can find the reason of this change in email "Propose to Move README.md to Swagger Repository". You can use this [readme.md](https://github.com/Azure/azure-rest-api-specs/blob/dw/llc-readme/specification/purview/data-plane/readme.md) to have a try. The command I use in my local is:
```
autorest --typescript --use=/home/weidong/Documents/projects/autorest.typescript --typescript-sdks-folder=/home/weidong/Documents/projects/azure-sdk-for-js ../azure-rest-api-specs/specification/purview/data-plane/readme.md
```

About setting multi-client declaratively, it's a very good idea, and I prefer it too. Is there a example that what flag you will use and how you link the swaggers?

PR #1292, opened 2022-01-26
remove unused import in clientDefinitions

dw511214992:
fix https://github.com/Azure/autorest.typescript/issues/1293

Issue description: The generated `clientDefinitions` contains credential related import, which is not used in the generated file `clientDefinions.ts`. The unused import leads to build failure when we build it by using azure-sdk-for-js 's tsconfig.

### DIFF
--- a/src/restLevelClient/generateClientDefinition.ts
+++ b/src/restLevelClient/generateClientDefinition.ts
@@ -116,14 +116,6 @@ export function generatePathFirstClient(model: CodeModel, project: Project) {
   });
 
   const clientName = getLanguageMetadata(model.language).name;
-  const uriParameter = getClientUriParameter();
-
-  const { addCredentials, credentialKeyHeaderName } = getAutorestOptions();
-  const credentialTypes = addCredentials ? ["TokenCredential"] : [];
-
-  if (credentialKeyHeaderName) {
-    credentialTypes.push("KeyCredential");
-  }
 
   const clientIterfaceName = `${clientName}RestClient`;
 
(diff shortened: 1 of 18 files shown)


joheredi:
Would you mind adding a description about what's causing the issue and the approach you took to fix the problem please?
Without this context, it is hard to make sense of the changes. For example, the first thing I see is removing some logic to check if there is clientUrl parameters. Without context I'm not sure how that's related to the issue

src/restLevelClient/generateClientDefinition.ts
```diff
@@ -116,14 +116,6 @@ export function generatePathFirstClient(model: CodeModel, project: Project) {
   });
 
   const clientName = getLanguageMetadata(model.language).name;
-  const uriParameter = getClientUriParameter();
```
joheredi, line 119: How is this related to the unused import?

src/restLevelClient/generateClientDefinition.ts
```diff
@@ -116,14 +116,6 @@ export function generatePathFirstClient(model: CodeModel, project: Project) {
...
-
-  const { addCredentials, credentialKeyHeaderName } = getAutorestOptions();
-  const credentialTypes = addCredentials ? ["TokenCredential"] : [];
-
-  if (credentialKeyHeaderName) {
-    credentialTypes.push("KeyCredential");
-  }
```
joheredi, line 126: Are we no longer supporting KeyCredential?

src/restLevelClient/generateClientDefinition.ts
dw511214992: This variable is not used in generating `clientDefinitions`. So I delete it.

src/restLevelClient/generateClientDefinition.ts
dw511214992: No. I delete it in generating `clientDefinitions`, which doesn't need to import credential related packages.

dw511214992:
> Would you mind adding a description about what's causing the issue and the approach you took to fix the problem please? Without this context, it is hard to make sense of the changes. For example, the first thing I see is removing some logic to check if there is clientUrl parameters. Without context I'm not sure how that's related to the issue

@joheredi I have added the context the the PR decription, and please review it. thanks

joheredi (approved):

PR #1298, opened 2022-01-31
Update rollup configuration for Azure SDK packages

witemple-msft:
Given that we recently migrated all packages to use Rollup 2, we need to also update the configuration here so that freshly-generated track 2 packages get it as well.

`namedExports` is no longer required, so we can shorten the default config as well. Even better would be if we could leverage dev-tool's default script for packages that are going to be part of the Azure SDK for JS monorepo.

I updated the static generators, but I could use some pointers about how to update tests/expectations etc. and validate that the implementation works and is free of warnings on a real spec.


joheredi:
> Given that we recently migrated all packages to use Rollup 2, we need to also update the configuration here so that freshly-generated track 2 packages get it as well.
> 
> `namedExports` is no longer required, so we can shorten the default config as well. Even better would be if we could leverage dev-tool's default script for packages that are going to be part of the Azure SDK for JS monorepo.
> 
> I updated the static generators, but I could use some pointers about how to update tests/expectations etc. and validate that the implementation works and is free of warnings on a real spec.

@witemple-msft you would need to run the following

> npm run clone:specs
npm run generate-swaggers
npm run smoke-test

This would regenerate all our tests. It takes quite a bit of time, so I went ahead and re-generated it for you. If you are interested to try it out feel free to run the commands above and I'll be happy to help if you have any questions

joheredi (approved):
Thanks for sending out this PR! Changes look good!

deyaaeldeen (approved):
🚀

witemple-msft:
Thank you @deyaaeldeen and @joheredi!

Issue #1276, opened 2022-01-04
There are 2 contributing guides in this repo


ramya-rao-a:
There are 2 contributing guides in this repo
- https://github.com/Azure/autorest.typescript/blob/main/docs/contributing.md
- https://github.com/Azure/autorest.typescript/blob/main/CONTRIBUTING.md

sarangan12:
Closing this issue with PR https://github.com/Azure/autorest.typescript/pull/1307

PR #1307, opened 2022-02-10
Merge Contributing Guides

sarangan12:
The TS Generator Repository has 2 contributing guides:

- https://github.com/Azure/autorest.typescript/blob/main/docs/contributing.md
- https://github.com/Azure/autorest.typescript/blob/main/CONTRIBUTING.md

This PR merges the two contributing guides to one and removes one of them. 

Associated with the Issue https://github.com/Azure/autorest.typescript/issues/1276


CONTRIBUTING.md
```diff
@@ -86,6 +87,14 @@ npm run build
 npm run test
 ```
 
+Running the command above will do the following things:
```
joheredi, line 90: We now have a new set of tests `Version Tolerance Tests` for Rest Clients. I think you can add them to the list above and briefly describe them in their own section like this:

Version tolerance tests provide coverage against unexpected **breaking changes** when generating a newer version of a swagger. RLCs are not expected to produce breaking changes unless the swagger itself has an API breaking change.

Version Tolerance tests would generate 2 clients one from an initial Swagger and a second from the `updated` swagger that contains changes that are not expected to generate breaking changes. The same set of tests is run against both generated clients to make sure no breaking changes resulted.

 Version Tolerance tests have the following npm scripts:

* `npm run validate-version-tolerance`: generates SDKs and tests for breaking changes between swagger versions
* `generate-version-tolerance`: Generates version tolerance before and after sdks
* `generate-version-tolerance:tests`: generates `updated` spec.ts based on hand written tests against the initial SDK
* `test-version-tolerance`: executes all .spec.ts tests under `test/version-tolerance/`

In CI we'll run `npm run validate-version-tolerance` which calls all the other npm scripts. The other individual npm scripts are useful for development.

For more details on the implementation see https://github.com/Azure/autorest.typescript/pull/1268

joheredi (approved):
Added a comment with info to add for Version Tolerance Tests

CONTRIBUTING.md
```diff
@@ -103,6 +113,23 @@ fatal: destination path './.tmp/specs' already exists and is not an empty direct
 
 delete the `.tmp` folder and then try again.
 
+### Version Tolerence Tests
```
joheredi, line 116: ❤️

joheredi (approved):
🚀

PR #1317, opened 2022-02-17
Fix sample filename fix in codegen

MaryGao:
This PR is trying to fix filename conflicts when there're duplicate example keys across different operationIds as swagger validation can only make sure they are identical within one operation. we are fixing this issue by changing the generated sample filename from the swagger example key to a normalized operationId with a suffix `Sample`. And for those operations with multiple examples, we will group them in the same file and each example will be generated as an independent function with the name of example key as before.


qiaozha:
please submit the related smoke test change. Thanks

qiaozha:
@joheredi @sarangan12  This change mostly looks good to me now. Can you also take a look at this PR ? The purpose of this PR is updated in the PR description. let us know if you need more information. Thanks

qiaozha (approved):

sarangan12 (approved):

joheredi (approved):

Issue #797, opened 2020-11-21
Test generated code in other node versions


joheredi:
Have a test matrix to make sure that the generated code works across node versions

ramya-rao-a:
@joheredi Let's make sure we do this for v4 of the code generator as well as it powers all the management plane libraries at the moment

joheredi:
We already test Node 10, 12 and 14 in CI.  We may need to update to include node 16.
@sarangan12 do you think we still need to port these to v4x as well, now that management is migrating to v6?

ramya-rao-a:
@joheredi, At this point, we can skip v4 as mgmt plane packages will move to v6 in the next few months.

@sarangan12, We can remove Node.js 10 now as it is out of support scope now. And we should add Node.js 16

PR #1318, opened 2022-02-17
Add Node 16 to test matrix and remove Node 10

JonathanCrd:
#### Related issues
- https://github.com/Azure/autorest.typescript/issues/797

#### Description
This PR removes Node 10 from the test matrix since it is no longer supported and adds Node 16. This changes only apply to v6 since mgmt. plane packages already moved or are moving out from v4.

It also removes the folder `eng\pipelines\` and all its contents since this is a duplicate of `scripts\` and is no longer in used.


JonathanCrd:
The changed `eng/pipelines/ci.yml` seems not to be the one that controls the CI, this folder seems to be a duplicate of `.scripts`.

Just committed a change in the file `.scripts/ci.yml` to test this. If CI runs the test for Node 16, I'll be removing the folder `eng/pipelines/` since it would be a duplicate.

JonathanCrd:
@joheredi @sarangan12 this PR is ready for review now :)

joheredi (approved):

Issue #1272, opened 2021-12-27
[RLC] Use title as client type


dw511214992:
I think we should change the client type to the value of `title`, instead of `{value_of_title} + RestClient`. The reasons of doing this change are:

- It is different from other language as they use `{value_of_title}` as client type directly. We should keep consistent with them.
- The value of  `title` is the name of a function which creates client. Most times, the value of `title` is in pascal, which doesn't meet the specification of Typescript code style. For example, [PurviewScanning](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/purview/purview-scanning-rest/src/purviewScanning.ts#L473).

Suggested Changes:
- Change the client type to the value of `title`.
- Change the name function, which creates client, to `create + {value_of_title}`.

joheredi:
Thanks for sharing your thoughts @dw511214992. In general, I like to use Pascal casing for these client factory functions as it looks cleaner for example

```typescript
const client = PurviewScanning();

//vs

const client = createPurviewScanning();
```

It seems to be common in the JavaScript community to Pascal case these functions, and also would feel more familiar for users familiar with classes, just without the `new`.

Though, this wouldn't impact end-users because they are the default exports, so they can choose how they want to import.

```typescript
import NameOfYourLiking from "@azure-rest/purview-scanning"
```

The way we write documentation and samples would likely inspire our customers.

@bterlson my suggestion is that we keep them as we have today and keep documenting them this way, what do you think?

```typescript
import PurviewScanning from "@azure-rest/purview-scanning"
```

dw511214992:
If we still use the `title` as the client factory name, I think it's better to trim `Client` because it's very strange if the `title` ends with `Client`.
For example, if the value of `title` is `FooClient`, then the client name will be `FooClientRestClient` and the client factory name will be `FooClient`. What do you think about it? @joheredi

joheredi:
@qiaozha do you remember if we added the `Client` suffix recently? I think it should be fine to not suffix with `Client`. @qiaozha do you know if there would be any issues around multi client if we do this?

qiaozha:
> @qiaozha do you remember if we added the `Client` suffix recently? I think it should be fine to not suffix with `Client`. @qiaozha do you know if there would be any issues around multi client if we do this?

I don't remember we added this suffix, I think it's fine to remove it if we did.

joheredi:
[...]

PR #1323, opened 2022-02-22
[RLC] fix-rest-level-client-title-issue

qiaozha:
to fix issue https://github.com/Azure/autorest.typescript/issues/1272


src/generators/indexGenerator.ts
```diff
@@ -36,6 +36,7 @@ export function generateIndexFile(
 function generateRLCIndexForMultiClient(file: SourceFile) {
   const { model } = getSession();
   const clientName = model.language.default.name;
+  const createClientFuncName = `create${normalizeName(clientName, NameType.Class)}`;
```
joheredi, line 39: I would like to keep the factory function name as we had before but without the "Client" suffix. For example `PurviewCatalogClient` -> `PurviewCatalog` (not createPurviewCatalog). It is a common convention to have these factory functions following the pascal casing similar to calling a constructor.

joheredi: If there is a collision with the interface name we could name the interface `PurviewCatalogLike` which seems to be a common convention in TypeScript

src/transforms/samplesTransforms.ts
```diff
@@ -139,13 +139,16 @@ export async function getAllExamples(codeModel: TestCodeModel, clientDetails: Cl
             );
             const parameterTypeName = parameterTypeDetails.typeName;
             let paramAssignment = "";
+            let isAnyTypeBody = false;
             if (methodParameter.parameter.protocol?.http?.["in"] === "body") {
               sample.hasBody = true;
-              sample.bodySchemaName = parameterTypeName;
-              sample.importedTypes?.push(parameterTypeName);
+              sample.bodySchemaName = parameterTypeName;
```
joheredi, line 145: What is the goal of this change?

src/generators/indexGenerator.ts
qiaozha: okay changed.

src/transforms/samplesTransforms.ts
qiaozha: This is trying to fix some bug in the sample generation when the body schema type is any type or any object type, it will still  trying to import this Record<string, unknown> from the client. which doesn't exist at all. 
I guess I will probably need to file another PR for this. Will undo the change in this PR.

joheredi (approved):

PR #1324, opened 2022-02-22
Add compat dependency

sarangan12:
This PR is to add the option flag `core-compat-http-mode` to support the SDK generation using the new `@azure/core-compat-http` package.


src/generators/modelsGenerator.ts
```diff
@@ -69,6 +69,10 @@ export function generateModels(clientDetails: ClientDetails, project: Project) {
       namespaceImport: "coreRestPipeline",
       moduleSpecifier: "@azure/core-rest-pipeline"
     });
+    modelsIndexFile.addImportDeclaration({
```
joheredi, line 72: Do we need to check for `compatMode` before importing this?

test/utils/test-swagger-gen.ts
```diff
@@ -851,6 +852,16 @@ const testSwaggers: { [name: string]: SwaggerConfig } = {
     addCredentials: false,
     isTestPackage: true
   },
+  corecompattest: {
```
joheredi, line 855: Since we don't have any actual tests for this package, I think it would make sense to make this a SmokeTest instead. What do you think?

joheredi (approved):
Looks good, I wonder if it would make more sense to have the test as a Smoke Test instead of an integration test because we are not adding any `spec.ts` files

src/generators/modelsGenerator.ts
sarangan12: Actually nope. All unused imports will be removed at the end by the prettier.

test/utils/test-swagger-gen.ts
sarangan12: Added the test

PR #1329, opened 2022-03-01
Upgrade target framework in v4x branch

sarangan12:
Upgrade target framework in v4x branch since the older version is no longer supported. 

@jeremymeng Please review and approve

### DIFF
--- a/src/autorest.typescript.csproj
+++ b/src/autorest.typescript.csproj
@@ -18,7 +18,8 @@
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
  
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+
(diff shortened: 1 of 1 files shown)


src/autorest.typescript.csproj
```diff
@@ -18,7 +18,7 @@
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
  
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
jeremymeng, line 21:
```suggestion
    <TargetFramework>netcoreapp3.1</TargetFramework>

```

joheredi:
It seems like we have a hard dependency on a package called `dotnet-sdk-2.0.0` which installs dotnet as an npm dependency. Looking at the CI error it seems like we need to do a more involved change to actually support v3

sarangan12:
No longer required.

PR #1335, opened 2022-03-03
try-fix-dpg-test-coverage

qiaozha:
To make test coverage data of autorest.typescript shows up correctly in the test server dash board http://azure.github.io/autorest/dashboard.html


qiaozha:
@joheredi @sarangan12 I think this PR is ready for review. And you can see the coverage data for typescirpt here http://azure.github.io/autorest/dashboard.html

joheredi (commented):
Would it be possible to split this change into 2 PRs? One that fixes pushing the code coverage and the other to change the structure of the CI. I'm not 100% certain I agree with the restructuring, one of the issues is that it would probably increase the time it takes for CI to complete, for example, the latest CI in this PR was canceled after 1 hour.

For the coverage publishing changes, I'm fine with them. So if you split we can merge the coverage fix right away

qiaozha:
> Would it be possible to split this change into 2 PRs? One that fixes pushing the code coverage and the other to change the structure of the CI. I'm not 100% certain I agree with the restructuring, one of the issues is that it would probably increase the time it takes for CI to complete, for example, the latest CI in this PR was canceled after 1 hour.
> 
> For the coverage publishing changes, I'm fine with them. So if you split we can merge the coverage fix right away

@joheredi I see the latest ci is called after 21 mins to finish. 
image

Actually this change is trying to 
1. split previous integration test into the generation jobs and test job
1. split the generation for integration into two jobs, hlc generation and rlc generation
2. combined the rlc generation with version-tolerance generation into the same job
3. merge the test for hlc, rlc, version-tolerance tasks into the same job. This is necessary to get the correct test coverage because the test coverage data is related with start-test-server and if we split them into different jobs, we can only get part of the test coverage for each job. Besides, merging them won't take that long, because all we did here is to call all kinds of http request in the test server.
 
I don't think this can cost longer than previous task. If anything, it should reduce the time cost.

joheredi:
I guess this split makes sense from a development perspective as it makes it easier to run only a subset of tests. @sarangan12 would you mind taking a look as well, do you have any concerns?

joheredi (approved):

sarangan12 (approved):
LGTM

Issue #1338, opened 2022-03-14
[Mgmt] Long running operation activate the poller before we return it to the customer.


qiaozha:
currently we don't generate the poller activate code in beginXXX method which is inconsistent with our track1 SDK.
This can be fix by adding await poller.poll() line after this line 
https://github.com/Azure/autorest.typescript/blob/7e4f45c0092e565780d81a62e3df42332e20a183/src/generators/operationGenerator.ts#L982-L985

qiaozha:
close this issue as PR merged.

Issue #1337, opened 2022-03-09
[Mgmt] Add baseUrl customization for track2 SDK


qiaozha:
Currently we are using 
`baseUri: options.endpoint || "https://management.azure.com/" ` 
 in mgmt plane track2 SDK [track2 link](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/keyvault/arm-keyvault/src/keyVaultManagementClient.ts#L83). which is different from what we did in the [track1 SDK link](https://github.com/Azure/azure-sdk-for-js/blob/feature/v4/sdk/keyvault/arm-keyvault/src/keyVaultManagementClientContext.ts#L57)
`this.baseUri = options.baseUri || this.baseUri || "https://management.azure.com";`
I wonder why we remove the support of baseUri customization and use endpoint instead ?

customer suggest us to change it into 

`baseUri: options.endpoint ?? options.baseUri ?? "https://management.azure.com/"`

see original issue here https://github.com/Azure/azure-sdk-for-js/issues/20651

qiaozha:
@sarangan12 and @joheredi do you have any comments about this issue ?

joheredi:
It is unfortunate that we have `baseUri` in core-client's [ServiceClientOptions](https://github.com/Azure/azure-sdk-for-js/blob/3a5d82089d2458ed21c37c572142e80f3561e27b/sdk/core/core-client/src/serviceClient.ts#L33). As it stands right now I think this would be the right fix to do since we are providing an option that we are ignoring.

However, I wonder if we should mark baseUri as deprecated in core-client and replace it with `endpoint`? @xirzec, @bterlson what do you think?

Note: the generator generates an `endpoint` property and extends the `ServiceClientOptions` interface

bterlson:
Having two properties for the same thing seems bad, but agree that step 1 in this situation is to make sure baseUri isn't ignored. The customer's fix makes sense to me (endpoint wins if it's set). Deprecating makes sense to me for longer term.  How did we end up with two properties for the same thing?

qiaozha:
@bterlson  I guess this is probably because data plane is using endpoint since track1 and mgmt plane on the other hand is using baseUri. @joheredi feel free to correct me if I am wrong. Thanks

joheredi:
@bterlson, I think `baseUri` was inherited from `core-http` and we didn't get rid of it 😞 in favor of endpoint. I think the way we can proceed right now is with 3 changes:

[...]

PR #1339, opened 2022-03-14
Fix baseuri and long running operation

qiaozha:
fix issue #1338  #1337

### DIFF
--- a/test/integration/generated/datafactory/src/operations/triggers.ts
+++ b/test/integration/generated/datafactory/src/operations/triggers.ts
@@ -271,10 +271,12 @@ export class TriggersImpl implements Triggers {
       { resourceGroupName, factoryName, triggerName, options },
       subscribeToEventsOperationSpec
     );
-    return new LroEngine(lro, {
+    const poller = new LroEngine(lro, {
       resumeFrom: options?.resumeFrom,
       intervalInMs: options?.updateIntervalInMs
     });
+    poller.poll();
+    return poller;
   }
 
   /**
@@ -380,10 +382,12 @@ export class TriggersImpl implements Triggers {
       { resourceGroupName, factoryName, triggerName, options },
       unsubscribeFromEventsOperationSpec
     );
-    return new LroEngine(lro, {
+    const poller = new LroEngine(lro, {
       resumeFrom: options?.resumeFrom,
       intervalInMs: options?.updateIntervalInMs
     });
+    poller.poll();
+    return poller;
   }
 
   /**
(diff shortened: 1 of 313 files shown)


test/integration/generated/datafactory/src/operations/triggers.ts
```diff
@@ -271,10 +271,12 @@ export class TriggersImpl implements Triggers {
       { resourceGroupName, factoryName, triggerName, options },
       subscribeToEventsOperationSpec
     );
-    return new LroEngine(lro, {
+    const poller = new LroEngine(lro, {
       resumeFrom: options?.resumeFrom,
       intervalInMs: options?.updateIntervalInMs
     });
+    await poller.poll();
```
joheredi, line 278: @deyaaeldeen, this change is related to the discussion about sending the initial request for LRO when creating the poller.

I believe this behavior change should be in `@azure/core-lro` either by default or based on an option, rather manually calling `.poll()` in the generated code. What do you think?

joheredi:
baseUri changes look good. I'm not sure if the LRO changes should live in `@azure/core-lro` instead. Would like to hear @deyaaeldeen's thoughts.

qiaozha:
> baseUri changes look good. I'm not sure if the LRO changes should live in `@azure/core-lro` instead. Would like to hear @deyaaeldeen's thoughts.

This LRO change is coming from the email that @deyaaeldeen replied.

test/integration/generated/datafactory/src/operations/triggers.ts
deyaaeldeen: @joheredi `LroEngine` is a class and it is not possible to make the constructor send requests because it cannot return a promise. The other thing we could do is to make core-lro export a factory function that creates an instance of the engine and activates it but that will require some discussion with @bterlson and @xirzec.

I am happy to explore this further but for now we can go with the current change until we decide on/create the factory.

deyaaeldeen (commented):
Looks good but I will defer approval to @joheredi.

nit: please consider breaking the PR into two since the two main changes here do not seem related.

test/integration/generated/datafactory/src/operations/triggers.ts
deyaaeldeen: eventually, in core-lro@v3, I would like to move away from the class and just have a function to create the poller, this will make it easier to create an activated poller.

qiaozha:
It takes too much time for running the test to just make the ci pass, much longer than the code change. As this is just two simple change and irrelevant with each other. I just think maybe can combine they to save time for running the test.

test/integration/generated/datafactory/src/operations/triggers.ts
joheredi: Thanks for the clarification @deyaaeldeen. I'm fine with the change then

joheredi (approved):

sarangan12 (approved):

Issue #1344, opened 2022-03-17
Core Compatibility Package is not added correctly


sarangan12:
The `@azure/core-http-compat` package is not included correctly in the client file. For example, in the Search swagger, the package is not included correctly

sarangan12:
Issue has been fixed

PR #1345, opened 2022-03-17
Code change to fix the compat addition bug

sarangan12:
This PR is to fix the issue reported in https://github.com/Azure/autorest.typescript/issues/1344.

@joheredi @JonathanCrd Please review and approve. Thanks


JonathanCrd (approved):
Thanks for fixing this 🚀

src/generators/clientFileGenerator.ts
```diff
@@ -94,6 +94,10 @@ export function generateClient(clientDetails: ClientDetails, project: Project) {
       namespaceImport: "coreClient",
       moduleSpecifier: "@azure/core-client"
     });
+    clientFile.addImportDeclaration({
```
joheredi, line 97: Do we want to check for the compat flag before adding the import?

joheredi (approved):

src/generators/clientFileGenerator.ts
sarangan12: The check is not necessary. If the import is not used, it will be removed at the end by the prettier

Issue #1342, opened 2022-03-17
[Mgmt][HLC] doesn't generate correct mapper for those renamed parameter because of reserved words.


qiaozha:
For example the parameter we have for this checkNameAvailibitity is 
```
  /**
   * Description for Check if a resource name is available.
   * @param name Resource name to verify.
   * @param typeParam Resource type used for verification.
   * @param options The options parameters.
   */
  checkNameAvailability(
    name: string,
    typeParam: CheckNameResourceTypes,
    options?: CheckNameAvailabilityOptionalParams
  ): Promise<CheckNameAvailabilityResponse> {
    return this.sendOperationRequest(
      { name, typeParam, options },
      checkNameAvailabilityOperationSpec
    );
  }
```
Here the typeParam is renamed from the reserved word type. but in the generated Mappers.
```
export const ResourceNameAvailabilityRequest: coreClient.CompositeMapper = {
  serializedName: "ResourceNameAvailabilityRequest",
  type: {
    name: "Composite",
    className: "ResourceNameAvailabilityRequest",
    modelProperties: {
      name: {
        serializedName: "name",
        required: true,
        xmlName: "name",
        type: {
          name: "String"
        }
      },
      type: {
        serializedName: "type",
        required: true,
        xmlName: "type",
        type: {
          name: "String"
        }
      },
      isFqdn: {
        serializedName: "isFqdn",
        xmlName: "isFqdn",
        type: {
          name: "Boolean"
        }
      }
    }
  }
};
```
we are still seeing type as the key, which cause the generated code can't work 
<img width="1267" alt="image" src="https://user-images.githubusercontent.com/55688292/158784461-d40fc5bd-2c7a-4a66-b0ff-f68972a74978.png">

This is original come from our customer reported https://github.com/Azure/azure-sdk-for-js/issues/20728

PR #1346, opened 2022-03-18
fix-mapper-parameter-rename-issue

qiaozha:
fix issue #1342 and issue https://github.com/Azure/azure-sdk-for-js/issues/20933


qiaozha:
Hi @sarangan12 @joheredi Could you take a look at this PR ? Thanks

test/smoke/generated/web-resource-manager/src/models/parameters.ts
```diff
@@ -908,7 +908,7 @@ export const name1: OperationParameter = {
 };
 
 export const typeParam: OperationParameter = {
-  parameterPath: "typeParam",
+  parameterPath: "type",
```
qiaozha, line 911: @joheredi  I am not sure if this is the right fix. In this case, we changed the parameterPath from typeParam to type.
And type is simple string type, we don't have it defined in the models/index.ts. Can you confirm?

test/integration/generated/datasearch/src/models/parameters.ts
```diff
@@ -54,7 +54,7 @@ export const indexName: OperationURLParameter = {
 };
 
 export const xMsClientRequestId: OperationParameter = {
-  parameterPath: ["options", "requestOptionsParam", "xMsClientRequestId"],
+  parameterPath: ["options", "requestOptions", "xMsClientRequestId"],
```
qiaozha, line 57: looks like this one still has a definition of requestOptionsParam. Here's how this parameter being called in the operation.
```
  /**
   * Queries the number of documents in the index.
   */
  count(
    options?: DocumentsCountOptionalParams
  ): Promise<DocumentsCountResponse> {
    return this.client.sendOperationRequest({ options }, countOperationSpec);
  }
```
Here's how the model being defined in models/index.ts
```
export interface DocumentsCountOptionalParams
  extends coreClient.OperationOptions {
  /** Parameter group */
  requestOptionsParam?: RequestOptions;
}
export interface RequestOptions {
  /** The tracking ID sent with the request to help with debugging. */
  xMsClientRequestId?: string;
}
```
@joheredi maybe we should keep requestOptions in this case ?

test/integration/generated/azureParameterGrouping/src/models/parameters.ts
```diff
@@ -124,7 +124,7 @@ export const fromParam: OperationQueryParameter = {
   parameterPath: [
     "options",
     "parameterGroupingPostReservedWordsParameters",
-    "fromParam"
+    "from"
```
qiaozha, line 127: In this case, we are generating the models like 
```
export interface ParameterGroupingPostReservedWordsParameters {
  /** 'from' is a reserved word. Pass in 'bob' to pass. */
  from?: string;
  /** 'accept' is a reserved word. Pass in 'yes' to pass. */
  accept?: string;
}
```
I guess this is a valid fix. But I am kind of confused at the case in datasearch.

test/integration/generated/datasearch/src/models/parameters.ts
joheredi: Yeah parameter path should match the structure of the generated object. For example here, accessing `xMsClientRequestId`, which is a property of `requestOptionsParam` in the interface  `DocumentsCountOptionalParams`. And contained inside the `options` parameter in the `count` operation in `documents.ts`. The parameter path should look like if you were accessing that property i.e. `options.requestOptionsParam.xMsClientRequestId` so in this case, the right parameter path would be `["options", "requestOptionsParam", "xMsClientRequestId"]`

So the new parameter path `requestOptions` would be incorrect. You'd either need to not change it or also change the interface  `DocumentsCountOptionalParams` to have `requestOptions` instead of `requestOptionsParam`

test/integration/generated/azureParameterGrouping/src/models/parameters.ts
joheredi: This is right if you wanted to access `from` of `options` you'd do `options.parameterGroupingPostReservedWordsParameters.from` whcih matches the new parameterPath

test/integration/generated/datasearch/src/models/parameters.ts
qiaozha: changed the parameterPath back to requestOptionsParam

test/integration/generated/deviceprovisioningservice/src/models/parameters.ts
```diff
@@ -275,7 +275,7 @@ export const asyncinfo: OperationQueryParameter = {
 };
 
 export const argumentsParam: OperationParameter = {
-  parameterPath: "argumentsParam",
+  parameterPath: "arguments",
```
joheredi, line 278: This needs to be called `argumentsParam`
image

test/smoke/generated/web-resource-manager/src/models/parameters.ts
joheredi: Parameter path always needs to match the parameter name passed to the operation in operations or client. In this clase this needs to be `typeParam`

image

test/integration/generated/deviceprovisioningservice/src/models/parameters.ts
qiaozha: changed back

test/smoke/generated/web-resource-manager/src/models/parameters.ts
qiaozha: changed back

qiaozha:
@joheredi I think all the comments should have been resolved. Could you help to take a look again ? Thanks

sarangan12 (approved):

joheredi (approved):

PR #1348, opened 2022-03-22
allow customize api version

qiaozha:
This PR is trying to fix  https://github.com/Azure/azure-sdk-for-js/issues/20590


joheredi:
Discussed offline, due to the side effects of setting apiVersion isConstant: false, we should explore the option of generating a pipelinePolicy that replaces the apiVersion if `options.apiVersion` is provided

qiaozha:
close this one in favor of #1358

PR #1349, opened 2022-03-22
rename client class name for rlc

qiaozha:
Currently we are adding a `Like` suffix for client class name, and keep the title name as client factory name. 
In this PR, we will add a `Client` suffix for client class name, if it's not end with `Client`. and use `createClient` as the client factory name. and remove the default export of client factory name for multi-clients.


test/smoke/generated/purview-administration-rest/review/purview-administration-rest.api.md
```diff
@@ -561,6 +561,12 @@ interface ComplexReplacerConfigOutput {
     typeName?: string;
 }
 
+// @public (undocumented)
+function createClient(Endpoint: string, credentials: TokenCredential, options?: ClientOptions): PurviewMetadataPoliciesClient;
+
+// @public (undocumented)
+function createClient_2(endpoint: string, credentials: TokenCredential, options?: ClientOptions): PurviewAccountClient;
```
xirzec, line 568: This shouldn't be a top level export right? I'd expect it to only show up under the namespaces.

test/smoke/generated/agrifood-data-plane/review/agrifood-data-plane.api.md
```diff
@@ -884,6 +880,10 @@ export interface CascadeDeleteJobOutput {
     status?: string;
 }
 
+// @public (undocumented)
+function createClient(Endpoint: string, options?: ClientOptions): AzureAgriFoodPlatformDataPlaneServiceClient;
```
xirzec, line 884: I think we should only export this as the default export to leave for later adding a second client, what do you think @joheredi ?

xirzec (commented):
Thanks for taking care of this! Just a couple of nits about how the factories are exported.

test/smoke/generated/agrifood-data-plane/review/agrifood-data-plane.api.md
joheredi: I think this only exports it as the default below, there is no access to createClient as a named import/export 

image

test/smoke/generated/purview-administration-rest/review/purview-administration-rest.api.md
```diff
@@ -1256,7 +1262,7 @@ interface PrivateLinkServiceConnectionStateOutput {
 
 declare namespace PurviewAccount {
     export {
-        PurviewAccountClient as PurviewAccount,
+        createClient_2 as PurviewAccount,
```
joheredi, line 1265: @qiaozha we need to change this so that this gets generated as `createClient_2 as createClient`. instead of PurviewAccount

test/smoke/generated/purview-administration-rest/review/purview-administration-rest.api.md
joheredi: I don't think these createClient functions are exported as toplevel, I only see 2 things exported as top level:

image

joheredi (approved):
Left another comment after another pass

joheredi (commented):
Please change the export to `as createClient` for multi-client packages

test/smoke/generated/agrifood-data-plane/review/agrifood-data-plane.api.md
xirzec: It would be nice if we could somehow make this clear in api-extractor or else api reviews are going to be misleading

Issue #1334, opened 2022-03-03
[RLC] Clear generated output folder in regenerating codes


dw511214992:
Currently, codegen doesn't clear src folder in regenerating codes. If there is an existed client file named `AClient.ts`, and the latest swagger changes the client name to `BClient`. Then the new generated client file will be named `BClient.ts`. However, codegen doesn't delete file `AClient.ts`, which leads to build failure.

qiaozha:
see  pr #1355

PR #1355, opened 2022-03-28
clear generated output folder for rlc and bump a new beta version

qiaozha:
to resolve issue https://github.com/azure/autorest.typescript/issues/1334 and release a newer version of typescript for a few bugfixes

### DIFF
--- a/src/restLevelClient/generateRestLevel.ts
+++ b/src/restLevelClient/generateRestLevel.ts
@@ -26,6 +26,8 @@ import { generateSampleTestFile } from "../generators/test/sampleTestGenerator";
 import { generateEsLintConfig } from "../generators/static/esLintConfigGenerator";
 import { generateRollupConfig } from "../generators/static/rollupConfigFileGenerator";
 import { generateReadmeFile } from "../generators/static/readmeFileGenerator";
+import * as path from "path";
+import * as fsextra from "fs-extra";
 
 /**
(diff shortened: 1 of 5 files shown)


src/restLevelClient/generateRestLevel.ts
```diff
@@ -77,6 +79,8 @@ export async function generateRestLevelClient() {
   // Save the source files to the virtual filesystem
   project.saveSync();
   const fs = project.getFileSystem();
+  const pathToClear = outputPath ? path.join(outputPath, srcPath) : srcPath;
```
joheredi, line 82: I think Autorest can do this if you pass `--clear-output-folder=true`, maybe we don't need to do this in here. Can you try removing these changes and in the generate-swaggers script add   `--clear-output-folder=true`?

joheredi: Talked about this offline. We do this same thing in HLC generator, we should be fine

joheredi (approved):

src/restLevelClient/generateRestLevel.ts
qiaozha: We had a similar discussion for HLC about this before https://github.com/Azure/autorest.typescript/issues/1166#issuecomment-905033088

sarangan12 (approved):

PR #1358, opened 2022-03-31
allow-customize-api-version-for-customer

qiaozha:
to resolve this issue https://github.com/Azure/azure-sdk-for-js/issues/20590


test/integration/generated/additionalProperties/src/additionalPropertiesClient.ts
```diff
@@ -46,6 +51,32 @@ export class AdditionalPropertiesClient extends coreClient.ServiceClient {
     // Assigning values to Constant parameters
     this.$host = options.$host || "http://localhost:3000";
     this.pets = new PetsImpl(this);
+    if (options.apiVersion) {
+      this.customApiVersion(options.apiVersion);
+    }
+  }
+
+  /**  A policy that sets the api-version (or equivalent) to reflect the library version. */
+  private customApiVersion(apiVersion: string) {
```
joheredi, line 60: nit: you could also make this `apiVersion?: string` and move your `if(options.apiVersion)` check inside this function and bail if no apiVersion is provided. With this, you'll keep the constructor simpler

test/integration/generated/additionalProperties/src/additionalPropertiesClient.ts
```diff
@@ -46,6 +51,32 @@ export class AdditionalPropertiesClient extends coreClient.ServiceClient {
...
+    const apiVersionPolicy = {
+      name: "replace api version",
```
joheredi, line 62: Let's use a single word name. "CustomApiVersionPolicy"

test/integration/generated/additionalProperties/src/additionalPropertiesClient.ts
```diff
@@ -46,6 +51,32 @@ export class AdditionalPropertiesClient extends coreClient.ServiceClient {
...
+      ): Promise<PipelineResponse> {
+        const param = request.url.split("?");
+        if (param.length > 1) {
+          const newParams = param[1].split("&").map((item) => {
+            if (item.indexOf("api-version") > -1) {
+              return item.replace(/(?<==).*$/, apiVersion);
+            }
+          });
+          request.url = param[0] + "?" + newParams.join("&");
+        }
```
joheredi, line 75: You can simplify this if you use URL. For example

```typescript
 const parsedUrl = new URL(request.url);
 parsedUrl.searchParams.set("api-version", apiVersion);
```

joheredi (approved):
Looks good, left a few comments

package.json
```diff
@@ -48,10 +48,10 @@
     "dependencies": {
         "@autorest/codemodel": "^4.17.0",
         "@autorest/extension-base": "^3.4.1",
-        "@azure-rest/core-client": "1.0.0-alpha.20220112.1",
         "@autorest/testmodeler": "^1.0.4",
+        "@azure-rest/core-client": "1.0.0-alpha.20220112.1",
         "@azure-tools/codegen": "^2.5.294",
-        "@azure/core-client": "^1.3.0",
+        "@azure/core-client": "file:../azure-sdk-for-js/sdk/core/core-client/azure-core-client-1.5.1.tgz",
```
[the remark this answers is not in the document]
qiaozha: We will need to release a new version of core-client-rest.

joheredi:
Maybe we don't need to add the policy to all the generated SDKs, just the ones that have `apiVersion` as a parameter. That way we don't need to add the apiVersion in core-client `ServiceClientOptions`.

You should be able to check if there is a client parameter called apiVersion and use that to decide if you need to generate the policy.

What do you think?

joheredi (approved):

qiaozha:
> Maybe we don't need to add the policy to all the generated SDKs, just the ones that have `apiVersion` as a parameter. That way we don't need to add the apiVersion in core-client `ServiceClientOptions`.
> 
> You should be able to check if there is a client parameter called apiVersion and use that to decide if you need to generate the policy.
> 
> What do you think?

@joheredi I am not sure. let's say If the service team add a new different api version file in the input files for some reason, and we will probably not consider this as a breaking change in previous practice, but now we will remove this allow custom api version feature in our next code generation generation which should be a breaking change.

qiaozha:
@joheredi @sarangan12  I have updated the PR to generate the custom api version policy only when the apiVersion exists in client properties. Can you help take a look again ? Thanks

joheredi (approved):
Looks good! Can you please add a test that verifies that the api-version is changed?

test/integration/azureSpecialProperties.spec.ts
```diff
@@ -426,4 +426,35 @@ describe("AzureSpecialProperties", () => {
...
+        name: bearerTokenAuthenticationPolicyName
+      });
+      await client.apiVersionDefault.getMethodGlobalValid({
+        onResponse: r => {
+          _response = r;
+        }
+      });
+      assert.isTrue(_response!.request.url.indexOf("2011-05-18") > -1);
+    });
+  })
```
qiaozha, line 459: @joheredi added the test here.

joheredi: Great, thanks! You should be able to move the assert inside the onResponse callback and remove _response. Other than that looks great. Feel free to merge

qiaozha: Thanks for the comment, updated.

PR #1360, opened 2022-04-06
Update codegen to use new version of tracing

joheredi:
Update codegen to use new version of tracing


src/generators/operationGenerator.ts
```diff
@@ -882,7 +855,8 @@ function writeSendOperationRequest(
     getTracingTryCatchStatement(
       sendRequestStatement,
       responseName,
-      !!tracingInfo
+      !!tracingInfo,
+      `${clientDetails.className}-${operationName}`
```
maorleger, line 859:
```suggestion
      `${clientDetails.className}.${operationName}`
```

test/integration/generated/bodyComplexWithTracing/src/operations/array.ts
```diff
@@ -44,25 +43,16 @@ export class ArrayImpl implements Array {
   async getValid(
     options?: ArrayGetValidOptionalParams
   ): Promise<ArrayGetValidResponse> {
-    const { span } = createSpan(
+    return tracingClient.withSpan(
       "BodyComplexWithTracing-getValid",
-      options || {}
+      options ?? {},
```
maorleger, line 48: I don't recall if that actually compiles or if typescript complains. What I've seen work well is on line 44 replacing `options?: ArrayGetValidOptionalParams` with `options: ArrayGetValidOptionalParams = {}` 

Do you think it's possible to generate the code that way?

test/integration/generated/bodyComplexWithTracing/src/operations/array.ts
```diff
@@ -44,25 +43,16 @@ export class ArrayImpl implements Array {
...
+      async () => {
```
maorleger, line 49:
```suggestion
      async (updatedOptions) => {
```

test/integration/generated/bodyComplexWithTracing/src/operations/array.ts
```diff
@@ -44,25 +43,16 @@ export class ArrayImpl implements Array {
...
+        return this.client.sendOperationRequest(
+          { options },
```
maorleger, line 51:
```suggestion
          { updatedOptions },
```
I think, or is it `updatedOptions` ? `updatedOptions` in this case will be of type `OperationOptions` but not sure how that ties into here

maorleger (commented):
Thank you for doing the codegen changes! It would have taken me a lot of time just to wrap my head around the code!

I looked at one of the generated samples and have a few comments - let me know if these are problematic

test/integration/generated/bodyComplexWithTracing/src/operations/array.ts
joheredi: It is not complaining. Tried to change options to have a default value of `= {}` but the change is a bit more involved than I expected. The reason is that we share logic to generate the interface and overloads which don't support default values, so there would be a bit more work to get that going.

If you feel strongly about this we can definitely, invest to make the change. I think we could have it like this and file an issue to improve this by adding the default values.

maorleger: If the compiler is happy I'm happy 😄 

But yeah, I agree there's no need to invest energy for this change if what you have here compiles fine. Thanks!

src/generators/operationGenerator.ts
```diff
@@ -894,22 +868,15 @@ function getSpanStatusCode() {
 function getTracingTryCatchStatement(
```
JonathanCrd, line 868: Since we don't longer use a try catch statement, should this function be renamed?

test/integration/generated/bodyComplexWithTracing/src/tracing.ts
```diff
@@ -6,9 +6,9 @@
...
  */
 
-import { createSpanFunction } from "@azure/core-tracing";
+import { createTracingClient } from "@azure/core-tracing";
 
-export const createSpan = createSpanFunction({
+export const tracingClient = createTracingClient({
   namespace: "Microsoft.Body.Complex",
-  packagePrefix: "Azure.Body.Complex"
+  packageName: "Azure.Body.Complex"
```
JonathanCrd, line 13: Isn't this call missing the package version?

test/integration/generated/lro/src/tracing.ts
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * Code generated by Microsoft (R) AutoRest Code Generator.
+ * Changes may cause incorrect behavior and will be lost if the code is regenerated.
+ */
+
+import { createTracingClient } from "@azure/core-tracing";
+
+export const tracingClient = createTracingClient({
+  namespace: "Microsoft.Autorest.LRO",
+  packageName: "Azure.Autorest.LRO"
```
JonathanCrd, line 13: Same here, no package version.

Actually, I don't recall if these samples have a dummy SDK_VERSION constant, is that why the property is not being generated?

JonathanCrd (changes requested):
I left a couple of comments and one suggestion regarding a function name.
Everything else looks good to me 👍🏼

src/generators/operationGenerator.ts
JonathanCrd: Maybe something like `getTracingClientWithSpanStatement`?

test/integration/generated/lro/src/tracing.ts
joheredi: Nah, we should have the package version name. I'll add that, thanks for keeping me honest :P

JonathanCrd (approved):
:shipit:

PR #1365, opened 2022-04-12
Update test recorder in code generator

MaryGao:


src/generators/static/packageFileGenerator.ts
```diff
@@ -55,7 +55,7 @@ function restLevelPackage(packageDetails: PackageDetails) {
   const hasLRO = hasPollingOperations(model);
   const packageInfo: Record<string, any> = {
     name: `${packageDetails.name}`,
-    "sdk-type": `${azureArm ? "mgmt" : "client"}`,
```
joheredi, line 58: why did we change this? are we no longer setting sdk-type as mgmt for management packages?

joheredi (approved):
Looks good! Please double check with @qiaozha if removing `mgmt` in the package file is intended before merging

src/generators/static/packageFileGenerator.ts
qiaozha: I think this is only impact the rest level client package.json generation. not the hlc.

MaryGao: There is an issue mentioned [here](#1367 ) with the option `azureArm` and this will cause the sdk-type always is `mgmt`. This is not expected for rest-level client generation, `client` would be more proper as the default value. So I directly change it to `client` as there is no `mgmt` client generated.

PR #1371, opened 2022-04-19
Update the test recorder in HLC code generator

MaryGao:
Update the test recorder in HLC code generator


src/generators/test/sampleTest.ts.hbs
```diff
@@ -1,32 +1,31 @@
...
-    AZURE_CLIENT_SECRET: "azure_client_secret",
-    AZURE_TENANT_ID: "88888888-8888-8888-8888-888888888888",
-    SUBSCRIPTION_ID: "azure_subscription_id"
-  },
-  customizationsOnRecordings: [
-    (recording: any): any =>
-      recording.replace(/"access_token":"[^"]*"/g, `"access_token":"access_token"`)
-  ],
-  queryParametersToSkip: []
+const replaceableVariables: { [k: string]: string } = {
```
joheredi, line 5: maybe we can use `Record<string, string>` instead of ` { [k: string]: string }`

test/integration/generated/additionalProperties/package.json
```diff
@@ -79,7 +81,7 @@
     "unit-test:node": "cross-env TEST_MODE=playback npm run integration-test:node",
     "unit-test:browser": "echo skipped",
     "integration-test": "npm run integration-test:node && npm run integration-test:browser",
-    "integration-test:node": "mocha -r esm --require ts-node/register --timeout 1200000 --full-trace test/*.ts --reporter ../../../common/tools/mocha-multi-reporter.js",
+    "integration-test:node": "dev-tool run test:node-ts-input -- --timeout 1200000 'test/*.ts'",
```
joheredi, line 84: We need to make sure that this is only generated like this when azure-sdk-for-js flag is true. Otherwise dev-tool is not going to be available and `dev-tool` can't be published

joheredi (commented):
I think it looks good, there are a few pipeline failing, also make sure that dev-tool is only used when azure-sdk-for-js is true

src/generators/test/sampleTest.ts.hbs
MaryGao: Sure, I will update it

qiaozha:
@MaryGao Please resolve the conflicts and consider my comments here https://github.com/Azure/azure-sdk-for-js/pull/21457#discussion_r854707286

joheredi (approved):

PR #1372, opened 2022-04-19
Audit fixes for @autorest/typescript

sarangan12:
Not ready for review


joheredi:
Thanks for taking care of the Component Governance issues! Would you mind filing an issue to re-enable samples mentioning that it was removed due to CG issues?

joheredi:
@qiaozha, for context. We need to drop sample generation since the `testmodeler` package is causing Component Governance alerts which is of critical priority right now. We'll have an issue tracking re-enabling it.

joheredi (approved):

PR #1374, opened 2022-04-20
Audit Fixes III for @autorest/typescript package

sarangan12:
Audit Fixes III for @autorest/typescript package

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
         "test-version-tolerance": "mocha -r ts-node/register --timeout 2000 './test/version-tolerance/**/!(sampleTest).spec.ts'",
         "clone:specs": "git clone https://github.com/Azure/azure-rest-api-specs.git ./.tmp/specs",
         "smoke-test": "ts-node ./test/utils/smoke-test.ts",
-        "copyFiles": "copyfiles src/**/*.hbs dist/ && copyfiles src/**/**/*.hbs dist/"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
         "test-version-tolerance": "mocha -r ts-node/register --timeout 2000 './test/version-tolerance/**/!(sampleTest).spec.ts'",
         "clone:specs": "git clone https://github.com/Azure/azure-rest-api-specs.git ./.tmp/specs",
         "smoke-test": "ts-node ./test/utils/smoke-test.ts",
-        "copyFiles": "copyfiles src/**/*.hbs dist/ && copyfiles src/**/**/*.hbs dist/"
+        "copyFiles": "cp src/**/*.hbs dist/ && cp src/**/**/*.hbs dist/ && cp src/generators/samples/*.hbs dist/src/generators/samples/ && cp src/generators/static/*.hbs dist/src/generators/static/ && cp src/generators/test/*.hbs dist/src/generators/test/ && cp src/restLevelClient/*.hbs dist/src/restLevelClient/"
```
joheredi, line 40: Does this work cross platform?

sarangan12: cp normally works for linux/mac boxes. I use a windows machine and it is working fine for that too

joheredi: I think it works in PowerShell but it doesn't work in windows cmd 😢 
image

Is there other package we can use that works cross platform? Or can we write a node script that does this for us?

sarangan12: I did not use the powershell. I used it in the regular command prompt

joheredi (approved):

package.json
joheredi: Awesome! Thanks for verifying

PR #1375, opened 2022-04-21
Improve test browser bundling

joheredi:
Currently, our browser tests are doing a few things that give us fake coverage:

1. We are pollyfilling some of Node's platform APIs. For example FS and Process (see webpack.config.js)
2. We are excluding quite a few tests from our test bundling
3. There is no way to set up browser mapping to tell the bundler to use a different implementation for Browsers.

This PR addresses the point above and adapts some of the tests to actually validate against the browser. The changes in this PR are mainly

1. Remove polyfills
2. Set up "web" target
3. Abstract FileSystem functions and provides mappings for the browser when appropriate
4. Remove the all the previously ignored test files
5. Fix some tests to actually validate browser behavior. For example, browsers set the user agent in the `x-ms-useragent` header while node uses `user-agent`. We were not validating the browser


qiaozha (approved):

sarangan12 (approved):
Great!!!

PR #1378, opened 2022-04-22
Remove Component Det from Test Stages

sarangan12:
While working on the Component Governance Issues, I have noticed that the Component Detection Job is run at every stage (Build, HLC Test, RLC Test, Smoke Test, Unit Test). This is not only time cosuming and unnecessary. Sometime, when it hangs it will create a component governance violations  that are not there. 

@praveenkuttappan suggested to remove the component governance from all steps other than build. This PR is to do that. You can see a test run here: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1524546&view=logs&j=462ec23f-14a8-561c-bd2a-3cc23815db36 - This one has Component Governance only under the Build Stage 

Compare it with the old run - https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1524441&view=logs&j=88cef286-44ca-5c67-b516-cbd63d8b0608 where every stage has Component Governance. 

@joheredi @praveenkuttappan  Please review and approve the PR

### DIFF
--- a/.scripts/ci.yml
+++ b/.scripts/ci.yml
@@ -54,7 +54,7 @@ stages:
       - job: HLC_Generation
         dependsOn: Build
         variables:
-          skipComponentGovernanceDetection: true
+          - template: globals.yml
         strategy:
           matrix:
             macOS_Node12:
@@ -86,7 +86,7 @@ stages:
       - job: RLC_Generation
         dependsOn: Build
         variables:
-          skipComponentGovernanceDetection: true
+          - template: globals.yml
         strategy:
           matrix:
             macOS_Node12:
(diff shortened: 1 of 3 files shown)


.scripts/ci.yml
```diff
@@ -53,6 +53,8 @@ stages:
               path: $(Build.ArtifactStagingDirectory)
       - job: HLC_Generation
         dependsOn: Build
+        variables:
```
praveenkuttappan, line 56: A better approach is to create a template global.yml with this variable and include that template in each job. You can refer similar implementation in js repo here.

global template in JS: `https://github.com/Azure/azure-sdk-for-js/blob/main/eng/pipelines/templates/variables/globals.yml`


And it's reference in job: https://github.com/Azure/azure-sdk-for-js/blob/main/eng/pipelines/templates/jobs/ci.yml#L31-L32

joheredi (approved):

PR #1383, opened 2022-05-10
Add chai and @types/chai in code gen

MaryGao:
Fix the two libs missing issue  both in RLC and HLC gen
- chai 
- @types/chai


qiaozha (approved):

joheredi (approved):
Thanks!

Issue #1373, opened 2022-04-20
Samples Generation Feature in @autorest/typescript package


sarangan12:
In the `@autorest/typescript` package, there was a module to generate samples. This feature was added recently before few months. But, for this feature to work, there is a dependency on the `@autorest/testmodeler` package. Unfortunately, this has caused a security Component Governance issue that needs to be resolved immediately. 

So, I have removed that feature with the PR: https://github.com/Azure/autorest.typescript/pull/1372. I am creating this issue to ensure that we discuss and decide if we should bring the feature back and if so do it without causing any Component Governance Issue.

qiaozha:
@sarangan12 @joheredi Would you mind share some information how the test modeler is violating the component governance issue ? we will work on the fix.
cc @changlong-liu  @lirenhe

sarangan12:
There are dependencies such as `async`, `validator`, `ansi-regex`, `yargs-parser` package dependencies which have Audit violations. You could go to a previous commit and run `npm audit` which will give you a complete list

ArcturusZhang:
This is done per @qiaozha

PR #1392, opened 2022-05-17
add back sample generation for mgmt

qiaozha:
For issue https://github.com/Azure/autorest.typescript/issues/1373


README.md
```diff
@@ -89,9 +89,9 @@ pipeline:
 
 ```
 
-```yaml $(generate-test)
+```yaml $(generate-sample)
 use-extension:
-  "@autorest/testmodeler": "1.0.4"
+  "@autorest/testmodeler": "2.2.5"
```
qiaozha, line 94: This version should be able to resolve the audit issue

qiaozha:
@joheredi @sarangan12  do you have any idea why the smoke test for rlc is failing ? it says package-lock.json has some uncommited changes but why does other tasks is passing ?

sarangan12:
@qiaozha You should have run the command `npm run rlc-generate-swaggers` That would solve the build failure

sarangan12 (approved):
Left a minor comment on build failure. Otherwise LGTM

qiaozha:
> @qiaozha You should have run the command `npm run rlc-generate-swaggers` That would solve the build failure

I tried in my local, but nothing is changed.

joheredi (approved):

.scripts/build.yml
```diff
@@ -1,8 +1,8 @@
 steps:
   - task: NodeTool@0
     inputs:
-      versionSpec: "14.x"
-    displayName: "Use Node 14"
+      versionSpec: "16.x"
+    displayName: "Use Node 16"
```
qiaozha, line 5: I have changed all the node version as 16.x. because we have used shrinkwrap to resolve the cg issue in testmodeler as we don't directly depend on the problematic package and this depdencies is from external, but shrinkwrap only get supported in npm v7+. Therefore, I upgrade the node version in our pipeline.

qiaozha: @joheredi @sarangan12 Please take a look at the change. If no other objections, I will merge this PR tomorrow.

PR #1410, opened 2022-06-07
fix-dictionary-type-sample-generation-bug

qiaozha:
This PR is trying to fix the sample generation error of Dictionary body object. 

Previously, the generated samples will try to import the Dictionary type from generated package like this 

```
import {
   { [propertyName: string]: any }
} from "@azure/arm-datafactory"
```

which is incorrect


sarangan12 (approved):
LGTM

joheredi (approved):

PR #1418, opened 2022-06-14
[RLC] Generate isUnexpected helper

joheredi:
Generate a helper function to aid in type narrowing. This helper function takes in response and checks if the status code for the given path is a success as defined in the REST API Spec.

With this helper function, these are the 2 options to check for success and narrow down the response type
image

Also as part of this change, we are generating `"default"` for the non-defined status codes. Before we generated `"500"` which is misleading because that may not be what the service returns at all


test/rlcIntegration/generated/bodyComplexRest/src/responses.ts
```diff
@@ -35,7 +35,7 @@ export interface BasicGetValid200Response extends HttpResponse {
 
 /** Get complex type {id: 2, name: 'abc', color: 'YELLOW'} */
 export interface BasicGetValiddefaultResponse extends HttpResponse {
-  status: "500";
+  status: "default";
   body: ErrorModelOutput;
```
MaryGao, line 39: Thanks for preparing this PR!
We all agree that the default `500` might not be a correct error code, but even if we switch to `default` I think it's misleading and hard to leverage the error type indication if we have condition like `response.status == `501` .

I was wondering it it is possible we could have below types:

1. Maybe define all status type: 
2. Define success status type
3. Define non-success type

```typescript
export type AllStatus = '200' | '201' | '202' | '400' | '401' | '404' | '500';
export type GetSuccessStatus = '200';
export type GetDefaultStatus = Exclude<AllStatus, GetSuccessStatus>;
```

MaryGao (approved):

test/rlcIntegration/bodyComplextRest.spec.ts
```diff
@@ -24,33 +25,32 @@ describe("BodyComplex Rest Client", () => {
         const result = await client.path("/complex/basic/valid").get();
 
         try {
-          if (result.status !== "200") {
+          if (isUnexpected(result)) {
```
lirenhe, line 28: I may lack some context but I am not quite sure adding this helper function 'isUnexpected()' would help as user still need to parse the error code themselves. 
And can we use 'result.IsSuccessful' instead?
I suggest we shall do some investigation about better error handling for RLC. 
So can we postpone this after RLC 1.0 GA?    cc @qiaozha

test/rlcIntegration/generated/bodyComplexRest/src/responses.ts
joheredi: I was exploring something like this however I found 2 big issues:

1. If we generate a list of all possible status codes, then intellisense suggestions for status codes become useless as the size of the list is huge
AllStatusCodes

2. Now narrowing down to all errors also becomes undoable since you will have to write an if check for all the non-success status codes i.e. `if(response.status !== "400" &&  response.status !== "401" &&  response.status !== "402" && response.status !== "403" && response.status !== "404" && response.status !== "405" && /* And so on */)` to get the type narrowed down correctly.

[Here's a sample playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggNnAysAhsArgZygXigIgEYAGY-AHyOMIqoCZaSBmR4gFlYFZWA2VgdlYAOVgE5GJCTUqFCDGYRYKOC7gr4LBCkQvEy6ZfdKJ15JpSZUm1JjSa0mdJvUSaHXxxWcUXFVxTaKdooOik6KLoRs7lGebN5svmz+bIFswWyhbOFskZwxnJ6c3py+nP6cgZzBnKGc4ZyRPDE8njzePL48-jyBPME8oTzhPJH8Mfye-N78vvz+-IH8wfyh-OH8kUIxQp5C3kK+Qv5CgULBQqFC4UKRojGinqLeor6i-qKBosGioaLhohcBncBmMBjMBgsBisBhsBjsBgcBicBiBkkodFktExEMU2Ki+LhhARhCRhBRhCBBmxmJpELoULoMLocLoCLoSLoKLoQLc2KYYKYEKYUKYMKYcKYCKYSKYKKYQOi2LYYPiyqhyWVcPSyqR2WVQPy2MKxohpWNMMqxoRtWNKMa2OajrB7UdUO6jrh-UdSOGjqB42xkyDENmQZhiyDCNWQZRm2x2wTYP2CahxwTcPOCaR1wTQPu2MehYhr0LMM+hYRv0LKMBtDc7jcxjcZjcFjcVjcNjcdjcDjcTjcLgFjaxlAFrbx44J0+7xPrpIXg8p9ep49p69bDNXnZZq97HNXg559b544Fp9bItPnYlp97MtPg4V9aV45Vr9biVfndSr97mSvoOuT1ka44muBrbmuBnZWuBva2uBg4OuOTqoc2rqoe2Hqod23qof2fqocOgbjsGZGtmGZGdpGZG9jGZGDvG46JixzYpix7bpix3ZZix-a5ixw4FuORaia2paiZ2Faib21aiYOdaUNE7jRMY0RmNEFjRFY0Q2NEdjRA40RONELgqqpY74CqmlTtZM72fp87KYuLmmSuylrtZG7eZp26ebpe6eYZh6eaZJ7KWe1kXpFmnXpFul3pFhmPpFpkvspb7WR+mWad+mW6X+mWGYBmWmSBylgdZEHVZp0HVbpcHVYZiHVaZKHWWhnXqZhnXaThnX6fhnXGURnXmaR1nkVNmlUVNum0VNhkMVNpnMdZrEbepHEbdp3EbfpfEbcZgkbeZInWWJl2aZJl26TJl2GfJl2mUp+D5O4+TGPkZj5BY+RWPkNj5HY+QOPkTj5C4hSfVZhS-XZhSA4QwPOe9rno5DHnvV5xTfaYtDFP9TKE0FOOg6FOOQxF71RaU31CoT8W04DSW06DqW05DGXvVl5TfWqlDlP9mpC0VvOg6VvOQxV71VZU33FIT9WVIDTWVKDrWVJDHXVJ9rSE711T-QN1TA8N1Tg2N1TQ5NtTfdMhNzbUgOLbUoMrbUkPrfUn27ITO31P9+31MDR31ODp31NDF2NN9zyE7djSAw9jSg89jSQ3WADcABQoCQFAADiEDAAAqpgEAAE6YIg6AAMb1xAmC16gGDYHg+DAvgUDkAQoI933XfsPgecF9AJfl5XNcAKJV1XAD2VfIGgWC4FAM8AB713A6AACYQAAPPASBt1gAA0xelxX1e1w3Tctyv7cAHx57nACWAB2wDVwAZigTcr5T1vnXRuzdMAACVm5gAXp-SuUAADeucoBQEwGfTAAAuIBN8a6gIfq3VemBz7IKgAAIwXnvEAWCEEADcUC7wgFgtBVcv4AHMAC+udOEf2-n-ABE9r7T0wAAEQgP-dAcBgBQMwDAuB0AkEoLQYQrBk8cGYDnovZe6DiEoPIZQ6h1dNHUPrhQxhUBP7oAALakOrtnKAljwEoFYWY5hbD2GcO4QfHeKAq7QF-ugT+9dgDv1gVAZxwCa4AAoACUKjBEgPvuA6Rsj4F91UUI0R4jJHJNgZXN+Ji4HACgL4zAEiil4HCWomJb936-ygJEkpZSAB0Sj264BwJ3fA0SgA) with the all status codes idea if you want to play around and see ther UX

In my experience, the most common scenario for error checking is to check if it is an error or not but, no the specific error as the OpenAPI standard doesn't provide a way to define each error, just a general shape.

I'm open to exploring possibilities with you and postponing this for v1.1. However, I'd really like to merge the change to generate "default" instead of "500" for v1.0

test/rlcIntegration/bodyComplextRest.spec.ts
joheredi: I was talking to @bterlson and I'd like to propose a slightly different approach. Which is:

* We don't generate a missleading status code. For the default case we give status a type of `string` which represents correctly an unknown set of strings
   * This will initially make TS collapse the type of status for both types the Success and Error to string, and intellisense provide no suggestions. Here's where `isUnexpected` comes to help.
* As soon as we validate with `isUnexpected` we can narrow down the types and get suggestions again.


Here are 2 playgrounds for the current vs proposed


[Current](https://www.typescriptlang.org/play?#code/PTAEBUAsEsGdTqAhqATgU1gBwPYDtZ010sNC8AXaPAc2VFgFcBjZzWAOgCgRQB5PABsAnqABMABgkAacRICMyPABM5AFmQZQzfLGjL0GVU1btMPMEjyicFSIYYUkFRvB0HQAd2iDBoAEZEOgT6huiqVqCGqDioXNQUhgBmSGygAOLoFACqhKiwAMosbLCwAEqYuAREAN5coI7OrgBcoABEkhJtoAA+7ZLy3X0dEmpt0vUBOMrCrTUAbkiCjOitsBSo1DQAvly7FhAw8IgoNOh4hktRqDGo2kiEsp5EyvgA5BSgANZ4OJ4B6CSsSIkCsygOnhgzEgjRcbmmRGeb1UZ0+sBwXiIriIbQMKUYggobQOkD+6HmDjxSAJn0Qv0+GCWInoi0E+lAkAoFCwsNc2gR3ASyVSREyOTysAAIoDqYSKthdLVJusmrBWriZTTxpN-NNZqAatFYnN3KtQHhGABbQKoADcoEt7CQZzWGy2212+wOAEkvFZPhQMYFkP5BERA6AaIwkKhVHZw8IsIjoHZQHZEElGHhmFR8EpVBgXKg8Gn7GnE+HQZ9QfAVXCAMII5DwKyidabWigIF3bH5NNBohIUPhnAHaHoZhfLuxZDMRhOPx1vmmzhcTPZ3MluDZC4ADyTOfCAAoyFVCK0xblDIViux5WeiH1LxLpfi5ZVFaAAJStU+fxDPter6yhQ96fnUDQNIWjDFqAACEADaIxdLIIyDKhkhiG0AC6HDUMwywGLAJ4ftUHBLrAX57FwBwALI4JOXZZjm0B5nYzjEEWBD0H+1RcAYBExkQ64sXmqJXvkR4-hkWQSTepilGB1S9DJ4pAZq74KtUto0bwYpUJ28bEFphBcME6zGQ+oAALyRrJEpSTpBwAJo4IwHJIBSvLwh4TA0Gc6ysdxSQxJapZEKikY4HBBzgKgojMKgFB+AA1I4WAiggJZGeglpYBQogAI6MLYmDcOZOBhhwgg4DQJEmeg5FOHCNnWbZbRtFR8RJKAR7bnuB6JMo9UPl+X4GpMDS8K57nMFYHzEEyCX2Ix3YMAe0BJNAzD8kRXgpqS873IFhkYq2HCgM0bxHpNCBJCNipNaqrXtQArAonUTZB338gQlWNTVdVtOkzjoJ4SCiEaqCdbd+w-bwUCIIgrLsmtRk6HlPiGLI-hHd4vjmuSDiglgSYlg8oBUjSCDwPSPHoFcAAS4DgAACt5u3oLdm0PWRFEve0VOEp9EE-Q0FVVYDR5tIj8D434FwUncJNkzD31w5BdgxP8vGEBwuozBwUM6V6EsA7VvN6wbwgcKyKxfk5QA) vs [Proposed](https://www.typescriptlang.org/play?#code/PTAEBUAsEsGdTqAhqATgU1gBwPYDtZ010sNC8AXaPAc2VFgFcBjZzWAOgFgAoEUAPJ4ANgE9QAJgAMUgDSSpARmR4AJgoAsyDKGb5Y0Vegzqmrdpl78kecTgqRjDCkgqN4eo6ADu0YcNAAIyI9AkNjdHUbUGNUHFReagpjADMkNlAAcXQKAFVCVFgAZRY2WFgAJUxcAiIAb15QZ1d3AC5QACJpKQ7QAB9O6UVega6pDQ7ZRqCcVVF2uoA3JGFGdHbYClRqGgBfXn2eKzAoREQUGnQ8YxWY1DjUXSRCeW8iVXwAcgpQAGs8HDeILoFLxIiQGyqY4+GDMSDNNweWZEN6fdSXH6wHA+IjuIgdIxpRjCCgdaGQQHoRZOQlIYk-RAAn4YFZiejLYSGUCQCgULAI9y6ZHcHhJVLpIjZPIFWAAERBdJJVWw+nq002LVgGy2OymPCagVm81AdVi8QWnnWoDwjAAtsFUABuUC29hIS7a7a0XaHQ68aEASR8Nh+FGxwWQgWERDDoBojCQqHUDhjoiwKOgDlADkQKUYeGYVHwKnUGDcqDw2cc2bTMYhPwh8A1iIAwsjkPAbOJNl66KDHnjCtnw0QkFGYzhoXD0MxfqB+8hmIwXAFm4LLZxeHmC0XK3BctcAB7pwuRAAUZBqhHaUvyxmKpXYyqvRAGt5l8qJSuqqtAAEp2kvX9EHfe9P0VChn1-Bp9SaYhy0rABCABtMYenkMZhgw6QJA6ABdDhqGYVYjFgC8f1qDg11gP8Dn9PgwAAWRwWd53zQtoGLBxXHgxgK07YgVVqXgjGIxMiG3DjiwxO9CjPACshyWSH3McooNqfpFOlMCFXpdTCEdej+ClKhaCrIggOEnhQk2QSX1AABeOMlJleTDKOBjQFEHBGFAYQcBwQhzIFJEvAwPRbVdNRXE4ghgX7Ig8ETOJfDMj5vDwaFwFQcRmFQCgAgAamcLAJQQSsUxiW0sAocQAEdGHsTAOFAAA5bEmBoS5Nli2BeBsnBow4fyaHIoTCColxEUchynI6DpaI86AUlAM99yPE9klUMaXz-P8TWmJp+AATR8p48G+YhWVyxxWIXbAZ2W6BmCFUifEzCllyeHqzNjLtDoQFIdtVSbNRmuaAFYlAWg7YLguCBqGkazw6TJXHQbwkHEM1UAWgG-Tho6ThgeBEA5LkF0qiKsD8Yx5ECL7fH8a0qScCEsHTStnlAWl6QQeAmXoFkAgACXAcAAAUQte9AAeW4HKOo8HOl5kkYZg+H4cR9BhpwUaOlOeAmYCa5qUednObxwmCfhhwUrskHDTmDgcfcgn+BFiIvLOiFqWly0GEYLrMF3eBWkW-hsty-KipKsrqGC9Bqtq0AGqazdrP0QadeRyyJqV2a5qt7XddGvOdad0QOA5NY-3c3ggA)



/cc - @bterlson. @lirenhe, @MaryGao, @qiaozha

test/rlcIntegration/generated/bodyComplexRest/src/responses.ts
MaryGao: >In my experience, the most common scenario for error checking is to check if it is an error or not but, no the specific error as the OpenAPI standard doesn't provide a way to define each error, just a general shape.

Yeah, I think the most common scenario is to refer a general shape and we have a swagger extension - [x-ms-error-response](https://github.com/Azure/autorest/tree/main/docs/extensions#x-ms-error-response) which could allow service team to define different error shape.

test/rlcIntegration/bodyComplextRest.spec.ts
MaryGao: Compared with the `default` type I would prefer the native `string` type more. That would be good to narrow down the error types without casting or any type checking errors with the helper func.

qiaozha: Just have one simple question, if isUnexpected, should the error and message be the properties of result not result.body ? 
Otherwise, this proposal looks good to me. @lirenhe  how do you think of it?

lirenhe: As I know, Go SDK uses result.properties but I can't say which one is better.
And overall, the proposal looks good to me. Thanks @joheredi for the update.

qiaozha (approved):
LGTM

test/rlcIntegration/generated/bodyComplexRest/src/responses.ts
joheredi: We'd still support x-ms-error-response. Customers will need to check for specific status codes if they want to handle each one specifically.

```ts
if (isUnexpected(response)) {
   if (response.status === "401") {
      throw response.body.authError;
   }

  if (response.status === "400") {
      throw response.body.userError;
  }

   throw response.body.generalError;
} 
```

PR #1419, opened 2022-06-15
[RLC] Remove typedoc and disable docs generation

joheredi:
Remove typedoc and disable docs generation

### DIFF
--- a/src/generators/static/packageFileGenerator.ts
+++ b/src/generators/static/packageFileGenerator.ts
@@ -134,8 +134,7 @@ function restLevelPackage(packageDetails: PackageDetails) {
       "unit-test": "echo skipped",
       "unit-test:node": "echo skipped",
       "unit-test:browser": "echo skipped",
-      docs:
-        "typedoc --excludePrivate --excludeExternals  --out ./dist/docs ./src"
+      docs: "echo skip"
     },
     sideEffects: false,
     autoPublish: false,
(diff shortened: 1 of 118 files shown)


src/generators/static/packageFileGenerator.ts
```diff
@@ -134,8 +134,7 @@ function restLevelPackage(packageDetails: PackageDetails) {
       "unit-test": "echo skipped",
       "unit-test:node": "echo skipped",
       "unit-test:browser": "echo skipped",
-      docs:
-        "typedoc --excludePrivate --excludeExternals  --out ./dist/docs ./src"
+      docs: "echo skip"
```
deyaaeldeen, line 137: Is there a need to keep the script?

deyaaeldeen (approved):

src/generators/static/packageFileGenerator.ts
joheredi: We can remove it. Do you know if rush would complain? If not I'll get rid of it

deyaaeldeen: I don't believe rush will complain about the lack of the docs script. I don't have it in my libraries and I didn't see any complaints regarding it. Also for context, we used to have that script but we decided to remove it at some point a few months ago.

joheredi: I'll remove it

qiaozha (approved):

PR #1420, opened 2022-06-15
Generate interface instead of intersection for object models

deyaaeldeen:
Generated object models will change from:

```ts
export type CatAPTrue = PetAPTrue & {
  friendly?: boolean;
};
```
to
```ts
export interface CatAPTrue extends PetAPTrue {
  friendly?: boolean;
}
```

This change addresses feedback received during review for the new API for the text analytics library where a lot of the generated models are exported as part of the public surface. Instead of recreating all those models as interfaces manually, this PR updates the code generator to generate them as interfaces from the get-go.


test/integration/generated/textanalytics/src/models/index.ts
```diff
@@ -525,51 +525,54 @@ export interface SentenceAssessment {
...
+  extends AnalyzeJobMetadata,
+    TasksState,
+    Pagination {
+  errors?: TextAnalyticsError[];
+  /** if includeStatistics=true was specified in the request this field will contain information about the request payload. */
+  statistics?: TextDocumentBatchStatistics;
+}
 
-export type TasksStateTasksDetails = TaskState;
+export interface TasksStateTasksDetails extends TaskState {}
```
deyaaeldeen, line 551: perhaps generate an interface only if the list of properties is not empty.

joheredi: I think practically it is the same as having the alias. So I think I prefer the simplicity of the new code over handling the case to generate the alias.

The ideal thing may be to not generate it at all and give anything that has a `TasksStateTasksDetails` type  `TaskState` instead. But I think that work would be a lot more involved.

I'm fine with the new generated interface

joheredi (approved):
I think this looks great! Thanks for taking the time to send out the fix.

@qiaozha would you mind taking a look as well please?

qiaozha (approved):
LGTM

witemple-msft (approved):
This will _extremely significantly_ improve the quality of generated documentation for packages that export generated types.

Thank you @deyaaeldeen for looking into this and LGTM.

test/integration/generated/textanalytics/src/models/index.ts
deyaaeldeen: Thanks for your valuable perspective! I agree with all what you said :)

PR #1423, opened 2022-06-17
fix readme h1 title in mgmt plane sdk

qiaozha:
fix https://github.com/Azure/azure-sdk-for-js/issues/22197
fix https://github.com/Azure/azure-sdk-for-js/issues/22203


qiaozha:
@joheredi @sarangan12 Can you help take a look at this PR ? Thanks

src/generators/static/readmeFileGenerator.ts
```diff
@@ -96,7 +96,14 @@ function createMetadata(
   const packageDirectoryName = names?.[1];
 
   const clientPackageName = packageDetails?.name;
-  const { clientClassName, serviceTitle } = getClientAndServiceName(codeModel.language, codeModel.info);
+  const { clientClassName, serviceTitle } = getClientAndServiceName(
+    codeModel.language,
+    codeModel.info
+  );
+  const simpleServiceNameTmp =
```
sarangan12, line 103: Can't we just name it `serviceName`/`simpleServiceName`/`heading` or anything without Tmp?

src/generators/static/readmeFileGenerator.ts
```diff
@@ -82,7 +80,9 @@ function createMetadata(
   } = getAutorestOptions();
   const { addCredentials } = getSecurityInfoFromModel(codeModel.security);
 
-  const azureHuh = packageDetails?.scopeName === "azure" || packageDetails?.scopeName === "azure-rest";
+  const azureHuh =
```
sarangan12, line 83: what exactly is azure`Huh`? Can we use a meaningful name?

sarangan12 (approved):
Left 2 nit comments. Otherwise, I am fine with the changes

src/generators/static/readmeFileGenerator.ts
qiaozha: azureHub : true means package names that are @azure or @azure-rest scope, in our smoke test, we are renaming all the packages with the  @msinternal scope. where the azureHub should be false.

qiaozha: Do you have any suggestions on a meaningful name ?

src/generators/static/readmeFileGenerator.ts
qiaozha: changed

PR #1424, opened 2022-06-17
Removing lookbehind RegEx in AutoRest

### LINKED ISSUE #1422: Browser: Lookbehind and Negative Lookbehind RegEx not supported in Safari
The following code that is generated in the `SearchClient`, the `addCustomApiVersionPolicy` uses a lookbehind RegEx, which is not supported on Safari via [CanIUse](https://caniuse.com/js-regexp-lookbehind).  This has caused an issue with the Azure SDK for JavaScript:

https://github.com/Azure/azure-sdk-for-js/issues/22274

mpodwysocki:
Fixes #1422

Lookbehind and negative Lookbehind regular expressions are [not supported in Safari](https://caniuse.com/js-regexp-lookbehind).  Removing in favor of simple string replacement.

cc @qiaozha


joheredi (approved):
Thanks for sending out this fix @mpodwysocki

Issue #930, opened 2021-05-07
[typescript] Endpoint with string content produces unbuildable code


craignicol:
With the following input:
```
openapi: 3.0.1
info:
  title: Data
  description: Services for general drinkaware data
  version: 1.0.0
paths:
  /:
    get:
      operationId: Home
      summary: Get Home Page
      responses:
        '200':
          description: Home page
          content:
            text/plain; utf-8:
              schema:
                type: string
                pattern: '^.*$'
                maxLength: 1000
                example: unauthorized 
```

The typescript emitter creates a `data.ts` file with the following specification:

```
const homeOperationSpec: coreHttp.OperationSpec = {
  path: "/",
  httpMethod: "GET",
  responses: {
    200: {
      bodyMapper: {
        type: { name: "String" },
        constraints: { MaxLength: 1000, Pattern: {} }
      }
    }
  },
  urlParameters: [Parameters.$host],
  headerParameters: [Parameters.accept],
  serializer
};
```

The `bodyMapper` property cannot be mapped to a valid type:

```
      TS2322: Type '{ type: { name: "String"; }; constraints: { MaxLength: number; Pattern: {}; }; }' is not assignable to type 'BaseMapper | CompositeMapper | SequenceMapper | DictionaryMapper | EnumMapper | undefined'.
  Type '{ type: { name: "String"; }; constraints: { MaxLength: number; Pattern: {}; }; }' is not assignable to type 'EnumMapper'.
    Types of property 'type' are incompatible.
      Property 'allowedValues' is missing in type '{ name: "String"; }' but required in type 'EnumMapperType'.
```

**Expected behavior**
Valid YAML produces a valid typescript file

**Additional context**
```
AutoRest code generation utility [cli version: 3.1.2; node: v12.18.2, max-memory: 2048 MB]
(C) 2018 Microsoft Corporation.
https://aka.ms/autorest

There is a new version of AutoRest available (3.2.0).
 > You can install the newer version with with npm install -g autorest@latest

   Loading AutoRest core      '/home/craignicol/.autorest/@autorest_core@3.0.6375/node_modules/@autorest/core/dist' (3.0.6375)
   Loading AutoRest extension '@autorest/typescript' (6.0.0-dev.20201105.2->6.0.0-dev.20201105.2)
   Loading AutoRest extension '@autorest/modelerfour' (4.15.421->4.15.421)
```

sarangan12:
[...]

PR #1426, opened 2022-06-20
Remove Constraints Generator for Response Objects

sarangan12:
This PR is a result of the issue https://github.com/Azure/autorest.typescript/issues/930. Basically, the user has created a swagger/yaml file, which has a `Pattern` constraint in the response object. This is not handled correctly which causes a build failure of the generated SDK. (The `Pattern` constraint in the request object is handled correctly).

On analysis, I have found that it is indeed the case. We simply use `JSON.stringify(mapper)` which does not work correctly on `Pattern` constraint. We should be using the `new RegExp(...)` expression for the the pattern. 

So, now I have 2 options in front of me:

1. Split the `Pattern` constraint (if available) and use the correct mapping of `new RegExp(...)`. 
2. On the other hand, we should not be handling the constraints check even on the request object (The validation code has already been removed from the `core-client` package) We do not even have a constraint validation on the response object anyway. So, adding any constraints on the response object is simply unnecessary. So, I believe the correct approach is to ignore the constraints from the response object, in the generation phase itself. 

This PR takes the 2nd approach and implements it. 

@joheredi @xirzec Please review and approve the PR. Thanks


xirzec (approved):
Thanks for driving this work upstream!

test/integration/generated/appconfiguration/package.json
```diff
@@ -77,7 +77,8 @@
     "unit-test:browser": "echo skipped",
     "integration-test": "echo skipped",
     "integration-test:node": "echo skipped",
-    "integration-test:browser": "echo skipped"
+    "integration-test:browser": "echo skipped",
+    "docs": "echo skipped"
```
joheredi, line 81: Is adding "docs": "echo skipped" an intentional change or a merge conflict resolution problem?

joheredi (approved):
The specific changes look good. There seem to be some merge conflicts in the generated files.

Issue #1414, opened 2022-06-10
Add test coverage for security scopes in testserver


qiaozha:
As we have supported security scopes in the autorest.tyoescript, though we have tests in our own, but we still need to support the test coverage in testserver.  there are already two in the testserver:
https://github.com/Azure/autorest.testserver/blob/main/swagger/security-aad.json
https://github.com/Azure/autorest.testserver/blob/main/swagger/security-key.json
The REST API doc for AAD is not quite right as it is missing the `scopes` … I’ll make a PR to fix that.
 
We should add two more cases:
_Both_ AAD and API Key (either one is allowed).  This is common for Cognitive Services.
Two different API Keys which are both required.  This is not common but Metrics Advisor has this.

MaryGao:
The first part is high priority but the second one we need to check offline maybe low priority

PR #1430, opened 2022-06-22
add-test-coverage-for-security-add-and-security-key

qiaozha:
fix https://github.com/Azure/autorest.typescript/issues/1414

### DIFF
--- a/test/rlcIntegration/securityAadRest.spec.ts
+++ b/test/rlcIntegration/securityAadRest.spec.ts
@@ -1,0 +1,9 @@
+import SecurityAADRestClient from "./generated/securityAADRest/src";
+import { assert } from "chai";
+
+describe("Security AAD Rest Client", () => {
+  it("should be able to instantiate a new client without Typescript errors", () => {
+    const client = SecurityAADRestClient("subscriptionId" as any);
+    assert.isDefined(client);
+  });
+});
--- a/test/rlcIntegration/securityAadRest.spec.ts
+++ b/test/rlcIntegration/securityAadRest.spec.ts
@@ -1,9 +1,15 @@
 import SecurityAADRestClient from "./generated/securityAADRest/src";
 import { assert } from "chai";
+import { now } from "lodash";
 
 describe("Security AAD Rest Client", () => {
   it("should be able to instantiate a new client without Typescript errors", () => {
-    const client = SecurityAADRestClient("subscriptionId" as any);
+    const client = SecurityAADRestClient({
+      getToken: async () => {
+        assert.equal(1, 1);
+        return { token: "FakeToken", expiresOnTimestamp: now() };
+      }
+    });
     assert.isDefined(client);
   });
 });
--- a/test/rlcIntegration/securityKeyRest.spec.ts
+++ b/test/rlcIntegration/securityKeyRest.spec.ts
@@ -1,0 +1,9 @@
+import SecurityKeyRestClient from "./generated/securityKeyRest/src";
+import { assert } from "chai";
+
+describe("Security AAD Rest Client", () => {
+  it("should be able to instantiate a new client without Typescript errors", () => {
+    const client = SecurityKeyRestClient("subscriptionId" as any);
+    assert.isDefined(client);
+  });
+});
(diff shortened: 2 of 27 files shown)


test/rlcIntegration/securityAadRest.spec.ts
```diff
@@ -0,0 +1,9 @@
+import SecurityAADRestClient from "./generated/securityAADRest/src";
+import { assert } from "chai";
+
+describe("Security AAD Rest Client", () => {
+  it("should be able to instantiate a new client without Typescript errors", () => {
+    const client = SecurityAADRestClient("subscriptionId" as any);
```
qiaozha, line 6: I wasn't very sure how to create a fake TokenCredential here.

test/rlcIntegration/securityKeyRest.spec.ts
```diff
@@ -0,0 +1,9 @@
+import SecurityKeyRestClient from "./generated/securityKeyRest/src";
+import { assert } from "chai";
+
+describe("Security AAD Rest Client", () => {
+  it("should be able to instantiate a new client without Typescript errors", () => {
+    const client = SecurityKeyRestClient("subscriptionId" as any);
```
qiaozha, line 6: Same here, I wasn't very sure how to create a fake KeyCredential here.

test/rlcIntegration/securityAadRest.spec.ts
joheredi: I think you should be able to pass an object like

{getToken: async () => {assert.equal(validate, something); return { accessToken: "FakeToken"}}}

joheredi (commented):
How do we handle anonymous credentials and combination of credentials?

I would think we could do that with overloads?

test/rlcIntegration/securityAadRest.spec.ts
qiaozha: Updated.

test/rlcIntegration/securityKeyRest.spec.ts
qiaozha: updated

qiaozha:
> How do we handle anonymous credentials and combination of credentials?
> 
> I would think we could do that with overloads?

What exactly is anonymous credentials ? we haven't get the case of combination of credentials swagger in testserver ready yet. but the generated code would look like 
```
  credentials: TokenCredential | KeyCredential,
  options: ClientOptions = {}
):
```
in that case, we will write a test to call createClient twice one for TokenCredential and one for KeyCredential.

joheredi:
> > How do we handle anonymous credentials and combination of credentials?
> > I would think we could do that with overloads?
> 
> What exactly is anonymous credentials ? we haven't get the case of combination of credentials swagger in testserver ready yet. but the generated code would look like
> 
> ```
> export default function createClient(
>   credentials: TokenCredential | KeyCredential,
>   options: ClientOptions = {}
> ):
> ```
> 
> in that case, we will write a test to call createClient twice one for TokenCredential and one for KeyCredential.

I think anonymous credentials mean that credentials are optional IIRC the spec allows AnonymousCredential as well. I would imagine for that we'd generate an overload that doesn't take any credentials

qiaozha:
In that case, we only need to set the add-credentials: false, I remember there's already a case for that, let me find one.

qiaozha:
https://github.com/Azure/autorest.typescript/blob/main/test/rlcIntegration/generated/urlRest/src/urlRestClient.ts#L7-L9 here's an example of we set add-credentials: false, and how the client factory function looks like

joheredi:
I'm thinking of the scenario where both TokenCredentials and AnonymousCredentials are defined in the swagger

joheredi (approved):
If reasonable, please add the Anonymous credential support in this PR. Otherwise, can you file an issue for that and maybe try to prioritize?

qiaozha:
> If reasonable, please add the Anonymous credential support in this PR. Otherwise, can you file an issue for that and maybe try to prioritize?

create an issue here https://github.com/Azure/autorest.typescript/issues/1447 to track the support of anonymous credential

Issue #1395, opened 2022-05-19
[mgmt] the generated mappers for discriminators is incorrect


qiaozha:
For consumption mgmt plane. we generate the discriminator mappers like this 
//mappers file
```
export let discriminators = {
  LegacyReservationRecommendationProperties: LegacyReservationRecommendationProperties,
  "Resource.UsageDetail": UsageDetail,
  "Resource.ReservationRecommendation": ReservationRecommendation,
  "ProxyResource.ChargeSummary": ChargeSummary,
  "LegacyReservationRecommendationProperties.Single": LegacySingleScopeReservationRecommendationProperties,
  "LegacyReservationRecommendationProperties.Shared": LegacySharedScopeReservationRecommendationProperties,
  "Resource.legacy": LegacyReservationRecommendation,
  "Resource.modern": ModernReservationRecommendation,
  "ProxyResource.legacy": LegacyChargeSummary,
  "ProxyResource.modern": ModernChargeSummary
};
```

Here the UsageDetail, ReservationRecommandation are both subclass of Resources and each of them has subclesses with x-ms-discriminator-value legacy and modern, 

Currently we are not generating the mapper key by using the direct parent class name. instead we are using the very first ancestors' name in this case Resource. which cause both LegacyUsageDetail and LegacyReservationRecommandation have the same key Resource.Legacy. 

we should generate the mappers like 
```
export const discriminators = {
  'UsageDetail': UsageDetail,
  'UsageDetail.legacy': LegacyUsageDetail,
  'UsageDetail.modern': ModernUsageDetail,
  'ReservationRecommendation': ReservationRecommendation,
  'ReservationRecommendation.legacy': LegacyReservationRecommendation,
  'ReservationRecommendation.Nodern': ModernReservationRecommendation,
  'ChargeSummary': ChargeSummary,
  'ChargeSummary.legacy': LegacyChargeSummary,
  'ChargeSummary.modern': ModernChargeSummary,
}
```

This is original from our customer reported issue here https://github.com/Azure/azure-sdk-for-js/issues/20026 

To do work:
1. we need to add some test case in the test server to cover such scenario 
2. we need to fix it in this typescript code generator 
3. re-release a new package of @azure/arm-consumption for our customer.

BoostSebastian:
Hi, do we have any future release on this?

qiaozha:
Hi @BoostSebastian I have a PR to fix it and will let you know when it finishes ?

PR #1432, opened 2022-06-27
fix-discriminator-mapper generation issue

qiaozha:
fix https://github.com/Azure/autorest.typescript/issues/1395


joheredi (approved):
Looks good to me @sarangan12 can you also take a look please?

packages/autorest.typescript/test/smoke/generated/graphrbac-data-plane/review/graphrbac-data-plane.api.md
```diff
@@ -283,7 +283,7 @@ export interface DirectoryObject {
     [property: string]: any;
     readonly deletionTimestamp?: Date;
     readonly objectId?: string;
-    objectType: "User" | "Application" | "Group" | "ServicePrincipal";
+    objectType: string;
```
qiaozha, line 286: This is expected. DirectoryObject has four subclasses, "User" , "Application",  "Group" , "ServicePrincipal", but they didn't specify x-ms-discriminator-value, in this case the discriminator value for this four subclasses should have the value of the  class name itself. but that doesn't mean objectType can only have this four values.

packages/autorest.typescript/test/smoke/generated/graphrbac-data-plane/src/models/mappers.ts
```diff
@@ -2282,7 +2282,7 @@ export const UserUpdateParameters: coreClient.CompositeMapper = {
 };
 
 export let discriminators = {
-  DirectoryObject: DirectoryObject,
+  "DirectoryObject.DirectoryObject": DirectoryObject,
```
qiaozha, line 2285: This may be unexpected.

sarangan12 (approved):

packages/autorest.typescript/test/smoke/generated/graphrbac-data-plane/src/models/mappers.ts
qiaozha: fixed

Issue #1389, opened 2022-05-12
[RLC] Sample generation: model design & transformation


MaryGao:
fixes https://github.com/Azure/autorest.testserver/pull/381
fixes https://github.com/Azure/autorest.typescript/pull/1435

We'd like to support sample generation in RLC codegen and it will support all cases in below:

- Click to view list: #1393 

In this issue we'll cover the body only has `string` type

- Design TestModel for RLC
- Support below actions
  - GET
  - POST
  - DELETE
  - PUT

PR #1435, opened 2022-06-30
Sample generation in phase 1

MaryGao:
In our first phase to implement sample auto-generation we'll provide features:

- Normal GET/POST/PUT/PATCH/... cases
- LRO
- Pagination

Include integration test for below folder:
- bodyStringRest
- dpgCustomization
- customUrl
- headerRest

Include smoke test folder:
- agrifood-data-plane

Have a dependency on the pr to update swagger examples: https://github.com/Azure/autorest.testserver/pull/395

fixes https://github.com/Azure/autorest.typescript/issues/1389

Gist here: https://gist.github.com/MaryGao/0949bb40c552dcd1d6866b671953d6e4


test/rlcIntegration/generated/lroRest/package.json
```diff
@@ -82,11 +82,5 @@
     "rollup": "^2.66.1",
     "rollup-plugin-sourcemaps": "^0.6.3",
     "uglify-js": "^3.4.9"
-  },
-  "//sampleConfiguration": {
```
qiaozha, line 86: Was this generated before ?

test/rlcIntegration/generated/lroRest/tsconfig.json
```diff
@@ -19,8 +19,7 @@
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
     "outDir": "./dist-esm",
-    "declarationDir": "./types",
-    "paths": { "@msinternal/lro-rest": ["./src/index"] }
+    "declarationDir": "./types"
   },
-  "include": ["./src/**/*.ts", "samples-dev/**/*.ts"]
```
qiaozha, line 25: why do we have this if we are removing sampleConfiguration in the package.json

src/generators/static/tsConfigFileGenerator.ts
```diff
@@ -81,9 +83,11 @@ export function generateTsConfig(project: Project) {
     highLevelTsConfig.include.push("samples-dev/**/*.ts");
     highLevelTsConfig.compilerOptions["paths"] = {};
     highLevelTsConfig.compilerOptions["paths"][clientPackageName] = ["./src/index"];
-    restLevelTsConfig.include.push("samples-dev/**/*.ts");
-    restLevelTsConfig.compilerOptions["paths"] = {};
-    restLevelTsConfig.compilerOptions["paths"][clientPackageName] = ["./src/index"];
+    if (hasRLCSamplesGenerated) {
```
qiaozha, line 86: why do we need one more condition for this ?

src/generators/samples/rlcSampleGenerator.ts
```diff
@@ -17,27 +29,392 @@ export function generateRLCSamples(model: TestCodeModel, project: Project) {
...
+function getAssignmentStrArray(parameters: SampleParameter[]) {
+    return parameters.filter(p => !!p.assignment).map(p => p.assignment!);
+}
+
+function getContactParameterNames(parameters: SampleParameter[]) {
+    return parameters.filter(p => p.name != null).map(p => p.name!).join(',');
 }
 
-function createSampleData(model: TestCodeModel) {
+function getParameterAssignment(exampleValue: ExampleValue) {
```
qiaozha, line 290: This function looks familiar. Do we have a similar one in high level sample generation ? Can we reuse ?

src/generators/samples/rlcSamples.ts.hbs
```diff
@@ -1,28 +1,47 @@
...
+  {{this}}
+  {{/each}}
+  {{#if this.isPaging}}
+  const initialResponse = await client.path({{this.pathParamNames}}).{{method}}({{methodParamNames}});
+  const pageData = paginate(client, initialResponse);
+  const result = [];
+  for await (const item of pageData) {
+    result.push(item);
+  }
+  {{else if this.isLRO}}
```
qiaozha, line 35: what if a method is both LRO and Paginate ?

src/generators/static/packageFileGenerator.ts
```diff
@@ -253,9 +253,7 @@ function restLevelPackage(packageDetails: PackageDetails) {
   }
 
   if (
-    generateSample &&
-    (model as TestCodeModel)?.testModel?.mockTest?.exampleGroups &&
-    (model as TestCodeModel).testModel!.mockTest!.exampleGroups.length > 0
+    generateSample && hasRLCSamplesGenerated
```
qiaozha, line 256: if no examples provided why do you we need to generate the sampleConfiguration ?

test/rlcIntegration/generated/lroRest/package.json
MaryGao: Yes, it is introduced by the first time I init the sample option with a prototype sample file. This time I disable this option for LRORest.

src/generators/static/tsConfigFileGenerator.ts
MaryGao: To ensure we do have samples to generated. Sometimes we enable sample option but don't have any sample to be generated. In HLC sample gen we use the condition directly `clientDetails?.samples?.length && clientDetails?.samples?.length > 0`

src/generators/static/packageFileGenerator.ts
MaryGao: The condition `hasRLCSamplesGenerated` will be truthy if and only if we have examples provided and we could transform to valid sample models. 

So we will not generate sampleConfiguration if there is no examples provided.

test/rlcIntegration/generated/headerRest/tsconfig.json
```diff
@@ -19,7 +19,8 @@
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
     "outDir": "./dist-esm",
-    "declarationDir": "./types"
+    "declarationDir": "./types",
+    "paths": { "@msinternal/header-rest": ["./src/index"] }
   },
-  "include": ["./src/**/*.ts"]
+  "include": ["./src/**/*.ts", "samples-dev/**/*.ts"]
 }
```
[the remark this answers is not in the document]
MaryGao: We have introduced that change last time when we enable the sample option

test/rlcIntegration/generated/lroRest/tsconfig.json
MaryGao: This is the case where we enable the sampleGenerated option but no examples provide in testserver. I'll add examples in phase 2 for this swagger.

src/generators/samples/rlcSampleGenerator.ts
MaryGao: Yes, this is copied directly from HLC sample gen with slightly change for property name. Considering there is no any UnitTest could I do the refactoring in Phase 2 where I'll introduce more coner cases?

src/generators/samples/rlcSamples.ts.hbs
MaryGao: Currently it will go to the first matched condition `this.isPaging` if both LRO and Paginate. Another way to handle this is to ignore this case in transformation because we don't support this now. Which one do you prefer?

qiaozha (approved):

PR #1436, opened 2022-07-01
POC for abstracting generation model

joheredi:
Extracting file generation to a separate library and scoping down the responsibilities of @autorest/typescript to transform M4 code-model into a new RLC Model.

A new CADL Emitter for RLC would also create a RLCModel and consume the "generation lib" to get RLC. With this we can guarantee that outputs from Autorest and CADL would be the same.


src/lib/ts-rlc-codegen/generateClientDefinitions.ts
```diff
@@ -0,0 +1,140 @@
...
+    signatures.push({
+      docs: [
+        `Resource for '${key
+          .replace(/}/g, "\\}")
+          .replace(
+            /{/g,
+            "\\{"
+          )}' has methods for the following verbs: ${Object.keys(
+          paths[key].methods
+        ).join(", ")}`
```
deyaaeldeen, line 115: nitpick: Perhaps consider pulling this logic into a helper function?

src/lib/ts-rlc-codegen/helpers/operationHelpers.ts
```diff
@@ -0,0 +1,47 @@
...
+  const methodDefinitions: OptionalKind<MethodSignatureStructure>[] = [];
+  for (const key of Object.keys(methods)) {
+    const method = methods[key];
+    const description = methods[key][0].description;
+
+    let areAllOptional = methods[key][0].hasOptionalOptions;
+
+    methodDefinitions.push({
+      name: key,
+      ...(description && { docs: [{ description }] }),
```
deyaaeldeen, line 21: Is it guaranteed that those methods will always be documented? otherwise I am afraid undocumented ones will be flagged as undocumented by the API extractor tool making the review file less pleasant to read. Did you consider creating a default documentation for undocumented methods? maybe just the method name could work?

deyaaeldeen (commented):
Looks good! Exciting to see ground work for a CADL front-end 🚀 I left a few nitpicky comments and can approve after CI passes.

PR #1443, opened 2022-07-12
fix-array-type-payload

qiaozha:
This PR is trying to resolve issues like this 
image
we should just import the element type

### DIFF
--- a/src/transforms/samplesTransforms.ts
+++ b/src/transforms/samplesTransforms.ts
@@ -1,349 +1,361 @@
 import {
-    ChoiceSchema,
-    CodeModel,
-    ConstantSchema,
-    ImplementationLocation,
-    Operation,
-    OperationGroup,
-    SchemaType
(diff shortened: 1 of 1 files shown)


src/transforms/samplesTransforms.ts
```diff
@@ -1,349 +1,361 @@
...
+                bodySchemaName = "Record<string, unknown>";
+              } else if (
+                methodParameter.exampleValue.schema.type !==
+                SchemaType.Dictionary
+              ) {
+                if (parameterTypeName.endsWith("[]")) {
+                  importedTypeSet.add(
+                    parameterTypeName.substring(0, parameterTypeName.length - 2)
+                  );
+                } else {
```
qiaozha, line 202: This is only part I changed. the rest part is changing because I format this file.

joheredi (approved):

PR #1471, opened 2022-07-27
add-widget-service-in-autorest-typescript-smoke-test

qiaozha:


qiaozha:
close it as no need now

PR #1475, opened 2022-08-02
Move common code gen parts to shared package

joheredi:


packages/rlc-codegen/src/helpers/camelCase.ts
```diff
@@ -0,0 +1,14 @@
...
+  const { uppercaseThreshold = 4 } = options;
+  const thresholdRegex = new RegExp(
+    `^(?<![A-Z])[A-Z]{1,${uppercaseThreshold}}(?![A-Z])`
+  );
+  if (!thresholdRegex.test(str)) {
+    return str;
+  }
+
+  return str.charAt(0).toLocaleLowerCase() + str.slice(1);
+}
```
qiaozha, line 14: Should we combine camelCase.ts and pascalCase.ts to the same file such as nameUtils.ts ?

packages/cadl-rlc-emitter/src/index.ts
```diff
@@ -1,55 +1,34 @@
...
-                        ),
-                        optionsName: 'options',
-                        responseTypes: { success: ['string'], error: [] },
-                        returnType: '',
-                        successStatus: ['200']
-                    }
-                ]
-            }
-        };
+  const paths: Paths = transformPaths(program);
```
qiaozha, line 7: Just wonder if we should have transform function wrap of the transformPaths and this transform function returns back the whole rlcModel.

packages/cadl-rlc-emitter/src/index.ts
```diff
@@ -1,55 +1,34 @@
...
+      options: {
+        // TODO: Read configuration from CADL
+        includeShortcuts: true
+      }
+    },
+    {
+      clientImports: new Set(),
+      importedParameters: new Set(),
+      importedResponses: new Set()
     }
```
qiaozha, line 22: the same here, I wonder if we should have an independent function emitClientDefinition and then call emitClientDefinition inside onEmit ?

packages/autorest.typescript/src/restLevelClient/generateClientDefinition.ts
```diff
@@ -1,386 +1,48 @@
...
-} from "./helpers/operationHelpers";
-import {
-  generateMethodShortcuts,
-  REST_CLIENT_RESERVED
-} from "./generateMethodShortcuts";
-import { Methods, PathParameter, Paths, ResponseTypes } from "./interfaces";
-import { isLongRunningOperation } from "./helpers/hasPollingOperations";
+import { transformPaths } from "./transforms/transformPaths";
+import { Paths, buildClientDefinitions } from "@azure-tools/rlc-codegen";
 export let pathDictionary: Paths = {};
```
qiaozha, line 11: I see pathDirectionary is being used in several places, should we make it part of RLCModel instead of export it from generateClientDefinition.ts file ? it looks a little strange if we need to import something from another generateXXX.ts file.

packages/autorest.typescript/src/generators/samples/rlcSampleGenerator.ts
```diff
@@ -1,371 +1,490 @@
...
+  OperationMethod
+} from "../../restLevelClient/interfaces";
 import { camelCase } from "@azure-tools/codegen";
 import { pathDictionary } from "../../restLevelClient/generateClientDefinition";
 import { Operation, ParameterLocation } from "@autorest/codemodel";
 import { isLongRunningOperation } from "../../restLevelClient/helpers/hasPollingOperations";
 import { isPagingOperation } from "../../utils/extractPaginationDetails";
 import { getSecurityInfoFromModel } from "../../utils/schemaHelpers";
 import { getParameterAssignment } from "../../utils/valueHelpers";
+import { Paths, PathMetadata } from "@azure-tools/rlc-codegen";
```
qiaozha, line 28: looks like the sample generation refactor is inevitable ?

packages/rlc-codegen/src/interfaces.ts
```diff
@@ -35,3 +37,14 @@ export interface OperationAnnotations {
   isLongRunning?: boolean;
   isPageable?: boolean;
 }
+
+export interface RLCOptions {
+  includeShortcuts?: boolean;
```
qiaozha, line 42: Currently cadl doesn't support we input command line agruments, I guess this is potential blocker for our August MVP ?

packages/autorest.typescript/src/generators/samples/rlcSampleGenerator.ts
MaryGao: I think the main change is that 

- `Paths` models changed from `autorest.typescript` to `rlc-codegen`
- Format changes

Right?

packages/rlc-codegen/src/interfaces.ts
MaryGao: I think the [sidecar ](https://github.com/microsoft/cadl/issues/218) feature could be used to comsume metadata for versining, any customizations as well as other configurations. 

Maybe the command line arguments could be fed into `typescript.cadl` file as input similar to the `readme.md` file.

@joheredi  How do you think?

packages/cadl-rlc-emitter/src/transformPaths.ts
```diff
@@ -0,0 +1,43 @@
+import { Paths } from "@azure-tools/rlc-codegen";
+import { Program } from "@cadl-lang/compiler";
+import { getAllRoutes } from "@cadl-lang/rest/http";
+
+export function transformPaths(model: Program): Paths {
```
MaryGao, line 5: Can we move the `transformPaths` to the folder `transform`? considering we have sereval files to do transformation.

packages/rlc-codegen/src/helpers/camelCase.ts
```diff
@@ -0,0 +1,14 @@
+export function camelCase(
+  str: string,
```
MaryGao, line 2: We have three folders: 

- a) autorest.typescript
- b) cadl-rlc-emitter 
- c) rlc-codegen.

Just want to confirm one thing: the common part which will be used in a) + b) will be defined in c), Right?

packages/autorest.typescript/src/restLevelClient/generateClientDefinition.ts
joheredi: Yeah it is now part of the RLC Model. I'll see if I can remove this global

packages/autorest.typescript/src/generators/samples/rlcSampleGenerator.ts
joheredi: @qiaozha yeah, I had to do it so we consume the common interfaces across the board
@MaryGao you're right!

packages/rlc-codegen/src/interfaces.ts
joheredi: Yeah it looks like sidecar will help here! TBH I hadn't looked into how to get those options :P I just assumed there will be a way to do so.

@qiaozha I don't think it will be a blocker, worst case scenario we set it to `false`, shortcuts are not on by default yet so no harm done for MVP

packages/rlc-codegen/src/helpers/camelCase.ts
joheredi: Your understanding is correct. Also I agree we'll need to rename our folders and packages. I think I'll do that in a PR of itself as a follow-up. This is my proposal

`cadl-rlc-emitter` => `cadl-typescript` (@azure-tools/cadl-typescript)
`rlc-codegen` => `typescript-codegen-common` (@azure-tools/typescript-codegen-common)

packages/cadl-rlc-emitter/src/index.ts
joheredi: Yeah, really everything under `cadl-rlc-emiter` needs more work. I'm thinking about moving that to a separate PR

packages/rlc-codegen/src/interfaces.ts
MaryGao: Yeah, I haven't saw any options to set them either, just assume finally we would have a way to do so.

If there is no way to set arguments now maybe we just set default values for them

MaryGao (approved):
LGTM, we will refine some items in future

qiaozha (approved):

PR #1476, opened 2022-08-02
add-model-utils-for-cadl-emitter

qiaozha:


packages/rlc-codegen/src/interfaces.ts
```diff
@@ -35,3 +35,54 @@ export interface OperationAnnotations {
...
+  srcPath: string;
+  paths: Paths;
+  schemas?: Schema[];
+}
+export interface File {
+  path: string,
+  content: string
+}
+
+export enum SchemaContext {
```
joheredi, line 49: I wonder if we should use `Model` instead of Schema. Schema is heavily used in autorest and I think Model is more generic term

joheredi (commented):
I think this is looking good. I would like to see how the new parts of the RLCModel plug into @autorest/typescript.

packages/rlc-codegen/src/interfaces.ts
```diff
@@ -35,3 +35,54 @@ export interface OperationAnnotations {
...
+    all?: ObjectSchema[];
+    immediate?: ObjectSchema[];
+  };
+  parents?: { 
+    all?: ObjectSchema[]
+    immediate?: ObjectSchema[];
+  }
+}
+
+export interface Property extends Schema {
```
MaryGao, line 82:
```suggestion
export type Property = Schema;
```

packages/rlc-codegen/src/interfaces.ts
```diff
@@ -35,3 +35,54 @@ export interface OperationAnnotations {
...
+
+}
+
+export interface Parameter extends Schema {
```
MaryGao, line 86:
```suggestion
export type Parameter = Schema;
```

packages/rlc-codegen/src/interfaces.ts
```diff
@@ -35,3 +35,54 @@ export interface OperationAnnotations {
...
+  type: string;
+  description?: string;
+  required?: boolean;
+  default?: any;
+  readOnly?: boolean;
+  usage?: SchemaContext[];
+}
+
+export interface ObjectSchema extends Schema {
+  properties?: Record<string, Schema>;
```
MaryGao, line 69:
```suggestion
  properties?: Record<string, Property >;
```

packages/rlc-codegen/src/interfaces.ts
```diff
@@ -35,3 +35,54 @@ export interface OperationAnnotations {
...
+  name: string;
+  type: string;
+  description?: string;
+  required?: boolean;
+  default?: any;
+  readOnly?: boolean;
+  usage?: SchemaContext[];
+}
+
+export interface ObjectSchema extends Schema {
```
MaryGao, line 68: @qiaozha Just wondering if this is an ObjectSchema, its name will be the same as type, right? For example the model is `Car` then its name and type will be "Car".

PR #1480, opened 2022-08-05
Enable responses.ts cadl emitter and do autorest refactoring

MaryGao:


packages/cadl-rlc-emitter/src/transform/transformResponses.ts
```diff
@@ -0,0 +1,77 @@
...
+        }
+        rlcResponses.push(rlcOperationUnit);
+    }
+    if (importedModels.size > 0) {
+        importDetails.set(ImportKind.ResponseOutput, importedModels);
+    }
+    return rlcResponses;
+}
+
+function transformHeaders(from: HttpOperationResponse, to: ResponseMetadata) {
```
qiaozha, line 37: Is from a reserved word ?

packages/autorest.typescript/src/restLevelClient/transforms/transformResponseTypes.ts
```diff
@@ -0,0 +1,144 @@
...
+        rlcResponses.push(rlcOperationUnit);
+    }
+    if (importedModels.size > 0) {
+        importDetails.set(ImportKind.ResponseOutput, importedModels);
+    }
+    return rlcResponses;
+}
+
+
+function transformHeaders(from: Response, to: ResponseMetadata) {
```
joheredi, line 44: I'd suggest to make this function return the ResponseMetadata instead of populating it in the parameter `to`.


```ts
function transformHeaders(from: Response, to: ResponseMetadata)
```

And use it
```ts
const headers = transformHeaders(response, rlcResponseUnit);
const body = transformBody(response)

rlcResponseUnit = {
...rlcResponseUnit,
headers,
body
}
```

Pure functions are preferred as they are easier to test and more predictable when they don't cause side effects. So rule of thumb is, whenever you can return and not modify parameters we should do that, just modify parameters when there is no other option

joheredi: This is one of the reasons I want to refactor at some point the functions that take the importedModels, importedParameters and clientImports sets

packages/cadl-rlc-emitter/src/index.ts
```diff
@@ -42,6 +44,11 @@ async function emitClientDefinition(rlcModels: RLCModel, program: Program) {
   await emitFile(clientDefinitionsFile, program);
 }
 
+async function emitResponseTypes(rlcModels: RLCModel, program: Program) {
+  const responsesFile = buildResponseTypes(rlcModels);
+  if (responsesFile) await emitFile(responsesFile, program);
```
joheredi, line 49: nit: it is preferred in general to add the `{}` even if it is a single line statement in the if block

packages/cadl-rlc-emitter/src/transform/transformResponses.ts
```diff
@@ -0,0 +1,77 @@
+import { HeaderMetadata, ImportKind, OperationResponse, ResponseMetadata } from "@azure-tools/rlc-codegen";
+import { Program } from "@cadl-lang/compiler";
+import { getAllRoutes, HttpOperationResponse } from "@cadl-lang/rest/http";
+
+export function transformToResponseTypes(program: Program, importDetails: Map<ImportKind, Set<string>>): OperationResponse[] {
+    const [routes, _diagnostics] = getAllRoutes(program);
```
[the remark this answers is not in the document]
joheredi: I guess we should make sure Prettier is configured correctly to avoid these missmatches

packages/cadl-rlc-emitter/src/transform/transformResponses.ts
joheredi: I don't think it would cause compile errors, but I agree we should try to use another name. Maybe `response` is a good enough name

packages/autorest.typescript/src/restLevelClient/transforms/transformResponseTypes.ts
```diff
@@ -0,0 +1,144 @@
...
+ * response's body has a reference to a model or if it is a primitive, and returns the Typescript type
+ * to generate
+ * @param response - response to get the body type from
+ * @param importedModels - track models to import
+ */
+function getBodyTypeName(
+    response: SchemaResponse,
+    importedModels: Set<string>
+): string | undefined {
+    return getElementType(
```
joheredi, line 92: Do we need to wrap this into getBodyTypeName? Or can we just use directly getElementType?

packages/autorest.typescript/src/restLevelClient/transforms/transformResponseTypes.ts
```diff
@@ -0,0 +1,144 @@
...
+        response.schema,
+        [SchemaContext.Output, SchemaContext.Exception],
+        importedModels
+    );
+}
+
+// Gets a list of all the available operations requests in the specification
+function getAllOperationRequests(model: CodeModel) {
+    let operations: Operation[] = [];
+    model.operationGroups.forEach(og =>
```
joheredi, line 102: nit: maybe this would be more readable

```ts
for (const og of model.operationGroups) {
   for (const operation of og.operations) {
      if (operation.requests?.length) {
        operations.push(operation); 
     }
   }
}
```

packages/cadl-rlc-test/cadl-output/src/responses.ts
```diff
@@ -0,0 +1,45 @@
+import { HttpResponse } from "@azure-rest/core-client";
+import { ErrorOutput, WidgetOutput } from "./outputModels";
+
+/** Ok */
+export interface WidgetServiceList200Response extends HttpResponse {
+    status: "200";
+}
+
+export interface WidgetServiceListdefaultResponse extends HttpResponse {
```
joheredi, line 9: Can we fix the lowercase `d` in default? I know it was like this before, but I think we should fix it

joheredi (approved):
Looks great! I left a couple suggestions

packages/cadl-rlc-emitter/src/transform/transformResponses.ts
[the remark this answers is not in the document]
MaryGao: Let's introduce the Prettier in future pr

packages/autorest.typescript/test/rlcIntegration/bodyFormDataRest.spec.ts
```diff
@@ -27,8 +27,7 @@ describe("Integration tests for BodyFormData", () => {
 
     const result = await client.path("/formdata/stream/uploadfile").post({
       contentType: "multipart/form-data",
-      body: { fileContent: content, fileName },
-      binaryResponse: true
```
MaryGao, line 31: `binaryResponse` is missing in latest GA version, how to fix this?

packages/autorest.typescript/package.json
```diff
@@ -55,7 +55,7 @@
         "@autorest/codemodel": "~4.18.2",
         "@autorest/extension-base": "^3.4.1",
         "@autorest/testmodeler": "^2.2.5",
-        "@azure-rest/core-client": "dev",
+        "@azure-rest/core-client": "1.0.0-beta.10",
```
MaryGao, line 58: Fixed the version

PR #1482, opened 2022-08-09
parameters.ts file refactoring & emitter

MaryGao:
This pr will generate parameters.ts file in CADL emitter and refactoring existing autorest.typscript to latest RLC model.

fixes https://github.com/Azure/autorest.typescript/issues/1457


packages/autorest.typescript/test/rlcIntegration/generated/headerRest/src/parameters.ts
```diff
@@ -287,7 +287,11 @@ export type ResponseDurationParameters = ResponseDurationHeaderParam &
 export interface ParamByteHeaders {
   /** Send a post request with header values "scenario": "valid" */
   scenario: string;
-  /** Send a post request with header values "啊齄丂狛狜隣郎隣兀﨩" */
+  /**
+   * Send a post request with header values "啊齄丂狛狜隣郎隣兀﨩"
+   *
+   * Value may contain base64 encoded characters
```
MaryGao, line 293: Just refine the comment with getDoc method

packages/rlc-codegen/src/interfaces.ts
```diff
@@ -112,10 +112,28 @@ export interface Parameter extends Schema {
...
+export interface ParameterBodyMetadata {
+  // In case of formData we'd get multiple properties in body marked as partialBody
+  isPartialBody?: boolean;
+  body?: ParameterBodySchema[];
+}
+
+export type ParameterBodySchema = Schema;
 export interface ParameterMetadata {
-  type: "query" | "path" | "header" | "body";
+  type: "query" | "path" | "header";
```
qiaozha, line 136: why remove body?

packages/cadl-rlc-emitter/src/transform/transformResponses.ts
```diff
@@ -33,19 +37,20 @@ export function transformToResponseTypes(program: Program, importDetails: Map<Im
...
+function transformHeaders(
+    response: HttpOperationResponse): ResponseHeaderSchema[] | undefined {
+    if (!response.responses.length) {
         return;
-    } else if (from.responses.length > 1) {
+    } else if (response.responses.length > 1) {
         // TODO: handle one status code map to multiple rsps
     }
-    const headers = from.responses[0]?.headers;
+    const headers = response.responses[0]?.headers;
```
qiaozha, line 47: Why we don't need other responses' headers , just the first one ?

packages/cadl-rlc-test/cadl-output/src/parameters.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type WidgetServiceListParameters = RequestParameters;
+export type WidgetServiceReadParameters = RequestParameters;
+
+export interface WidgetServiceCreateBodyParam {
+    body: Widget;
+}
+
+export type WidgetServiceCreateParameters = WidgetServiceCreateBodyParam & RequestParameters;
+export type WidgetServiceCustomGetParameters = RequestParameters;
```
qiaozha, line 12: Compare with the file autorest generated https://github.com/Azure/autorest.typescript/pull/1471/files#diff-ab653ac0d8bb20bf748634d4ea6e4debc829adfae78ee1ee525eb8bbac25e6ff. Seems we have extra prefix of WidgetService of each parameters.  Also, there's some parameters definition missing, like CreateMediaTypeParam and ReadParameters.

qiaozha (changes requested):
Can you format all files to use 2 indents ? so that we are consistent.

packages/rlc-codegen/src/interfaces.ts
MaryGao: body needs extra information so I use separate model to handle it

packages/cadl-rlc-emitter/src/transform/transformResponses.ts
MaryGao: @joheredi How do you think below case? Do you think we need to adjust our response structure?
This is case we need to do further investigation between CADL and Swagger. 

#### Swagger Response Structure
One operation_id has different responses
However in the same response we will share the status code, body and headers etc
#### CADL Response Structure
One operation_id has different responses
In the same response we'll share the status code but it may have diff body and headers(e.g content-type)
#### RLC Current Response
Currently RLC response has similar structure with swagger one so we can't handle the different bodies with below case.
```
"responses": {
          "200": {
            "description": "The request has succeeded.",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/Widget"
                },
              "text/plain": {
                "schema": {
                  "$ref": "#/components/schemas/Widget2"
                }
              }
            }
          },
```

packages/cadl-rlc-test/cadl-output/src/parameters.ts
MaryGao: @qiaozha Thanks for pointing this out!

I will fix the extra prefix issue here. The ReadParameters is not missing one and we have `WidgetServiceReadParameters`.

But the `CreateMediaTypeParam` is missing in our case. This is introduced by the swagger code(https://github.com/Azure/autorest.typescript/blob/2e9066f7870e06032ac19626b2ba7a0d6d7c16ca/packages/autorest.typescript/test/smoke/swagger/widget.json#L46) generated from CADL. And I think our CADL one is correct.

packages/cadl-rlc-test/cadl-output/src/responses.ts
```diff
@@ -1,45 +1,45 @@
...
+    body: any;
 }
 
 /** Ok */
-export interface WidgetServiceCustomGet200Response extends HttpResponse {
+export interface CustomGet200Response extends HttpResponse {
     status: "200";
-    body: WidgetOutput;
+    body: any;
 }
```
MaryGao, line 40: Will enhance to detailed type once getSchemaFromType method is refined

packages/rlc-codegen/src/buildResponseTypes.ts
```diff
@@ -1,163 +1,163 @@
 import { InterfaceDeclarationStructure, OptionalKind, Project, PropertySignatureStructure, StructureKind } from "ts-morph";
-import { HeaderMetadata, ImportKind, ResponseMetadata, RLCModel } from "./interfaces.js";
+import { ImportKind, ResponseHeaderSchema, ResponseMetadata, RLCModel } from "./interfaces.js";
```
MaryGao, line 2: Just format files

packages/cadl-rlc-emitter/src/transform/transformResponses.ts
joheredi: It looks like we should make our RLC model able to support multiple response types depending on the statusCode, which seems to be an improvement from CADL

packages/cadl-rlc-test/cadl-output/src/parameters.ts
joheredi: I'm thinking that we may always want to add the prefix regardless of how many operation groups we have. The reason is to avoid breaking changes if the swagger/cadl adds a new operation group in the future. With the current design, the old Parameters and Responses will now have a Prefix (because now there are 2 operation groups). This will be a breaking change. This may be a good opportunity to change this

packages/autorest.typescript/src/restLevelClient/transforms/transformParameterTypes.ts
```diff
@@ -0,0 +1,208 @@
...
+import { ImportKind, OperationParameter, ParameterBodyMetadata, ParameterBodySchema, ParameterMetadata, ParameterMetadatas, Schema } from "@azure-tools/rlc-codegen";
+import { getLanguageMetadata } from "../../utils/languageHelpers";
+import { NameType, normalizeName } from "../../utils/nameUtils";
+import { getDocs } from "../getPropertySignature";
+import { getOperationParameters } from "../helpers/operationHelpers";
+import { getElementType, primitiveSchemaToType } from "../schemaHelpers";
+
+export function transformParameterTypes(model: CodeModel, importDetails: Map<ImportKind, Set<string>>) {
+  const rlcParameters: OperationParameter[] = [];
+  let importedModels = new Set<string>();
```
joheredi, line 14: can we use Map<ImportKind, Set<string>> for this as well?

packages/cadl-rlc-test/cadl-output/src/parameters.ts
MaryGao: Good point Jose. Do you think we need to adjust autorest.typescript this time?

qiaozha: Should we also change the name in autorest typescript to remove this ?

packages/autorest.typescript/src/restLevelClient/transforms/transformParameterTypes.ts
MaryGao: Maybe I will change the name here 
- importedModels: Set 
- the inputted one - importDetails: Map<ImportKind, Set<string>>

So back to your question I think we use the Map one already importedModels is the internal Set

packages/cadl-rlc-emitter/src/transform/transformResponses.ts
MaryGao: @joheredi  @qiaozha  I create an issue: https://github.com/Azure/autorest.typescript/issues/1486. I prefer we do more investigation after we have basic version of other files. 

How do you think?

packages/cadl-rlc-test/package.json
```diff
@@ -7,7 +7,8 @@
     "@cadl-lang/ts-rlc": "0.0.1",
```
lirenhe, line 7: I think we have a decision to change the name of the emitter.

packages/cadl-rlc-test/package.json
```diff
@@ -7,7 +7,8 @@
...
     "@cadl-lang/rest": "latest",
     "@cadl-lang/openapi": "latest",
-    "@azure-tools/cadl-autorest": "latest"
+    "@azure-tools/cadl-autorest": "latest",
```
lirenhe, line 10: Use 'latest' may cause breaking change with the latest version of CADL.

packages/cadl-rlc-test/cadl-output/src/parameters.ts
```diff
@@ -0,0 +1,21 @@
+import { RequestParameters } from "@azure-rest/core-client";
+import { Pet } from "./models";
+
+export type PetsDeleteParameters = RequestParameters;
+export type PetsReadParameters = RequestParameters;
+
+export interface PetsCreateBodyParam {
```
lirenhe, line 7: Will the @doc decorator in CADL be the comments of those exports?

MaryGao: Will enhance it in future pr

packages/cadl-rlc-test/package.json
MaryGao: Currently we are still in feature branch `cadl-integration` in dev mode so we want to keep latest. We will update to fixed versions when merging back to main branch.

packages/cadl-rlc-test/package.json
MaryGao: @joheredi  Do you have any chance to confirm the name `@azure-tools/cadl-typescript` with Brian?
https://github.com/Azure/cadl-azure/issues/1759#issuecomment-1201857743

packages/cadl-rlc-test/cadl-output/src/parameters.ts
MaryGao: updated

qiaozha (approved):

qiaozha:
@joheredi can you take a look at this PR again ? Thanks

MaryGao:
Due to the permission issue we can't merge this pr. We open a new one to apply these change: https://github.com/Azure/autorest.typescript/pull/1492.

So close this pr considering the new one is merged.

PR #1497, opened 2022-08-16
Generate index.ts file from CADL and refactor autorest.typescript

MaryGao:
Generate index.ts file from CADL and refactor autorest.typescript

fixes https://github.com/Azure/autorest.typescript/issues/1458


qiaozha (approved):
LGTM

PR #1503, opened 2022-08-18
Build generator helper function to make the generation simpler

MaryGao:
Build generator helper function to make the generation simpler.

This time I only focus the refactoring in autorest.typescript part and rlc-codegen one will be in future pr.


packages/autorest.typescript/src/restLevelClient/generateSchemaTypes.ts
```diff
@@ -1,23 +1,15 @@
...
-  // Get all paths
-  const importedParameters = new Set<string>();
-  const importedResponses = new Set<string>();
-  const clientImports = new Set<string>();
-  const rlcModels = transform(model, {
-    importedParameters,
-    importedResponses,
-    clientImports
-  });
+export function generateSchemaTypes(rlcModels: RLCModel, project: Project) {
```
qiaozha, line 13: why do we still need this function instead of importing it from rlc-codegen directly ?

MaryGao: I plan to refactor this to remove project parameter in rlc-codegen after you finish the upgrading. 

Or I could directly refactor this both rlc-codegen & autorest.typescript.

qiaozha (approved):

packages/autorest.typescript/src/restLevelClient/generateSchemaTypes.ts
qiaozha: ok

PR #1514, opened 2022-08-25
fix-model-generation-failure-in-cadl-ranch

qiaozha:


packages/cadl-rlc-emitter/src/modelUtils.ts
```diff
@@ -548,10 +549,7 @@ function applyIntrinsicDecorators(
     const values = getKnownValues(program, cadlType);
     if (values) {
       const enumSchema = { ...newTarget, ...getSchemaForEnum(program, values) };
-      enumSchema["x-ms-enum"].modelAsString = true;
```
MaryGao, line 551: I'd like to confirm that the `x-ms-enum` extension will NOT appear in CADL model, Right?

I think this is only useful for swagger input.

qiaozha: yes, we don't need this.

packages/cadl-rlc-emitter/src/modelUtils.ts
```diff
@@ -65,6 +65,7 @@ export function getSchemaForType(
     const schema = getSchemaForModel(program, type, usage, needRef) as any;
     if (usage && usage.includes(SchemaContext.Output)) {
       schema.outputTypeName = `${schema.name}Output`;
+      schema.typeName = `${schema.name}`;
```
MaryGao, line 68: Is it possible we could define returnType for `getSchemaForType` by leveraging the authering file and our integration testing?

Currently all communications are based on `any` type, that would introduce error like `TypeError: Cannot read properties of undefined (reading 'type')` also we have to guess what will return from end user side.

MaryGao (approved):
Thanks for fixing this, the pr looks good for me and I left some comments we could improve in future.

packages/cadl-rlc-emitter/test/integration/generated/hello/src/clientDefinitions.ts
```diff
@@ -4,11 +4,13 @@ import { Client, StreamableMethod } from "@azure-rest/core-client";
...
 export interface HelloOperations {
-  world(options: HelloWorldParameters): StreamableMethod<HelloWorld200Response>;
+  world(
+    options?: HelloWorldParameters
+  ): StreamableMethod<HelloWorld200Response>;
 }
 
 export interface World {
-  get(options: HelloWorldParameters): StreamableMethod<HelloWorld200Response>;
+  get(options?: HelloWorldParameters): StreamableMethod<HelloWorld200Response>;
```
MaryGao, line 13: Glad to see this is fixed!

Issue #1511, opened 2022-08-24
HLC LRO integraion tests will fail when upgrading @azure/core-lro to the latest


MaryGao:
HLC LRO integraion tests will fail when upgrading @azure/core-lro to the latest and please take a look at the failure details.

https://dev.azure.com/azure-sdk/public/_build/results?buildId=1799698&view=logs&j=c46da42c-2794-5fc1-ff45-518534d5e680&t=10c7c4f0-012d-5a7c-b59c-f34223e6bb5e

deyaaeldeen:
The failure is due to a difference in the literal string of the error messages thrown when the operation/poller is canceled/failed. I recommend updating the test expectations to match the new error messages.

MaryGao:
@deyaaeldeen  Thanks for this information! I think most of these wording changes works for me except below one:

```
      + expected - actual

      -should have thrown instead
      +Received unexpected HTTP status code 204 while polling. This may indicate a server issue.
```
Could you suggest what the message `should have thrown instead` means? Could you provide any more specific information?

deyaaeldeen:
That old error message was thrown by the test and not by core-lro. This is a change in behavior in core-lro where a 204 response no longer raises an exception. Please update that test to check the response object instead, but I don't remember what fields and values it contains.

PR #1517, opened 2022-08-26
modify error message in lro

kazrael2119:
https://github.com/Azure/autorest.typescript/issues/1511


packages/autorest.typescript/test/integration/lro.spec.ts
```diff
@@ -256,7 +256,7 @@ describe("LROs", () => {
       } catch (e) {
         assert.equal(
           e.message,
-          "Received unexpected HTTP status code 204 while polling. This may indicate a server issue."
+          "should have thrown instead"
```
deyaaeldeen, line 259: please update those tests to no longer check for an error. It should check the response object instead.

PR #1520, opened 2022-08-29
Use authoring as cadl test

qiaozha:
fix https://github.com/Azure/autorest.typescript/issues/1522


packages/cadl-rlc-test/spec/main.cadl
```diff
@@ -1,208 +1,339 @@
...
+// - Security definition for apiKey
 
-// TODO: Use `Uri` type when available: https://github.com/microsoft/cadl/issues/654
-@serviceTitle("Confidential Ledger Service")
-@serviceVersion("2022-05-13")
-@useAuth(OAuth2Auth<[MyFlow]>)
-@versionedDependency(Azure.Core.Versions.v1_0_Preview_1)
 @server(
-  "{ledgerUri}",
-  "Confidential Ledger Service",
```
MaryGao, line 23: I think we could keep the `Confidential Ledger Service` as our smoke test. So we could guarantee the changes will not break that RP.

I could help add into smoke test if necessary.

packages/cadl-rlc-emitter/src/transform/transformPaths.ts
```diff
@@ -15,25 +15,16 @@ import {
...
-        format: { path: route.path },
-        target: route.operation
-      });
-      continue;
-    }
     const respNames = [];
     for (const resp of route.responses) {
       const respName = getResponseTypeName(
-        route.operation.namespace?.name ?? "",
+        route.container.name,
```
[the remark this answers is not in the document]
qiaozha: I asked Weidong, he mentioned java is using container.name.

packages/cadl-rlc-test/spec/main.cadl
qiaozha: Maybe we should move them both to the smoke-test and remove the cadl-rlc-test folder @joheredi how do you think of it ?

MaryGao (approved):
LGTM and left some comments

qiaozha:
@joheredi I will merge this PR first and feel free to left some comments if you found anything. Thanks

Issue #1515, opened 2022-08-25
[Cadl RLC] Design the test cases and add unit test for getSchemaForType in modelUtil.ts


MaryGao:
Currently there is no any UT in cadl-emitter and this is not a good practice. So in this issue we could resolve this by

- Enable the UT framework in cadl-rlc-emitter project
- Add UTs for function getSchemaForType in modelUtil.ts

Issue #1525, opened 2022-08-30
[CADL in RLC]fail to generate doc correctly


qiaozha:

MaryGao:
Cases need more discussion:

- should generate apiVersion description with 'api-version'
- should generate header description with custom name

Issue #1524, opened 2022-08-30
[RLC Cadl] Error thrown when generating some cadl types to typescript types


MaryGao:
When we write our test cases from cadl file to typescript code some types could not be handled correctly.  We use this issue to track these cases to ensure they could work correctly.

- should handle type_literals:boolean -> boolean_literals fixed
- should handle extensible_enum -> string fixed
- should handle anonymous model -> effictive type/interface fixed
- should handle inheritance model -> multiple types/interfaces fixed

Epic: https://github.com/Azure/cadl-azure/issues/1942

fixes https://github.com/Azure/autorest.typescript/pull/1552

PR #1526, opened 2022-08-30
Enable unit test for every single file

MaryGao:
Enable the test for every single file and it could take CADL as input and examine if the final typescript as expected.

Generally I ignore below two cases

- The behavior is unexpected 1) maybe incorrect output 2) error throw in our code ==> We need to fix our code to pass them

- The behavior is un-decided ==> We need to take time to re-think how to design better

model issue tracked here  https://github.com/Azure/autorest.typescript/issues/1524
doc issue tracked here https://github.com/Azure/autorest.typescript/issues/1525

fixes https://github.com/Azure/autorest.typescript/issues/1515


packages/cadl-rlc-emitter/test/unit/generate/models.spec.ts
```diff
@@ -0,0 +1,452 @@
...
+      const cadlTypeDefinition = `
+      @knownValues(TranslationLanguageValues)
+      model TranslationLanguage is string;
+      enum TranslationLanguageValues {
+        English,
+        Chinese,
+      }`;
+      const cadlType = "TranslationLanguage";
+      // TODO: ensure this trick is working
+      const typeScriptType = `"English" | "Chinese" | string & {}`;
```
qiaozha, line 140: I don't think this is decided, it's just a proposal still needs some discussion about it.  Currently, we are still using string for extensible enum. @joheredi  right ?

MaryGao: Yes I just ignore this case because it needs effort to finalize the current design. 

Generally I ignore below two cases 

- The behavior is unexpected 1) maybe incorrect output 2) error throw in our code ==> We need to fix our code to pass them
- The behavior is un-decided ==> We need to take time to re-think how to design better

qiaozha (approved):

PR #1528, opened 2022-08-31
fix dictionary type in cadl emitter

### LINKED ISSUE #1527: [CADL in RLC]Map type is not generate correctly
Currently we are generating Record<string, unknown> for map type, but in cadl, they have already specify the value type.
for example Record<string> in cadl represents Record<string, string>.

qiaozha:
fix #1527


MaryGao (approved):
Thanks for fixing it! LGTM

Issue #1539, opened 2022-09-02
Fix the issues in parameters and responses when generating authoring


MaryGao:
Response.ts
Header parameter name inconsistent with swagger generated, should generate operation-location not operationlocation 
Parameters.ts
Missing content type parameter in the cadl output 
Had extra api-version parameter in the cadl-output

MaryGao:
Close it considering fixed

PR #1540, opened 2022-09-02
fix-missing-body-param-and-doc

qiaozha:

### DIFF
--- a/packages/cadl-rlc-emitter/src/transform/transformParameters.ts
+++ b/packages/cadl-rlc-emitter/src/transform/transformParameters.ts
@@ -135,11 +135,11 @@ function transformBodyParameters(
   parameters: HttpOperationParameters,
   importedModels: Set<string>
 ): ParameterBodyMetadata | undefined {
-  const bodyParameters = parameters.bodyParameter;
-  if (!bodyParameters) {
+  const bodyType = parameters.bodyType;
+  if (!bodyType) {
     return undefined;
   }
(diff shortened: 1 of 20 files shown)


packages/cadl-rlc-emitter/src/transform/transformParameters.ts
```diff
@@ -154,8 +154,10 @@ function transformBodyParameters(
       {
         name: "body",
         type,
-        required: !Boolean(bodyParameters.optional),
-        description: getDoc(program, bodyParameters)
+        required: parameters.bodyParameter
+          ? !Boolean(parameters.bodyParameter.optional)
+          : bodySchema.required,
+        description: getDoc(program, bodyType)
```
MaryGao, line 160: I don't think we should get description from bodyType because this way we'll get the model description not property description

packages/cadl-rlc-emitter/src/transform/transformParameters.ts
```diff
@@ -135,11 +135,11 @@ function transformBodyParameters(
   parameters: HttpOperationParameters,
   importedModels: Set<string>
 ): ParameterBodyMetadata | undefined {
-  const bodyParameters = parameters.bodyParameter;
-  if (!bodyParameters) {
+  const bodyType = parameters.bodyType;
```
MaryGao, line 138: Not sure if it's better to get bodyType first if no then get bodyParameter.type

MaryGao (approved):
LGTM

PR #1541, opened 2022-09-02
Fix reponses docs issue

MaryGao:


packages/cadl-rlc-emitter/test/unit/generate/responses.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+    );
+  });
+
+  it("should generate property name without custome name", async () => {
+    const responses = await emitResponsesFromCadl(`
+        @doc("Metadata for long running operation status monitor locations")
+        model LongRunningStatusLocation {
+            @doc("The location for monitoring the operation state.")
+            @header
+            operationLocation: ResourceLocation<string>;
```
qiaozha, line 42: why are we generate "operation-location" for this case ?

MaryGao: In this case we could both get the property name `operationLocation` and `operation-location` I just follow the openapi way to get the name here. It would be the former one.

see code below:
https://github.com/microsoft/cadl/blob/bb5847a37d4a7064d2445cc488c47448f99da6f4/packages/openapi3/test/return-types.test.ts#L6-L20

qiaozha (approved):

PR #1542, opened 2022-09-02
Filter out api version in parameter

MaryGao:
Filter out api version in parameter due to it is client-level parameter


qiaozha (approved):
LGTM

Issue #1481, opened 2022-08-08
The binary type is not workable with the latest @azure-rest/core-client version


MaryGao:
We have RLC integration testing for binary type : https://github.com/Azure/autorest.typescript/blob/main/packages/autorest.typescript/test/rlcIntegration/bodyFormDataRest.spec.ts.

However if we upgrade @azure-rest/core-client to the latest GA or dev one we will meet below issue:

```
TSError: ⨯ Unable to compile TypeScript:
test/rlcIntegration/bodyFormDataRest.spec.ts(31,7): error TS2345: Argument of type '{ contentType: "multipart/form-data"; body: { fileContent: Uint8Array; fileName: string; }; binaryResponse: boolean; }' is not assignable to parameter of type 'UploadFileParameters'.
  Object literal may only specify known properties, and 'binaryResponse' does not exist in type 'UploadFileParameters'.
test/rlcIntegration/bodyFormDataRest.spec.ts(63,7): error TS2345: Argument of type '{ body: Uint8Array; binaryResponse: boolean; contentType: "application/octet-stream"; }' is not assignable to parameter of type 'UploadFileViaBodyParameters'.
  Object literal may only specify known properties, and 'binaryResponse' does not exist in type 'UploadFileViaBodyParameters'.

    at createTSError (/Users/runner/work/1/s/common/temp/node_modules/.pnpm/ts-node@10.9.1_bea9319c18027313ff9f3aa6f8c8e70e/node_modules/ts-node/src/index.ts:859:12)

```

Epic: https://github.com/Azure/cadl-azure/issues/1942

MaryGao:
close as merged

PR #1551, opened 2022-09-06
Support bytes type in responses.ts and parameters.ts

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1481


packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/outputModels.ts
```diff
@@ -164,7 +164,7 @@ export interface SharkOutputParent extends FishOutputParent {
 
 export interface SawsharkOutput extends SharkOutputParent {
   /** Value may contain base64 encoded characters */
-  picture?: string;
+  picture?: Uint8Array;
```
MaryGao, line 167: @joheredi Even we change to `Uint8Array` but the real result is still a string returned from the core client

packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/models.ts
```diff
@@ -129,7 +133,11 @@ export interface SharkParent extends FishParent {
 
 export interface Sawshark extends SharkParent {
   /** Value may contain base64 encoded characters */
-  picture?: string;
+  picture?:
+    | string
+    | Uint8Array
+    | ReadableStream<Uint8Array>
+    | NodeJS.ReadableStream;
```
MaryGao, line 140: @joheredi This would be fine because the new type is compatible with old `string` one.

packages/cadl-rlc-emitter/test/integration/generated/clients/interfaces/typescript.json
```diff
@@ -7,5 +7,6 @@
     "description": "Client Interface Test Service",
     "version": "1.0.0"
   },
-  "isCadlTest": true
+  "isCadlTest": true,
+  "overrideTitle": "InterfaceClient"
```
qiaozha, line 11: I don't think we should have override here, just put title here. and then use it to override whatever we defined in the cadl. the same as what we do in typescript.md.

.scripts/ci.yml
```diff
@@ -10,13 +10,13 @@ parameters:
     default: true
   - name: OSVmImage
     type: string
-    default: ubuntu-18.04
+    default: ubuntu-20.04
```
qiaozha, line 13: why are we upgrading the image versions ?

MaryGao: There would be 1) one [error ](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1837591&view=logs&j=39c431fc-b8dd-5dce-1f4e-2e5be8576b0f)happened if we didn't upgrade macOS image versions. Also 2) one [warning ](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1837647&view=logs&j=6e6cd934-b249-5120-4299-21f24be20cc5)to upgrade ubuntu versions.

So I resolved them by upgrading both systems

PR #1563, opened 2022-09-15
support multiple parameters in url

### LINKED ISSUE #1413: support multiple paramters in x-ms-parameterized-host
Curently if we set the x-ms-parameterized-host as 
```
  "x-ms-parameterized-host": {
    "hostTemplate": "{Endpoint}/atlas/{serviceVersion}",
    "useSchemePrefix": false,
    "parameters": [
      {
        "$ref": "#/parameters/Endpoint"
      },
      {
        "$ref": "#/parameters/serviceVersion"
      }
    ]
  },
```
and generated client constructor factory function look like this 
```
  Endpoint: string,
  credentials: TokenCredential,
  options: ClientOptions = {}
): PurviewCatalogClient {
  const baseUrl =
    options.baseUrl ?? `${Endpoint}/atlas/{serviceVersion}/catalog/api`;
  options.apiVersion = options.apiVersion ?? "2022-03-01-preview";
  options = {
    ...options,
    credentials: {
      scopes: ["https://purview.azure.net/.default"]
    }
  };
 ...
}
```

This would be a problem once #1399 is enabled.

qiaozha:
fix #1413

### DIFF
--- a/packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/bodyComplexRestClient.ts
+++ b/packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/bodyComplexRestClient.ts
@@ -7,9 +7,8 @@ import { BodyComplexRestClient } from "./clientDefinitions";
 export default function createClient(
   options: ClientOptions = {}
 ): BodyComplexRestClient {
-  const baseUrl = options.baseUrl ?? "http://localhost:3000";
+  const baseUrl = options.baseUrl ?? `http://localhost:3000`;
   options.apiVersion = options.apiVersion ?? "2016-02-29";
-
   const userAgentInfo = `azsdk-js-body-complex-rest/1.0.0-preview1`;
   const userAgentPrefix =
     options.userAgentOptions && options.userAgentOptions.userAgentPrefix
--- a/packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/bodyComplexRestClient.ts
+++ b/packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/bodyComplexRestClient.ts
@@ -4,6 +4,10 @@
 import { getClient, ClientOptions } from "@azure-rest/core-client";
 import { BodyComplexRestClient } from "./clientDefinitions";
 
+/**
+ * Initialize a new instance of the class BodyComplexRestClient class.
+ *
+ */
 export default function createClient(
   options: ClientOptions = {}
 ): BodyComplexRestClient {
--- a/packages/cadl-rlc-emitter/test/unit/generate/clientFactory.spec.ts
+++ b/packages/cadl-rlc-emitter/test/unit/generate/clientFactory.spec.ts
@@ -1,0 +1,145 @@
+import { assert } from "chai";
+import { emitClientFactoryFromCadl } from "../emitUtil.js";
+import { assertEqualContent } from "../testUtil.js";
+
+describe("Client Factory generation testing", () => {
+  describe("handling url parameters", () => {
+    it("should handle zero parameter", async () => {
+      const models = await emitClientFactoryFromCadl(`
+      @server(
+        "localhost",
+        "Language Service"
+      )
+      @serviceTitle("PetStoreClient")
+      namespace PetStore;
+      `);
+      assert.ok(models);
+      assertEqualContent(
+        models!.content,
+        `
+        import { getClient, ClientOptions } from "@azure-rest/core-client";
+        import { testClient } from "./clientDefinitions";
+        
+        export default function createClient(options: ClientOptions = {}): testClient {
+        const baseUrl = options.baseUrl ?? \`localhost\`;
+        
+        const userAgentInfo = \`azsdk-js--rest/1.0.0-beta.1\`;
+        const userAgentPrefix =
+            options.userAgentOptions && options.userAgentOptions.userAgentPrefix
+            ? \`\${options.userAgentOptions.userAgentPrefix} \${userAgentInfo}\`
+            : \`\${userAgentInfo}\`;
+        options = {
+            ...options,
+            userAgentOptions: {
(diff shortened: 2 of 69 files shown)


packages/cadl-rlc-emitter/test/unit/generate/clientFactory.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+      );
+    });
+
+    it("should handle two parameters", async () => {
+        const models = await emitClientFactoryFromCadl(`
+            @server(
+              "{Endpoint}/language/{Version}",
+              "Language Service",
+              {
+                Endpoint: Endpoint,
```
MaryGao, line 97: Do you know if we could set the parameter as optional?

packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/bodyComplexRestClient.ts
```diff
@@ -7,9 +7,8 @@ import { BodyComplexRestClient } from "./clientDefinitions";
 export default function createClient(
   options: ClientOptions = {}
 ): BodyComplexRestClient {
-  const baseUrl = options.baseUrl ?? "http://localhost:3000";
```
MaryGao, line 10: Do we have any test case for this in swagger repo?

packages/cadl-rlc-emitter/test/unit/generate/clientFactory.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+            }
+            `);
+        assert.ok(models);
+        assertEqualContent(
+          models!.content,
+          `
+            import { getClient, ClientOptions } from "@azure-rest/core-client";
+            import { testClient } from "./clientDefinitions";
+            
+            export default function createClient(
```
MaryGao, line 118: One suggestion: maybe we could add document for the factory method and include relevant doc

MaryGao (approved):
Left small comments! 
@joheredi Could you take a look also?

packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/bodyComplexRestClient.ts
qiaozha: ut no, other integration test or smoke test yes.

MaryGao: I didn't find any `createClient` signature changes for multiple parameter cases

packages/rlc-codegen/src/interfaces.ts
```diff
@@ -11,8 +11,13 @@ export interface RLCModel {
   responses?: OperationResponse[];
   importSet?: Map<ImportKind, Set<string>>;
   pageInfo?: PageInfo;
+  uriInfo?: UriInfo
 }
 
+export interface UriInfo {
```
joheredi, line 17: if possible I would recommend not to use Uri but use Url instead

packages/cadl-rlc-emitter/test/unit/generate/clientFactory.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+              Endpoint: string,
```
joheredi, line 119: is there any way we can camel case these parameters?

packages/cadl-rlc-emitter/test/unit/generate/clientFactory.spec.ts
joheredi: Not sure I understand the question. The path parameters that @qiaozha is working on in this PR are different in that these are part of the "host", not the operation resource. These are shared across all requests, however, the operation path parameters are specific to each operation. Hope this answers your question

joheredi (approved):
Looks good, left a few minor comments. It would be great to have a smoke test that covers these changes.

packages/cadl-rlc-emitter/test/unit/generate/clientFactory.spec.ts
MaryGao: Yes, thanks for your explanations!

packages/rlc-codegen/src/interfaces.ts
qiaozha: updated.

packages/cadl-rlc-emitter/test/unit/generate/clientFactory.spec.ts
qiaozha: we should not normalize the parameter as they should be exactly the same as the serializedName.

packages/cadl-rlc-emitter/test/unit/generate/clientFactory.spec.ts
qiaozha: updated..

packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/bodyComplexRestClient.ts
qiaozha: I thought you were asking whether we have test for single url parameter in case of any regression error. 
add the integration test for multiple url parameters here https://github.com/Azure/autorest.typescript/pull/1563/files#diff-87bf906cad32684f3972ea059bae2326eab6bc4eeb2a1de7aa149e5789450dea

qiaozha:
> Looks good, left a few minor comments. It would be great to have a smoke test that covers these changes.

Added the integration test here https://github.com/Azure/autorest.typescript/pull/1563/files#diff-87bf906cad32684f3972ea059bae2326eab6bc4eeb2a1de7aa149e5789450dea, I just take a few path from purview catalog and change their swagger definition a little bit to show our scenario as I can't find there's current valid case of multiple url parameters.

Issue #1488, opened 2022-08-12
Please review the links in README.md files in this repo as they are broken


sharathmalladi:
Reference Links from this page are broken: https://github.com/Azure/autorest.typescript/blob/main/docs/samples/readme.md
Link referencing [typescript-specific docs](https://github.com/Azure/autorest.typescript/blob/main/docs/docs/migrate/readme.md) from this page is broken: https://github.com/Azure/autorest.typescript

It will be helpful to generally verify the links referenced from README.md files work so the help docs and samples are easier to find to newbie developers.

qiaozha:
close it as this should have been resolved. @sharathmalladi feel free to reopen it if you still found issues in the readme.md link.

PR #1574, opened 2022-09-22
fix readme.md broken links

qiaozha:
fix https://github.com/Azure/autorest.typescript/issues/1488


qiaozha:
close it in favor of #1575

PR #1518, opened 2022-08-26
modify lro

kazrael2119:
There are two issues introduced when we upgrade the current dependencies and we'll try to fix in this pr.

fixes https://github.com/Azure/autorest.typescript/issues/1511
fixes https://github.com/Azure/autorest.typescript/issues/1516
fixes https://github.com/Azure/autorest.typescript/issues/1544

PR #1586, opened 2022-09-27
Update LRO generation

joheredi:
Porting changes from ##1568 after merging cadl-integration into main.

Updating LRO generation to the new API and adding a flag to generate using the legacy API


deyaaeldeen (commented):
Thanks a lot for porting the changes! 🚀

I see there are two main failures in CI:
- clients that still need to be regenerated, would you mind regenerating them?
- test failures and these should be fixed once https://github.com/Azure/autorest.typescript/pull/1518 is merged in.

packages/autorest.typescript/src/utils/autorestOptions.ts
```diff
@@ -73,10 +74,16 @@ export async function extractAutorestOptions(): Promise<AutorestOptions> {
...
     coreHttpCompatMode,
-    dependencyInfo
+    dependencyInfo,
+    useLegacyLro
   };
 }
 
+async function getUseLegacyLro(host: AutorestExtensionHost): Promise<boolean> {
+  const useLegacyLroOption = await host.getValue("use-legacy-lro");
+  return useLegacyLroOption === null ? false : Boolean(useLegacyLroOption);
```
deyaaeldeen, line 84: Enabling this flag will generate the old LRO implementation. However, the new LRO implementation, added here, is not backward compatible so this flag shouldn't be used to re-generate already GA'd libraries.

The flag is disabled by default, so to re-generate already GA'd libraries, the flag will have to be set explicitely in the generation command.

This experience is not the best for customers of the code generator, but I believe it will help us migrate to the new implementation faster. If that doesn't work, I am open to set it to true by default.

/cc @joheredi, @qiaozha, @MaryGao

deyaaeldeen: I talked to @xirzec offline about it and he is ok with the idea in principle. I wonder if anyone else has any other thoughts?

deyaaeldeen (approved):
Thanks @joheredi for porting the changes!

We still need to update RLC LROs to not throw by enabling the core-lro flag but I can do that in another PR.

packages/autorest.typescript/test/rlcIntegration/dpgCustomizationRest.spec.ts
```diff
@@ -44,7 +44,7 @@ describe("DPGCustomization Client", () => {
       const initialResponse = await client
         .path("/customization/lro/{mode}", "raw")
         .put();
-      const poller = getLongRunningPoller(client, initialResponse);
+      const poller = await getLongRunningPoller(client, initialResponse);
```
qiaozha, line 47: why is this not a breaking ?

qiaozha: I wonder if the changes made in RLC pollingHelper are also behind the use-legacy-lro flag ?

deyaaeldeen: Good point, addressed in https://github.com/Azure/autorest.typescript/pull/1586/commits/49a3d53dc802dbb6a7ad1ae29fcb85ca15909512.

packages/rlc-common/src/static/pollingContent.ts
```diff
@@ -61,9 +61,7 @@ export async function getLongRunningPoller<TResult extends HttpResponse>(
   };
 
   {{#if useLegacyLro}}
-  const poller = new LroEngine(poller, options);
-  await poller.poll();
```
qiaozha: We don't need this if it's useLegacyLro now ? I remember previously the poller instance won't call the inital request if we don't add poller.poll() right after creating it.

deyaaeldeen: As you mentioned earlier, changing it to call the `poll` method will cause a breaking change, because the helper will be async and will need to be awaited. The `useLegacyLro` flag should be used to not introduce breaking changes but for the majority of the cases, introducing the breaking change is fine because most RLC libraries are still in beta.

packages/rlc-common/src/metadata/buildPackageFile.ts
```diff
@@ -141,7 +141,8 @@ function restLevelPackage(model: RLCModel, hasSamplesGenerated: boolean) {
         "@azure/core-paging": "^1.2.0"
       }),
       ...(hasLRO && {
-        "@azure/core-lro": "^2.2.0"
```
MaryGao, line 144: Do we need to consider the flag `useLegancyLRO` in package.json?

deyaaeldeen: No need to, the package versions are backward compatible.

packages/rlc-common/src/static/pollingContent.ts
MaryGao: Two follow-up questions 

- Is the latest version `2.5.0` backward compatible with lower version like `2.2.0`? Do we need to bind the version to older one in package.json when the `useLegacyLro` enabled
- Why we remove the code line `await poller.poll();` the `useLegacyLro` enabled here?

MaryGao (approved):
LGTM

qiaozha (approved):

PR #1592, opened 2022-10-03
Bump node version to v14

jeremymeng:
We are planning to drop NodeJS v12 support.  This PR updates codegen to use v14 as template.


qiaozha (approved):
LGTM

Issue #1544, opened 2022-09-02
RLC new error by using check status function


kazrael2119:
After fixing the bug where check status function does not work，on pipeline ，now there are other errors are occurred.

In **Http infrastructure rest Client** ( **function : error message** ){
delete503 should retry and return 200: AssertionError: expected '503' to equal '200'
get502 should retry and return 200: AssertionError: expected '502' to equal '200'
head408 should retry and return 200: AssertionError: expected '408' to equal '200'
patch500 should retry and return 200: AssertionError: expected '500' to equal '200'
patch504 should retry and return 200: AssertionError: expected '504' to equal '200'
post503 should retry and return 200:  AssertionError: expected '503' to equal '200'
put500 should retry and return 200: AssertionError: expected '500' to equal '200'
put504 should retry and return 200: AssertionError: expected '504' to equal '200'
}

In **LRO Rest Client**  ( **function : error message** ){
should handle PutNonRetry400 : AssertionError: expected undefined to equal 400
should handle putNonRetry201Creating400 : AssertionError: expected undefined to equal 400
should get 400 with putNonRetry201Creating400InvalidJson : AssertionError: expected undefined to equal 400
should handle putAsyncRelativeRetry400 : AssertionError: expected undefined to equal 400
should handle delete202NonRetry400 : AssertionError: expected undefined to equal 400
should handle deleteNonRetry400 : AssertionError: expected undefined to equal 400
should handle postNonRetry400 : AssertionError: expected undefined to equal 400
should handle post202NonRetry400 : AssertionError: expected undefined to equal 400
should handle putAsyncRelativeRetryInvalidHeader: AssertionError: expected undefined to equal 404
should handle delete202RetryInvalidHeader : AssertionError: expected undefined to equal 404
should handle deleteAsyncRelativeRetryInvalidHeader : AssertionError: expected undefined to equal 404
should handle post202RetryInvalidHeader : AssertionError: expected undefined to equal 404
should handle postAsyncRelativeRetryInvalidHeader : AssertionError: expected undefined to equal 404
should retry put201CreatingSucceeded200: AssertionError: Unexpected status code 500
should retry putAsyncRelativeRetrySucceeded: AssertionError: Unexpected status code 500
should retry deleteProvisioning202Accepted200Succeeded: AssertionError: Unexpected status code 500

should retry delete202Retry200:  AssertionError: expected '500' to equal '200'
[...]

PR #1596, opened 2022-10-09
Update dependencies to the latest

MaryGao:
There are two issues introduced when we upgrade the current dependencies and we'll try to fix in this pr.

fixes https://github.com/Azure/autorest.typescript/issues/1511
fixes https://github.com/Azure/autorest.typescript/issues/1544

@joheredi @deyaaeldeen I closed previous pr and reopen this one based on main branch. I updated the core-lro to 2.4.0 and  upgraded other packages to the latest. And fixed test case failures in ci and above issues.

However I didn't change any codegen logic especially relevant to lro option and I plan to do it in this [issue](https://github.com/Azure/autorest.typescript/issues/1598) once your existing lro prs merge into main.

### DIFF
--- a/packages/autorest.typescript/test/integration/generated/paging/src/models/index.ts
+++ b/packages/autorest.typescript/test/integration/generated/paging/src/models/index.ts
@@ -27,6 +27,10 @@ export interface ProductResult {
   nextLink?: string;
 }
 
+export interface BodyParam {
+  name?: string;
+}
+
 export interface OdataProductResult {
   values?: Product[];
   odataNextLink?: string;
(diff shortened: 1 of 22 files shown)


packages/autorest.typescript/test/integration/generated/paging/src/models/index.ts
```diff
@@ -27,6 +27,10 @@ export interface ProductResult {
   nextLink?: string;
 }
 
+export interface BodyParam {
+  name?: string;
```
MaryGao, line 31: These files changed because we upgraded the test server to the latest one.

packages/autorest.typescript/test/integration/lro.spec.ts
```diff
@@ -199,23 +193,13 @@ describe("LROs", () => {
         await client.lROs.beginPost202NoRetry204AndWait(LROOptions);
         throw new Error("should have thrown instead");
       } catch (e) {
-        assert.equal(
-          e.message,
-          "Received unexpected HTTP status code 204 while polling. This may indicate a server issue."
-        );
+        assert.equal(e.message, "should have thrown instead");
```
joheredi, line 196: This error was thrown from line 194. We should adjust the expectation here. If we expect to no longer throw on 204 I think we should update the test to verify the status code and remove the expectation for throw

qiaozha (approved):

PR #1610, opened 2022-10-17
add @types/node in packagefilegenerator.ts

kazrael2119:
https://github.com/Azure/azure-sdk-for-js/issues/23519


packages/autorest.typescript/src/generators/static/packageFileGenerator.ts
```diff
@@ -195,6 +195,7 @@ function regularAutorestPackage(
     packageInfo.devDependencies["@types/chai"] = "^4.2.8";
     packageInfo.devDependencies["chai"] = "^4.2.0";
     packageInfo.devDependencies["cross-env"] = "^7.0.2";
+    packageInfo.devDependencies["@types/node"] = "^18.11.0";
```
jeremymeng, line 198: can we use ^14.0.0, to be consistent with `node: ">=14.0.0"` in the engines section?

qiaozha (approved):

PR #1613, opened 2022-10-18
support-sdk-folder-in-emiiter

### LINKED ISSUE #1609: [CADL DPG] Generate Codes in sdk-folder defined in cadl-project.yaml
Example `cadl-project.yaml`:
```
emitters:
  "@azure-tools/cadl-typescript":
    "sdk-folder": sdk/confidentialledger/azure-confidentialledger-rest
```
When generating codes, the emitter should generate codes in `sdk-folder` when the option is defined.

qiaozha:
fix #1609

### DIFF
--- a/packages/cadl-typescript/src/transform/transfromRLCOptions.ts
+++ b/packages/cadl-typescript/src/transform/transfromRLCOptions.ts
@@ -5,6 +5,7 @@ import {
   Program
 } from "@cadl-lang/compiler";
 import { getAuthentication } from "@cadl-lang/rest/http";
+import { isAbsolute } from "path";
 
 export function transformRLCOptions(program: Program, options: RLCOptions): RLCOptions {
   const config = options;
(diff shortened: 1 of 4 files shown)


dw511214992 (approved):

packages/cadl-typescript/src/transform/transfromRLCOptions.ts
```diff
@@ -18,6 +19,9 @@ export function transformRLCOptions(program: Program, options: RLCOptions): RLCO
       config.packageDetails.scopeName = nameParts[0]?.replace("@", "");
     }
   }
+  if (options["sdk-folder"] && isAbsolute(options["sdk-folder"])) {
```
MaryGao, line 22: I'd like to confirm that this is only allowed for absolute path?

qiaozha: @dw511214992 will replace the value with absolute path before run cadl compile

MaryGao (approved):
LGTM

PR #1621, opened 2022-10-20
example-on-refactor-rlc-implementation

qiaozha:


packages/autorest.typescript/test/rlcIntegration/generated/urlRest/src/shortCuts/paths.ts
```diff
@@ -0,0 +1,17 @@
+import { StreamableMethod } from "@azure-rest/core-client";
```
qiaozha, line 1: @joheredi do you have any concerns about the folder and filename ?

packages/autorest.typescript/test/rlcIntegration/urlRest.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { assert } from "chai";
 import UrlRest, {
+  getBooleanTrue,
```
qiaozha, line 3: @joheredi  what if we have two operations with the same name from different operation group ?  How are customers going to make it work if they want to use both of them inside the same test file ?

packages/autorest.typescript/test/rlcIntegration/generated/urlRest/src/index.ts
```diff
@@ -9,5 +9,6 @@ export * from "./responses";
 export * from "./clientDefinitions";
 export * from "./isUnexpected";
 export * from "./outputModels";
+export * from "./shortCuts/paths";
```
qiaozha, line 12: should remove this line

packages/autorest.typescript/test/rlcIntegration/generated/urlRest/src/shortCuts/paths.ts
qiaozha: src/shortCuts/paths.ts
src/shortCuts/foo.ts

PR #1629, opened 2022-10-24
fix constant array type

qiaozha:


packages/autorest.typescript/test/smoke/swagger/anomaly-detector-rest.md
```diff
@@ -11,7 +11,7 @@ description: AnomalyDetector Rest Client
 license-header: MICROSOFT_MIT_NO_VERSION
 output-folder: ../
 source-code-folder-path: ./src/
-require: https://raw.githubusercontent.com/Azure/azure-rest-api-specs/main/specification/cognitiveservices/data-plane/AnomalyDetector/readme.md
+require: https://raw.githubusercontent.com/Azure/azure-rest-api-specs/021e5ab6ec840d651d32a2c6bdf832800cdf8eb6/specification/cognitiveservices/data-plane/AnomalyDetector/readme.md
```
qiaozha, line 14: This main cause the change of anomaly detector smoke test change. fix it so that it won't cause problems in the future.

joheredi (approved):

packages/cadl-typescript/src/modelUtils.ts
```diff
@@ -536,6 +536,13 @@ function mapCadlTypeToTypeScript(
     case "ModelProperty":
       return mapCadlIntrinsicModelToTypeScript(program, cadlType, usage);
   }
+  if (cadlType.kind === undefined) {
```
MaryGao, line 539: Does that mean the inputted `type` have no `kind` information from CADL?

If so it could be a bug from cadl, maybe we need to create bug from them to track this.

MaryGao (approved):
Approved but I think we need to follow up this issue with CADL team

packages/cadl-typescript/src/modelUtils.ts
qiaozha: I already mentioned this in the chat and Jose said he will confirm with cadl team.

Issue #1547, opened 2022-09-05
[Cadl RLC]Handle form data type with binary type


MaryGao:
Handle form data type with binary type

MaryGao:
See discussion here: https://github.com/Azure/cadl-azure/issues/2006

PR #1630, opened 2022-10-24
Support multiple form data

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1547


qiaozha:
Is there a case where the response is specify with media type or form data ? Or is that moved into the GA scope ?

qiaozha (approved):

MaryGao:
> Is there a case where the response is specify with media type or form data ? Or is that moved into the GA scope ?

I notice that in autorest test we don't have that kind of test cases. And I am not sure whether the client core lib supports or not.

Do you know any real cases? If no I would prefer not to take effort on that.

PR #1634, opened 2022-10-25
use-stable-core-client-version

### LINKED ISSUE #1633: update autorest.typescript dependency @azure/core-client version to 1.6.1
Current version is an alpha one.

now that the version has been published, we should update it to use GA version.

qiaozha:
fixes #1633


MaryGao (approved):
LGTM

PR #1639, opened 2022-10-26
Paging enhancements for allowing consumers to participate in continuation

### LINKED ISSUE #1199: Paging support: byPage not exposing settings parameter
When generating an operation with paging, it mostly works well, except the `byPage` method that is generated does not pass the settings parameter down as defined on the `PagedAsyncIterableIterator` interface.

```typescript
export interface PagedAsyncIterableIterator<T, PageT = T[], PageSettingsT = PageSettings> {
  next(): Promise<IteratorResult<T, T>>;
  [Symbol.asyncIterator](): PagedAsyncIterableIterator<T, PageT, PageSettingsT>;
  byPage: (settings?: PageSettingsT) => AsyncIterableIterator<PageT>;
}
```
Generate code:
```typescript
public listAvailableWorkers(
    options?: RouterGetAvailableWorkersOptionalParams
  ): PagedAsyncIterableIterator<Worker> {
    const iter = this.getAvailableWorkersPagingAll(options);
    return {
      next() {
        return iter.next();
      },
      [Symbol.asyncIterator]() {
        return this;
      },
      byPage: () => {
        return this.getAvailableWorkersPagingPage(options);
      }
    };
  }
```

### LINKED ISSUE #1326: [Paging] Expose nextLink/Continuation Token as part of the returned items
Some customers need access to it to suspend and resume iterating. See https://github.com/Azure/azure-sdk-for-js/issues/20380#issuecomment-1051410026

xirzec:
This issue changes generated methods for pagination to take the `PageSettings` parameter in order to allow consumers to pass the continuationToken from a previous request. Similarly, it exposes a new helper function `getContinuationToken` to allow the consumer to extract a continuationToken from a previous page result.

I don't currently try to do anything with the maxPageSize field of PageSettings, since it seems difficult to generically handle that for clients without knowing which parameter the option affects (e.g. `top`).

Having some issues running the integration tests, but I wanted to get this up for feedback in case I'm implementing things in an inefficient manner.

Fixes #1326
Fixes #1199


packages/autorest.typescript/test/integration/generated/appconfigurationexport/src/appConfigurationClient.ts
```diff
@@ -152,22 +153,31 @@ export class AppConfigurationClient extends coreClient.ServiceClient {
...
+      byPage: (settings?: PageSettings) => {
+        return this.getKeysPagingPage(options, settings);
       }
     };
   }
 
   private async *getKeysPagingPage(
-    options?: GetKeysOptionalParams
+    options?: GetKeysOptionalParams,
+    settings?: PageSettings
```
deyaaeldeen, line 164: `maxPageSize` is not implemented, right? should `{ continuationToken?: string }` be used instead of `PageSettings`?

packages/autorest.typescript/src/generators/indexGenerator.ts
```diff
@@ -24,6 +24,10 @@ function generateHLCIndex(clientDetails: ClientDetails, file: SourceFile) {
   const { disablePagingAsyncIterators } = getAutorestOptions();
   if (clientDetails.options.hasPaging && !disablePagingAsyncIterators) {
     file.addStatements([`/// <reference lib="esnext.asynciterable" />`]);
```
deyaaeldeen, line 26: nitpick: unrelated but wanted to highlight it: I don't believe this is needed anymore since our minimum supported node version is 14 and async iterables is part of node 10.

packages/autorest.typescript/test/integration/generated/appconfigurationexport/src/appConfigurationClient.ts
xirzec: I think you're right that we shouldn't overpromise here.

packages/autorest.typescript/src/generators/indexGenerator.ts
xirzec: Certainly not needed from a platform compat perspective, I think it's more a question of if our tsconfig build target pulls these in or not, have to check. It's a good idea to update the target now that we have a new baseline, so I'll file an issue for it.

xirzec: https://github.com/Azure/azure-sdk-for-js/issues/23641

joheredi (approved):
Thanks for taking on this change 😄

packages/autorest.typescript/test/integration/generated/appconfigurationexport/src/appConfigurationClient.ts
xirzec: Revisiting this, it's a little tricky to use an anonymous type here since it gets referenced in multiple places. To avoid silently ignoring the parameter, I've added a runtime check to fail fast if someone passes it. A convenience client that wants to support it can still pull the parameter off and do the right thing before calling into the generated method.

Issue #1635, opened 2022-10-25
Support optional client-level parameter derived from x-ms-parameterized-host


MaryGao:
### Background

In recent RLC release we notice the requirement to handle optional client-level parameters in codegen. 
- One is for optional apiVersion in [anomalydetecter pr](https://github.com/saintzou66/azure-sdk-for-js/pull/2/files#diff-1aebd09d4f0df4e52c51dcbc6bfa057e8c5cda083c94bff5fff1ebb26ddc07c8)
- The other is a [reported issue](https://github.com/Azure/autorest.typescript/issues/1615) in devcenter

In our current implementation we put these parameters in the signature of client factory. This could work fine for required parameter without default value. However if there exists default value we couldn't leverage it defined in swagger.

### Current implementation
If we set the x-ms-parameterized-host as 
```
  "x-ms-parameterized-host": {
    "hostTemplate": "{Endpoint}/atlas/{serviceVersion}",
    "useSchemePrefix": false,
    "parameters": [
      {
        "$ref": "#/parameters/Endpoint"
      },
      {
        "$ref": "#/parameters/serviceVersion"
      }
    ]
  },
```
and generated client constructor factory function look like this 
```
export default function createClient(
  Endpoint: string,
  credentials: TokenCredential,
  serviceVersion: string,
  options: ClientOptions = {}
): PurviewCatalogClient {
  const baseUrl =
    options.baseUrl ?? `${Endpoint}/atlas/{serviceVersion}/catalog/api`;
    // ...
}
```

#### Option 1(preferred) - Put necessary parameter into XXXClientOption interface

```
export interface AzureDevCenterClientOption extends ClientOptions {
   devCenterDnsSuffix: string;
}

export default function createClient(
  tenantId: string,
  devCenter: string,
  credentials: TokenCredential,
  options: AzureDevCenterClientOption = {}
): AzureDevCenterClient { 
  // Handle the default value here
}
```

#### Option 2 - Put the default value in method signature

```
export default function createClient(
  tenantId: string,
  devCenter: string = "asia",  credentials: TokenCredential,
  devCenterDnsSuffix: string = "devcenter.azure.com",  options: ClientOptions = {}
): AzureDevCenterClient {
  const baseUrl = options.baseUrl ?? `https://${tenantId}-${devCenter}.${devCenterDnsSuffix}`
```
We have two concerns 
[...]

PR #1643, opened 2022-10-29
Support multiple parameters for parametrized host

joheredi:
This PR addresses the lack of support for multiple parameters in the parametrized host. This covers the following cases.

* When host has 2 parameters `{endpoint}/api/{apiVersion}`
* When host has many parameters `{host}.{subDomain}.{suffix}/api/{apiVersion}`

fixes https://github.com/Azure/autorest.typescript/issues/1635


packages/cadl-rlc-test/test/parametrizedHost/spec/main.cadl
```diff
@@ -0,0 +1,63 @@
...
+@useAuth(OAuth2Auth<[MyFlow]>)
+@versionedDependency(Azure.Core.Versions.v1_0_Preview_1)
+@server(
+  "{host}.{subdomain}.{sufix}.com/{apiVersion}",
+  "Confidential Ledger Service",
+  {
+    host: "one",
+    subdomain: "two",
+    sufix: "three",
+    apiVersion: "v1"
```
qiaozha, line 27: I wonder in this case, if these parameters are string type with a default value or just const strings? It looks like we are not treating it as const string here https://github.com/Azure/autorest.typescript/pull/1643/files#diff-26365d8da06b69db64937725fd469585f163940932d720de759bc478d5be1705R20-R23

packages/cadl-rlc-test/test/playfab/cadl-output/src/playFab.ts
```diff
@@ -10,7 +10,6 @@ export default function createClient(
   options: ClientOptions = {}
 ): PlayFabClient {
   const baseUrl = options.baseUrl ?? `${endpoint}`;
-  options.apiVersion = options.apiVersion ?? "0000-00-00";
```
qiaozha, line 13: why is this line missing ?

packages/cadl-typescript/src/transform/transform.ts
```diff
@@ -69,7 +73,7 @@ export async function transformRLCModel(program: Program, emitterOptions: RLCOpt
 
 function transformApiVersionParam(program: Program): Parameter | undefined {
   const apiVersion = getServiceVersion(program);
-  if (apiVersion) {
+  if (apiVersion && apiVersion !== "0000-00-00") {
```
qiaozha, line 76: why add this filter `apiVersion !== "0000-00-00"` ?

packages/cadl-typescript/src/transform/transform.ts
```diff
@@ -96,7 +100,8 @@ export function transformUrlInfo(program: Program): UrlInfo | undefined {
           urlParameters.push({
             name: key,
             type: getSchemaForType(program, type).type,
-            description: getDoc(program, type)
+            description: getDoc(program, type),
+            value: type.kind === "String" ? type.value : undefined
```
qiaozha, line 104: wonder if we should consider there're other types of const values in the urlParameters. like boolean or number ?

packages/rlc-common/src/buildClient.ts
```diff
@@ -123,10 +172,29 @@ function getClientFactoryBody(
...
+    }
+  }
+
+  // Special handling for API Version as query param
+  const apiVersion = getApiVersion(model);
+  let apiVersionStatement: string = "";
+  if (apiVersion) {
+    apiVersionStatement = `options.apiVersion = options.apiVersion ?? "${apiVersion}"`;
+  }
+
```
qiaozha, line 197: wonder why move this up ?

packages/autorest.typescript/test/commands/run.ts
```diff
@@ -141,7 +145,8 @@ export async function runAutorest(
     `--output-folder=${outputPath}`,
     `--use=.`,
     `--package-name=${packageDetails.name}`,
-    `--memory=8g`
+    `--memory=8g`,
+    "--use:https://tinyurl.com/23odwxl9"
```
qiaozha, line 149: Is this temporary for debugging or we actually need this ?

packages/autorest.typescript/test/commands/smoke-test-list.ts
```diff
@@ -243,6 +243,15 @@ export const readmes: SpecDefinition[] = [
...
     buildTag: "ci_rlc"
+  },
+  {
+    path: joinPath(
+      `${__dirname}`,
+      "..",
+      "./smoke/swagger/anomaly-detector-mv-rest.md"
+    ),
+    params: [AutorestParams.RestClient, AutorestParams.AzureSdkForJs],
+    buildTag: "debug"
```
qiaozha, line 254: Is this just for debug ?

joheredi: Good catch!

packages/cadl-rlc-test/test/parametrizedHost/spec/main.cadl
MaryGao: +1 I think the more proper way to indicate default value would be:
```
host?: string = "one";
```

packages/rlc-common/src/interfaces.ts
```diff
@@ -14,9 +14,15 @@ export interface RLCModel {
...
 
+export interface PathTemplateApiVersion {
+  value: string;
+  templateName: string;
+}
+
 export interface UrlInfo {
   endpoint?: string;
   urlParameters?: PathParameter[];
+  pathTemplateApiVersion?: PathTemplateApiVersion;
```
MaryGao, line 25: What is this param `pathTemplateApiVersion` used for?

packages/autorest.typescript/test/smoke/swagger/anomaly-detector-mv-rest.md
```diff
@@ -0,0 +1,21 @@
+# Azure Form Recognizer Protocol Layer
+
+> see https://aka.ms/autorest
+
+## Configuration
+
+```yaml
+package-name: "@azure-rest/ai-anomaly-mv-detector"
```
MaryGao, line 8: I didn't find the generated client for this one.

packages/rlc-common/src/buildClient.ts
```diff
@@ -123,10 +172,29 @@ function getClientFactoryBody(
   if (!model.options || !model.options.packageDetails || !model.urlInfo) {
     return "";
   }
+  const pathTemplateParameter = model.urlInfo?.pathTemplateApiVersion;
```
MaryGao, line 175: Where to use the param pathTemplateParameter?

packages/cadl-typescript/src/transform/transform.ts
MaryGao: Just wondering if you think the `0000-00-00` is not a valid apiVersion, could we move the logic to method `getServiceVersion`?

```
if(apiVersion !== "0000-00-00") {
 return undefine;
}
```

joheredi: Yeah it looks like if the CADL definition doesn't define a service version, the compiler gives us `"0000-00-00"` I'll move this to `getServiceVersion` and bring this up to the CADL team as it may be an issue in the compiler

packages/cadl-rlc-test/test/playfab/cadl-output/src/playFab.ts
joheredi: I removed because `"0000-00-00"` seems to be a bug in the compiler when the CADL doesn't define a service version

packages/cadl-typescript/src/transform/transform.ts
joheredi: Yeah, that is a good point, let me see what I can do here

packages/autorest.typescript/test/commands/run.ts
```diff
@@ -136,7 +140,7 @@ export async function runAutorest(
 
   commandArguments.push(
     inputFileCommand,
-    "--version=3.6.6",
+    "--version=3.9.3",
```
qiaozha, line 143: I found if I don't bump this autorest core version, the ci will just hang in there and CI will keep get timeout error

packages/autorest.typescript/test/integration/generated/nameChecker/src/models/index.ts
```diff
@@ -373,10 +367,7 @@ export type AutocompleteMode = "oneTerm" | "twoTerms" | "oneTermWithContext";
 
 /** Optional parameters. */
 export interface DocumentsCountOptionalParams
-  extends coreClient.OperationOptions {
-  /** Parameter group */
-  requestOptionsParam?: RequestOptions;
```
qiaozha, line 378: Here also has been removed.

packages/autorest.typescript/test/integration/generated/readmeFileChecker/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@msinternal/keyvault-secrets",
+  "name": "@azure/keyvault-secrets",
```
qiaozha, line 2: This looks strange

qiaozha: Guess this is also because the configuration in the readme.md has higher priority than the command line.

qiaozha (changes requested):
A lot of smoke test generated code has been removed. I don't think that's expected.

qiaozha (approved):

packages/cadl-typescript/test/unit/generate/models.spec.ts
```diff
@@ -354,7 +354,7 @@ describe("Input/output model type", () => {
       });
     });
 
-    it("should handle anonymous model -> effective type/interface", async () => {
+    it.skip("should handle anonymous model -> effective type/interface", async () => {
```
qiaozha, line 357: Skip it as this is probably not a valid case.

packages/autorest.typescript/test/integration/generated/nameChecker/src/models/index.ts
```diff
@@ -272,12 +272,6 @@ export interface AutocompleteRequest {
   top?: number;
 }
 
-/** Parameter group */
-export interface RequestOptions {
-  /** The tracking ID sent with the request to help with debugging. */
-  xMsClientRequestId?: string;
-}
-
```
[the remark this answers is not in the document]
qiaozha: leave as it is. If there's an real issue we will fix it later.

qiaozha (approved):

Issue #1628, opened 2022-10-24
[RLC] Add test cases for dictionary of cadl-ranch


MaryGao:
Now cadl-ranch has dictionary test cases.  we need to support in typescript emitter.

Issue #1617, opened 2022-10-19
[RLC] Add test cases to verify the support of array payload in methods for CADL


lirenhe:
Cadl:
op world(): string[];
op world2(): modelA[];
...

Need to verify whether JS RLC could generate correct code for the above operation.

- [x] Add unit test in `responses.spec.ts`
- [x] Add unit test in `parameters.spec.ts`
- [x] Support `array` integration test cases in cadl-ranch

MaryGao:
@kazrael2119 Could you help implement this?

lirenhe:
Here is the cadl ranch file for array type:
https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/array/item-types/main.cadl

PR #1644, opened 2022-10-31
Add cadl ranch [array & dictionary] integration test

MaryGao:
In this pr we'll cover below items:
- Upgrade the cadl-ranch to the latest
- Support array cases 
- Support dictionary cases

fixes https://github.com/Azure/autorest.typescript/issues/1628
fixes https://github.com/Azure/autorest.typescript/issues/1617


packages/cadl-typescript/test/integration/arrayItemTypes.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  beforeEach(() => {
+    client = ArrayItemTypesClientFactory({
+      allowInsecureConnection: true
+    });
+  });
+
+  matrix([testedTypes], async (params: TypeDetail) => {
+    it(`should get a ${params.type} value`, async () => {
+      try {
```
joheredi, line 62: do we need to try catch? I think if something throws it will automatically fail won't it?

packages/cadl-typescript/test/integration/generated/array/itemTypes/cadl-project.yaml
```diff
@@ -0,0 +1,12 @@
+emitters:
+  "../../../../../../../../../../../packages/cadl-typescript/dist/src/index.js":
```
joheredi, line 2: I wonder if we can just have @azure-tools/cadl-typescript and rush resolve it to the local project?

packages/cadl-typescript/test/unit/generate/models.spec.ts
```diff
@@ -199,6 +199,18 @@ describe("Input/output model type", () => {
...
       await verifyPropertyType(cadlType, typeScriptType);
     });
+
+    it("should handle never -> never", async () => {
+      const cadlType = "never";
+      const typeScriptType = "never";
+      await verifyPropertyType(cadlType, typeScriptType);
+    });
+
+    it("should handle unknown -> unknown", async () => {
```
MaryGao, line 209: Treat unknown as unknown

packages/cadl-typescript/test/integration/generated/array/itemTypes/cadl-project.yaml
qiaozha: Actually we have tried before, As this path is resolved by cadl compiler which doesn't belong to this project, rush can not help us.

qiaozha (approved):

packages/cadl-typescript/test/integration/arrayItemTypes.spec.ts
MaryGao: It will skip the rest of testing but here we want to continue testing

Issue #1555, opened 2022-09-08
[Cadl RLC] Design how to support read-only in properties


MaryGao:
In Cadl we could define properties' visibility as below:

```cadl
model Person {
  @visibility("read") id: string;
  @visibility("write") secret: string;
  name: string;

  @visibility("read", "write")
  test: string;

  @visibility("create", "read")
  create: string;
}
```

In swagger we could define read-only as below or by [extension](https://github.com/Azure/autorest/tree/main/docs/extensions#x-ms-mutability):
```json
"etag": {
    "type": "string",
    "readOnly": true,
    "description": "A unique read-only string that changes whenever the resource is updated."
  }
```

- [ ] Confirm with Cadl team about possible values in @visibility
- [ ] Ensure the supported cases in RLC
- [ ] Design how to handle read-only/write-only/... cases in RLC both for input & output models

MaryGao:
@timotheeguerin Could you suggest possible values for @visibility decorator? 

I notice we have ["create", "read", "write"] in our samples. Are they all values? What the `create` mean?
https://github.com/microsoft/cadl/blob/4d5d3b0647b741e09516140fcf8a4d5d1ed55402/packages/samples/visibility/visibility.cadl

timotheeguerin:
Those are the ones we have as knonw visibility 
```
export enum Visibility {
  Read = 1 << 0,
  Create = 1 << 1,
  Update = 1 << 2,
  Delete = 1 << 3,
  Query = 1 << 4,

  All = Read | Create | Update | Delete | Query,

  /**
   * Additional flag to indicate when something is nested in a collection
   * and therefore no metadata is applicable.
   */
  Item = 1 << 20,
}
```

See this PR too for how visibility is taken into account for rest api https://github.com/microsoft/cadl/pull/1034/files#diff-947587669197236eb5f8cb7bb2d6815c0c872aa8d2275c70e40231b1424cc648

`create` means it is available during the create process vs write which is create + update

MaryGao:
@timotheeguerin One follow-up question for these verbs in Visibility enum, do these actions map to the HTTP verbs to indicate 
detailed process? For example `create` process means a  `POST/PUT` REST call.

MaryGao:
Design concluded here: https://github.com/Azure/azure-sdk/issues/4772#issuecomment-1269051265

My understanding is 

- If readonly is marked in output model, we'll tag as `readonly`
- If readonly is marked in input model, we'll remove these properties from input models

Issue #1553, opened 2022-09-08
[Cadl RLC] Design the generic/template models and provide user-friendly name for them


MaryGao:
This issue may split into two issues. We'd like to cover two parts:

- Rethink the current design for generic/template models from Cadl
- Fix the naming issue if underlying template issue don't provide a friendly name

MaryGao:
Regarding the second one we may reference this issue: https://github.com/Azure/cadl-azure/issues/1995

weidongxu-microsoft:
Note that according to Johan, there should not be a new model for request (at least for PUT), as he expect a single routd-trip model. E.g. request and response model both as "Project".

PR #1647, opened 2022-11-02
Try fix modle name in merge and patch

qiaozha:
fixes https://github.com/Azure/autorest.typescript/issues/1555
fixes https://github.com/Azure/autorest.typescript/issues/1553


packages/autorest.typescript/test/smoke/generated/agrifood-data-plane/src/outputModels.ts
```diff
@@ -33,23 +33,23 @@ export interface ApplicationDataOutput {
   /** End date-time of the operation data, sample format: yyyy-MM-ddTHH:mm:ssZ. */
   operationEndDateTime?: string;
   /** Link for attachments. */
-  attachmentsLink?: string;
+  readonly attachmentsLink?: string;
   /** Optional boundary ID of the field for which operation was applied. */
```
MaryGao, line 37: Do we have an example or testing to cover readOnly in input models?
My understanding is 

- If readonly is marked in output model, we'll tag as `readonly`
- If readonly is marked in input model, we'll remove these properties from input models

qiaozha: This model is used in input models, https://github.com/qiaozha/autorest.typescript/blob/try-fix-modle-name-in-merge-and-patch/packages/autorest.typescript/test/smoke/generated/agrifood-data-plane/src/models.ts#L5-L54 we don't change anything which is the correct behavior.

qiaozha: readonly has nothing to do with the http method. not sure why you mention PATCH.

joheredi (approved):

PR #1657, opened 2022-11-05
Upgrade to TypeScript 4.8

xirzec:
This PR updates our TS dependencies for all generator packages and newly generated packages.


jeremymeng (approved):
Looks good to me! Thanks for the upgrade! There's formatting changes but they look better.

xirzec:
> Looks good to me! Thanks for the upgrade! There's formatting changes but they look better.

I undid the formatting changes since I think it'll differ from what gets autogenerated. Can fix that separately another time. 😄

PR #1664, opened 2022-11-16
Extensible enum in host parameter

MaryGao:
In this [pr ](https://github.com/joheredi/autorest.typescript/pull/1/files#r1022105262) we handle the cadl extensible enum as string literal, I think the more consistent way is handled as string. See our current type map test cases [here](https://github.com/Azure/autorest.typescript/blob/93b44b67c19319d1ce8bec78c8b258a6afe916c5/packages/cadl-typescript/test/unit/generate/models.spec.ts#L94).

previous implementation:
https://github.com/Azure/autorest.typescript/blob/35480556a076e0d82c902a376a02e813b3b20c72/packages/cadl-rlc-test/test/anomalyDetector/cadl-output/src/anomalyDetector.ts#L11-L16

Current pr:
https://github.com/Azure/autorest.typescript/blob/93b44b67c19319d1ce8bec78c8b258a6afe916c5/packages/cadl-typescript/test/unit/generate/clientFactory.spec.ts#L188-L197


qiaozha:
can we close this one ?

PR #1665, opened 2022-11-18
Remove cadl azure core models in generated code

qiaozha:
fixes https://github.com/Azure/autorest.typescript/issues/1452


.scripts/ci.yml
```diff
@@ -100,20 +100,16 @@ stages:
           matrix:
             macOS_Node16:
               Pool:
-              OSVmImage: "macOS-12"
-              NodeTestVersion: "16.x"
-            Browser_Linux_Node16:
-              Pool: ${{ parameters.LinuxPool }}
-              OSVmImage:
-              NodeTestVersion: "16.x"
-            Linux_Node14:
```
MaryGao, line 109: Why we remove the above two envs?

qiaozha: they are duplicated

packages/cadl-typescript/src/modelUtils.ts
```diff
@@ -289,6 +293,35 @@ function isSchemaProperty(program: Program, property: ModelProperty) {
   return !(headerInfo || queryInfo || pathInfo || statusCodeinfo);
 }
 
+function extractPagedMetadataNested(
+  program: Program,
```
MaryGao, line 297: I don't think we need to define this func again.
https://github.com/Azure/autorest.typescript/blob/3ee5d2733b8be991c7787937a6fd4ad554f1a3ea/packages/cadl-typescript/src/operationUtil.ts#L160

packages/rlc-common/src/buildObjectTypes.ts
```diff
@@ -46,6 +49,40 @@ export function buildObjectInterfaces(
   return objectInterfaces;
 }
 
+export function buildObjectAliases(
```
MaryGao, line 52:
```suggestion
export function buildPagedObjectAliases(
```

packages/rlc-common/src/buildResponseTypes.ts
```diff
@@ -20,10 +20,12 @@ import {
   getResponseTypeName
 } from "./helpers/nameConstructors.js";
 
+let hasErrorResponse = false;
```
MaryGao, line 23: Not sure there will be a bug in multi-client package.

When we have two clients but only one of them imports error response, but the other will not.

qiaozha: already initate the value as false at line 28. each client will get a false at first.

packages/rlc-common/src/buildObjectTypes.ts
MaryGao, line 52: Do we still need to build aliases expect the paged cases?

qiaozha: Not sure I get your point, Paged has already been added into the core-paging.

packages/cadl-rlc-test/test/contoso/cadl-output/review/contosowidgetmanager-rest.api.md
```diff
@@ -7,8 +7,11 @@
 import { Client } from '@azure-rest/core-client';
 import { ClientOptions } from '@azure-rest/core-client';
 import { CreateHttpPollerOptions } from '@azure/core-lro';
+import { ErrorModel } from '@azure-rest/core-client';
+import { ErrorResponse } from '@azure-rest/core-client';
 import { HttpResponse } from '@azure-rest/core-client';
 import { OperationState } from '@azure/core-lro';
+import { Paged } from '@azure/core-paging';
```
qiaozha, line 14: see examples of Paged here

packages/rlc-common/src/interfaces.ts
```diff
@@ -141,6 +141,9 @@ export interface Schema {
   default?: any;
   readOnly?: boolean;
   usage?: SchemaContext[];
+  alias?: string;
+  outputAlias?: string;
```
MaryGao, line 145: I feel like our alias logic is sperated into different places, sometimes in schema, sometimes in responses.

I smelled that we need to refactor the alias logic if the modelUtil could have it, it's better to move necessary logic into there.

packages/rlc-common/src/buildObjectTypes.ts
MaryGao: I mean the `buildObjectAliases` is to get all alias definition.

Currently we have one case to generate alias, that is paged.

qiaozha: no, for polymorphism, we also use alias.

packages/rlc-common/src/interfaces.ts
qiaozha: when do we use alias in response ?

MaryGao: Maybe I remember wrongly but we have alias model in merge+path, right?

MaryGao (commented):
I just left small comments.

@joheredi Could you help review this pr?

packages/rlc-common/src/interfaces.ts
qiaozha: sync offline, would like keep merge and patch as it is.

joheredi (approved):

Issue #1667, opened 2022-11-22
[Cadl RLC] Fix the extensible enum issue and support apiVersion default value suggested from cadl-dpg


MaryGao:
- [x] Fix the extensible enum issue 
- [ ] Support apiVersion default value suggested from cadl-dpg

In this [pr ](https://github.com/joheredi/autorest.typescript/pull/1/files#r1022105262) we handle the cadl extensible enum as string literal, I think the more consistent way is handled as string. See our current type map test cases [here](https://github.com/Azure/autorest.typescript/blob/93b44b67c19319d1ce8bec78c8b258a6afe916c5/packages/cadl-typescript/test/unit/generate/models.spec.ts#L94).

previous implementation:
https://github.com/Azure/autorest.typescript/blob/35480556a076e0d82c902a376a02e813b3b20c72/packages/cadl-rlc-test/test/anomalyDetector/cadl-output/src/anomalyDetector.ts#L11-L16

Current pr:
https://github.com/Azure/autorest.typescript/blob/93b44b67c19319d1ce8bec78c8b258a6afe916c5/packages/cadl-typescript/test/unit/generate/clientFactory.spec.ts#L188-L197

draft pr is here: https://github.com/Azure/autorest.typescript/pull/1664

PR #1676, opened 2022-11-28
Fix the optional url parameter issue and set default value for apiversion

MaryGao:
Fix the optional url parameter issue and set default value for apiversion

Bump the versions

- rlc-common
- cadl-typescript
- autorest.typescript

fixes https://github.com/Azure/autorest.typescript/issues/1667


packages/autorest.typescript/src/restLevelClient/transforms/transform.ts
```diff
@@ -42,7 +42,7 @@ export function transform(model: CodeModel): RLCModel {
     schemas: transformSchemas(model),
     responses: transformResponseTypes(model, importDetails),
     importSet: importDetails,
-    apiVersionParam: transformApiVersionParam(model),
+    apiVersionInQueryParam: transformApiVersionParam(model),
```
MaryGao, line 45: Rename the property because it only indicates the api-version in query param. If we have api-version in url it will save in UrlInfo

qiaozha (approved):

Issue #1677, opened 2022-11-26
Openapi with a query param with name 'type' is incorrect generated


jrpedrianes:
- **Package Name**:  @autorest/typescript
- **Package Version**:  6.0.0-rc.3
- **Operating system**: MacOS
- [ x ] **nodejs**
    - **version**: v16.18.1
- [  ] **browser**
    - **name/version**: 
- [  ] **typescript**
    - **version**:
- Is the bug related to **documentation** in
  - [ ] README.md
  - [ ] source code documentation
  - [ ] SDK API docs on https://docs.microsoft.com

**Describe the bug**
I think that when a query parameter called 'type' appears in an openapi.yaml file the code generated has an issue.

**To Reproduce**

Given the next openapi  file:

```yaml
openapi: 3.0.3

info:
  title: Test
  version: '1.0'

servers:
  - url: 'https://api.example.com/v1.0'
    description: Live environment

paths:
  /echo:
    get:
      summary: 'summary'
      operationId: requestEcho
      parameters:
        - name: type
          in: query
          schema:
            type: string
      responses:
        '200':
          description: description.
          content:
            application/json:
              schema:
                type: object
                properties:
                  prop1:
                    type: string
````

And the following autorest configuration:

```yaml
v3: true
typescript: true

input-file: ./openapi.yaml
output-folder: ./generated

openapi-type: data-plane
add-credential: true
package-name: 'example'
package-version: '0.0.1'

generate-metadata: false
generate-test: false
generate-sample: false
hide-clients: true

use-extension:
  '@autorest/typescript': '6.0.0-rc.3'
  '@autorest/modelerfour': '4.25.0'

modelerfour:
  seal-single-value-enum-by-default: false
```

When generating the code, the `type` query param is named `typeParam` in the related interface:
```javascript
/** Optional parameters. */
export interface RequestEchoOptionalParams extends coreClient.OperationOptions {
  typeParam?: string
}
```

But then the mapper is wrong:
```javascript
export const typeParam: OperationQueryParameter = {
  parameterPath: ['options', 'type'],
  mapper: {
    serializedName: 'type',
    type: {
      name: 'String',
    },
  },
}
```

Due to in the `interface` the query parameter is `typeParam` but then when must be used to build the final URL the parameter path used is `['options', 'type']`, with this the REST library never will add it to the URL.

**Expected behavior**

- The parameter is added to the URL
[...]

PR #1684, opened 2022-12-05
fix-parameter-rename-in-options

qiaozha:
fixes https://github.com/Azure/autorest.typescript/issues/1677 
refer https://github.com/Azure/autorest.typescript/pull/1346


packages/autorest.typescript/test/unit/transforms/parameterTransforms.spec.ts
```diff
@@ -362,5 +362,123 @@ describe("parameterTransforms", () => {
       assert.equal(p1.parameter, param1);
       assert.equal(p2.parameter, param2);
     });
+
+    it("should guard required query parameters name", () => {
```
MaryGao, line 366: I didn't get the point here, why do we need to guard the name when it is required?

MaryGao (approved):
LGTM

Issue #1648, opened 2022-11-02
[RLC] Handle the unknown and never type from CADL in codegen


MaryGao:
Currently the unknown and never type are directly mapped to TypesScript unknown, never. We have two concerns for this implementation: 
- `never` type tagged on property is wired because end user cannot refer or assign value to it
- `unknown` type is meaningful for input model not for output model, after getting the model details from server end user may adjust values and then re-assign to other models.

So we have new propersals for these two types:
| Cadl Type |  Input Type | Output Type   | 
|---|---|---|
| never| ignore   | ignore  | 
| unknown| unknown | any|

Here we have examples for them:

- `never` in OperationStatus from azure-core: https://github.com/Azure/cadl-azure/blob/307d4c72fe43f87c5eadda5c02901a7760d3cdb9/packages/cadl-azure-core/lib/foundations.cadl#L20-L33
- `unknown` in Cadl-ranch example: https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/arrays/item-types/main.cadl#L85

@joheredi Could you confirm the above design with Brian?

qiaozha:
for unknown, we can treat it as any in the output.

PR #1685, opened 2022-12-05
Fix unknown never type

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1648


packages/cadl-typescript/src/modelUtils.ts
```diff
@@ -103,7 +103,12 @@ export function getSchemaForType(
     return type;
   }
   if (type.kind === "Intrinsic" && type.name === "unknown") {
```
qiaozha, line 105: can we use isUnknownType here ?

packages/cadl-typescript/src/modelUtils.ts
```diff
@@ -103,7 +103,12 @@ export function getSchemaForType(
...
-    return { type: "unknown" };
+    const returnType: any = { type: "unknown" };
+    if (usage && usage.includes(SchemaContext.Output)) {
+      returnType.outputTypeName = "any";
+      returnType.typeName = "unknown";
+    }
+    return returnType;
   }
   if (type.kind === "Intrinsic" && type.name === "never") {
```
qiaozha, line 113: isNeverType ?

packages/cadl-typescript/src/modelUtils.ts
```diff
@@ -693,6 +698,13 @@ function mapCadlIntrinsicModelToTypeScript(
             schema.outputTypeName = `Record<string, ${valueType.name}Output>`;
             schema.outputValueTypeName = `${valueType.name}Output`;
           }
+        } else if (isUnknownType(indexer.value!)) {
+          schema.typeName = `Record<string, ${valueType.type}>`;
+          schema.valueTypeName = valueType.name;
```
qiaozha, line 703: Is this valueTypeName used for imported models ? If yes, I am not sure what we should import if the type is Record<string, unknown> or Record<string, any>

packages/cadl-typescript/src/modelUtils.ts
```diff
@@ -693,6 +698,13 @@ function mapCadlIntrinsicModelToTypeScript(
...
+          if (usage && usage.includes(SchemaContext.Output)) {
+            schema.outputTypeName = `Record<string, ${valueType.outputTypeName}>`;
+            schema.outputValueTypeName = `${valueType.outputTypeName}`;
```
qiaozha, line 706: same here.

packages/cadl-typescript/src/modelUtils.ts
```diff
@@ -874,6 +886,12 @@ function getPriorityName(schema: Schema): string {
   return schema.outputTypeName ?? schema.typeName ?? schema.name;
 }
 function getDictionaryValueName(schema: DictionarySchema): string | undefined {
+  if (
+    schema.outputValueTypeName === "any" ||
+    schema.outputValueTypeName === "unknown"
+  ) {
+    return undefined;
+  }
```
qiaozha, line 894: don't think we should set it at the first.

packages/cadl-typescript/test/unit/generate/responses.spec.ts
```diff
@@ -104,6 +104,24 @@ describe("Responses.ts", () => {
   });
 
   describe("body generation", () => {
+    it("should generate Record<unknown> as body property", async () => {
```
qiaozha, line 107: Can we have a unit test for request body as Record<unknown> ?

qiaozha (approved):

PR #1690, opened 2022-12-08
Fix custom wrapper issue in adp

### LINKED ISSUE #1679: [Cadl RLC] Can't get correct body type with wrapped body as parameter
We have below Cadl definition:

https://github.com/Azure/azure-rest-api-specs-pr/blob/2d0503cc27e26d1c3e0c24bce53d539c5095f938/specification/adp/DataManagement/ingestion/discovery/operations.cadl#L12-L17

Currently the generated body was `Discovery` the correct body type would be `DiscoveryCreationParameters`.

- [ ] transformBodyParameters should return correct body type
- [ ] transformSchemas should generate `DiscoveryCreationParameters` model

qiaozha:
fixes #1679


packages/cadl-typescript/src/transform/transformSchemas.ts
```diff
@@ -30,13 +33,18 @@ export function transformSchemas(program: Program, client: Client) {
     transformSchemaForRoute(route);
   }
   function transformSchemaForRoute(route: HttpOperation) {
+    if (route.operation.name === "UploadBatchServiceLogs") {
```
MaryGao, line 36: Is this line debugging code?

packages/cadl-typescript/src/transform/transformSchemas.ts
```diff
@@ -30,13 +33,18 @@ export function transformSchemas(program: Program, client: Client) {
...
+      route;
+    }
     if (route.parameters.bodyType) {
-      let bodyModel = route.parameters.bodyType;
-      const operation = getResourceOperation(program, route.operation);
-      if (operation) {
-        bodyModel = operation.resourceType;
+      if (route.parameters.bodyType.kind === "Model" && route.parameters.bodyType.name === "UploadBatchServiceLogsResult") {
+        route;
```
MaryGao, line 41: +1

packages/cadl-typescript/src/transform/transformSchemas.ts
qiaozha: removed

packages/cadl-typescript/src/transform/transformSchemas.ts
qiaozha: removed

packages/cadl-rlc-test/test/authoring/spec/main.cadl
```diff
@@ -100,6 +102,12 @@ model Deployment {
   name: string;
 }
 
+@doc("Deployment resource creation parameters.")
+@withVisibility("create")
+model DeploymentCreationParameters {
+  ...Deployment;
+}
```
MaryGao, line 109: I was wondering if we could add this part into testing not in smoke-test.

Because we sync the authoring cadl from cadl project. If we add here we would manully maintain the effort.

qiaozha: Because this needs additional decorator not sure how to do that in the unit test

packages/cadl-typescript/src/transform/transformSchemas.ts
```diff
@@ -31,11 +34,7 @@ export function transformSchemas(program: Program, client: Client) {
   }
   function transformSchemaForRoute(route: HttpOperation) {
     if (route.parameters.bodyType) {
-      let bodyModel = route.parameters.bodyType;
-      const operation = getResourceOperation(program, route.operation);
-      if (operation) {
-        bodyModel = operation.resourceType;
-      }
+      const bodyModel = getBodyType(program, route);
```
MaryGao, line 37:
```suggestion
      const bodyModel = getBodyType(program, route);
      if (bodyModel && bodyModel.kind === "Model") {
        getGeneratedModels(bodyModel, SchemaContext.Input);
      }
```

packages/cadl-typescript/src/modelUtils.ts
```diff
@@ -905,3 +907,41 @@ export function getFormattedPropertyDoc(
   }
   return propertyDoc ?? enhancedDocFromType;
 }
+
+export function getBodyType(program: Program, route: HttpOperation) {
```
MaryGao, line 911: Not sure we could specify the return type

packages/cadl-typescript/src/modelUtils.ts
```diff
@@ -120,6 +106,22 @@ export function getSchemaForType(
   });
   return undefined;
 }
+export function getEffectiveModelFromType(program: Program, type: Type) {
```
MaryGao, line 109: Not sure we could specify the return type

packages/cadl-typescript/src/modelUtils.ts
```diff
@@ -905,3 +907,41 @@ export function getFormattedPropertyDoc(
   }
   return propertyDoc ?? enhancedDocFromType;
 }
+
+export function getBodyType(program: Program, route: HttpOperation) {
+  let bodyModel = route.parameters.bodyType;
+  if (bodyModel && bodyModel.kind === "Model" && route.operation) {
+    const resourceType = getResourceOperation(
```
MaryGao, line 914: I notice that it's hard to get the body type, not sure you know some context and then we could add some comments for why we do this way.

packages/cadl-rlc-test/test/authoring/spec/main.cadl
MaryGao: Could we keep it to a sperate RP?

packages/cadl-typescript/src/modelUtils.ts
qiaozha: In the case where we are using resource operation from cadl-azure-core, there's a resource type, if they don't use any custom wrapper we can use resourceType to replace the body type, but in the customization case, we need to keep the body type not replace it.

joheredi (approved):

PR #1715, opened 2023-01-03
Add openai smoke test for cadl

qiaozha:


MaryGao (approved):
LGTM and please fix the build error

PR #1716, opened 2023-01-05
Cadl openai generation

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1706

apiView: https://apiview.dev/Assemblies/Review/0ab4f095a315468ba416ae3cafe0f08f


packages/cadl-typescript/src/transform/transfromRLCOptions.ts
```diff
@@ -79,6 +80,12 @@ function processAuth(program: Program) {
           if (!securityInfo.credentialScopes) {
             securityInfo.credentialScopes = [];
           }
+          if (flow.scopes.length === 0) {
+            reportDiagnostic(program, {
+              code: "no-credential-scopes",
```
MaryGao, line 85: Report warning when no scope defined

packages/cadl-rlc-test/test/openai/cadl-output/src/parameters.ts
```diff
@@ -1,17 +1,50 @@
...
+
+export interface EmbeddingsHeaders {
+  /** The ID of the end-user, for use in tracking and rate-limiting. */
+  user?: string;
+}
 
 export interface EmbeddingsBodyParam {
-  body?: EmbeddingsRequest;
+  body?: {
+    input_type: string;
```
MaryGao, line 14: @joheredi We implement the CADL spread alias as anonymous model, because the alias info would be erased so there is no model name for this body type.

Do you have any concern for this design?

https://github.com/Azure/autorest.typescript/blob/b6734113ecaef100844d0231c89fa453cd142145/packages/cadl-rlc-test/test/openai/spec/models/completions.create.cadl#L9-L26

.vscode/launch.json
```diff
@@ -84,7 +84,7 @@
     {
       "name": "Generate code for CadlRlcTest",
       "request": "launch",
-      "cwd": "${workspaceFolder}/packages/cadl-rlc-test",
+      "cwd": "${workspaceFolder}/packages/cadl-rlc-test/test/openai",
```
joheredi, line 87: do we need this change?

packages/cadl-rlc-test/test/openai/cadl-output/src/parameters.ts
joheredi: Yeah, if we can't get a model name from CADL I prefer to have the anonymous model definition. This will protect us from generating awkward model names

joheredi (approved):
Thanks for adding the unit tests! I'm looking forward to see some usage samples for this library

.vscode/launch.json
MaryGao: will revert this change

PR #1719, opened 2023-01-11
support-collection-format-in-rlc

### LINKED ISSUE #1687: RLC support collection format
In maps-render, we have an API Get Map Static Image, which accept a query parameter pinsto create pins on the maps.

We can make multiple pins by adding specifying pins key/value pair multiple times like following:
```
?pins=<pin query 1>&pins=<pin query 2>
```
But in RLC, the generated type for pins is an array, and it will serialize the query as following:
```
?pin=<qpin query 1>,<pin query 2>
```
here's the api-spec for pins: https://github.com/Azure/azure-rest-api-specs/blob/83d42b77e751deafcf46784877922ec1538f41c6/specification/maps/data-plane/Render/stable/2022-08-01/render.json#L670

The proposal is to generate different collectionFormat helper to convert the array into a string, and the generated type for those array type should be `array | string`
```
client.path("/foo").get({
  queryParameters: {
  foo: createMultiCollection([1, 2, 3, 4, 5]), // foo=1&foo=2&foo=3&foo=4&foo=5 for this case, we should return the foo: 1&foo=2&foo=3&foo=4 and give a hint to our customer that they should  also pass skipUrlEncoding at the same time.
  baz: createCsvCollection([1,2,3,4,5]) // baz=1,2,3,4,5
  pipe: createPipeCollection([1,2,3,4,5]) // pipe=1|2|3|4|5
}
})
```

FYI for other kinds of collection format: https://swagger.io/docs/specification/2-0/describing-parameters/#:~:text=csv%20(-,default,-). 

Action items, we also need to support this in cadl emitter after cadl supports collectionFormat.

qiaozha:
fixes #1687


joheredi (commented):
its looking good, make sure to un comment the commented tests

packages/rlc-common/src/static/serializeHelper.ts
```diff
@@ -0,0 +1,29 @@
+export const buildMultiCollectionContent = `
+export function buildMultiCollection(
+  oriArray: string[],
```
qiaozha, line 3: queryParameters

qiaozha: parameterName=v1&parameterName=v2

packages/autorest.typescript/test/rlcIntegration/generated/multipleUrlParameters/src/parameters.ts
```diff
@@ -38,14 +38,14 @@ export interface EntityListByGuidsPathParameters {
 }
 
 export interface EntityListByGuidsQueryParamProperties {
-  /** An array of GUIDs of entities to list. */
-  guid: Array<string>;
+  /** An array of GUIDs of entities to list.This type has special collectionFormat, we provide buildMultiCollection from serializeHelper.ts to help */
```
qiaozha, line 41: This parameter needs to be formated as multi collection. we provide buildMultiCollection from serializeHelper.ts to help

packages/autorest.typescript/test/rlcIntegration/generated/multipleUrlParameters/src/parameters.ts
```diff
@@ -38,14 +38,14 @@ export interface EntityListByGuidsPathParameters {
...
+  guid: Array<string> | string;
```
qiaozha, line 42: remove Array<string> type.

joheredi (approved):

packages/autorest.typescript/test/rlcIntegration/urlRest.spec.ts
```diff
@@ -478,17 +478,17 @@ describe("Integration tests for UrlRest", () => {
         .path("/queries/array/pipes/string/valid")
         .get({
           queryParameters: {
-            arrayQuery: testArray.join("|") as any
+            arrayQuery: testArray.join("|")
```
MaryGao, line 481: Can we use the helper function here?

packages/autorest.typescript/test/rlcIntegration/urlRest.spec.ts
```diff
@@ -478,17 +478,17 @@ describe("Integration tests for UrlRest", () => {
...
           }
         });
       const result3 = await client.path("/queries/array/ssv/string/valid").get({
         queryParameters: {
-          arrayQuery: testArray.join(" ") as any
+          arrayQuery: testArray.join(" ")
```
MaryGao, line 486: +1

MaryGao (approved):
LGTM and left small comments!

PR #1727, opened 2023-01-17
Upgrade dev dependency `mkdirp` to `^2.1.2`

jeremymeng:


jeremymeng:
@MaryGao any tips on fixing the smoke tests Check Git Tree failures?

joheredi:
> @MaryGao any tips on fixing the smoke tests Check Git Tree failures?

I think these need to be updated as well: 

https://github.com/Azure/autorest.typescript/blob/d51bd1c7507f4a6630923ddf62fc8e2d20b5ab00/packages/rlc-common/src/metadata/buildPackageFile.ts#L153

https://github.com/Azure/autorest.typescript/blob/d51bd1c7507f4a6630923ddf62fc8e2d20b5ab00/packages/autorest.typescript/src/generators/static/packageFileGenerator.ts#L116

jeremymeng:
> I think these need to be updated as well:

Thanks @joheredi! Not sure how my search missed these

deyaaeldeen (approved):
Thanks for upgrading us👍

Issue #1598, opened 2022-10-09
[RLC] Upgrade core-lro & Turn on the option resolveOnUnsuccessful in LRO helper function


MaryGao:
- [ ] Resolve the failure in pr https://github.com/Azure/autorest.typescript/pull/1586
- [ ] Enable the option resolveOnUnsuccessful

## Enable the option resolveOnUnsuccessful

In [core-lro pr](https://github.com/Azure/azure-sdk-for-js/pull/23311) we intro a new flag and we need to enable it in this issue.

> `resolveOnUnsuccessful` flag that controls whether to throw an error if the operation failed or was canceled. This flag is turned off by default but REST-level clients should turn it on to maintain the direction of not raising exceptions when receiving unexpected responses.

Simply speaking it would change our behavior from throw Exception to not throw in Rest LRO helper function below is an example from our test case.

### My concern to enable this flag

If we will not throw this execption from LRO helper function the customers would leverage the `switch` statement to differentiate the success or failure cases. It would introduce more code lines to handle the conditions. I am not sure if this is elegant way to handle failures compared with `throw` way.

@joheredi @deyaaeldeen @qiaozha  How do you think ? From REST client perspective it's better to enable it or not?

``` typescript
type Response = { result?: any; error?: any };
  const poller = {} as SimplePollerLike<OperationState<Response>, Response>;
  const response = await poller.pollUntilDone();
  switch (poller.getOperationState().status) {
    case "canceled":
    case "failed":
      console.log(response.error);
      break;
    case "succeeded":
      console.log(response.result)
      break;
  }

```

### Before - Throw case

```typescript
it("should handle deleteAsyncRelativeRetryInvalidHeader ", async () => {
  try {
    const initialResponse = await client
      .path("/lro/error/deleteasync/retry/invalidheader")
      .delete();

    const poller = await getLongRunningPoller(client, initialResponse, {
      intervalInMs: 0
    });

    await poller.pollUntilDone();
    assert.fail("Scenario should throw");
  } catch (error) {
    //  =============> the LRO helper would throw exception <===============
    assert.equal(error.message, "The long-running operation has failed");
  }
});
```

### After - Failure case

```typescript

it.only("should handle deleteAsyncRelativeRetryInvalidHeader ", async () => {
  try {
    const initialResponse = await client
      .path("/lro/error/deleteasync/retry/invalidheader")
      .delete();

    const poller = await getLongRunningPoller(client, initialResponse, {
      intervalInMs: 0
[...]

PR #1739, opened 2023-02-02
Update the LRO resolveOnUnsuccessful  flag in RLC

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1598


packages/autorest.typescript/test/rlcIntegration/lroRest.spec.ts
```diff
@@ -273,13 +286,13 @@ describe("LRO Rest Client", () => {
...
         intervalInMs: 0
       });
-
       const result = await poller.pollUntilDone();
       if (isUnexpected(result)) {
         const error = `Unexpected status code ${result.status}`;
         assert.fail(error);
-        throw new Error(error);
       }
+      // FIXME
```
MaryGao, line 294: @deyaaeldeen We have 8 cases failure after enabling this flag. Could you help on that?



https://dev.azure.com/azure-sdk/public/_build/results?buildId=2156796&view=logs&j=71cc2e7d-6454-5bf0-71b5-f1eef07a6c23&t=f2857310-5d68-5640-1409-0ba88dcfe076&l=1727

deyaaeldeen: I am looking now but I suspect the test expectation should be updated.
For the first failure, the operation succeeded, see [the corresponding core-lro test](https://github.com/Azure/azure-sdk-for-js/blob/6a76e6ac3f67fdf7b2d22c7b3a242538b0c60f54/sdk/core/core-lro/test/lro.spec.ts#L302).

MaryGao: I updated these test cases and then the relevant behaviors are not as my expectation.

packages/autorest.typescript/test/rlcIntegration/lroRest.spec.ts
```diff
@@ -273,13 +286,13 @@ describe("LRO Rest Client", () => {
...
+      assert.equal(poller.getOperationState().status, "failed");
```
MaryGao, line 295: Considering the `result.body.properties?.provisioningState` is failed status so I assume the operation state is also `falied` not `succeeded`

packages/autorest.typescript/test/rlcIntegration/lroRest.spec.ts
```diff
@@ -872,24 +868,19 @@ describe("LRO Rest Client", () => {
...
+        .delete();
+
+      assert.equal(initialResponse.status, "202");
+      assert.isNotTrue(isUnexpected(initialResponse));
+      const poller = await getLongRunningPoller(client, initialResponse, {
+        intervalInMs: 0
+      });
+      const result = await poller.pollUntilDone();
+      // FIXME: Polling was unsuccessful. Expected status to have a string value or no value but it has instead: 400. This doesn't necessarily indicate the operation has failed. Check your Azure subscription or resource status for more information.
+      assert.equal(result.status, "400");
```
MaryGao, line 882: I assume there will be no exception thrown if I enable the flag.

But here I met the ERROR.

packages/autorest.typescript/test/rlcIntegration/lroRest.spec.ts
deyaaeldeen: I updated expectations in https://github.com/Azure/autorest.typescript/commit/c989dd5e7044e2b9c2c2173751faf53a22cb52f8

packages/autorest.typescript/test/rlcIntegration/lroRest.spec.ts
deyaaeldeen: This test operates in the resource location mode where success or failure is determined solely based on the status code of the response and since it is 200 in this case then the operation should be considered succeeded.

packages/autorest.typescript/test/rlcIntegration/lroRest.spec.ts
deyaaeldeen: This is a failure related to unexpected body format where the status field is not a string. There is no way to understand this body so an error is thrown.

MaryGao: So could you clarify the idea of resolveOnUnsuccess flag here? Which cases could be handled by lib? Which could not?


Previously I naively thought there will be no exception thrown when enabling this flag

MaryGao: @joheredi Could you also review this pr?

deyaaeldeen: You're right that we are trying to minimize the number of thrown exceptions in RLCs and we provide the `isUnexpected` helper to inspect response objects for errors instead of throwing a `RestError`. However, there are situations where there is nothing you can do but to throw and one of those situations is failed parsing of the response. `@azure-rest/core-client` itself throws an exception in this case, see [[here]](https://github.com/Azure/azure-sdk-for-js/blob/ffdbb47c298497623af4790e1f085d19f7b36c50/sdk/core/core-client-rest/src/sendRequest.ts#L233). This failed parsing exception can't be handled by core-lro in any meaningful way because it could be the case that the operation is still running but the response wasn't a proper JSON for some reason. This is very unlikely to happen in reality though.

`resolveOnUnsuccessful` instructs the poller to not throw an error if the operation has been determined to be failed or canceled. Thrown errors originating from any other situation will still be surfaced, e.g. parse errors, abort errors, etc.

MaryGao: Thanks for clarification, it's clear now! Some cases still failed with wordings see commit: https://github.com/Azure/autorest.typescript/pull/1739/commits/d4795e753fd23819a59b9dca9f6478c61a21f0ef

MaryGao:
@deyaaeldeen @joheredi  @qiaozha The ci passed and pls help review the pr!

packages/autorest.typescript/test/rlcIntegration/lroRest.spec.ts
```diff
@@ -77,20 +80,22 @@ describe("LRO Rest Client", () => {
...
+      });
 
-        await poller.pollUntilDone();
-        assert.fail("should have thrown instead");
-      } catch (e) {
-        assert.equal(e.message, "Operation was canceled");
+      const response = await poller.pollUntilDone();
+      if (isUnexpected(response)) {
+        const error = `Unexpected status code ${response.status}`;
+        assert.fail(error);
```
deyaaeldeen, line 94: nitpick:
```suggestion
        assert.fail(`Unexpected status code ${response.status}`);
```

deyaaeldeen: We could also use `pathUnchecked` instead so we don't have to make this useless call to `isUnexpected`.

packages/autorest.typescript/test/rlcIntegration/generated/dpgCustomization/src/pollingHelper.ts
```diff
@@ -46,6 +46,8 @@ export async function getLongRunningPoller<TResult extends HttpResponse>(
     }
   };
 
+  // enable resolveOnUnsuccessful flag by default
```
deyaaeldeen, line 49: nitpick: is generating this comment necessary?

deyaaeldeen (approved):
LGTM, thanks a lot for taking on this!

packages/autorest.typescript/test/rlcIntegration/generated/dpgCustomization/src/pollingHelper.ts
MaryGao: will remove that

PR #1740, opened 2023-02-02
Cadl collection format support

### LINKED ISSUE #1687: RLC support collection format

qiaozha:
fixes #1687


MaryGao (approved):
LGTM

PR #1744, opened 2023-02-06
Bump the emitter version

MaryGao:
Bump the emitter version


lirenhe:
@marygao, could you check the build failure?

lirenhe (approved):

packages/cadl-rlc-test/test/contoso/cadl-output/2022-11-01-preview/openapi.json
```diff
@@ -458,10 +454,7 @@
           "description": "The faked shared model."
         }
       },
-      "description": "A widget.",
-      "required": [
-        "manufacturerId"
-      ]
```
qiaozha, line 464: why is this getting removed ?

MaryGao: I guess this is a fix in openapi, because this model is only used in patch method. there is no reason to add any required parameter

qiaozha:
Please add the change log for the PR merged today.

PR #1745, opened 2023-02-08
fix-path-separator-issue

### LINKED ISSUE #1720: [RLC] Top-level index file could not be generated in Windows
When we extract the top-level index file path we use `/src/` to match the path and get the proper index. However this may not work in Windows env because our path looks like `C:\\Users\\marygao\\...\\cadl-rlc-test\\test\\openai\\cadl-output\\src\\inference` then the file could not be generated.

https://github.com/Azure/autorest.typescript/blob/main/packages/rlc-common/src/buildTopLevelIndexFile.ts#L45

Another small bug is when we clear our src path the top-level index could not be cleared.

https://github.com/Azure/autorest.typescript/blob/main/packages/cadl-typescript/src/index.ts#L45

qiaozha:
fixes #1720

### DIFF
--- a/packages/rlc-common/src/buildTopLevelIndexFile.ts
+++ b/packages/rlc-common/src/buildTopLevelIndexFile.ts
@@ -44,7 +44,7 @@ export function buildTopLevelIndex(model: RLCModel) {
     });
     const content = indexFile.getFullText();
     const filePath = path.join(
-      srcPath.substring(0, srcPath.indexOf(path.sep + "src" + path.sep) + 5),
+      srcPath.substring(0, srcPath.indexOf(path.sep + "src") + 4),
       `index.ts`
     );
     return { path: filePath, content };
--- a/packages/rlc-common/src/buildTopLevelIndexFile.ts
+++ b/packages/rlc-common/src/buildTopLevelIndexFile.ts
@@ -44,7 +44,7 @@ export function buildTopLevelIndex(model: RLCModel) {
     });
     const content = indexFile.getFullText();
     const filePath = path.join(
-      srcPath.substring(0, srcPath.indexOf("/src/") + 5),
+      srcPath.substring(0, srcPath.indexOf(path.sep + "src" + path.sep) + 5),
       `index.ts`
     );
     return { path: filePath, content };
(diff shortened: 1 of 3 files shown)


packages/rlc-common/src/buildTopLevelIndexFile.ts
```diff
@@ -44,7 +44,7 @@ export function buildTopLevelIndex(model: RLCModel) {
     });
     const content = indexFile.getFullText();
     const filePath = path.join(
-      srcPath.substring(0, srcPath.indexOf("/src/") + 5),
+      srcPath.substring(0, srcPath.indexOf(path.sep + "src" + path.sep) + 5),
```
MaryGao, line 47: Could you also change the logic here?
https://github.com/Azure/autorest.typescript/blob/main/packages/rlc-common/src/helpers/pathUtils.ts#L3

packages/rlc-common/src/buildTopLevelIndexFile.ts
```diff
@@ -44,7 +44,7 @@ export function buildTopLevelIndex(model: RLCModel) {
     });
     const content = indexFile.getFullText();
     const filePath = path.join(
-      srcPath.substring(0, srcPath.indexOf("/src/") + 5),
+      srcPath.substring(0, srcPath.indexOf(path.sep + "src") + 4),
```
MaryGao, line 47: Why the index is changed from 5 to 4?

packages/rlc-common/src/helpers/pathUtils.ts
```diff
@@ -1,7 +1,9 @@
+import * as path from "path";
+
 // FIXME https://github.com/Azure/autorest.typescript/issues/1720
 export function getRelativePartFromSrcPath(srcPath: string) {
-  const relativePart = srcPath.substring(srcPath.indexOf("/src/") + 4);
-  return relativePart.startsWith("/")
+  const relativePart = srcPath.substring(srcPath.indexOf(path.sep + "src") + 4);
```
MaryGao, line 5: Do we need to change the index here?

packages/rlc-common/src/helpers/pathUtils.ts
```diff
@@ -1,7 +1,9 @@
+import * as path from "path";
+
 // FIXME https://github.com/Azure/autorest.typescript/issues/1720
```
MaryGao, line 3:
```suggestion
```

MaryGao (approved):
LGTM

Issue #1601, opened 2022-10-11
Query parameter default not parsed correctly


benjaminbolbrinker:
#### Context

I have the following endpoint in the `openapi.yaml` and try to call it using a client generated via autorest.

```
"/":
    get:
      tags:
      - Pet
      description: Returns a list of pets
      operationId: getPets
      x-eov-operation-handler: index
      parameters:
      - name: ids
        in: query
        description: List of pet ids
        style: form
        explode: false
        allowReserved: true
        schema:
          type: array
          items:
            type: string
         default: ['id1', 'id2'] # NOTE: This does not seem to work
```
#### Issue

When setting default parameters for Arrays, as described above, I run into the following:

```
      "type": "Error",
      "message": "options.ids must be of type Array.",
      "stack":
          Error: options.ids must be of type Array.
              at serializeSequenceType (/app/node_modules/@azure/core-client/src/serializer.ts:557:11)
              at SerializerImpl.serialize (/app/node_modules/@azure/core-client/src/serializer.ts:176:19)
              at calculateQueryParameters (/app/node_modules/@azure/core-client/src/urlHelpers.ts:167:56)
              at getRequestUrl (/app/node_modules/@azure/core-client/src/urlHelpers.ts:50:43)
              at petServiceSdk.sendOperationRequest (/app/node_modules/@azure/core-client/src/serviceClient.ts:148:17)
              at petServiceSdk.getPets (/app/node_modules/@olf-de/users-service-sdk/src/petServiceSdk.ts:276:17)
```

#### Expected behavior

After generating the typescript sdk I try to use it via:

```
const { petServiceSdk } = require('petSdk');         
                                                                                                        
const petSdk = new petServiceSdk({                                                                   
  endpoint: 'https://...', 
  allowInsecureConnection: true,                                                                    
});                                                                                                 
                                                                                                                                                                                                  
petSdk.getPets()
```

The call `petSdk.getPets()`  should perform the HTTP request and either:
- set the `ids` to the default of ['id1', 'id2'] or
- ignore the default and leave handling of default parameters to the server-side (**preferred**)

#### Proposed solution

The problem seems to be the conversion of the default parameter specified in the openapi. 
In the `sdk/src/models/parameters.ts` the following code is generated via autorest:

```
export const ids: OperationQueryParameter = {
  parameterPath: ["options", "ids"],
  mapper: {
     defaultValue: "id1,id2",  // NOTE: Unexpected comma separated string 
    serializedName: "ids",
[...]

PR #1757, opened 2023-02-15
Fix the default value for array

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1601

### DIFF
--- a/packages/autorest.typescript/src/utils/valueHelpers.ts
+++ b/packages/autorest.typescript/src/utils/valueHelpers.ts
@@ -28,7 +28,8 @@ export enum MapperTypes {
   Uuid = "Uuid",
   Number = "Number",
   Sequence = "Sequence",
-  any = "any"
+  any = "any",
+  Dictionary = "Dictionary"
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/autorest.typescript/src/utils/valueHelpers.ts
```diff
@@ -83,6 +84,21 @@ export function getStringForValue(
...
+  if (Array.isArray(value)) {
+    const valuesArr = [];
+    for (let item of value) {
+      valuesArr.push(getStringForValue(item, valueType));
+    }
+    return `[${valuesArr.join(", ")}]`;
+  }
+  const valueStr = getStringForValue(value, valueType);
+  return `[${valueStr}]`;
+}
```
qiaozha, line 100: Can you consider the way we build value for samples ?

MaryGao: Yes we handle this in a similar way in samples but the input is different because the sample we use Example as input.

qiaozha (approved):

PR #1765, opened 2023-02-17
Try work with named unions

### LINKED ISSUE #1753: [RLC] Support namedUnion
Cadl has the[ named union](https://microsoft.github.io/cadl/language-basics/unions) type like below：
```
model SimpleModel {
  prop?: Breed;
}

union Breed {
  beagle: Beagle,
  shepherd: GermanShepherd,
  retriever: GoldenRetriever,
}
```

We should support this in emitter. I think we have two options here.

## Option 1: Take it as general union

```
  prop?: Beagle | GermanShepherd | GoldenRetriever;
}
```

## Option 2: Take it as type alias (preferred)

```

export type Breed = Beagle | GermanShepherd | GoldenRetriever;

  prop?: Breed;
}
```

I prefer the option 2 considering we could have a name for this union for better expression.

/cc @qiaozha @joheredi

qiaozha:
fixes #1753


MaryGao (approved):
LGTM

PR #1766, opened 2023-02-17
fix-api-version-parameter-in-query

qiaozha:
fixes https://github.com/Azure/autorest.typescript/issues/1769
translator service has a query api version parameter and they don't set client level api version, which we should generate the api version.


packages/cadl-rlc-test/test/parametrizedHost/cadl-output/src/parameters.ts
```diff
@@ -3,4 +3,14 @@
 
 import { RequestParameters } from "@azure-rest/core-client";
 
-export type ListCollectionsParameters = RequestParameters;
+export interface ListCollectionsQueryParamProperties {
+  /** The API version to use for this operation. */
+  "api-version": string;
```
MaryGao, line 8: So because there is no default value so we generate the parameter for every operation, right?

qiaozha, line 8: The cadl has two parameter called apiVersion ?  
```
@useDependency(Azure.Core.Versions.v1_0_Preview_1)
@server(
  "{host}.{subdomain}.{sufix}.com/{apiVersion}",
  "Confidential Ledger Service",
  {
    @path
    host?: string = "one",
    @path
    subdomain?: string = "two",
    @path
    sufix?: string = "three",
    @path
    apiVersion?: string = "v1"
  }
)
namespace Azure.Test.ParametrizedHost;
```

qiaozha: In this case, they have a apiVersion in @server, will need to check how cadl deal with it.

packages/cadl-rlc-test/test/translator/cadl-output/src/parameters.ts
```diff
@@ -0,0 +1,325 @@
...
+export interface GetLanguagesQueryParamProperties {
+  /**
+   * A comma-separated list of names defining the group of languages to return.
+   * Allowed group names are: `translation`, `transliteration` and `dictionary`.
+   * If no scope is given, then all groups are returned, which is equivalent to passing
+   * `scope=translation,transliteration,dictionary`. To decide which set of supported languages
+   * is appropriate for your scenario, see the description of the [response object](#response-body).
+   */
+  scope?: string;
+  "api-version": string;
```
MaryGao, line 36: Just confirm that we generate api-version as operation-level parameter and will not prompt to client level one. Right?


@joheredi do you think we need to set it as client-level in this case?

packages/cadl-rlc-test/test/translator/cadl-output/src/parameters.ts
```diff
@@ -0,0 +1,325 @@
...
+   * Possible values are: true (default) or false.
+   *
+   * allowFallback=false specifies that the translation should only use systems trained for the category specified
+   * by the request. If a translation for language X to language Y requires chaining through a pivot language E,
+   * then all the systems in the chain (X → E and E → Y) will need to be custom and have the same category.
+   * If no system is found with the specific category, the request will return a 400 status code. allowFallback=true
+   * specifies that the service is allowed to fall back to a general system when a custom system doesn't exist.
+   */
+  allowFallback?: boolean;
+  "api-version": string;
```
MaryGao, line 130: Another comment is we didn't set the default value here, right?

MaryGao (approved):
I will approve first and then re-implement this part

Issue #1748, opened 2023-02-09
[RLC] Add support `null` type from Cadl


MaryGao:
## Background

In OpenAI service we have JSON that is a list of floats that may or may not contain a null item. Roughly speaking, something that looks like [null, 0.1, 0.2, ...] (see [issue](https://github.com/Azure/autorest.csharp/issues/3084)).  But currently our implementation is just ignore the `null` type, I don't think this is correct because we will lose ability to support `null` case.

See we have below Cadl definition we need to design the relevant output
```
model MyModel {
  item?:  float32 | null;   
}
```

Current generated ts:
```
model MyModel {
  item?:  float32;   
}
```

## From swagger

In RLC the null is not supported either, if it is generated from swagger.
https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-nullable

## My proposal

I think we should keep the `null` type as it is in generated code.

```
model MyModel {
  item?:  number | null;   
}
```

So that we could support below cases:

- null is a valid value in array
- partial update may use explicit null to indicate the value is not present https://kdrozd.pl/how-to-perform-a-partial-update-patch-with-explicit-null/

## Other consideration

Some guideline may not recommend to use null. Think it is a bad design - https://basarat.gitbook.io/typescript/recap/null-undefined#final-thoughts. The question is if we don't use null how to express these valid cases. 

/cc @qiaozha @joheredi

joheredi:
I agree we should generate nullable properties, null is meaningful and semantically different from `undefined`, especially in REST.

undefined === the absence of
null === an actual value.

One real life example of the difference in REST is on a PATCH merge operation where the initial entity is:

```js
   {
      name: "Bar",
      description: "This is a foo entity"
   }
```

Sending this
```js
   {
      name: "Foo",
      description: undefined
   }
```

Results in only name being updated but description unaltered.

```js
   {
      name: "Foo",
     description: "This is a foo entity"
   }
```

On the other hand, null tells the service to delete the description field

```js
   {
      name: "Foo",
      description: null
   }
```

resulting in 

```js
   {
      name: "Foo"
   }
```


I think we should generate the following 2 scenarios

```cadl
model MyModel {
  item?:  number | null;   
}
```
```ts
interface MyModel {
   item?: number | null;
}
```

And


```cadl
model MyModel {
  item:  number | null;   
}
```
```ts
interface MyModel {
   item: number | null;
}
```

@bterlson does this make sense?

PR #1767, opened 2023-02-21
Null support

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1748


packages/cadl-typescript/test/unit/modelsGenerator.spec.ts
```diff
@@ -83,9 +83,37 @@ describe("Input/output model type", () => {
...
+      const typeScriptType = "null";
+      await verifyPropertyType(cadlType, typeScriptType);
+    });
+
+    it("should generate nullable union", async () => {
       const cadlType = "string | null";
-      const typeScriptType = "string";
+      const typeScriptType = "string | null";
+      await verifyPropertyType(cadlType, typeScriptType);
+    });
```
qiaozha, line 96: Can you add a test of nullable model union  and named nullable model union ？

packages/cadl-typescript/test/unit/modelsGenerator.spec.ts
```diff
@@ -917,6 +995,55 @@ describe("Input/output model type", () => {
        export type MyNamedUnionOutput = Model1Output | Model2Output;`
       });
     });
+
+    it.only("should handle nullable named unions", async () => {
```
MaryGao, line 999:
```suggestion
    it.skip("should handle nullable named unions", async () => {
```

qiaozha (approved):

Issue #1737, opened 2023-02-01
Support @projectedName for CADL+DPG


lirenhe:
There is a feedback from archboard review meeting to rename a model in dotnet SDK for OpenAI.
This should be a general requirement to enable to support for @projectedName to handle this requirement.

Please check https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/projection/main.cadl for the test project.

CADL-dpg support is already done in https://github.com/Azure/cadl-azure/pull/2462.

MaryGao:
We may need to integrate @projectedName with JSON format.

PR #1777, opened 2023-03-10
Projected name fix

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1737

Over the team's [discussion](https://teams.microsoft.com/l/message/19:906c1efbbec54dc8949ac736633e6bdf@thread.skype/1678424797151?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1678417960707&teamName=Azure%20SDK&channelName=TypeSpec%20(Cadl)%20Discussion%20%F0%9F%90%AE&createdTime=1678424797151&allowXTenantAccess=false) we know

- For model name we would prefer the projectedName over friendlyName
- For property we would only accept the projectedName for json format


packages/cadl-typescript/src/modelUtils.ts
```diff
@@ -334,10 +335,13 @@ function getSchemaForModel(
   usage?: SchemaContext[],
   needRef?: boolean
 ) {
-  const friendlyName = getFriendlyName(program, model);
+  const overridedModelName =
+    getProjectedName(program, model, "javascript") ??
```
joheredi, line 339: Why do we need  `getProjectedName(program, model, "javascript")`? We shouldn't allow renames on the REST API names.

packages/cadl-typescript/src/modelUtils.ts
```diff
@@ -334,10 +335,13 @@ function getSchemaForModel(
...
+    getProjectedName(program, model, "client") ??
+    getFriendlyName(program, model);
```
joheredi, line 341: I forgot, what is a friendlyName?

packages/cadl-typescript/src/modelUtils.ts
joheredi: I think it would be safer if we removed both `getFriendlyName(program, model);` and `getFriendlyName(program, model);`

MaryGao: Sure we could remove the logic for getFriendlyName. 

I added this because the model name will not be used for any serialization and deserialization. So I think it would be better to name the model with language specific name or friendly name than json one.

packages/cadl-typescript/src/modelUtils.ts
MaryGao: https://microsoft.github.io/typespec/standard-library/built-in-decorators#friendlyname

image

joheredi (approved):

PR #1781, opened 2023-03-14
Prepare packages to generate ESModule imports

joheredi:
This is the first PR for Modular client generation. Preparing the packages to be able to generate ESModule imports when needed


packages/cadl-typescript/src/modelUtils.ts
```diff
@@ -334,10 +334,10 @@ function getSchemaForModel(
   usage?: SchemaContext[],
   needRef?: boolean
 ) {
-  const friendlyName = getFriendlyName(program, model);
+  const restApiName = getProjectedName(program, model, "json");
```
MaryGao, line 337: Please remember to revert this change

PR #1790, opened 2023-03-23
[Modular] Generate Classical client and root index

joheredi:
This PR introduces the Classical client generation which is based off the api/ layer.


packages/typespec-test/eng/smoke-test.js
```diff
@@ -37,7 +38,7 @@ async function main() {
   for (const folder of folders) {
     const path = join(root, "test", folder);
     console.log(`================Start ${folder}===============`);
-    generate(path);
+    // generate(path);
```
MaryGao, line 41: Why do we comment out this line?

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/index.ts
```diff
@@ -1,4 +1,380 @@
...
+  CertificatesGetCertificateOptions,
+} from "./api/Certificates.js";
+export {
+  getComputeNodeExtensions,
+  listComputeNodeExtensions,
+  ComputeNodeExtensionsGetComputeNodeExtensionsOptions,
+  ComputeNodeExtensionsListComputeNodeExtensionsOptions,
+} from "./api/ComputeNodeExtensions.js";
+export {
+  addUser,
```
MaryGao, line 234: How do you think if we have duplicated method names under diff folder?

packages/typespec-ts/src/modular/helpers/typeHelpers.ts
```diff
@@ -63,3 +63,19 @@ export function getType(type: Type): TypeMetadata {
       throw new Error(`Unsupported type ${type.type}`);
   }
 }
+
+export function buildParameterType(
```
MaryGao, line 67: Could you also add comment for this? Which parameter is it to build?

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/BatchService.ts
```diff
@@ -0,0 +1,815 @@
...
+      return addPool(this._client, options);
+    },
+    listPools: (
+      options: PoolListPoolsOptions = { requestOptions: {} }
+    ): Promise<BatchPoolListResult> => {
+      return listPools(this._client, options);
+    },
+    deletePool: (
+      poolId: string,
+      options: PoolDeletePoolOptions = { requestOptions: {} }
```
MaryGao, line 267: Why not below statement?

```suggestion
      options?: PoolDeletePoolOptions
```

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/BatchService.ts
```diff
@@ -0,0 +1,815 @@
...
+        options
+      );
+    },
+    listComputeNodeExtensions: (
+      poolId: string,
+      nodeId: string,
+      options: ComputeNodeExtensionsListComputeNodeExtensionsOptions = {
+        requestOptions: {},
+      }
+    ): Promise<NodeVMExtensionList> => {
```
MaryGao, line 811: I may not get the point here, why the classic method would return the REST model?

According to my understanding the model `NodeVMExtensionList` would only accept the json property names not any renaming.

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/BatchService.ts
```diff
@@ -0,0 +1,815 @@
...
+import {
+  PoolStatistics,
+  BatchPool,
+  AutoScaleRun,
+  JobStatistics,
+  BatchJob,
+  Certificate,
+  BatchJobSchedule,
+  BatchTask,
+} from "./rest/models.js";
```
MaryGao, line 217: Not sure why we have dependency for REST models?

I assume all models are imported from api folder.

packages/typespec-test/eng/smoke-test.js
joheredi: Oops my bad, thanks for catching it

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/index.ts
joheredi: This is a good point, let's address it later on.
Maybe we can start by failing and show a meaningful error to the users asking to get some projectedNames in client.tsp

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/BatchService.ts
joheredi: `{ requestOptions: {} }`  is creating a default options object with an empty request options object as a porperty. This saves us to have to nullcheck and create an object when we need to add properties to request options. Otherwise in the body we'll ahve to do:

```ts

   if(!options) {
      options = {requestOptions: {}};
   }

   if (!options.requestOptions) {
      options.requestOptions = {};
   }

   options.requestOptions.foo = "foo";

```

vs


```ts
   options.requestOptions.foo = "foo";
```

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/BatchService.ts
joheredi: Good catch fixing this. It looks like the auto fix missing imports from ts-morph is getting the wrong module

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/BatchService.ts
```diff
@@ -0,0 +1,815 @@
...
+
+export class BatchService {
```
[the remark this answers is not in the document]
joheredi: There shouldn't be confict in the filename. We need to be careful about the exports though

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/index.ts
MaryGao: Make sure we improve to faile the code gen not generated but compiler error.

PR #1797, opened 2023-03-31
Modular eventgrid

joheredi:


packages/rlc-common/src/helpers/nameConstructors.ts
```diff
@@ -108,3 +108,16 @@ export function getImportModuleName(name: ModuleName, codeModel: RLCModel) {
...
 }
+
+export function getClientName(model: RLCModel) {
+  const clientName = model.libraryName;
+  let clientInterfaceName = clientName.endsWith("Client")
+    ? `${clientName}`
+    : `${clientName}Client`;
+
+  if (model.options?.isModularLibrary) {
+    clientInterfaceName = `${clientName.replace("Client", "")}Context`;
```
deyaaeldeen, line 119: nitpick: is this equivalent?
```suggestion
    clientInterfaceName = clientName.replace("Client", "Context");
```

packages/typespec-test/test/openai_modular/generated/typespec-ts/package-files/package.json
```diff
@@ -0,0 +1,133 @@
+{
```
qiaozha, line 1: just out of curious why we have package.json both in the typespec-ts folder and in the  typespec-ts/package-file folder

joheredi: This is temporary, we need to generate a propper package.json file that follows the modular configuration and that can take the azure-sdk-for-js flag to generate slightly different one. Right now the one in the folder generates what we need for azure-sdk-for-js: true. For now the root package.json file is the same as RLC, once we have this correctly we'll move the real one to the root and stop generating the RLC

packages/typespec-ts/src/modular/static/scripts/renameOutput.mjs
```diff
@@ -0,0 +1,17 @@
...
+
+const __dirname = url.fileURLToPath(new URL('.', import.meta.url));
+const destinationPath = path.join(__dirname, "../dist/");
+
+const inputFile = path.join(destinationPath, "index.js");
+const inputMapFile = path.join(destinationPath, "index.js.map");
+
+if (fs.existsSync(inputFile) && fs.existsSync(inputMapFile)) {
+  fs.renameSync(inputFile, path.join(destinationPath, "index.cjs"));
+  fs.renameSync(inputMapFile, path.join(destinationPath, "index.cjs.map"));
```
qiaozha, line 16: why we need to rename the output ?

qiaozha (commented):
what does this file packages/typespec-ts/src/modular/buildStaticFiles.ts do ?

packages/typespec-test/test/openai_modular/generated/typespec-ts/package-files/package.json
qiaozha: Got it, thanks.

packages/typespec-ts/src/modular/static/scripts/renameOutput.mjs
joheredi: This is required to make it a hybrid package and we can export both versions cjs and ESModules

joheredi:
> what does this file packages/typespec-ts/src/modular/buildStaticFiles.ts do ?

good catch, it doesn't do anything. I have removed it

packages/typespec-ts/src/modular/static/scripts/renameOutput.mjs
joheredi: Otherwise, users would be forced to migrate their projects to ESModules

qiaozha: get it

qiaozha (approved):
LGTM

Issue #1771, opened 2023-02-22
[RLC] should not generate body: Record<string, unknown> for 204 response.


qiaozha:
In purview workflow swagger, https://github.com/Azure/azure-rest-api-specs/blob/a207ec1754c20e5d601c08274bc33d40167968c2/specification/purview/data-plane/Azure.Analytics.Purview.Workflow/preview/2022-05-01-preview/purviewWorkflow.json#L181 we generated body: Record<string, unknown> for it, but we should not generate body at all.

PR #1801, opened 2023-04-03
remove-undefined-body-in-response

qiaozha:
fixes https://github.com/Azure/autorest.typescript/issues/1771

### DIFF
--- a/packages/autorest.typescript/test/smoke/generated/agrifood-data-plane/src/responses.ts
+++ b/packages/autorest.typescript/test/smoke/generated/agrifood-data-plane/src/responses.ts
@@ -103,7 +103,6 @@ export interface ApplicationDataCreateOrUpdateDefaultResponse
 /** Deletes a specified application data resource under a particular farmer. */
 export interface ApplicationDataDelete204Response extends HttpResponse {
   status: "204";
-  body: Record<string, unknown>;
 }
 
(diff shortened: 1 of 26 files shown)


packages/autorest.typescript/test/smoke/generated/agrifood-data-plane/src/responses.ts
```diff
@@ -169,8 +167,6 @@ export interface AttachmentsDeleteDefaultResponse extends HttpResponse {
 /** Downloads and returns attachment as response for the given input filePath. */
 export interface AttachmentsDownload200Response extends HttpResponse {
   status: "200";
-  /** Value may contain any sequence of octets */
-  body: Uint8Array;
```
MaryGao, line 173: Why we remove this? Not sure it's possible to add UT.

MaryGao (approved):
LGTM, btw could you check if this exists in TypeSpec ?

packages/autorest.typescript/test/smoke/generated/agrifood-data-plane/src/responses.ts
qiaozha: it's deleted by mistake, added back.

qiaozha:
> LGTM, btw could you check if this exists in TypeSpec ?

Checked, should not exist on the typespec side.

PR #1809, opened 2023-04-11
add-internal-cadl-ranch-test-case

### LINKED ISSUE #1795: [Feature]: Support the decorator of @internal
Please check the definition of this decorator:
https://azure.github.io/typespec-azure/docs/libraries/azure-dpg/reference/decorators#@Azure.ClientGenerator.Core.internal

The test was already added into Cadl-ranch:
https://github.com/Azure/cadl-ranch/pull/237

Please note: this decorator may cause breaking change for the SDK.

### LINKED ISSUE #1800: [DPG] Full CADL-Ranch coverage
Title is self explanatory :)

qiaozha:
fixes #1800 
fixes #1795

### DIFF
--- a/packages/typespec-ts/test/integration/generated/internal/src/clientDefinitions.ts
+++ b/packages/typespec-ts/test/integration/generated/internal/src/clientDefinitions.ts
@@ -1,0 +1,27 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { GetInternalParameters, PostInternalParameters } from "./parameters";
+import { GetInternal200Response, PostInternal200Response } from "./responses";
+import { Client, StreamableMethod } from "@azure-rest/core-client";
+
+export interface GetInternal {
+  get(options: GetInternalParameters): StreamableMethod<GetInternal200Response>;
+}
+
+export interface PostInternal {
+  post(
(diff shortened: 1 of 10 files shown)


packages/typespec-ts/test/integration/generated/internal/src/clientDefinitions.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export interface PostInternal {
+  post(
+    options: PostInternalParameters
+  ): StreamableMethod<PostInternal200Response>;
+}
+
+export interface Routes {
+  /** Resource for '/internal/getInternal' has methods for the following verbs: get */
+  (path: "/internal/getInternal"): GetInternal;
```
MaryGao, line 20: I would suggest to differentiate the interal code and publish ones by document or naming convention.

Or we could enable a flag to not generate internal codes.

joheredi: For RLC we shouldn't do anything for internal. If it is in the Service Typespec it should be generated. AFAIK this is a feature that would be used in client.tsp to shape clients rather than for describing services. So for RLC we should just ignore this decorator.

MaryGao (approved):
LGTM

PR #1810, opened 2023-04-11
Api version test cases

MaryGao:


packages/rlc-common/src/helpers/apiVersionUtil.ts
```diff
@@ -0,0 +1,54 @@
...
+  const param = urlInfo.urlParameters?.filter(
+    (p) =>
+      p.name.toLowerCase() === "api-version" ||
+      p.name.toLowerCase() === "apiversion"
+  );
+  if (!param || param?.length < 1) {
+    return;
+  }
+  const detail: ApiVersionInfo = {
+    definedPosition: "query",
```
qiaozha, line 23: should the definedPosition for this scenario be `path` ?

packages/typespec-ts/src/modelUtils.ts
```diff
@@ -54,6 +60,11 @@ import {
 } from "@typespec/http";
 import { getPagedResult, isFixed } from "@azure-tools/typespec-azure-core";
 import { extractPagedMetadataNested } from "./operationUtil.js";
+import {
+  DpgContext,
+  getDefaultApiVersion,
```
qiaozha, line 65: I remember we need to do something in the tcgc about the getDefaultApiVersion logic ? Has that been done ?

packages/rlc-common/src/helpers/apiVersionUtil.ts
MaryGao: Good catch!

qiaozha (approved):

Issue #1331, opened 2022-03-01
[RLC] Support loggingOption in rest level client contructor


qiaozha:
In high level client, we can add logginOptions like 
```
    const internalPipelineOptions: InternalPipelineOptions = {
      ...options,
      loggingOptions: {
        logger: logger.info,
        additionalAllowedHeaderNames: MetricsAdvisorLoggingAllowedHeaderNames,
        additionalAllowedQueryParameters: MetricsAdvisorLoggingAllowedQueryParameters,
      },
    };
    this.client = new GeneratedClient(this.endpointUrl, internalPipelineOptions);
```

I wonder whether we need to add the support in Rest Level Client

PR #1827, opened 2023-04-24
add logger support for rlc

qiaozha:
fixes https://github.com/Azure/autorest.typescript/issues/1331


packages/autorest.typescript/test/rlcIntegration/generated/bodyFormDataRest/src/bodyFormData.ts
```diff
@@ -21,6 +22,9 @@ export default function createClient(
     ...options,
     userAgentOptions: {
       userAgentPrefix
+    },
+    loggingOptions: {
+      logger: logger.info
```
MaryGao, line 27: My understanding is the customer could feed their own logger method, if that's undefined we would use default one. It would be like:

```typescript
logger: options.logger || logger.info
```

joheredi (approved):

MaryGao (approved):

PR #1830, opened 2023-04-26
Multiclient in modular support

qiaozha:
This PR is mainly focus on how to implement multiple subclients in the Modular library and see the design https://github.com/Azure/azure-sdk-for-js/pull/26164/files  

And the following items are non-goals of this PR:
1. multiple endpoints scenario in RLC layer. Issue tracked #1894 
2. identify the shared models and only generate it once. See more details in this issue https://github.com/Azure/typespec-azure/issues/3035
3. Split the RLC layer from Modular. 

In this `packages/typespec-test/test/loadtesting_modular` smoke test case, we have two subclients `LoadTestRunClient` and `LoadTestAdministrationClient` in the Modular layer, and they both point to the same @service, which means the RLC layer is single endpoint. 
 
our current code structure is 
image

On the subpath exports, we export the following subpaths 
```
   "exports": {
    ".": {
      "types": "./types/src/index.d.ts",
      "require": "./dist/index.cjs",
      "import": "./dist-esm/src/index.js"
    },
    "./loadTestAdministration": {
      "types": "./types/src/loadTestAdministration/index.d.ts",
      "import": "./dist-esm/src/loadTestAdministration/index.js"
    },
    "./loadTestAdministration/api": {
      "types": "./types/src/loadTestAdministration/api/index.d.ts",
      "import": "./dist-esm/src/loadTestAdministration/api/index.js"
    },
    "./loadTestAdministration/models": {
      "types": "./types/src/loadTestAdministration/models/index.d.ts",
      "import": "./dist-esm/src/loadTestAdministration/models/index.js"
    },
    "./loadTestRun": {
      "types": "./types/src/loadTestRun/index.d.ts",
      "import": "./dist-esm/src/loadTestRun/index.js"
    },
    "./loadTestRun/api": {
      "types": "./types/src/loadTestRun/api/index.d.ts",
      "import": "./dist-esm/src/loadTestRun/api/index.js"
    },
    "./loadTestRun/models": {
      "types": "./types/src/loadTestRun/models/index.d.ts",
      "import": "./dist-esm/src/loadTestRun/models/index.js"
    }
  },
```
Based on our latest discussion, "We only need one way to import these things without there being ambiguity about which way is correct". I have removed the top level `api/index.ts` and `models/index.ts` in the multi-client modular case.

Remove `./rest` subpath export as discussed in the Q&A session.


packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/api/bar/BarContext.ts
```diff
@@ -0,0 +1,19 @@
...
+export { Client } from "../../rest/bar/index.js";
+
+/** Bar */
+export function createBar(
+  endpoint: string,
+  apiVersion: string,
+  options: ClientOptions = {}
+): Client.BarContext {
+  const baseUrl = endpoint;
+  const clientContext = getClient(baseUrl, options);
```
qiaozha, line 17: In this case, the api version has a default value 1.0.0 and we put it into options in RLC client factory, should we also put it in the options as give it a default value if it's not set ? 
Another question is what if there're other client level parameters which doesn't exist in the core-client-rest ServiceOptions, does this mean we need common folder even after we migrate some of the types into core-client-rest ?

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/index.ts
```diff
@@ -0,0 +1,20 @@
...
+export { Resource, CustomPage } from "./api/foo/models.js";
+export {
+  CreateOrUpdateOptions,
+  GetOptions,
+  DeleteOptions,
+  ListOptions,
+} from "./api/foo/operations.js";
+export { FooClient } from "./FooClient.js";
+export { ClientOptions, RequestOptions } from "./common/interfaces.js";
+export { Resource, CustomPage } from "./api/bar/models.js";
```
qiaozha, line 13: In this case, because both subclients has used shared models in the typespec, and we export both models.ts,  whch is duplicated.  I wonder what is the right thing to do here ?  should we generate a common models.ts or should we try to rename one of them ?

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/rest/bar/index.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import * as Parameters from "./parameters.js";
+import * as Responses from "./responses.js";
+import * as Client from "./clientDefinitions.js";
+import * as OutputModels from "./outputModels.js";
+
+export { createClient } from "./barClient.js";
+export { Parameters, Responses, Client, OutputModels };
```
qiaozha, line 10: In this case, we don't have isUnexpected generated, but in the bar subclient, and I wonder if we should use code like `if(isUnexpected)` in the modular level when calling the rest level ?

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/rest/foo/index.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import * as Parameters from "./parameters.js";
+import * as Responses from "./responses.js";
+import * as Client from "./clientDefinitions.js";
+import * as Models from "./models.js";
+import * as OutputModels from "./outputModels.js";
+import * as PaginateHelper from "./paginateHelper.js";
+
+export { createClient } from "./fooClient.js";
+export { Parameters, Responses, Client, Models, OutputModels, PaginateHelper };
```
qiaozha, line 12: Another question about isUnexpected is that, in the foo subclient, even if we generated isUnexpected.ts file, we don't export it in the index file, and however it's being used in the api level, should we add this into the index file ? What if both bar and foo subclients have this isUnexpected ?

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/api/foo/operations.ts
```diff
@@ -0,0 +1,126 @@
...
+
+/** Creates a new resource or updates an existing one. */
+export async function createOrUpdate(
+  context: Client.FooContext,
+  name: string,
+  type: string,
+  name: string,
+  options: CreateOrUpdateOptions = { requestOptions: {} }
+): Promise<Resource> {
+  const result = await context.path("/cadl-foo/resources/{name}", name).put({
```
qiaozha, line 25: In this case, the body parameter is a Resource type which has two required properties name and type, however name is already a path level parameter, I wonder if purely exlude the path parameter from the required body flatten parameter is enough ？ 
Here's the typespec definition   
```
  @doc("Creates a new resource or updates an existing one.")
  createOrUpdate is ResourceCreateOrReplace<Resource>;
```

packages/rlc-common/src/helpers/nameUtils.ts
```diff
@@ -19,70 +19,70 @@ export enum NameType {
 const Newable = [NameType.Class, NameType.Interface, NameType.OperationGroup];
 
 const ReservedModelNames: ReservedName[] = [
-  { name: "any", reservedFor: [NameType.Parameter] },
```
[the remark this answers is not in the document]
joheredi: I hit a similar issue today, we should guard Operation as well

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/index.ts
joheredi: If we can identify common models it would be great just to generate and export them once

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/rest/bar/index.ts
joheredi: Yeah, though we should see why isUnexpected was not generated?

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/rest/foo/index.ts
joheredi: I don't think so, users that import the api/ operations don't need to use isUnexpected right?

qiaozha: the api folder will use it right ? does that count as internally ?

qiaozha: I think in the single client case, we do export the isUnexpectedHelper to our customer.

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/api/foo/index.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export { createFoo, Client, FooClientOptions } from "./FooContext.js";
+export { Resource } from "./models.js";
```
qiaozha, line 5: need to figure out why custom page is missing.

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/api/bar/BarContext.ts
qiaozha: As rest layer will handle this kind of parameter default value, I think it would be better if we keep this clean and filter out any parameter that has a default value in the operation signature.

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/api/foo/operations.ts
qiaozha: turns out I should set the visibility of name as read because of key.

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/api/foo/index.ts
qiaozha: temporarily disable the removeUnusedInterfaces as in this case, CustomPage is not being used in bar sub client, however it has been used in foo sub client. and removeUnusedInterfaces would just remove all of them, and I am not very sure how to fix that.

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/review/eventgrid.api.md
```diff
@@ -13,6 +13,11 @@ export interface AcknowledgeCloudEventsOptions extends RequestOptions {
     contentType?: string;
 }
 
+// @public
+export interface AcknowledgeOptions {
+    lockTokens: string[];
+}
```
qiaozha, line 19: I guess this is because I removed the removeUnusedInterfaces for now

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -40,17 +40,15 @@ export function _publishCloudEventSend(
...
 > {
-  return context
-    .path("/topics/{topicName}:publish", topicName)
-    .post({
-      allowInsecureConnection: options.requestOptions?.allowInsecureConnection,
-      skipUrlEncoding: options.requestOptions?.skipUrlEncoding,
-      contentType:
-        (options.contentType as any) ??
-        "application/cloudevents+json; charset=utf-8",
-      headers: { ...options.requestOptions?.headers },
```
qiaozha, line 51: probably need to figure out why headers is removed here

packages/typespec-ts/src/modular/helpers/namingHelpers.ts
```diff
@@ -55,71 +56,98 @@ enum NameType {
...
-  { name: "catch", reservedFor: [NameType.Parameter] },
-  { name: "class", reservedFor: [NameType.Parameter] },
-  { name: "const", reservedFor: [NameType.Parameter] },
-  { name: "constructor", reservedFor: [NameType.Parameter] },
-  { name: "continue", reservedFor: [NameType.Parameter] },
-  { name: "date", reservedFor: [NameType.Parameter, ...Newable] },
-  { name: "debugger", reservedFor: [NameType.Parameter] },
-  { name: "declare", reservedFor: [NameType.Parameter] },
-  { name: "default", reservedFor: [NameType.Parameter] },
+  { name: "any", reservedFor: [NameType.Parameter, NameType.Operation] },
```
qiaozha, line 59: I am thinking if we should use the one from rlc-common for it

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -235,27 +273,29 @@ function getRequestParameters(operation: Operation): string {
     }
   }
 
-  let paramStr = "";
```
qiaozha, line 238: Maybe this is why headers are missing when sending the request.

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/api/bar/operations.ts
```diff
@@ -0,0 +1,139 @@
...
+  result: GetBinary200Response
+): Promise<any> {
+  return result.body;
+}
+
+/** */
+export async function getBinary(
+  context: Client.BarContext,
+  options: GetBinaryOptions = { requestOptions: {} }
+): Promise<any> {
```
MaryGao, line 39: Why the return type is `any`?

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/api/bar/BarContext.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { ClientOptions } from "@azure-rest/core-client";
+import { Client } from "../../rest/bar/index.js";
+import { createClient as getClient } from "../../rest/bar/index.js";
+
+export interface BarClientOptions extends ClientOptions {}
+
+export { Client } from "../../rest/bar/index.js";
```
MaryGao, line 10: Duplicated with Line5

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/package-files/package.json
```diff
@@ -0,0 +1,133 @@
...
+      "types": "./types/src/index.d.ts",
+      "require": "./dist/index.cjs",
+      "import": "./dist-esm/src/index.js"
+    },
+    "./api": {
+      "types": "./types/src/api/index.d.ts",
+      "import": "./dist-esm/src/api/index.js"
+    },
+    "./rest": {
+      "types": "./types/src/rest/index.d.ts",
```
MaryGao, line 21: I was wondering if we could export the sub-client directly, so we could avoid to rename anything.

```
 "./rest/bar": {
      "types": "./types/src/rest/bar/index.d.ts",
       "import": "./dist-esm/src/rest/bar/index.js"
 },
 "./rest/foo": {
      "types": "./types/src/rest/foo/index.d.ts",
       "import": "./dist-esm/src/rest/foo/index.js"
 }
```

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/api/bar/BarContext.ts
qiaozha: Line 5 is import, this is export

packages/rlc-common/src/helpers/nameUtils.ts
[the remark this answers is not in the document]
qiaozha: I moved the modular to use the rlc-common ReservedModelNames. And only guard the operation name for delete for now.

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/index.ts
```diff
@@ -1,7 +1,11 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+export { EventGridClient } from "./EventGridClient.js";
 export {
+  createEventGrid,
+  EventGridClientOptions,
+  EventGridContext,
```
qiaozha, line 8: will need to check on this whether this should be exported.

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/index.ts
```diff
@@ -10,14 +14,17 @@ export {
...
   ReleaseResult,
   RejectResult,
-} from "./api/models.js";
-export {
+  publishCloudEvent,
+  publishCloudEvents,
+  receiveCloudEvents,
+  acknowledgeCloudEvents,
+  releaseCloudEvents,
+  rejectCloudEvents,
```
qiaozha, line 22: I don't think we should export the operations in the api layer to the top classical client layer, right? @bterlson  @mpodwysocki

qiaozha:
I will split some changes into a spread PR, and make this clearer.

packages/typespec-test/test/multiclient_modular/spec/main.tsp
```diff
@@ -0,0 +1,87 @@
...
+using Azure.Core;
+using Azure.ClientGenerator;
+using TypeSpec.Versioning;
+
+@service({
+  title: "MultiClient",
+  version: "1.0.0",
+})
+@useDependency(Azure.Core.Versions.v1_0_Preview_1)
+namespace Cadl.MultiClient;
```
lirenhe, line 16: Let's use TypeSpec as namespace.

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/src/api/bar/operations.ts
qiaozha: Because the TypeSpec is defined like this 
```
  @route("/get-binary")
  @get
  getBinary(): OkResponse & {
    @header("content-type") contentType: "image/png";
    @body body: bytes;
  };
```
And previously, we had some issues in handling the Anonymous model,  Jose has this temporary workaround for this https://github.com/Azure/autorest.typescript/blob/main/packages/typespec-ts/src/modelUtils.ts#L159-L165

packages/typespec-test/test/multiclient_modular/generated/typespec-ts/package-files/package.json
qiaozha: Not a fan of this proposal, even if the package is called multi client, that multiple sub clients are act as a whole package with full functionalities. We use /api or /rest to provide different customer with different user experiences at their needs, I don't think we want to differentiate our customer because of which sub client they are using. @bterlson @mpodwysocki do you have any suggestions on this ?

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/index.ts
```diff
@@ -10,14 +12,11 @@ export {
...
-export {
   PublishCloudEventOptions,
   PublishCloudEventsOptions,
   ReceiveCloudEventsOptions,
   AcknowledgeCloudEventsOptions,
   ReleaseCloudEventsOptions,
   RejectCloudEventsOptions,
-} from "./api/operations.js";
-export { EventGridClient, EventGridClientOptions } from "./EventGridClient.js";
+} from "./api/index.js";
```
qiaozha, line 21: I merge the previous exports from different files in the api folder into exports from `api/index.js` directly. Let me know if you are okay with this ?

xirzec (commented):
The code changes themselves look good for the situation of exporting multiple clients at the top level (based on type), but I'm a bit confused by how this differs from the linked proposal: https://gist.github.com/joheredi/93d3dda69d0b71b803cb5e7cb9066798

1. In the proposal, it appears that one client is chosen as the "default" rather than having multiple clients be at the top. Is this an intentional change? 
2. In the proposal, there are additional exports for each subclient, such as "./foo", "./foo/api", and "./foo/rest" -- are we planning to allow those as well for consumers who want a specific subclient?


The main concern I have with the first point is that in the case that multiclient is being used not because there are separately versioned endpoints, but because there are *multiple* versions of a single endpoint, combining them would be a bad idea.

For example, say that one package has a `FooClient` and a `FooExperimentalClient` that are essentially the same client modulo a different endpoint query parameter, but `FooExperimentalClient` has additional preview API operations.

In such a case we'd only want `FooClient` to be exposed as the default, but we would want consumers to be able to optionally import the experimental client such as 

```ts
import FooExperimentalClient from "@azure/foo/experimental";
```

qiaozha:
@xirzec Thanks for the comment.
For the default Client, I had a thinking about it, but then I realize that we actually don't have a default client in the past RLC multi-client implementation, I somehow remember there's some issue when I tried to build or run the code. I tried to recall more details, but the only detail I can found is that Jose specifically asked me to change the named exports for single client back https://github.com/Azure/autorest.typescript/pull/1267#pullrequestreview-843028560. I will try to get a default export in both RLC and Modular see if it works. If yes, I will change on the PR.

For the additional exports for each sub client, I remember I showed to Jose And he suggest me to change, but I don't recall if I have changed it or not, after tracing back to the first inital draft, I found that I haven't, sorry about that. 

For when should we use multi-client, your example makes me feel like it's more like a multi-api case. https://github.com/Azure/azure-sdk-for-js/blob/c369610dd5161a9a22827a39fb1eb30f4eebe121/design/multi-version-design.md, I also feel like that multi-api is kind of against the principal one package, one service version that we have been talking about these days. In fact, we haven't get a real case for multi-api yet, and at least we have a case for multi-client, which may because we used to tell service team that we don't support multi-api. And in RLC we are using some approach like what Jose suggested in this case https://teams.microsoft.com/l/message/19:344f6b5b36ba414daa15473942c7477b@thread.skype/1685512112110?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1684229932399&teamName=Azure%20SDK&channelName=Language%20%E2%80%93%20JS%E2%80%89%EF%BC%86%E2%80%89TS%20%F0%9F%A5%B7&createdTime=1685512112110 I am not even sure how to make the modular layer work if the RLC layer is using separate packages.

packages/typespec-ts/src/modular/buildRootIndex.ts
```diff
@@ -16,72 +18,38 @@ export function buildRootIndex(
...
 function exportClassicalClient(client: Client, indexFile: SourceFile) {
   const clientName = `${getClientName(client)}Client`;
   indexFile.addExportDeclaration({
-    namedExports: [clientName, `${clientName}Options`],
+    namedExports: [clientName],
     moduleSpecifier: `./${clientName}.js`
   });
 }
 
-function exportOptionsInterfaces(
```
qiaozha, line 36: Kind of forget why we remove this.

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/LoadTestAdministrationClient.ts
```diff
@@ -0,0 +1,163 @@
...
+  GetTestFileOptions,
+  ListTestFilesOptions,
+  ListTestsOptions,
+  UploadTestFileOptions,
+  DeleteTestFileOptions,
+  DeleteTestOptions,
+} from "./api/loadTestAdministration/index.js";
+
+export class LoadTestAdministrationClient {
+  private _client: LoadTestAdministrationContext;
```
qiaozha, line 42: See no LoadTestAdministrationContext get imported now, compile will error.

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/api/loadTestAdministration/operations.ts
```diff
@@ -0,0 +1,1104 @@
...
+}
+
+export function _createOrUpdateTestSend(
+  context: Client,
+  testId: string,
+  options: CreateOrUpdateTestOptions = { requestOptions: {} }
+): StreamableMethod<
+  | CreateOrUpdateTest200Response
+  | CreateOrUpdateTest201Response
+  | CreateOrUpdateTestDefaultResponse
```
qiaozha, line 80: This is the rest layer response type name.

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/package-files/package.json
```diff
@@ -0,0 +1,133 @@
...
+      "import": "./dist-esm/src/index.js"
+    },
+    "./api": {
+      "types": "./types/src/api/index.d.ts",
+      "import": "./dist-esm/src/api/index.js"
+    },
+    "./rest": {
+      "types": "./types/src/rest/index.d.ts",
+      "import": "./dist-esm/src/rest/index.js"
+    }
```
qiaozha, line 23: The subpath is incorrect. for multiclient case like loadtesting.

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/api/index.ts
```diff
@@ -0,0 +1,149 @@
...
+  TimeSeriesElement as LoadTestRunClientTimeSeriesElement,
+  MetricValue as LoadTestRunClientMetricValue,
+  DimensionValue as LoadTestRunClientDimensionValue,
+  Interval as LoadTestRunClientInterval,
+  DimensionValueList as LoadTestRunClientDimensionValueList,
+  testRun,
+  createOrUpdateAppComponents as LoadTestRunClientcreateOrUpdateAppComponents,
+  createOrUpdateServerMetricsConfig as LoadTestRunClientcreateOrUpdateServerMetricsConfig,
+  deleteTestRun,
+  getAppComponents as LoadTestRunClientgetAppComponents,
```
qiaozha, line 125: There's operation name conflicts in the api layer. In RLC layer we use operation group as the prefix, wonder what's the solution in API layer. 
Probably need to check out the operation name in client.tsp

qiaozha (commented):
Still have a lot of work to do

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/index.ts
xirzec: I assume we'll need to update again once #1915 is merged?

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/package-files/package.json
```diff
@@ -0,0 +1,133 @@
+{
+  "name": "@azure/load-testing",
+  "version": "1.0.0-beta.1",
+  "description": "\"This package contains Microsoft Azure LoadTestingClient client library.\"",
```
xirzec, line 4: extra double quotes?

qiaozha:
> I assume we'll need to update again once https://github.com/Azure/autorest.typescript/pull/1915 is merged?

@xirzec exactly, that's why I have been trying to get PR #1915 merged, kind of resolving conflicts all the time.

packages/typespec-ts/src/transform/transformResponses.ts
```diff
@@ -75,6 +75,7 @@ export function transformToResponseTypes(
     const rlcOperationUnit: OperationResponse = {
       operationGroup: getOperationGroupName(dpgContext, route),
       operationName: getOperationName(program, route.operation),
+      path: route.path,
```
qiaozha, line 78: we need to use path information to identify which response types are for  this operation. pure operationGroup and operationName isn't enough, because in the case that modular layer and rest layer don't apply the same operation group name, the only condition we can get is operationName, but operationName can be the same across different sub clients.

packages/typespec-ts/src/modular/buildRootIndex.ts
```diff
@@ -5,83 +5,90 @@ import { Client } from "./modularCodeModel.js";
...
-    overwrite: true
-  });
+  exportClassicalClient(client, rootIndexFile, subfolder);
 
-  exportModels(file, srcPath);
-  exportOptionsInterfaces(client, file, srcPath);
-  exportClassicalClient(client, file);
+  if (isLast) {
+    // exportApiIndex(rootIndexFile, srcPath);
+    exportModels(rootIndexFile, project, srcPath);
```
qiaozha, line 28: We need isLast, to build the models, because those the models are just ready by that time.

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -688,6 +719,31 @@ function emitBasicOperation(
...
+  );
+  const sourceRoutePath = ignoreDiagnostics(
+    getHttpOperation(context.program, operation)
+  ).path;
+  const rlcResponses = rlcModels.responses?.filter((op) => {
+    return (
+      (sourceOperationGroupName === "" ||
+        op.operationGroup === sourceOperationGroupName) &&
+      op.operationName === sourceOperationName &&
+      op.path === sourceRoutePath
```
qiaozha, line 743: Maybe operationName + path is enough?

qiaozha (commented):
Still have a few places to improve.

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -167,7 +167,7 @@ export async function _receiveCloudEventsDeserialize(
         id: p.event["id"],
         source: p.event["source"],
         data: p.event["data"],
-        dataBase64: p.event["data_base64"],
+        dataBase64: Buffer.from(p.event["data_base64"] ?? ""),
```
qiaozha, line 170: As in the RLC output, we generate byte-array as string, and in the modular layer we generate it as Uint8Array, we need to use Buffer.from to convert the type.

packages/typespec-test/test/loadTest/spec/models.tsp
```diff
@@ -501,7 +501,6 @@ model ResourceMetric {
 model TestRuns {
   // FIXME: is testRunId optional or required in the response? Resource key and path params cannot be optional
   @key
-  @path
```
qiaozha, line 504: This will cause testRunId being filtered in the RLC response type. which will cause compile errors and if we need to fix it, it will require us to support visibility which is a lot of work. so I removed @ path  temporarily, and also confirmed with Catalina offline, she will rewrite this part.

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/models/index.ts
```diff
@@ -0,0 +1,117 @@
...
+  GetTestOptions,
+  GetTestFileOptions,
+  ListTestFilesOptions,
+  ListTestsOptions,
+  UploadTestFileOptions,
+  DeleteTestFileOptions,
+  DeleteTestOptions,
+} from "../loadTestAdministration/models/index.js";
+export {
+  Test as LoadTestRunClientTest,
```
qiaozha, line 63: This is the shared model between the two sub clients loadTestRun and loadTestAdministration, This is temporarily rename it with a subclient prefix. I am going to address it in #1892

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/api/index.ts
```diff
@@ -0,0 +1,37 @@
...
+  listTests,
+  uploadTestFile,
+  deleteTestFile,
+  deleteTest,
+} from "../loadTestAdministration/api/index.js";
+export {
+  createLoadTestRun,
+  LoadTestRunClientOptions,
+  testRun,
+  createOrUpdateAppComponents as createOrUpdateAppComponentsInLoadTestRunClient,
```
qiaozha, line 24: This is the real api layer operation name conflicts. They tsp is defined as 
```
@client({
    name: "LoadTestAdministrationClient",
    service: AzureLoadTesting,
})
interface AdministrationOperations {
    ...,
    CreateOrUpdateAppComponents is LoadTestAdministration.CreateOrUpdateAppComponents;
    ...
};

@client({
    name: "LoadTestRunClient",
    service: AzureLoadTesting,
})
interface TestRunOperations {
    ...,
    CreateOrUpdateAppComponents is LoadTestRun.CreateOrUpdateAppComponents;
    ...
};
```
It's not a problem for other DPG, because they don't need to export the api directly to the top level, as it belongs to two different sub clients.

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/loadTestAdministration/api/operations.ts
```diff
@@ -0,0 +1,1058 @@
...
+  TestAppComponents,
+  AppComponent,
+  TestServerMetricConfig,
+  FileInfoList,
+  TestsList,
+} from "../models/models.js";
+import {
+  isUnexpected,
+  AzureLoadTestingContext as Client,
+  LoadTestAdministrationCreateOrUpdateAppComponents200Response,
```
qiaozha, line 16: The RLC layer response type name is weird because we have enabled operation group as the response type prefix to resolve the RLC layer operation name conflicts.

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/api/index.ts
qiaozha: Currently, I am using operation name + `In${subClientName}` as the suffix to resolve the conflicts. Not sure if there's other ideas.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -258,6 +262,7 @@ export async function _getImageOperationStatusDeserialize(
   result:
     | GetImageOperationStatus200Response
     | GetImageOperationStatusDefaultResponse
+    | GetImageOperationStatusLogicalResponse
```
qiaozha, line 265: I think this is the correct change, as for LRO, we have a LogicalResponse in the RLC layer for type inference.

packages/typespec-ts/src/index.ts
```diff
@@ -140,13 +235,18 @@ export async function $onEmit(context: EmitContext) {
...
 }
 
-function removeUnusedInterfaces(project: Project) {
-  const allInterfaces = project
-    .getSourceFiles()
-    .flatMap((file) => file.getInterfaces());
+/**
+ * Removing this for now, as it has some problem when we have two models with the same name and only one of them is unused, this function will end up removing the other used models.
+ */
+export function removeUnusedInterfaces(project: Project) {
```
qiaozha, line 241: This is for removing unused models. it fixes the bug that previous the same model name in two sub clients, and one is unused, the other one is being referred, and this function will end up remove both of them. 
  It also fixes that bug when we try to use rename to resolve the conflicts in the `src/models/index.ts`, the top level `src/index.ts` will export those after rename models which has been removed.

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -940,7 +997,7 @@ function emitStdScalar(
 ): Record<string, any> {
   switch (scalar.name) {
     case "bytes":
-      return { type: "Uint8Array", format: "byte" };
+      return { type: "byte-array", format: "byte" };
```
qiaozha, line 1000: bugfixing for byte-array type.

packages/typespec-ts/src/modular/helpers/clientHelpers.ts
```diff
@@ -14,7 +14,11 @@ export function getClientParameters(
 
   const params: OptionalKind<ParameterDeclarationStructure>[] = [
     ...parameters
-      .filter((p) => p.type.type !== "constant")
+      .filter(
+        (p) =>
+          p.type.type !== "constant" &&
+          (p.clientDefaultValue === null || p.clientDefaultValue === undefined)
```
qiaozha, line 20: Really don't remember what is this for, just remember it's a bugfix which has been fixed in TCGC.

qiaozha (commented):
Left some comments, hope this can accelerate the review.

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/api/index.ts
xirzec: I wonder if having a top level `/api` export is useful for multiclient packages. Perhaps we should instead force consumers who wish to import the api methods directly to import the one specific to the subclient?

So one proposal would be remove `api/index.ts` for multiclient generation and only generate `/loadTestRun/api/index.js` and `/loadTestAdministration/api/index.js`

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/models/index.ts
```diff
@@ -0,0 +1,117 @@
+// Copyright (c) Microsoft Corporation.
```
xirzec, line 1: same thought here, maybe we don't need to unify the models for multiple clients into one top level models export?

xirzec (commented):
A couple thoughts

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/api/index.ts
qiaozha: I am totally pro to remove this and remove the @azure/load-testing/api subpath in the scenario of multiple clients.

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/models/index.ts
qiaozha: The story with models is a little different, I am totally okay to remove the subpath of @azure/load-testing/models, but if the intention of removing it is to avoid conflicts, we will probably not get this benefit even remove it, because we still export those models in the src/index.ts https://github.com/Azure/autorest.typescript/pull/1830/files#diff-84b4020119a80f9e0d214073a464a463e61cdc9802e29f5f0b00b7bbc37880e5R69

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/package-files/package.json
qiaozha: Thanks for found this, it should have been resolved :)

qiaozha:
@xirzec @mpodwysocki I have removed the top level `api/index.ts` and `models/index.ts` in the multi clients case. Could you help review again and let me know if you have any comments ? Thanks

packages/typespec-ts/src/modular/buildProjectFiles.ts
```diff
@@ -39,10 +40,6 @@ export function emitPackage(
       "./models": {
         types: "./types/src/models/index.d.ts",
         import: "./dist-esm/src/models/index.js"
-      },
-      "./rest": {
-        types: "./types/src/rest/index.d.ts",
-        import: "./dist-esm/src/rest/index.js"
```
qiaozha, line 45: remove the rest layer subpath exports here.

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
mane-dev: Isn't Buffer.from only supported in Node but not in browser?

xirzec: Yeah I think we need to have some environment specific handling here. Found some prior art in keyvault :

https://github.com/Azure/azure-sdk-for-js/blob/9c9c709d5827b28ebdb9d09c2c1c82fe46d23442/sdk/keyvault/keyvault-certificates/src/utils.ts#L35

As well as special handling for base64 specificaly in core-client:
https://github.com/Azure/azure-sdk-for-js/blob/9c9c709d5827b28ebdb9d09c2c1c82fe46d23442/sdk/core/core-client/src/base64.browser.ts
https://github.com/Azure/azure-sdk-for-js/blob/9c9c709d5827b28ebdb9d09c2c1c82fe46d23442/sdk/core/core-client/src/base64.ts

qiaozha: @xirzec Thank you for the pointers, I will take a look and see if it's necessary to file another PR to fix this issue.

qiaozha:
Pending on https://github.com/Azure/autorest.typescript/issues/1933 to be fixed.

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
joheredi: I'm fine fixing this in a follow up pr

joheredi (approved):

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
qiaozha: Thank you :)

PR #1845, opened 2023-05-15
Bump emitter after support compiler v0.44

MaryGao:
- Bump the emitter version
- Switch to pnpm workspace
- Switch to peer dependency in emitter


packages/autorest.typescript/package.json
```diff
@@ -74,7 +74,7 @@
     "source-map-support": "^0.5.16",
     "ts-morph": "^15.1.0",
     "@azure/core-auth": "^1.3.2",
-    "@azure-tools/rlc-common": "^0.12.0"
+    "@azure-tools/rlc-common": "workspace:^0.13.0"
```
qiaozha, line 77: why ? Does it still work if we released autorest.typescript and typespec-ts ?

packages/typespec-ts/package.json
```diff
@@ -58,19 +58,20 @@
     "pacote": "15.1.3",
     "@types/pacote": "11.1.5"
   },
+  "peerDependencies": {
+    "@azure-tools/typespec-azure-core": ">=0.30.0 <1.0.0",
+    "@azure-tools/typespec-client-generator-core": ">=0.30.0 <1.0.0",
+    "@typespec/compiler": ">=0.44.0 <1.0.0",
+    "@typespec/http": ">=0.44.0 <1.0.0",
+    "@typespec/rest": ">=0.44.0 <1.0.0",
+    "@typespec/versioning": ">=0.44.0 <1.0.0"
```
qiaozha, line 67: why these libraries are not necessarily required for typescript emiter ? https://docs.npmjs.com/cli/v9/configuring-npm/package-json#peerdependencies  @joheredi Could you help confirm if this is ok with you ?

xirzec: We normally recommend against peer deps for most packages, due to the complexity and confusion surrounding how to satisfy them. I'm not sure if TypeSpec decided differently for the purposes of emitters.

@bterlson can you help offer guidance?

.scripts/nightly.yml
```diff
@@ -1,29 +1,30 @@
+trigger: none
 schedules:
   - cron: "0 0 * * *"
```
qiaozha, line 3: maybe 0 0 is not a very good time to start the task.

MaryGao: This would be triggered at midnight in a daily schedule.

MaryGao, line 3:
```suggestion
  - cron: "0 16 * * *"
```

.scripts/nightly.yml
```diff
@@ -1,29 +1,30 @@
+trigger: none
 schedules:
-  - cron: "0 0 * * *"
+  - cron: "0 16 * * *"
```
MaryGao, line 3:
```suggestion
  - cron: "0 12 * * *"
```

qiaozha (approved):
I will approve this for now, @xirzec  let me know if you have other concerns.

Issue #1760, opened 2023-02-15
Add support for overload in RLC


MaryGao:
Add support for overload in RLC

PR #1848, opened 2023-05-17
Support @overload & @sharedRoute

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1760 https://github.com/Azure/autorest.typescript/issues/1763

Leverage the TypeScript overload to support the decorators @sharedRoute and @overload. 

Adding below test cases:
- @sharedRoute for different positions including query, body, header etc
- @overload for the same route, different routes and different actions
- different media types supporting by union, overload and sharedRoute


packages/typespec-ts/test/integration/typespec/overload/routes.tsp
```diff
@@ -0,0 +1,67 @@
...
+using TypeSpec.Http;
+
+namespace TypeScript.Test.Overalod;
+
+@doc("Resource to process.")
+model Resource {
+  id: string;
+}
+@route("/overload")
+interface OveralodOperations {
```
MaryGao, line 12: @joheredi Could you help review the overload cases and their generated code?

packages/typespec-ts/src/operationUtil.ts
```diff
@@ -74,11 +75,15 @@ export function isDefinedStatusCode(statusCode: StatusCode) {
   return statusCode !== "*";
 }
 
-export function isBinaryPayload(body: Type, contentType: string) {
+export function isBinaryPayload(
```
joheredi, line 78: Now that TypeSpec supports [@encode ](https://microsoft.github.io/typespec/standard-library/built-in-decorators#@encode) I wonder if we can leverage it to get rid of this complex logic.

packages/typespec-ts/test/integration/typespec/overload/routes.tsp
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * Overload with same routes
+   */
+  @route("/get")
+  getThing(@body param: string | int32): string | int32;
+
+  @overload(OveralodOperations.getThing)
+  getString(@body param: string): string;
+
+  @overload(OveralodOperations.getThing)
```
joheredi, line 22: @m-nash, did we get to an agreement on how we'll do overloads? Does this align with that?

joheredi (approved):

packages/typespec-ts/src/operationUtil.ts
MaryGao: https://github.com/Azure/autorest.typescript/blob/main/packages/typespec-ts/src/modelUtils.ts#L837-L847 Currently we already have this implemented.

This check helps on binary body not byte in a normal model.

packages/typespec-ts/test/integration/typespec/overload/routes.tsp
m-nash: I believe we agreed we would use sharedRoute in the main.tsp and remove the @overload decorator from azure.core and potentially add it for client.tsp only.

@srnagar can you confirm if that is correct?

packages/typespec-ts/test/integration/generated/mediaTypes/tspconfig.yaml
```diff
@@ -0,0 +1,13 @@
...
+  "@azure-tools/typespec-ts":
+    "emitter-output-dir": "{project-root}"
+    generateMetadata: false
+    generateTest: false
+    addCredentials: false
+    azureSdkForJs: false
+    isCadlTest: true
+    title: MediaTypesClient
+    packageDetails:
+      name: "@msinternal/media-types"
```
qiaozha, line 13: Can you add create client test for these three cases ? just to make sure they can build successfully

qiaozha (commented):
Do you mentioned that you will use the autorest testserver to write test cases for media types ?

MaryGao:
> Do you mentioned that you will use the autorest testserver to write test cases for media types ?

I re-organizated these test cases and their path can't be shared.

packages/typespec-ts/test/integration/typespec/overload/routes.tsp
MaryGao: I moved the current overload cases to client.tsp. I notice that we have an issue https://github.com/Azure/typespec-azure/issues/2738 in Project TOM but not started yet.

srnagar: > I believe we agreed we would use sharedRoute in the main.tsp and remove the @overload decorator from azure.core and potentially add it for client.tsp only.
> 
> @srnagar can you confirm if that is correct?

Yes, that is correct.

qiaozha (approved):

Issue #1841, opened 2023-05-12
[Modular] Some descriptions are missing in models.ts


MaryGao:
Like below definition for `ReceiveResponse` the relevant description is empty.
https://github.com/Azure/autorest.typescript/blob/4c924c0dc5172e18202054f1308b95f462d89e1a/packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/models.ts#L28-L32

Actually we have this in tsp file:
https://github.com/Azure/autorest.typescript/blob/4c924c0dc5172e18202054f1308b95f462d89e1a/packages/typespec-test/test/eventgrid_modular/spec/main.tsp#L105

So we need to fix this.

MaryGao:
Close it as fixed

PR #1861, opened 2023-05-19
Dpg code gen fix docs

joheredi:
The classical client was missing method names. Adding them, also a couple of other fixes:

- Remove calculated type for deserialize operation parameter
- Added return type to send function
- renamed client options after the client


MaryGao:
may fix this https://github.com/Azure/autorest.typescript/issues/1841

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -15,6 +15,17 @@ import { buildType } from "./typeHelpers.js";
 import { getOperationName } from "./namingHelpers.js";
 import { getFixmeForMultilineDocs } from "./fixmeHelpers.js";
 
+function getRLCResponseType(operation: Operation) {
+  const { name } = getOperationName(operation, { casing: "pascal" });
+  const statusCodes: (string | number)[] = Array.from(
+    new Set(operation.responses.flatMap((r) => r.statusCodes)).values()
+  );
+
+  statusCodes.push("Default");
```
qiaozha, line 24: Should we consider LogicalResponse here ? @MaryGao

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/EventGridClient.ts
```diff
@@ -25,18 +25,21 @@ import {
...
+
 export class EventGridClient {
   private _client: EventGridContext;
 
   /** Azure Messaging EventGrid Client */
   constructor(
     endpoint: string,
     credential: KeyCredential,
-    options: ClientOptions = {}
+    options: EventGridClientOptions = {}
```
qiaozha, line 37: Just a reminder, may need to check the ClientOptions in the rest level. Also, should we generate EventGridClientOptions if there's no actual difference between EventGridClientOptions and ClientOptions ? For RLC, the answer is no, right now.

qiaozha (approved):

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/EventGridClient.ts
joheredi: Let's add the logic to remove if no client level parameters in a separate PR

PR #1863, opened 2023-05-19
Upgrade the cadl-ranch version to v15

MaryGao:
- Upgrade the cadl-ranch version to v15
- Add duration test in cadl-ranch


packages/typespec-ts/test/integration/encode.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+      type: "int32-seconds",
+      defaultValue: 36
+    },
+    {
+      type: "iso8601",
+      defaultValue: "P40D"
+    }
+  ];
+
+  matrix([durationData], async (params: TypeDetail) => {
```
qiaozha, line 42: Personally, I don't like using matrix and use as any to do the type casting to work around the compiler error. though it looks more succinct. @joheredi how do you think ?

MaryGao: Generally I would leverage the first generation code to ensure the type info is expected and later using any change if happened to check type changes.  I agree this is not perfect but easy to extend.

qiaozha (approved):

Issue #1864, opened 2023-05-22
subscriptionId is not a required parameter in all operations samples


kazrael2119:
issue:https://github.com/Azure/azure-sdk-for-js/issues/25857

https://github.com/Azure/azure-sdk-for-js/pull/25889#discussion_r1199521374
in swagger , there is no subscriptionId parameter. 
And after my test, this operation can work well without sending subscriptionId in client level.

we can fix it to enhance the customer experience， create an issue to track this

Issue #1772, opened 2023-02-22
[HLC] generate subscriptionId as optional if there're tenant level operations.


qiaozha:
Monitor recently add a spec of tenant level APIs https://github.com/Azure/azure-rest-api-specs-pr/pull/11065, but in our generated code, the client constructor still require a subscriptionId parameter, because subscriptionId has been configured as required and x-ms-parameter-location is client. 
```
  constructor(
    credentials: coreAuth.TokenCredential,
    subscriptionId: string,
    options?: MonitorClientOptionalParams
  ) {
```
Which will cause problems if some customers who are tenant level admins only want to use apis inside tenant level AG apis. because they will need to pass a subscriptionId to create the client anyway. 

The expected behavior is that our codegen could do a required check calculation for client level parameter and client level parameter can only be required when this parameter are required at all operations.

One point: Operations_List may not need a subscriptionId, but I think we should not change subscriptionId as non required if this is the only operation that doesn't have a required subscriptionId parameter.

PR #1869, opened 2023-05-25
make subscirption id optional if there're tenant level operations

qiaozha:
fixes https://github.com/Azure/autorest.typescript/issues/1864
fixes https://github.com/Azure/autorest.typescript/issues/1772


packages/autorest.typescript/src/transforms/operationTransforms.ts
```diff
@@ -607,3 +610,21 @@ function getMapperForSchema(
...
+  const subscriptionIdParameter = operation.parameters?.find(param => {
+    return (
+      param.protocol.http?.in === ParameterLocation.Path &&
+      param.language.default.name.toLowerCase() === "subscriptionid" &&
+      param.implementation === "Client"
+    );
+  });
+  if (
+    subscriptionIdParameter ||
+    operation.operationId === "Operations_List" ||
```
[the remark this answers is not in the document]
qiaozha: no, because operations_list and check name availability or check name exist operation don't have a subscriptionId, and if one RP only has this two operations without client level subscriptionId, then it should not be considered as a RP that has tenant level operations.

packages/autorest.typescript/src/transforms/transforms.ts
```diff
@@ -157,3 +160,45 @@ async function getUberParents(codeModel: CodeModel): Promise<ObjectDetails[]> {
...
+      return;
+    }
+  });
+  operationGroups.forEach(opGroup => {
+    opGroup.operations.forEach(op => {
+      if (op.isTenantLevel) {
+        hasTenantLevelOperation = true;
+      }
+      if (hasTenantLevelOperation) {
+        return;
```
MaryGao, line 196: +1 as above

packages/autorest.typescript/src/models/operationDetails.ts
```diff
@@ -77,6 +77,7 @@ export interface OperationDetails {
   lroOptions?: { "final-state-via": string };
   scope?: Scope;
   namePrefix?: string;
+  isTenantLevel?: boolean;
```
MaryGao, line 80: I didn't see a lot of value of operation-level property, what's the value of this?

packages/autorest.typescript/src/transforms/samplesTransforms.ts
```diff
@@ -162,12 +162,15 @@ export async function getAllExamples(
             clientParameterNames.push(parameterName);
           }
 
-          handleSubscriptionIdParameter(
-            subscriptionIdValue,
-            clientParameterNames,
-            sample.clientParamAssignments,
-            requiredParams
-          );
+          if (!opDetails.isTenantLevel) {
```
MaryGao, line 165: Oh I see this is for sample generation.

packages/autorest.typescript/src/transforms/operationTransforms.ts
```diff
@@ -607,3 +610,21 @@ function getMapperForSchema(
     })
   );
 }
+
+function isOperationTenantLevel(operation: Operation) {
```
MaryGao, line 614: Is my understanding correct?

If one operation has method-level subid defined, the method would return `true`?

MaryGao (approved):
LGTM

PR #1875, opened 2023-06-02
Add encode and empty model cases

MaryGao:
Add encode and empty model cases


packages/typespec-ts/test/integration/encode.spec.ts
```diff
@@ -15,56 +14,230 @@ describe("EncodeDurationClient Rest Client", () => {
...
+      }
+    });
+
+    it(`should get iso8601-array header`, async () => {
+      try {
+        const result = await client
+          .path(`/encode/duration/header/iso8601-array`)
+          .get({
+            headers: {
+              duration: ["P40D", "P50D"] as any
```
MaryGao, line 210: @qiaozha I notice that `RawHttpHeadersInput` only allows for `string|number|boolean` which means there would exist type inference conflict for this property.

https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-rest-pipeline/src/interfaces.ts#L12-L15

I feel like `as any` is just a walkaround and would suggest to update the `RawHttpHeadersInput` header to allow some values like `string[]` or also build a helper function for the default csv format.

How do you think?

qiaozha: Don't think we should change the RawHttpHeadersInput to add string[] type. let's add a helper for it.

qiaozha (approved):

packages/typespec-ts/test/integration/encode.spec.ts
MaryGao: Sure I'll apply that change.

PR #1876, opened 2023-06-02
respect client decorator

qiaozha:


packages/typespec-ts/src/transform/transform.ts
```diff
@@ -63,7 +63,10 @@ export async function transformRLCModel(
   const libraryName = normalizeName(
     options.batch && (options.isModularLibrary || options.batch.length > 1)
       ? client.name
-      : options?.title ?? getDefaultService(program)?.title ?? "",
+      : options?.title ??
+          client.name ??
```
qiaozha, line 67: this name come from tcgc and it also has some logic, I guess we should just follow. https://github.com/Azure/typespec-azure/blob/0456896c5d1d1907c2225125f7df294d613c11b1/packages/typespec-client-generator-core/src/decorators.ts#L59

MaryGao: do we need to fallback to getDefaultService(program)?.title if client.name is empty

MaryGao (approved):
LGTM

PR #1878, opened 2023-06-05
Fix collectionFormat issue in header

MaryGao:
When touching the collectionFormat integration cases we generate string[] for header with format defined, but I notice that `RawHttpHeadersInput` only allows for `string|number|boolean` which means there would exist type inference conflict for this property.

https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-rest-pipeline/src/interfaces.ts#L12-L15

To fix that we need to build a helper function for the default csv format for header.

### DIFF
--- a/packages/autorest.typescript/src/restLevelClient/transforms/transformParameterTypes.ts
+++ b/packages/autorest.typescript/src/restLevelClient/transforms/transformParameterTypes.ts
@@ -171,12 +171,7 @@ function transformBodyParameters(
     );
   } else {
     rlcBodyParam.body = [
-      getParameterSchema(
-        bodyParameters[0],
-        importedModels,
-        false,
-        contentTypeParam
-      )
+      getParameterSchema(bodyParameters[0], importedModels, false, contentTypeParam)
     ];
   }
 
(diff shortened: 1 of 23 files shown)


packages/autorest.typescript/src/restLevelClient/transforms/transformParameterTypes.ts
```diff
@@ -304,4 +304,4 @@ function getAllOperations(model: CodeModel): Operation[] {
   }
 
   return operations;
-}
+}
```
MaryGao, line 307: I didn't support header with array case in swagger side, because I didn't find a valid case anywhere and when I try to build one the autorest would complain `Unknown primitive schema array`.

packages/rlc-common/src/static/serializeHelper.ts
```diff
@@ -1,9 +1,9 @@
 export const buildMultiCollectionContent = `
 export function buildMultiCollection(
-  queryParameters: string[],
```
MaryGao, line 3: Change the param name to `item` because it is not limited to query params.

qiaozha (approved):

PR #1879, opened 2023-06-06
Spilt out some small changes from multiclient in modular

qiaozha:
pick up some small changes in https://github.com/Azure/autorest.typescript/pull/1830 to make it clearer


MaryGao (approved):
LGTM

PR #1884, opened 2023-06-13
Fix the query datetime support issue

MaryGao:
Generate the datetime as `Date | string` when the parameter is in query & model property, as `string` in header.

### DIFF
--- a/packages/typespec-ts/test/unit/parametersGenerator.spec.ts
+++ b/packages/typespec-ts/test/unit/parametersGenerator.spec.ts
@@ -3,92 +3,153 @@ import { emitParameterFromCadl } from "./util/emitUtil.js";
 import { assertEqualContent } from "./util/testUtil.js";
 
 describe("Parameters.ts", () => {
-  describe("handle query apiVersion", () => {
-    it("should't generate apiVersion if there's a client level apiVersion", async () => {
-      const parameters = await emitParameterFromCadl(
-        `
-        model ApiVersionParameter {
-          @query
-          "api-version": string;
-        }
-        op test(...ApiVersionParameter): string;
-        `
-      );
-      assert.ok(parameters);
-      assertEqualContent(
-        parameters?.content!,
-        `
-          import { RequestParameters } from "@azure-rest/core-client";
-          
-          export type TestParameters =  RequestParameters;
+  describe("query parameters", () => {
+    describe("apiVersion in query", () => {
+      it("should't generate apiVersion if there's a client level apiVersion", async () => {
+        const parameters = await emitParameterFromCadl(
           `
-      );
-    });
+          model ApiVersionParameter {
+            @query
+            "api-version": string;
+          }
+          op test(...ApiVersionParameter): string;
+          `
+        );
+        assert.ok(parameters);
+        assertEqualContent(
+          parameters?.content!,
+          `
+            import { RequestParameters } from "@azure-rest/core-client";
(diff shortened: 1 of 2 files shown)


packages/typespec-ts/test/unit/parametersGenerator.spec.ts
```diff
@@ -3,92 +3,153 @@ import { emitParameterFromCadl } from "./util/emitUtil.js";
...
-        `,
-        false,
-        true
-      );
-      assert.ok(parameters);
-      assertEqualContent(
-        parameters?.content!,
-        `
+    describe("other parameters in query", () => {
+      it("should generate user-custom-query ", async () => {
```
qiaozha, line 61: Not sure what's the point of adding this test ?

packages/typespec-ts/test/unit/parametersGenerator.spec.ts
```diff
@@ -3,92 +3,153 @@ import { emitParameterFromCadl } from "./util/emitUtil.js";
...
-            queryParameters: TestQueryParamProperties;
-        }
-        
-        export type TestParameters = TestQueryParam & RequestParameters;
-        `
+        ` import { RawHttpHeadersInput } from "@azure/core-rest-pipeline";
+          import { RequestParameters } from "@azure-rest/core-client";
+          
+          export interface TestHeaders {
+            "execution-to"?: string;
```
qiaozha, line 139: why is this different with what has been defined in the typespec ?

packages/typespec-ts/test/unit/parametersGenerator.spec.ts
MaryGao: this is existing test case I just re-org this one.

packages/typespec-ts/test/unit/parametersGenerator.spec.ts
MaryGao: This handled by tsp compiler not us.

packages/typespec-ts/test/unit/parametersGenerator.spec.ts
```diff
@@ -3,92 +3,153 @@ import { emitParameterFromCadl } from "./util/emitUtil.js";
...
+              queryParameters: TestQueryParamProperties;
           }
           
           export type TestParameters = TestQueryParam & RequestParameters;
           `
-      );
+        );
+      });
+
+      it("should generate offsetDateTime as Date | string", async () => {
```
MaryGao, line 90: This is the one added in this pr.

packages/typespec-ts/test/unit/parametersGenerator.spec.ts
qiaozha: @timotheeguerin Hi, do you have any context on this ? The only document I can found about header is that they are case insensitive, but it doesn't say this dash mark can also be ignored. https://datatracker.ietf.org/doc/html/rfc7230#section-3.2

timotheeguerin: The header decorator automatically convert camelCase to camel-case as the header value if you don’t specify an explicit value

qiaozha: Is this documented somewhere? I mean in TypeSpec documentation.

qiaozha (approved):

PR #1885, opened 2023-06-13
Issues in health insight support

qiaozha:


packages/typespec-ts/src/modelUtils.ts
```diff
@@ -453,6 +453,8 @@ function getSchemaForModel(
     true /** shouldGuard */
   );
 
+  modelSchema.typeName = modelSchema.name;
```
qiaozha, line 456: This is the only change. not sure why Health Insight fail to generate the propertyType name if that property is a Model, I tried to repro it in the unit-test, still unable to repo.

MaryGao: All models in property failed or just some specific one?

qiaozha: all.

MaryGao (approved):
LGTM

packages/typespec-ts/src/modelUtils.ts
MaryGao: Approved but still prefer to find the root cause to repro in UT.

PR #1888, opened 2023-06-15
[DPG] Add sourceOnly flag

deyaaeldeen:
Adds `--sourceOnly` flag which instructs the DPG to generate source files only in the root directory.

### DIFF
--- a/packages/rlc-common/src/interfaces.ts
+++ b/packages/rlc-common/src/interfaces.ts
@@ -246,6 +246,5 @@ export type ResponseHeaderSchema = Schema;
 export type ResponseBodySchema = Schema;
 
 export type ContentBuilder = {
-  (model: RLCModel): File | undefined;
-  (model: RLCModel, hasSampleGenerated?: boolean): File | undefined;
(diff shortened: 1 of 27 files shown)


packages/rlc-common/src/interfaces.ts
```diff
@@ -123,6 +123,7 @@ export interface RLCOptions {
   azureArm?: boolean;
   sourceFrom?: "Cadl" | "Swagger";
   isModularLibrary?: boolean;
+  sourceOnly?: boolean;
```
MaryGao, line 126: I assume we could disable the option `generateMetadata` and then we could only generate src folder.

qiaozha: `generateMetaData: false` should equals `sourceOnly: true`

qiaozha: Oh yeah, got your point, because we are putting the generated code into sources folder. Thanks for the explain :)

xirzec:
I want to rethink this PR a bit. Rather than making this behavior an option, I think the emitter should be smart enough to recognize when the project destination is using customization and place the correct files where they belong.

PR #1889, opened 2023-06-15
fix-is-unexpected-helper-in-healthinsight

qiaozha:
fix the healthinsight isUnexpected helper issue.

### DIFF
--- a/packages/rlc-common/src/buildIsUnexpectedHelper.ts
+++ b/packages/rlc-common/src/buildIsUnexpectedHelper.ts
@@ -116,7 +116,7 @@ export function buildIsUnexpectedHelper(model: RLCModel) {
   });
 
   if (allErrorTypes.size) {
-    const hasTemplate = hasParametrizedPath(pathDictionary);
+    const hasTemplate = true;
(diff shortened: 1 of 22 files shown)


packages/rlc-common/src/buildIsUnexpectedHelper.ts
```diff
@@ -116,7 +116,7 @@ export function buildIsUnexpectedHelper(model: RLCModel) {
   });
 
   if (allErrorTypes.size) {
-    const hasTemplate = hasParametrizedPath(pathDictionary);
+    const hasTemplate = true;
```
MaryGao, line 119: If this is always true, could we remove this logic?

packages/typespec-ts/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@azure-tools/typespec-ts",
-  "version": "0.13.2",
+  "version": "0.13.3",
```
MaryGao, line 3: Changelog?

MaryGao (approved):
lGTM

PR #1896, opened 2023-06-16
Should ignore client.tsp config in rlc

### LINKED ISSUE #1891: [DPG] Ignore the client.tsp in RLC generation
Currently we are using client.tsp @client decorator as the indicator of whether to generate multi-client, which is incorrect, we should generate RLC multi-client based on @service decorator.  Also, we should ignore any kinds of customization in the client.tsp in the RLC generation.

qiaozha:
fixes #1891 
And refactored the code in typespec-ts a little to move all utils related file into utils folder.
Also, removed the multiclient test case in the typespec-test folder, giving we have a real case of loadTest in RLC as the multi-client in test case.


common/config/rush/pnpm-lock.yaml
```diff
@@ -179,7 +179,7 @@ importers:
     specifiers:
       '@azure-tools/typespec-autorest': '>=0.31.0 <1.0.0'
       '@azure-tools/typespec-azure-core': '>=0.31.0 <1.0.0'
-      '@azure-tools/typespec-client-generator-core': '>=0.31.0 <1.0.0'
+      '@azure-tools/typespec-client-generator-core': file:/Users/zhangqiaoqiao/work/code/typespec-azure/packages/typespec-client-generator-core
```
qiaozha, line 182: This is just for temporary, will need to update it when tcgc PR get merged and released.

packages/typespec-test/test/loadTest/spec/client.tsp
```diff
@@ -0,0 +1,49 @@
+import "./main.tsp";
```
qiaozha, line 1: This file is added, and there are two @client decorators, we are using this file as the entrance of code, there's no other generated code change in the loadTest case, which means we have ignored it.

packages/typespec-test/test/multiclient/tspconfig.yaml
```diff
@@ -1,17 +0,0 @@
-emit:
```
qiaozha, line 1: removing this multiclient fake test, as we have testcase like loadTest now.

packages/typespec-ts/src/utils/clientUtils.ts
```diff
@@ -0,0 +1,16 @@
...
+export function getRLCClients(dpgContext: SdkContext): SdkClient[] {
+  const clientsInSdk = listClients(dpgContext);
+  const clients = listClients(dpgContext, false);
+  // to respect the boundaries that is defined by one @client
+  if (clientsInSdk.length === 1) {
+    return clientsInSdk;
+  }
+  // return the original @service clients
+  return clients;
+}
```
qiaozha, line 16: This is the part how we get the RLCClient now.

qiaozha: If the clientsInSdk.length === 1, it means, there's only one @client decorator, and we should respect the bondaries defined by it.  This is working because we already has test cases like healthinsights https://github.com/Azure/autorest.typescript/tree/main/packages/typespec-test/test/healthinsight and there are no changes from healthinsight in this PR which is expected.

packages/typespec-ts/src/utils/clientUtils.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function getRLCClients(dpgContext: SdkContext): SdkClient[] {
+  const clientsInSdk = listClients(dpgContext);
+  const clients = listClients(dpgContext, false);
+  // to respect the boundaries that is defined by one @client
+  if (clientsInSdk.length === 1) {
+    return clientsInSdk;
+  }
+  // return the original @service clients
+  return clients;
```
xirzec, line 15: In light of the feedback on https://github.com/Azure/typespec-azure/pull/3190, I think we should avoid the change to listClients and do something like this:


```suggestion
  const clientsInSdk = listClients(dpgContext);
  // to respect the boundaries that is defined by one @client
  if (clientsInSdk.length === 1) {
    return clientsInSdk;
  }
  
  // Otherwise use service as the client
  const services = listServices(context.program);

  return services.map((service) => {
    const clientName = service.type.name + "Client";
    return {
      kind: "SdkClient",
      name: clientName,
      service: service.type,
      type: service.type,
    };
  });
```

xirzec (commented):
Left some comments on how we calculate the clients, I don't think we actually need any changes to generator-core.

The other refactorings (moving utils and removing the other test case) look good.

packages/typespec-ts/src/utils/clientUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+    } else {
+      // otherwise use service as the client
+      validClients.push({
+        kind: "SdkClient",
+        name: service.name + "Client",
+        service: service,
+        type: service
+      });
+    }
+  });
```
qiaozha, line 33: @xirzec @mpodwysocki  I just realize the previous implementation of getRLCClients won't work if we have two @service  endpoints and each endpoint has only one @client, In this case, I think we should also respect the boundaries defined by the @client. Let me know how you think ?

packages/typespec-ts/test/unit/clientUtil.spec.ts
```diff
@@ -0,0 +1,152 @@
+import { assert } from "chai";
+import { getRLCClientsFromCadl } from "./util/emitUtil.js";
+
+describe("client utils get rlc clients", () => {
```
qiaozha, line 4: added the Unit test of getRLCClients here.

packages/typespec-ts/test/unit/clientUtil.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+      }
+    }
+    `;
+    const clients = await getRLCClientsFromCadl(content);
+    assert.equal(clients.length, 2);
+    assert.equal(clients[0]?.name, "MyClient");
+    assert.equal(clients[1]?.name, "MySecondClient");
+  });
+
+  it("should handle both 1:1 and 1:N mappings between @service and @client", async () => {
```
qiaozha, line 112: This case will probably not happen in reality, but still would like to make sure the function works.

packages/typespec-ts/src/utils/clientUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+  SdkContext,
+  listClients
+} from "@azure-tools/typespec-client-generator-core";
+import { Namespace } from "@typespec/compiler";
+
+export function getRLCClients(dpgContext: SdkContext): SdkClient[] {
+  const clientsInSdk = listClients(dpgContext);
+  const sdkClientsNamespacesMap = new Map<Namespace, SdkClient[]>();
+  const validClients: SdkClient[] = [];
+  clientsInSdk.forEach((client) => {
```
xirzec, line 12: style nit, prefer for/of to forEach

```ts
for (const client of clientsInSdk) {
```

packages/typespec-ts/src/utils/clientUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+    if (!sdkClientsNamespacesMap.has(client.service)) {
```
xirzec, line 13: since you are using `?? []` when getting the array, there's no need to do this extra `set`.

packages/typespec-ts/src/utils/clientUtils.ts
xirzec: If I am understanding correctly, this makes sense to me. 

So, for each service namespace, we respect the client if there is only one or we manufacture a "serviceNameClient" if there are mutliple. Is that correct?

My only question is what happens if initially there is only a single client, and then later on a second one is added? With our current design we would have a breaking change when the single client name is replaced with a service client name, I think.

xirzec (approved):
I'm good with this approach, modulo the breaking change concern when a second client is added to a service namespace.

packages/typespec-ts/src/utils/clientUtils.ts
qiaozha: we have provide a `title` configuration in the tspconfig.yaml to customiza the client name. see the example in health insight https://github.com/Azure/autorest.typescript/blob/main/packages/typespec-test/test/healthinsight/tspconfig.yaml#L23, if customers don't want a breaking change, they can always use title to change it back to the previous name. And this title has a higher priority than what's defined in the @client decorators.

packages/typespec-ts/src/utils/clientUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+  const validClients: SdkClient[] = [];
+  for (const client of clientsInSdk) {
+    const clients = sdkClientsNamespacesMap.get(client.service) ?? [];
+    clients.push(client);
+    sdkClientsNamespacesMap.set(client.service, clients);
+  }
+  sdkClientsNamespacesMap.forEach((value, service) => {
+    if (value.length === 1 && value[0]) {
+      // to respect the boundaries that is defined by one @client
+      validClients.push(value[0]);
```
MaryGao, line 20: Not sure this is the right way to do, because when we only have single client for single service with client customization. 
- How could we only get the scope defined in client.tsp?
- How could we ignore the customization in client.tsp?

qiaozha: I want to clarify something here.

The confusion part is that although our user experience for RLC is like `client.path("/somepath", pathParam1, ...,).put({ })` it looks completely irrelavant with the defined Operation name, the truth is, we have to use that operation name to generate the parameter type name and response type name.  

What @MaryGao is trying to suggest is that If we don't respect operation rename in the multi-client, then we shouldn't respect the operation rename in single client either.  Or we support the rename for both.

The potential issue here is that TypeSpec allows two different operations with the same operation name under different interfaces right now, And they are kind of replying on client.tsp rename to resolve the naming conflicts.

I have pointed this out in my previous email with Brian, but I guess he may have overlooked it. 
image

qiaozha: Just come up with another point of view, If we are actually both respect the rename or both not respect the rename in client.tsp, it could be a liitle weird for both cases.

because in the case of respecting the rename in client.tsp, we will end up with a multi-client user experience that doesn't use the @client name at all, but somehow use the operation name defined in the interfaces which the @client decorated on. 

and in the case of not respecting the rename in client,tsp, we will end up with a single-client user experience that uses the @client name, but somehow ignore the operation name defined in the interface which the @client decorated on. 

About the operation name breaking change concern when changing from single-client to multi-client, I found we will have breaking change anyway because of the client name change, if we have to use title to avoid that breaking, maybe we should also use projectedName json for this https://github.com/Azure/azure-rest-api-specs/pull/24481#discussion_r1234322861

xirzec: This is a great discussion and thank you both for raising the issue. I agree that having different behavior based on the number of input clients is not ideal, as any service might later update to switch from a single to a multi-client approach (though perhaps the reverse is less common.)

The main decision to me is if we want RLC to be **service**-first or **client**-first. My understanding from our previous discussion is that @bterlson felt strongly that we should be service-first.

Assuming we want to be service-first, I think we should *always* ignore the `@client` name and the only way to preserve it would be to set the name explicitly using `title`. This still has the issue of naming collisions that are @qiaozha mentions for operation names, but again we can on configuration overrides like `@projectedName` to address this.

The alternative, of course, is to switch to a client-first approach, but this would go against our previous agreement with @bterlson. 

@mpodwysocki @witemple-msft any thoughts or other solutions I've missed?

qiaozha: I am okay with whatever decision @bterlson would make.  Just provide a few more informations here.

Based on my understanding, transitioning from a single client to a multi-client scenario, or vice versa, is an uncommon occurrence, perhaps even rarer than the multi-endpoint scenario. Such decisions are often the result of compromises rather than unanimous agreement. For instance, there may be different opinions on whether utilizing multiple packages with a single client for health insights is the right approach, or if using a single package with multiple clients for load testing is more suitable. I don't think the SDK arch board wants to go through this review process all over again, and If any service team has any very strong business justifications to change the scenario, and the arch board agree with their decision, the service team should be the one pay for the breaking change not us. 

From user experience perspective, we will only have breaking changes if customers are referring the type by name. i.e. type annotating  for example
```typescript
    const glossaryParameters: GlossaryCreateGlossaryParameters = {
      body: {
        name: glossaryName,
        shortDescription: "Example Short Description",
        longDescription: "Example Long Description",
        language: "en",
        usage: "Example Glossary",
      },
    };
    const glossary = await client.path("/atlas/v2/glossary").post(glossaryParameters);
```
If customers don't use the refer by name, like 
```typescript
    const glossary = await client.path("/atlas/v2/glossary").post({
      body: {
        name: glossaryName,
        shortDescription: "Example Short Description",
        longDescription: "Example Long Description",
        language: "en",
        usage: "Example Glossary",
      },
    });
```
There will be no breaking at all, which should be our recommended way of how to use RLC.  I am thinking maybe we should figure out a way to guide customers to not use type annotating style in RLC.

xirzec: Great point about the usage patterns! Also the type-only name changes at least will not affect pure JS customers

packages/typespec-test/test/healthinsight/generated/typespec-ts/src/parameters.ts
```diff
@@ -5,21 +5,23 @@ import { RawHttpHeadersInput } from "@azure/core-rest-pipeline";
...
+
+export interface CreateJobHeaders {
   /** An opaque, globally-unique, client-generated string identifier for the request. */
   "Repeatability-Request-ID"?: string;
   /** Specifies the date and time at which the request was first created. */
   "Repeatability-First-Sent"?: string;
 }
 
-export interface MatchTrialsBodyParam {
+export interface CreateJobBodyParam {
```
qiaozha, line 17: This diff shows that we are using the original operation names.

packages/typespec-test/test/healthinsight/generated/typespec-ts/src/isUnexpected.ts
```diff
@@ -2,31 +2,39 @@
...
+  CreateJob200Response,
+  CreateJob202Response,
+  CreateJobLogicalResponse,
+  CreateJobDefaultResponse,
 } from "./responses";
 
 const responseMap: Record<string, string[]> = {
+  "GET /trialmatcher/jobs/{jobId}": ["200"],
   "POST /trialmatcher/jobs": ["200", "202"],
   "GET /trialmatcher/jobs": ["200", "202"],
```
qiaozha, line 16: I wonder if this is expected, in this case, it has a pollingOperation decorator in createJob operation that points to the getJob operation, I don't get why we have a `"GET /trialmatcher/jobs": ["200", "202"],` generated in the expected response map ?

packages/typespec-ts/src/utils/clientUtils.ts
```diff
@@ -0,0 +1,18 @@
...
+  const services = listServices(dpgContext.program);
+
+  return services.map(service => {
+    return {
+      kind: "SdkClient",
+      name: service.type.name + "Client",
+      service: service.type,
+      type: service.type
+    }
+  });
```
qiaozha, line 17: I found what Brian means about respect the package boundary is not to the respect the boundary that are defined by the @client decorator, but are the boundaries that this client.tsp refers to. so using `listServices` directly is the right choice here.

packages/typespec-ts/test/integration/generated/mediaTypes/src/clientDefinitions.ts
```diff
@@ -2,28 +2,26 @@
 // Licensed under the MIT license.
 
 import {
-  GetByUnionParameters,
+  GetByUnionOnlyParameters,
```
qiaozha, line 5: The changes here are because of client.tsp which is expected to ignore it. https://github.com/Azure/autorest.typescript/blob/main/packages/typespec-ts/test/integration/typespec/media-types/client.tsp

packages/typespec-ts/test/integration/generated/overload/src/clientDefinitions.ts
```diff
@@ -2,58 +2,41 @@
 // Licensed under the MIT license.
 
 import {
-  GetStringParameters,
-  GetNumberParameters,
-  UploadStringParameters,
-  UploadBytesParameters,
-  ProcessStringParameters,
-  ProcessBytesParameters,
+  GetThingParameters,
```
qiaozha, line 5: The same for the overload test case. https://github.com/Azure/autorest.typescript/blob/main/packages/typespec-ts/test/integration/typespec/overload/client.tsp

packages/typespec-test/test/healthinsight/generated/typespec-ts/src/isUnexpected.ts
qiaozha: Just realize this is probably for querying the resource after the lro succeed.

packages/typespec-ts/src/utils/clientUtils.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  SdkClient,
+  SdkContext
+} from "@azure-tools/typespec-client-generator-core";
+import { listServices } from "@typespec/compiler";
+
+export function getRLCClients(dpgContext: SdkContext): SdkClient[] {
```
xirzec, line 7: I like how simple this is now.

xirzec (approved):
Looks good to me and I believe this matches the feedback.

@bterlson @witemple-msft @mpodwysocki any concerns with merging this?

packages/typespec-ts/src/utils/clientUtils.ts
qiaozha: me too, sorry for any confusions I may have caused. I thought the boundary of client.tsp == the boundary of @client decorator.

MaryGao (approved):
LGTM

PR #1898, opened 2023-06-16
Update load testing service with projected name

MaryGao:
Update json renaming in client.tsp with @@projectedName decorator for load testing service.

### DIFF
--- a/packages/typespec-test/test/loadTest/generated/typespec-ts/src/outputModels.ts
+++ b/packages/typespec-test/test/loadTest/generated/typespec-ts/src/outputModels.ts
@@ -284,7 +284,138 @@ export interface ResourceMetricOutput {
 }
 
 /** Load test run model */
-export interface TestRunOutput {}
+export interface TestRunOutput {
+  /** Pass fail criteria for a test. */
+  passFailCriteria?: PassFailCriteriaOutput;
+  /**
+   * Secrets can be stored in an Azure Key Vault or any other secret store. If the
+   * secret is stored in an Azure Key Vault, the value should be the secret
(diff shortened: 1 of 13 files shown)


packages/typespec-test/test/loadTest/generated/typespec-ts/src/outputModels.ts
```diff
@@ -381,7 +512,7 @@ export interface DimensionValueOutput {
 }
 
 export interface DimensionValueListOutput {
-  value: Record<string, any>;
+  value: string[];
```
qiaozha, line 515: why?

packages/typespec-test/test/loadTest/spec/models.tsp
```diff
@@ -836,13 +836,11 @@ model DimensionValue {
 @parentResource(Dimension)
 model DimensionValueList {
   @key
-  value: {
-    items: DimensionValue[]
-  };
```
qiaozha, line 841: is this change come from service team ?

qiaozha (approved):
LGTM

packages/typespec-test/test/loadTest/spec/models.tsp
MaryGao: The change is not from me, yes, this mainly corrects the wrong definition.

packages/typespec-test/test/loadTest/generated/typespec-ts/src/outputModels.ts
MaryGao: Currently we model anonymous model as `Record<string, any>`

Issue #1910, opened 2023-07-04
RLC: enable a flag for tsp interface as operation group name


MaryGao:
Enable a flag for tsp interface as operation group name

PR #1911, opened 2023-07-04
RLC: enable a flag for tsp interface as operation group name

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1910


.vscode/launch.json
```diff
@@ -84,7 +84,7 @@
     {
       "name": "[TypeSpec] Smoke Test Debug",
       "request": "launch",
-      "cwd": "${workspaceFolder}/packages/typespec-test/test/openai_modular",
+      "cwd": "${workspaceFolder}/packages/typespec-test/test/loadTest",
```
MaryGao, line 87:
```suggestion
      "cwd": "${workspaceFolder}/packages/typespec-test/test/openai_modular",

```

.vscode/launch.json
```diff
@@ -85,6 +85,7 @@
       "name": "[TypeSpec] Smoke Test Debug",
       "request": "launch",
       "cwd": "${workspaceFolder}/packages/typespec-test/test/openai_modular",
+
```
MaryGao, line 88:
```suggestion
```

packages/typespec-ts/src/transform/transfromRLCOptions.ts
```diff
@@ -12,6 +12,8 @@ import { reportDiagnostic } from "../lib.js";
 import { getDefaultService } from "../utils/modelUtils.js";
 import { getRLCClients } from "../utils/clientUtils.js";
 
+export let ENABLE_GROUP_NAME_PREFIX = false;
```
qiaozha, line 15: don't understand why we need to have this ?

packages/typespec-ts/src/transform/transfromRLCOptions.ts
```diff
@@ -104,6 +108,16 @@ function processAuth(program: Program) {
   return securityInfo;
 }
 
+function getEnableOperationGroup(emitterOptions: RLCOptions) {
+  if (
+    emitterOptions.enableOperationGroup === true ||
+    emitterOptions.enableOperationGroup === "true"
```
qiaozha, line 114: I don't think it's necessary to consider "true".

packages/typespec-ts/src/lib.ts
```diff
@@ -71,7 +71,8 @@ export const RLCOptionsSchema: JSONSchemaType<RLCOptions> = {
     },
     azureArm: { type: "boolean", nullable: true },
     sourceFrom: { type: "string", nullable: true },
-    isModularLibrary: { type: "boolean", nullable: true, default: false }
+    isModularLibrary: { type: "boolean", nullable: true, default: false },
+    enableOperationGroup: { type: "boolean", nullable: true }
```
qiaozha, line 75: this is boolean, not sure where can you get a string ?

packages/rlc-common/src/interfaces.ts
```diff
@@ -123,6 +123,7 @@ export interface RLCOptions {
   azureArm?: boolean;
   sourceFrom?: "Cadl" | "Swagger";
   isModularLibrary?: boolean;
+  enableOperationGroup?: boolean | string;
```
qiaozha, line 126:
```suggestion
  enableOperationGroup?: boolean;
```

qiaozha (commented):
As we previously discussed, we should let the loadtesting case breaking, not sure choosing this one as the smoke test case is a good idea.

packages/typespec-ts/src/transform/transfromRLCOptions.ts
MaryGao: this is to save the global info to check if we enable op prefix. it would be used when getting operation name

qiaozha: why not get the value from rlc options? otherwise, why define it in the rlc options ?

packages/typespec-ts/src/lib.ts
MaryGao: There is a bug in compiler, https://github.com/microsoft/typespec/issues/2148

packages/typespec-ts/src/transform/transfromRLCOptions.ts
MaryGao: It's in options now sure let me switch to options.

MaryGao: Removed.

packages/typespec-test/test/loadTest/tspconfig.yaml
```diff
@@ -23,6 +23,7 @@ options:
     generateMetadata: true
     generateTest: true
     azureSdkForJs: false
+    enableOperationGroup: true
```
MaryGao, line 26: We have to enable the flag because there are duplicate operation name in diff interface.

qiaozha (approved):

PR #1915, opened 2023-07-06
add-models-subpath-in-modular-generation

### LINKED ISSUE #1913: [DPG] Should have models subpath export in Modular.
We want to avoid exporting all our models to the top level, as this would obscure some key information about the API. Instead, we want a separate subpath for models, so that they don’t clutter the API document and can still be imported by customers if needed.

qiaozha:
This PR is meant to reflect the suggested changes on adding a models subpath export in the modular generation.
In the previous implementation, we generated `src/api/models.ts` and export all models to both the api layer `index.ts`, and the classical client layer. 
This PR: 
1. moves all the models into `src/models/models.ts`, and all the operation options into `src/models/options.ts`.
2. removes the export for models in api layer `index.ts`
3. changes all the places that refer to models and operation options. 

fixes #1913


packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/index.ts
```diff
@@ -1,25 +1,6 @@
...
-  ChatMessage,
-  ChatRole,
-  ChatCompletions,
-  ChatChoice,
-  ImageOperationResponse,
-  ImageResponse,
-  ImageLocation,
-  State,
-  ImageSize,
-} from "./models.js";
```
qiaozha, line 22: This is the part where I mentioned to remove the export for models in api layer index.ts

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -16,14 +24,6 @@ import {
...
 } from "../rest/index.js";
 import { StreamableMethod } from "@azure-rest/core-client";
-import {
-  Embeddings,
-  Completions,
-  ChatMessage,
-  ChatCompletions,
-  ImageOperationResponse,
-  ImageSize,
-} from "./models.js";
```
qiaozha, line 26: This just changed the places.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/index.ts
```diff
@@ -19,7 +19,7 @@ export {
   ImageLocation,
   State,
   ImageSize,
-} from "./api/models.js";
+} from "./models/index.js";
```
qiaozha, line 22: change the reference

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/models/index.ts
qiaozha, line 1: This is the part of moving models from api/models.ts into models/index.ts

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/OpenAIClient.ts
```diff
@@ -22,6 +17,13 @@ import {
   GetImageOperationStatusOptions,
   StartGenerateImageOptions,
 } from "./api/index.js";
+import {
+  Embeddings,
+  Completions,
+  ChatMessage,
+  ChatCompletions,
+  ImageOperationResponse,
+} from "./models/index.js";
```
qiaozha, line 26: This is the part switching the reference,

packages/typespec-test/test/openai_modular/generated/typespec-ts/package-files/package.json
```diff
@@ -17,6 +17,10 @@
       "types": "./types/src/api/index.d.ts",
       "import": "./dist-esm/src/api/index.js"
     },
+    "./models": {
+      "types": "./types/src/models/index.d.ts",
+      "import": "./dist-esm/src/models/index.js"
+    },
```
qiaozha, line 23: This is the part to add model subpath in package.json. 
Please note, in the codegen, we currently have the real package.json under package-files folder, and generate a package.json without all kinds of SDK repo related dependencies so that the generated code can be build.

qiaozha (commented):
@xirzec @mpodwysocki @witemple-msft  You may mainly focus on the changes in openai_modular smoke test which are in 
`packages/typespec-test/test/openai_modular/generated/typespec-ts` folder of this repo.

xirzec (commented):
This looks correct to me. @mpodwysocki does this follow the same conventions as notification hub?

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/OpenAIClient.ts
```diff
@@ -22,6 +17,13 @@ import {
   GetImageOperationStatusOptions,
   StartGenerateImageOptions,
 } from "./api/index.js";
```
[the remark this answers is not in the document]
qiaozha: https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/notificationhubs/notification-hubs/src/models/options.ts Giving this is how notification-hubs organize it, I will update this PR to reflect the change.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/models/options.ts
```diff
@@ -0,0 +1,181 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { OperationOptions } from "@azure-rest/core-client";
+import { ImageSize } from "./models";
+
+export interface GetEmbeddingsOptions extends OperationOptions {
```
qiaozha, line 7: This is the part to move all the operation options into `src/models/options.ts`

packages/typespec-ts/src/modular/buildSubpathIndex.ts
```diff
@@ -1,8 +1,14 @@
...
-  const indexFile = project.createSourceFile(`${srcPath}/src/api/index.ts`);
+export function buildSubpathIndexFile(
+  project: Project,
+  srcPath: string,
+  subpath: string
+) {
+  const apiFiles = project.getSourceFiles(`**/src/${subpath}/**`);
+  const indexFile = project.createSourceFile(
+    `${srcPath}/src/${subpath}/index.ts`
+  );
```
qiaozha, line 11: I changed this because I found it's almost the same logic tol build `src/api/index.ts` and `src/models/index.ts`

packages/typespec-ts/src/modular/emitModels.ts
```diff
@@ -10,7 +10,9 @@ export function buildModels(
   project: Project,
   srcPath: string = "src"
 ): SourceFile {
-  const modelsFile = project.createSourceFile(`${srcPath}/src/api/models.ts`);
+  const modelsFile = project.createSourceFile(
+    `${srcPath}/src/models/models.ts`
```
qiaozha, line 14: Not sure if `models/models.ts` is a good filename, but I want to put all those models into one separate non-index.ts file.

qiaozha: If I divide it into duplicate files based on operation group name etc, it will potentially have the issue of the same models duplicated copies in different files.

packages/typespec-test/test/widget_dpg/generated/typespec-ts/src/models/options.ts
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { OperationOptions } from "@azure-rest/core-client";
+import { ColorType } from "./models";
```
qiaozha, line 5: should be models.js

packages/typespec-test/test/widget_dpg/generated/typespec-ts/src/api/operations.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+import "../models/index.js";
```
qiaozha, line 4: will need to remove this. or replace the the import from individual file under models folder.

packages/typespec-ts/src/modular/buildOperations.ts
```diff
@@ -53,33 +56,33 @@ export function buildOperationFiles(
...
+    operationGroupFile.fixMissingImports();
   }
-
-  sourceFile.addImportDeclaration({
-    moduleSpecifier: "./models.js",
-    namedImports: models
-  });
-
-  // Import all models and then let ts-morph clean up the unused ones
-  sourceFile.fixUnusedIdentifiers();
```
qiaozha, line 82: Remove this because this is actually not working, because by the time we generate operations.ts file, models folder has not been ready yet. and we use ts-morph `operationGroupFile.fixMissingImports();` which is smart enough to import those dependencies.

packages/typespec-ts/src/modular/emitModels.ts
xirzec: Yeah I think putting everything in one place for now is simpler. Not sure what a better name would be since these models could be used for both input and output. I think the generic term for these things are simply `values` but that name is just as generic as `models` anyway.

xirzec (approved):
Looks good to me! The generated examples are what I expected and this seems to be well aligned with what we have done for NotificationHubs.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -4,6 +4,7 @@
 import {
   OpenAIContext as Client,
   isUnexpected,
+  ChatMessage,
```
MaryGao, line 7: I may not understand why we have model's dependency from RLC not api layer.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -18,30 +19,20 @@ import {
...
-   */
-  user?: string;
-  /**
-   * The model name to provide as part of this embeddings request.
-   * Not applicable to Azure OpenAI, where deployment information should be included in the Azure
-   * resource URI that's connected to.
-   */
-  model?: string;
-}
+} from "../models/models.js";
```
MaryGao, line 28: I may prefer to import from index.js directly and it is simpler and in case we have some changes happen in index.js.
```suggestion
} from "../models/index.js";
```

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts
qiaozha: Because those imports are automactially fixed by ts-morph, it will "helpfully" add the first one it can find, I should move models/models.ts to the up.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/OpenAIClient.ts
```diff
@@ -16,12 +11,19 @@ import {
...
+  ChatMessage,
+  ChatCompletions,
+  ImageOperationResponse,
   GetEmbeddingsOptions,
   GetCompletionsOptions,
   GetChatCompletionsOptions,
   GetImageOperationStatusOptions,
   StartGenerateImageOptions,
-} from "./api/index.js";
+} from "./models/index.js";
```
MaryGao, line 26: Here we import from index.js and this is what I preferred.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts
qiaozha: notification hub is not doing this way https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/notificationhubs/notification-hubs/src/api/cancelScheduledNotification.ts#L6

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/OpenAIClient.ts
qiaozha: Should change this to models/options.

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -4,6 +4,7 @@
 import {
   AcknowledgeCloudEvents200Response,
   AcknowledgeCloudEventsDefaultResponse,
+  CloudEvent,
```
MaryGao, line 7: Same concern why it is imported from RLC?

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts
MaryGao: I am ok with this if our index.js would only `export * from xxx.js` no any renaming stuff.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/OpenAIClient.ts
qiaozha: Found we still need to import from models/models.ts when the body parameter is required. so model/index.js should be the case.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts
qiaozha: fixed

packages/typespec-ts/src/modular/buildOperations.ts
```diff
@@ -41,45 +44,70 @@ export function buildOperationFiles(
       ]);
     });
 
+    // Import models used from ./models.ts
+    // We SHOULD keep this because otherwise ts-morph will "helpfully" try to import models from the rest layer when we call fixMissingImports().
+    importModels(srcPath, operationGroupFile, project);
+
```
qiaozha, line 50: Added comments here about why we can't remove this and why the order matters.

packages/typespec-ts/src/modular/buildOperations.ts
```diff
@@ -41,45 +44,72 @@ export function buildOperationFiles(
...
 
   sourceFile.addImportDeclaration({
-    moduleSpecifier: "./models.js",
+    moduleSpecifier: "../models/models.js",
     namedImports: models
   });
 
   // Import all models and then let ts-morph clean up the unused ones
-  sourceFile.fixUnusedIdentifiers();
+  // we can't fixUnusedIdentifiers here because the operaiton files are still being generated.
```
qiaozha, line 111: Also commented here, why we should not call fixUnusedIdentifiers() here.

packages/typespec-ts/src/modular/buildOperations.ts
```diff
@@ -41,45 +44,72 @@ export function buildOperationFiles(
...
+      .map((id) => {
+        id.setModuleSpecifier(id.getModuleSpecifierValue() + ".js");
+        return id;
+      });
 
-    // Import models used from ./models.ts
-    importModels(operationGroupFile, project);
     operationGroupFile.fixMissingImports();
+    // have to fixUnusedIdentifiers after everything get generated.
+    operationGroupFile.fixUnusedIdentifiers();
```
qiaozha, line 89: This is the right place to call fixUnusedIdentifiers

packages/typespec-ts/src/modular/emitModels.ts
qiaozha: will just keep models/models.ts then :)

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/OpenAIClient.ts
MaryGao: You mean except the options we also have dependencies for models, why not import from both like api layer?
```typescript 
import {...} from "../models/options.js";
import {...} from "../models/models.js";
```

MaryGao (approved):
Overall LGTM, only have one small concern for the inconsistancy import between the api layer and   classical layer for models.

qiaozha:
@bterlson I am going to merge this PR now, feel free to let me know if you found anything, we can always change back :blush:

Issue #1916, opened 2023-07-06
Modular bug: undefined allowInsecureConnection in operation-level would override client-level truthy value


MaryGao:
### Issue

For modular send operation we would always set the default value for options as empty object, this would cause the operation-level option `undefined` e.g `allowInsecureConnection`.  This would introduce issues if we have client-level option set. Because the operation-level option would have high priority than client one https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-client-rest/src/getClient.ts#L64.

```typescript
export function _createOrReplaceSend(
  context: Client,
  role: string,
  name: string,
  options: CreateOrReplaceOptions = { requestOptions: {} }
): StreamableMethod<
  | CreateOrReplace200Response
  | CreateOrReplace201Response
  | CreateOrReplaceDefaultResponse
> {
  return context
    .path("/azure/core/lro/standard/users/{name}", name)
    .put({
      allowInsecureConnection: options.requestOptions?.allowInsecureConnection,
      skipUrlEncoding: options.requestOptions?.skipUrlEncoding,
      headers: { ...options.requestOptions?.headers },
      body: { role: role },
    });
}
```

MaryGao:
Close as fixed

PR #1917, opened 2023-07-06
fix the allowInsecureConnection issue

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1916


MaryGao:
close this as qiaoqiao would fix it in core side.

Issue #1922, opened 2023-07-11
Model type as an enum member


heycalmdown:
Let's say we have an enum `A`;

```typespec
enum A {
  AA,
  BB,
}
```

And a model `B`;

```typespec
model B {
  a: A,
}
```

Then, a model `C` which extends `B`;

```typespec
model C extends B {
  a: A.AA,
}
```

According to the [Type relations](https://microsoft.github.io/typespec/language-basics/type-relations), `a` can be `A.AA` because it can be assignable to `A`. It's a legit TypeSpec code, and can be compiled with `@typespec/openapi3`. But `@azure-tools/typespec-ts` emits the following error;

```
error @azure-tools/typespec-ts/invalid-schema: Couldn't get schema for type EnumMember
```

MaryGao:
@heycalmdown Does that mean you want the sub model C has more specific value than base model B? Could you share which version and command you use to generate this code?
I tried in my local with latest emitter, and didn't find the error but with below ouput.

```typescript
export interface C extends B {}

export interface B {
 /** Possible values: AA, BB */
 a: string;
}
```

heycalmdown:
```typespec
// main.tsp
import "@typespec/rest";

using TypeSpec.Http;

// PAYMENT_METHODS
enum PAYMENT_METHODS {
  CREDIT_CARD: "01",
}

model CommonRegistrationRequest {
  payMethod: PAYMENT_METHODS,
}

model RequestRegisterCC extends CommonRegistrationRequest {
  payMethod: PAYMENT_METHODS.CREDIT_CARD,
}

@service({})
@route("/")
namespace NicepayPlatform {
  @post
  @route("/registration")
  op Registration(@body request: RequestRegisterCC): {};
}
```

```yaml
# tspconfig.yaml
emit:
  - '@typespec/openapi3'
  - '@azure-tools/typespec-ts'
```

- @azure-tools/typespec-ts" - "version": "0.13.3"
- @typespec/compiler" - "version": "0.45.2"

```bash
$ tsp compile ./typespec/test
TypeSpec compiler v0.45.2

Diagnostics were reported during compilation:

/asdf/typespec/test/main.tsp:7:3 - error @azure-tools/typespec-ts/invalid-schema: Couldn't get schema for type EnumMember
> 7 |   CREDIT_CARD: "01",
    |   ^^^^^^^^^^^^^^^^^
[...]

PR #1924, opened 2023-07-12
Support enum member as type

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1922

### DIFF
--- a/packages/typespec-ts/test/integration/modelInheritance.spec.ts
+++ b/packages/typespec-ts/test/integration/modelInheritance.spec.ts
@@ -1,7 +1,8 @@
 import TypeModelInheritanceClientFactory, {
   InheritanceClient,
   Salmon,
-  SharkOutput
+  Shark,
+  Siamese
 } from "./generated/models/inheritance/src/index.js";
 import { assert } from "chai";
 
@@ -14,7 +15,7 @@ describe("ModelsInheritance Rest Client", () => {
     });
   });
 
-  const validBody = { name: "abc", age: 32, smart: true };
+  const validBody: Siamese = { name: "abc", age: 32, smart: true };
   it("should get valid", async () => {
     try {
       const result = await client.path("/type/model/inheritance/valid").get();
@@ -60,7 +61,9 @@ describe("ModelsInheritance Rest Client", () => {
       assert.strictEqual(result.status, "200");
       assert.strictEqual(result.body.age, 1);
       assert.strictEqual(result.body.kind, "shark");
-      assert.strictEqual((result.body as SharkOutput).sharktype, "goblin");
+      if (result.body.kind === "shark") {
+        assert.strictEqual(result.body.sharktype, "goblin");
+      }
     } catch (err) {
       assert.fail(err as string);
     }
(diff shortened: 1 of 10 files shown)


packages/typespec-ts/test/unit/clientFactoryGenerator.spec.ts
```diff
@@ -302,44 +305,19 @@ describe("Client Factory generation", () => {
...
-        
-        return client;
-    }
-    `
-      );
+      `,
+          false,
+          false
+        );
+        assert.fail("Should throw diagnostic errors");
```
qiaozha, line 312: Is it come from anywhere that we should throe diagnostic errors ?

packages/typespec-ts/test/unit/modelsGenerator.spec.ts
```diff
@@ -712,14 +712,110 @@ describe("Input/output model type", () => {
...
+            assertEqualContent(
+              outputModelFile?.content!,
+              `
+            export interface COutput extends BOutput {
+              a: 1.1;
+            }
+    
+            export interface BOutput {
+              /** Possible values: 1.1, 2.2 */
+              a: string;
```
qiaozha, line 813: Should this be number in this case ?

packages/typespec-ts/test/integration/modelInheritance.spec.ts
```diff
@@ -14,7 +15,7 @@ describe("ModelsInheritance Rest Client", () => {
     });
   });
 
-  const validBody = { name: "abc", age: 32, smart: true };
+  const validBody: Siamese = { name: "abc", age: 32, smart: true };
```
qiaozha, line 18: not sure if we shoud add this, as we are recommending our customers to not use type annotation now.

MaryGao: I think this is only for inheritance test case, we could add the interface name explicitly for type checking.

packages/typespec-ts/test/unit/modelsGenerator.spec.ts
MaryGao: number type is not allowed for tsp compiler so we would always fail in this case and in UT we have no way to catch this error.

packages/typespec-ts/test/unit/util/emitUtil.ts
```diff
@@ -108,6 +113,11 @@ export async function emitClientFactoryFromCadl(
   if (clients && clients[0]) {
     apiVersionInfo = transformApiVersionInfo(clients[0], dpgContext, urlInfo);
   }
+  if (isEmptyDiagnostic) {
+    expectDiagnosticEmpty(dpgContext.program.diagnostics);
+  } else {
+    throw dpgContext.program.diagnostics;
```
MaryGao, line 119: here to throw error.

packages/typespec-ts/test/unit/clientFactoryGenerator.spec.ts
MaryGao: https://github.com/Azure/autorest.typescript/pull/1924/files/3ab05f96704bfa7e4ff074e474e63808e1e23963#r1265061076

packages/typespec-ts/test/unit/modelsGenerator.spec.ts
```diff
@@ -239,28 +239,52 @@ describe("Input/output model type", () => {
...
+          cadlType,
+          typeScriptType,
+          {
+            additionalCadlDefinition: cadlTypeDefinition
+          },
+          true
+        );
+      });
+
+      it("with enum value is xx.xx", async () => {
```
MaryGao, line 266: Add one UT

qiaozha (approved):

Issue #1906, opened 2023-06-29
RLC: Handle Non-standard special headers for client-Request-Id and return-Client-Request-Id


lirenhe:
This involves 2 parts of the work:

a. Azure.Core should have a policy to append return-client-request-id and client-request-id.
b. Other SDKs plan to leverage codegen to automatically add this policy to pipeline in client constructors and remove those two head parameters from method signature.   JS RLC should decide what the API signature and payload would look like.

MaryGao:
Batch service has the requirement to custom their request id header.

- `return-client-request-id` ([tsp here](https://github.com/Azure/azure-rest-api-specs/blob/b37ad7f957f0c6f9dd149546e46399308f83b366/specification/batch/Batch/routes.tsp#L2729-L2738)) 
We need to confirm the detailed usage https://github.com/Azure/azure-rest-api-specs/pull/23291/files#r1251427423
- `client-request-id` ([tsp here](https://github.com/Azure/azure-rest-api-specs/blob/b37ad7f957f0c6f9dd149546e46399308f83b366/specification/batch/Batch/routes.tsp#L2719-L2727))

We have a `setClientRequestIdPolicy` with default header `x-ms-client-request-id` for tracing, so I prefer we could add a option `clientRequestIdHeaderName` in PipelineOptions.

setClientRequestIdPolicy: https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-rest-pipeline/src/policies/setClientRequestIdPolicy.ts#L19
PipelineOptions: https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-rest-pipeline/src/createPipelineFromOptions.ts#L24

MaryGao:
For another header return-client-request-id we need more confirmation.
- `return-client-request-id` ([tsp here](https://github.com/Azure/azure-rest-api-specs/blob/b37ad7f957f0c6f9dd149546e46399308f83b366/specification/batch/Batch/routes.tsp#L2729-L2738)) 

Still pending on confirming the detailed usage case. https://github.com/Azure/azure-rest-api-specs/pull/23291/files#r1251427423

PR #1928, opened 2023-07-19
Enable custom header name with client request id

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1906

In this pr, we would work on:
- Upgrade the @azure/core-rest-pipeline to v1.12.0 for both HLC and RLC
- Allow to set request header name as `client-request-id` in RLC
- Enable cadl-ranch test case


packages/typespec-ts/test/integration/clientRequestIdHeader.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+        .path("/special-headers/client-request-id")
+        .get();
+      assert.isNotNull(result.request.headers.get("client-request-id"));
+      assert.strictEqual(result.status, "204");
+    } catch (err) {
+      assert.fail(err as string);
+    }
+  });
+
+  it("should override client-request-id headers correctly", async () => {
```
MaryGao, line 29: The override could work as expected.

packages/typespec-ts/test/integration/clientRequestIdHeader.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+  beforeEach(() => {
+    client = ClientRequestIdClientFactory({
+      allowInsecureConnection: true,
+      retryOptions: {
+        maxRetries: 0
+      }
+    });
+  });
+
+  it("should set client-request-id headers correctly", async () => {
```
MaryGao, line 17:
```suggestion
  it("should append client-request-id header and value if no any input", async () => {
```

packages/typespec-ts/test/integration/clientRequestIdHeader.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+        .path("/special-headers/client-request-id")
+        .get();
+      assert.isNotNull(result.request.headers.get("client-request-id"));
+      assert.strictEqual(result.status, "204");
+    } catch (err) {
+      assert.fail(err as string);
+    }
+  });
+
+  it("should override client-request-id headers correctly", async () => {
```
MaryGao, line 29:
```suggestion
  it("should set client-request-id header and value if input is provided", async () => {
```

packages/rlc-common/src/interfaces.ts
```diff
@@ -12,6 +12,11 @@ export interface RLCModel {
   importSet?: Map<ImportKind, Set<string>>;
   helperDetails?: HelperFunctionDetails;
   urlInfo?: UrlInfo;
+  telemetryOptions?: TelemetryOptions;
+}
+
+export interface TelemetryOptions {
+  customRequestIdHeaderName?: string;
 }
```
qiaozha, line 20: why not import this TelemetryOptions from core-rest-pipeline ?

packages/typespec-ts/src/transform/transform.ts
```diff
@@ -103,14 +107,76 @@ export async function transformRLCModel(
...
+    urlInfo,
+    telemetryOptions
   };
 }
 
-export function transformUrlInfo(
-  program: Program,
-  dpgContext: SdkContext
-): UrlInfo | undefined {
+function transformTelemetryOptions(
```
qiaozha, line 115: Can you put it somewhere else and make the transform.ts file cleaner?

packages/typespec-ts/src/transform/transform.ts
```diff
@@ -103,14 +107,76 @@ export async function transformRLCModel(
...
+function getCustomRequestHeaderNameForOperation(
+  route: HttpOperation
+): string | undefined {
+  const CUSTOM_REQUEST_HEADER_NAME = "client-request-id";
+  const params = route.parameters.parameters.filter(
+    (p) =>
+      p.type === "header" && p.name.toLowerCase() === CUSTOM_REQUEST_HEADER_NAME
+  );
+  if (params.length > 0) {
+    return CUSTOM_REQUEST_HEADER_NAME;
```
qiaozha, line 172: feel a liitle weird that an operation level header parameter need to be implemented in the client level.

What if there're some operations has this header and some operation doesn't have this header ?

packages/typespec-ts/test/integration/clientRequestIdHeader.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  it("should set client-request-id header and value if input is provided", async () => {
+    try {
+      const overrideReqId = "86aede1f-96fa-4e7f-b1e1-bf8a947cb804";
+      const result = await client
+        .path("/special-headers/client-request-id")
+        .get({
+          headers: {
+            "client-request-id": overrideReqId
```
qiaozha, line 36: Is this feature not trying to allow customization of the header name that is used as client request id ? This is just pass a value to client-request-id, I don't see any customization of the header name ?

qiaozha: If you really want to test that you are using the customized header name, the place you should pass the new header name is where you create the client.

packages/rlc-common/src/interfaces.ts
MaryGao: I don't think adding the dependency `core-rest-pipeline` in rlc-common is a good idea, they are different things. One is the interface for client option, the other is the option for codegen to build factory method. Maybe I could try to think out a more specific name here.

packages/typespec-ts/src/transform/transform.ts
MaryGao: Sure.

packages/typespec-ts/src/transform/transform.ts
MaryGao: This is current agreement cross languages, and I think JS could also follow this.

packages/typespec-ts/test/integration/clientRequestIdHeader.spec.ts
MaryGao: The customization is done in client factory side, https://github.com/Azure/autorest.typescript/pull/1928/files/47be9347ea3ec92bed16bccc42cbaa7035f86062#diff-f8f59fb0ca54835e863a4d0d428934e6d47ce3338fad4eeaaf191a089fa34b63R30-R35

MaryGao: The test case is because our current implementation keeps their header as a parameter so customer may leverage their own request id to override our default ones. So in this testing I'd like to verify the overriding works.

qiaozha: Maybe add a test that we can customize the header name not the predefined one from clientoptions and then assert result.request has this new customized name

packages/typespec-ts/src/transform/transform.ts
qiaozha: it's still here ? do you forget to move it ?

packages/typespec-ts/test/integration/clientRequestIdHeader.spec.ts
MaryGao: Yes, sure we could add that test case.

MaryGao: https://github.com/Azure/autorest.typescript/pull/1928/files#diff-d6ec5b3bce965c3900cd5096805dc992c8fad53a46963c77400a948eba854bf7R49-R69

packages/typespec-ts/test/integration/clientRequestIdHeader.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+        result.request.headers.get("client-request-id"),
+        overrideReqId
+      );
+      assert.strictEqual(result.status, "204");
+    } catch (err) {
+      assert.fail(err as string);
+    }
+  });
+
+  it("should custom the header name in client options", async () => {
```
qiaozha, line 49: Can you add tests for modular layer? or you will add the modular test in another PR ?

qiaozha (approved):
LGTM, if you decide to add modular layer test in another PR.

packages/typespec-ts/test/integration/clientRequestIdHeader.spec.ts
MaryGao: Tracked in another pr: https://github.com/Azure/autorest.typescript/issues/1957

joheredi (approved):

PR #1930, opened 2023-07-19
upgrade-m4-to-4.26.2 to support final state-schema

qiaozha:
upgrade m4 version to 4.26.2 to support final-state-schema


lirenhe (approved):

qiaozha:
all tests are passing, should have no impact on the generated code.

Issue #1554, opened 2022-09-08
[Cadl RLC] Fix the byte array converted to string issue


MaryGao:
Currently we generate the `string` type for byte array but it's not perfect so we need to rethink about it.

1. Which type is proper for input?
2. Which type is proper for output?
3. How to handle encodings when serializing?

MaryGao:
Below are confirmed with Time

> The convention we have been following for http:
bytes as a response or body parameter means you are passing a binary body
bytes inside a json object means you need to base64 encode it

MaryGao:
Move to MQ considering we achieve parity now

qiaozha:
This is a probably a candidate for HRLC

PR #1934, opened 2023-07-25
Add converter for uint8array map to string and serialization for models

### LINKED ISSUE #1933: We are missing a serialize function in the modular layer
In the eventgrid_modular, test case,
we have a CloudEvent model. it's looking like 
```typescript
/** Properties of an event published to an Azure Messaging EventGrid Namespace topic using the CloudEvent 1.0 Schema. */
export interface CloudEvent {
  /** An identifier for the event. The combination of id and source must be unique for each distinct event. */
  id: string;
  /** Identifies the context in which an event happened. The combination of id and source must be unique for each distinct event. */
  source: string;
  /** Event data specific to the event type. */
  data?: any;
  /** Event data specific to the event type, encoded as a base64 string. */
  dataBase64?: Uint8Array;
  /** Type of event related to the originating occurrence. */
  type: string;
  /** The time (in UTC) the event was generated, in RFC3339 format. */
  time?: Date;
  /** The version of the CloudEvents specification which the event uses. */
  specversion: string;
  /** Identifies the schema that data adheres to. */
  dataschema?: string;
  /** Content type of data value. */
  datacontenttype?: string;
  /** This describes the subject of the event in the context of the event producer (identified by source). */
  subject?: string;
}
```
here we are using the dataBase64 as the property name. But in the rest layer, 
```typescript
/** Properties of an event published to an Azure Messaging EventGrid Namespace topic using the CloudEvent 1.0 Schema. */
export interface CloudEvent {
  /** An identifier for the event. The combination of id and source must be unique for each distinct event. */
  id: string;
  /** Identifies the context in which an event happened. The combination of id and source must be unique for each distinct event. */
  source: string;
  /** Event data specific to the event type. */
  data?: unknown;
  /** Event data specific to the event type, encoded as a base64 string. */
  data_base64?: string;
  /** Type of event related to the originating occurrence. */
  type: string;
  /** The time (in UTC) the event was generated, in RFC3339 format. */
  time?: Date | string;
  /** The version of the CloudEvents specification which the event uses. */
  specversion: string;
  /** Identifies the schema that data adheres to. */
  dataschema?: string;
  /** Content type of data value. */
  datacontenttype?: string;
  /** This describes the subject of the event in the context of the event producer (identified by source). */
  subject?: string;
}
```
the property name is data_base64, and in the send function, we just assign the modular layer model CloudEvent to the rest layer model
```typescript
export function _publishCloudEventSend(
  context: Client,
  event: CloudEvent,
  topicName: string,
  options: PublishCloudEventOptions = { requestOptions: {} }
): StreamableMethod<
  PublishCloudEvent200Response | PublishCloudEventDefaultResponse
> {
  return context
    .path("/topics/{topicName}:publish", topicName)
    .post({
      contentType:
        (options.contentType as any) ??
        "application/cloudevents+json; charset=utf-8",
      body: { event: event },
    });
}
```

The reason it could build now, it's because the dataBase64 and data_base64 property are optional.  
We should add an internal serialize function for each modular api if there's some propery rename and type mapping. 

In this case, there's still another issue, 
In the serialize we need to convert the Unit8Array into string. 
And in the deserialize, we need to convert the string into Uint8Array.

Another question need to mention is that although the property name and description says it's base64 encoded, its typespec doesn't really add this information.
```typescript
    @doc("Event data specific to the event type, encoded as a base64 string.")
    data_base64?: bytes;
```
if we want to test against base64 encoding, we should define a property like 
```
    @doc("Event data specific to the event type, encoded as a base64 string.")
    @encode("base64")
    data_base64?: bytes;
```

qiaozha:
fixes #1933  https://github.com/Azure/autorest.typescript/issues/1554
This PR:
1. update the dependencies for core-util
2. add converter from uint8array map to string and string to uint8array
3. add serialization for models type in Send function of modular layer.


packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -52,7 +53,20 @@ export function _publishCloudEventSend(
...
+          data: event["data"],
+          data_base64: uint8ArrayToString(event["dataBase64"] ?? "", "base64"),
+          type: event["type"],
+          time: new Date(event["time"] ?? ""),
+          specversion: event["specversion"],
+          dataschema: event["dataschema"],
+          datacontenttype: event["datacontenttype"],
+          subject: event["subject"],
+        },
+      },
```
qiaozha, line 69: This is serialize part I was mentioned in #1933. @joheredi @xirzec

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -28,6 +28,7 @@ import {
   StreamableMethod,
   operationOptionsToRequestParameters,
 } from "@azure-rest/core-client";
+import { uint8ArrayToString, stringToUint8Array } from "@azure/core-util";
```
qiaozha, line 31: this is pending a release of core-util https://github.com/Azure/azure-sdk-for-js/pull/26592

packages/typespec-ts/src/modular/buildOperations.ts
```diff
@@ -25,6 +26,7 @@ export function buildOperationFiles(
   needUnexpectedHelper: boolean = true
 ) {
   for (const operationGroup of client.operationGroups) {
+    utilImports.clear();
```
xirzec, line 29: it feels odd to have a module-level Set that gets cleared by this method. Does the set need to live outside this method or could it be created when buildOperationFiles is called?

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -583,7 +664,46 @@ function deserializeResponseValue(type: Type, restValue: string): string {
       } else {
         return restValue;
       }
+    case "byte-array":
+      utilImports.add("stringToUint8Array");
```
qiaozha, line 668: @xirzec  we will use the utilImports to add the helper function names here. And this function deserializeResponseValue(type: Type, restValue: string) doesn't take a parameter to add imports, and it's being called with a long chain, if we use utilImports as the parameter, we will need to change those functions all the way up. which I am not sure if it's worth it.

packages/typespec-ts/src/modular/buildOperations.ts
qiaozha: I put this outside the scope of this buildOperationFiles functions as explained here  https://github.com/Azure/autorest.typescript/pull/1934/files#r1282500160 

I need to clear this set because the apis per each operationGroup should be generated in an independent file.

xirzec: oh I see. I'm not wholly against it, but it does create some complexity that could cause unexpected behavior later if someone isn't careful.

Maybe we could later refactor this a bit - I remember when I wrote a prototype TS emitter I had a global bag of context that I passed through every generation method that handled things like tracking deps, etc.

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -167,7 +198,13 @@ export async function _receiveCloudEventsDeserialize(
...
         source: p.event["source"],
         data: p.event["data"],
-        dataBase64: Buffer.from(p.event["data_base64"] ?? ""),
+        dataBase64:
+          typeof p.event["data_base64"] === "string"
+            ? stringToUint8Array(
+                p.event["data_base64"] ?? new Uint8Array(),
+                "base64"
+              )
+            : p.event["data_base64"],
```
qiaozha, line 207: This is because the rest layer p.event["data_base64"] is Uint8Array | string, and the modular layer is Uint8Array. So I use the typeof to decide if I need to convert it.

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/loadTestAdministration/api/operations.ts
```diff
@@ -76,11 +76,42 @@ export function _createOrUpdateTestSend(
       contentType:
         (options.contentType as any) ?? "application/merge-patch+json",
       body: {
-        passFailCriteria: options?.passFailCriteria,
+        passFailCriteria: {
+          passFailMetrics: options?.passFailCriteria?.["passFailMetrics"],
```
qiaozha, line 80: added some serialization logic to expand the model type properties.

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/package.json
```diff
@@ -58,7 +58,8 @@
     "tslib": "^2.2.0",
     "@azure/core-paging": "^1.5.0",
     "@azure/core-lro": "^2.5.4",
-    "@azure/abort-controller": "^1.0.0"
+    "@azure/abort-controller": "^1.0.0",
+    "@azure/core-util": "^1.4.0"
```
MaryGao, line 62: Does that mean we would always have this dependency no matter it is used?

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -52,7 +53,23 @@ export function _publishCloudEventSend(
...
+        event: {
+          id: event["id"],
+          source: event["source"],
+          data: event["data"],
+          data_base64: uint8ArrayToString(
+            event["dataBase64"] ?? new Uint8Array(),
+            "base64"
+          ),
+          type: event["type"],
+          time: new Date(event["time"] ?? ""),
```
MaryGao, line 66: I know the `new Date` is not introduced in this pr, but personally I would prefer not to init the optional date with `new Date()`. This would mislead the customers because we give them a default value not undefined.

@joheredi How do you think?

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -52,7 +53,23 @@ export function _publishCloudEventSend(
       contentType:
         (options.contentType as any) ??
         "application/cloudevents+json; charset=utf-8",
-      body: { event: event },
+      body: {
+        event: {
+          id: event["id"],
+          source: event["source"],
+          data: event["data"],
+          data_base64: uint8ArrayToString(
```
MaryGao, line 61: The same concern with the default value for encoded string, why not just return undefined if `event["dataBase64"]`  is undefined.

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
qiaozha: resolved.

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
qiaozha: the same

joheredi (approved):

packages/typespec-ts/src/modular/buildOperations.ts
qiaozha: Thanks for the suggestion, will take that into consideration when I try to adopt to TCGC.

PR #1952, opened 2023-08-02
enable cadl-ranch multi-client-test case

qiaozha:
The cadl-ranch test cases for multi-client has been added via https://github.com/Azure/cadl-ranch/pull/334

This PR is trying to enable the cadl-ranch multi-client test cases in both RLC and Modular. As RLC doesn't take the customization of client.tsp, we will test it's actually ignoring it. In Modular, we will need to respect whatever customization inside the client.tsp. 

currently pending a new release of cadl-ranch.


packages/rlc-common/src/interfaces.ts
```diff
@@ -79,6 +79,7 @@ export interface PathMetadata {
 export type Paths = Record<string, PathMetadata>;
 
 export type PathParameter = {
+  oriName?: string;
```
MaryGao, line 82: What does this mean? Could we have a more meaningful name?

packages/typespec-ts/test/integration/modelInheritance.spec.ts
```diff
@@ -6,7 +6,7 @@ import TypeModelInheritanceClientFactory, {
 } from "./generated/models/inheritance/src/index.js";
 import { assert } from "chai";
 
-describe("ModelsInheritance Rest Client", () => {
+describe.skip("ModelsInheritance Rest Client", () => {
```
MaryGao, line 9: Why skip?

qiaozha: Because those path doesn't exist now, and you already updated the test in your custom header name PR.

packages/rlc-common/src/interfaces.ts
qiaozha: originalName, this is because we have parameter named client in the client constructor, and in RLC client factory function client already been used. that's why we need to rename it as clientParam.  but client parameter also need to be used to build the base url. we need to know the original parameter name that we need to replace.
image

packages/typespec-ts/test/integration/generated/client/structure/default/src/clientDefinitions.ts
```diff
@@ -0,0 +1,63 @@
...
+  /** Resource for '/two' has methods for the following verbs: patch */
+  (path: "/two"): Two;
+  /** Resource for '/three' has methods for the following verbs: patch */
+  (path: "/three"): Three;
+  /** Resource for '/four' has methods for the following verbs: patch */
+  (path: "/four"): Four;
+  /** Resource for '/five' has methods for the following verbs: patch */
+  (path: "/five"): Five;
+  /** Resource for '/six' has methods for the following verbs: patch */
+  (path: "/six"): Six;
```
MaryGao, line 58: The changes look good to me, just curious why we need 6 routes? It seems two routes could satisfy all cases.

qiaozha: I don't think it matters.

MaryGao (approved):
LGTM

PR #1953, opened 2023-08-02
Add `ts-node` to dev dependencies when generating tests

jeremymeng:
as tests relies on it to run .ts test files.


qiaozha (approved):
LGTM

Issue #1388, opened 2022-05-12
[HLC] Fix hard-code parameter issue of client example in readme 


MaryGao:
Currently in `hlcREADME.md.hbs` we hard coded the client example generation with fixed paramters, but it is not truly for all services.

Sometimes the service constructor don't have the required `subscriptionId`, sometimes others have additional `endpoint` in their contstuctor.

```javascript
const subscriptionId = "00000000-0000-0000-0000-000000000000";
const client = new {{ clientClassName }}(new DefaultAzureCredential(), subscriptionId);
```

So we need to fix this with actual parameter states not fixed ones.

lirenhe:
Need to add a new unit test for this.

PR #1954, opened 2023-08-03
fix the parameter issue of client example in readme

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/1388

Add a boolean value to the metadata to determine whether the subscription id needs to be added to the client example in readme.md


MaryGao (approved):
lgtm

Issue #1724, opened 2023-01-16
Invalid client constructor generated.


mvrius:
Hi,

I used the sample YAML from here: https://github.com/Azure/autorest/tree/main/Samples/openapi-v3/1a-code-generation-minimal but tried to generate the Typescript version.

```yaml
input-file: pets.yaml

typescript:
  namespace: Petstore
  output-folder: Client
```
This generates the following code for constructor:

```ts
  constructor(options?: SwaggerPetstoreOptionalParams) {
    // Initializing default values for options
    if (!options) {
      options = {};
    }
    const defaults: SwaggerPetstoreOptionalParams = {
      requestContentType: "application/json; charset=utf-8"
    };

    const packageDetails = `azsdk-js-swaggerPetstore/1.0.0-beta.1`;
    const userAgentPrefix =
      options.userAgentOptions && options.userAgentOptions.userAgentPrefix
        ? `${options.userAgentOptions.userAgentPrefix} ${packageDetails}`
        : `${packageDetails}`;

    const optionsWithDefaults = {
      ...defaults,
      ...options,
      userAgentOptions: {
        userAgentPrefix
      },
      endpoint:
        options.endpoint ?? options.baseUri ?? "http://petstore.swagger.io/v1"
    };
    super(optionsWithDefaults);

    let bearerTokenAuthenticationPolicyFound: boolean = false;
    if (options?.pipeline && options.pipeline.getOrderedPolicies().length > 0) {
      const pipelinePolicies: coreRestPipeline.PipelinePolicy[] = options.pipeline.getOrderedPolicies();
      bearerTokenAuthenticationPolicyFound = pipelinePolicies.some(
        (pipelinePolicy) =>
          pipelinePolicy.name ===
          coreRestPipeline.bearerTokenAuthenticationPolicyName
      );
    }
    if (
      !options ||
      !options.pipeline ||
      options.pipeline.getOrderedPolicies().length == 0 ||
      !bearerTokenAuthenticationPolicyFound
    ) {
      this.pipeline.removePolicy({
        name: coreRestPipeline.bearerTokenAuthenticationPolicyName
      });
      this.pipeline.addPolicy(
        coreRestPipeline.bearerTokenAuthenticationPolicy({
          credential: credentials,
          scopes:
            optionsWithDefaults.credentialScopes ??
            `${optionsWithDefaults.endpoint}/.default`,
          challengeCallbacks: {
            authorizeRequestOnChallenge:
[...]

PR #1956, opened 2023-08-04
Fix credential issue in hlc

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1724

- Add default credential info if the `addCredentials` is set as true by customers
- Add UT for readme generation in HLC


packages/autorest.typescript/src/generators/clientFileGenerator.ts
```diff
@@ -647,11 +654,8 @@ function getTrack2DefaultContent(
   packageDetails: PackageDetails,
   clientDetails: ClientDetails
 ) {
-  const {
-    azureArm,
-    allowInsecureConnection,
-    addCredentials
-  } = getAutorestOptions();
+  const { azureArm, allowInsecureConnection } = getAutorestOptions();
+  const { addCredentials } = getSecurityInfoFromModel(clientDetails.security);
```
MaryGao, line 658: Here is the only change for this pr, others are formattings. We get the option `addCredentials` from `getSecurityInfoFromModel` not `getAutorestOptions`.

qiaozha: I remember there's a case where we set add-credentials true for future manual customization and don't have a security schema defined.

MaryGao: Currently fix we would ignore the credential info if there is security schema defined. If we'd like to support this case we need to decide how to generate the code do you have more info for that case?

packages/autorest.typescript/src/utils/schemaHelpers.ts
```diff
@@ -258,13 +258,21 @@ export function getSecurityInfoFromModel(security: Security) {
...
   }
-  const scopes: string[] = [];
-  credentialScopes.forEach(item => {
-    scopes.push(item);
-  });
+  const scopes: string[] = [...credentialScopes];
+  let refinedAddCredentials =
+    addCredentials === false ? false : security.authenticationRequired;
+  // Add default info if explictly set the addCredentials as true
+  if (
```
MaryGao, line 265: Set the default value as token credential with scope `https://management.azure.com/.default`.

qiaozha (approved):

Issue #1892, opened 2023-06-15
[DPG] Shared Models in case of Multi-Client and Multi-Endpoint


qiaozha:
There are cases where we can have shared models in different sub clients, we will need to figure out a way to indentify those shared models, will need to work in the TCGC to get the information I need first. 

Pending on #https://github.com/Azure/typespec-azure/issues/3035

qiaozha:
Should consider serialize/deserial utils for shared models in multi-client and multi-endpoint scenario.

PR #1962, opened 2023-08-09
generate serialize & deserialize util for special unions 

### LINKED ISSUE #1921: Mapped property for one member of result union not supported
We have an interesting case in OpenAI where one side of a result type had a mapped property, so the generated deserialization function didn't compile:

```ts
export async function _getAzureBatchImageGenerationOperationStatusDeserialize(
  result:
    | GetAzureBatchImageGenerationOperationStatus200Response
    | GetAzureBatchImageGenerationOperationStatusDefaultResponse
): Promise<BatchImageGenerationOperationResponse> {
  if (isUnexpected(result)) {
    throw result.body;
  }

  return {
    id: result.body["id"],
    created: result.body["created"],
    expires: result.body["expires"],
    result: !result.body.result
      ? undefined
      : {
          created: result.body.result?.["created"],
          data: result.body.result?.["data"],
        },
    status: result.body["status"],
    error: !result.body.error ? undefined : result.body.error,
  };
}
```

The problem is that in the case that `result.body.result.data` is a `ImagePayload[]`, the wire property of `b64_json` needs to be converted to the friendly name of `base64Data`.

We worked around this for now by writing a manual conversion helper:

```ts
/**
 * Convert REST-level ImageGenerationsOutput.data to HLC-level ImageGenerations.data
 * @internal
 */
function toImageGenerationsData(
  input: ImageGenerationsOutput["data"]
): ImageLocation[] | ImagePayload[] {
  return input.map(function (locationOrPayload) {
    if ("url" in locationOrPayload) {
      return locationOrPayload as ImageLocation;
    } else {
      return {
        base64Data: locationOrPayload.b64_json,
      } as ImagePayload;
    }
  }) as ImageLocation[] | ImagePayload[];
}
```

This is a tricky problem to resolve since I think if the result was just one of these types we could unwind it and map it in a literal, but since it's a union we have to somehow generate code that only applies to the case where the property to be mapped exists.

qiaozha:
fixes #1921  

In this PR, we only consider generate deserializeUtil for special unions. If the unions in being used in request, we also need to generate the serializeUtils for them. 

Also, there're other cases that will need special serialize and deserialize utils like circular reference of the types as mentioned in #1982, which is not in the scope of this PR to fix.


packages/typespec-test/test/openai_modular/generated/typespec-ts/src/utils/deserializeUtil.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { ImageLocationOutput, ImagePayloadOutput } from "../rest/index.js";
+import { ImageLocation, ImagePayload } from "../models/models.js";
+
+/** type predict function fpr ImagePayloadOutput array from ImageLocationOutput[] | ImagePayloadOutput[] */
+function isImagePayloadArray(
```
qiaozha, line 8: The type predict function for special union variant.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/utils/deserializeUtil.ts
```diff
@@ -0,0 +1,35 @@
...
+): obj is ImagePayloadOutput[] {
+  if (Array.isArray(obj) && obj.length > 0) {
+    return (obj as ImagePayloadOutput[])[0].b64_json !== undefined;
+  }
+
+  return false;
+}
+
+/** deserialize function for ImagePayload array */
+function deserializeImagePayloadArray(
```
qiaozha, line 19: the deserialize function for special uinon variants.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/utils/deserializeUtil.ts
```diff
@@ -0,0 +1,35 @@
...
+  obj: ImagePayloadOutput[]
+): ImagePayload[] {
+  return (obj || []).map((item) => {
+    return { base64Data: item["b64_json"] };
+  });
+}
+
+/** deserialize function for ImageLocationOutput[] | ImagePayloadOutput[] */
+export function deserializeImageLocationArrayAndImagePayloadArrayUnion(
```
qiaozha, line 28: deserializethe whole union. no idea how to give it a better name.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/utils/deserializeUtil.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation.
```
qiaozha, line 1: Not sure if put the deserializeUtil into src/utils folder is correct.

packages/typespec-test/test/unionwidget_dpg/spec/main.tsp
```diff
@@ -0,0 +1,174 @@
...
+  data: WidgetData0 | WidgetData4[]
+}
+
+model Widget23 extends Widget {
+  data: WidgetData0[] | WidgetData5
+}
+
+model Widget24 extends Widget {
+  data: WidgetData0 | WidgetData5[]
+}
```
qiaozha, line 139: In general, we have four kinds of special union variants.
1. model with property of datetime type.
2. model with property of binary array type.
3. model that has different property name between rest layer and modular layer.
4. list that has model with above conditions as the element type.

The openAI case is a combination with case 3 and a normal model. But here we have listed 24 kinds of `Widget${num}` model basically cover all kinds of combinations of unions by combining with the four basical types and regular models with each other, which I think should be enough for now.

qiaozha (commented):
Self reviews.

packages/typespec-test/test/unionwidget_dpg/generated/typespec-ts/src/utils/deserializeUtil.ts
```diff
@@ -0,0 +1,512 @@
...
+    | WidgetData2Output
+    | WidgetData2Output
+    | WidgetData3Output
+    | WidgetData0Output[]
+    | WidgetData2Output
+    | WidgetData1Output[]
+    | WidgetData2Output
+    | WidgetData2Output
+    | WidgetData3Output[]
+): obj is WidgetData2Output {
```
qiaozha, line 61: As in this test case, we have a lot of different combinations of union types that have the same type of union variant. and we generate overloads for the type predict function of this union variant.

qiaozha: Just found out there's duplicate types in the implementation. will need to update it.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/utils/deserializeUtil.ts
MaryGao: I remember in our last review meeting we mentioned that we could put shared components in `common` folder.

packages/typespec-test/test/unionwidget_dpg/spec/main.tsp
MaryGao: Could we also add a case for union with three parts? `A | B | C`

packages/typespec-test/test/unionwidget_dpg/generated/typespec-ts/src/utils/deserializeUtil.ts
```diff
@@ -0,0 +1,512 @@
...
+export function deserializeWidgetData2ArrayAndWidgetData3Union(
+  obj: WidgetData2Output[] | WidgetData3Output
+): WidgetData2[] | WidgetData3 {
+  if (isWidgetData2Array(obj)) {
+    return deserializeWidgetData2Array(obj);
+  }
+  if (isWidgetData3(obj)) {
+    return deserializeWidgetData3(obj);
+  }
+  return obj;
```
MaryGao, line 458: I notice that in `deserializeWidgetData2ArrayAndWidgetData3Union` we have two `if`s, but in `deserializeWidgetData1AndWidgetData2ArrayUnion` we only one `if`.

packages/typespec-test/test/unionwidget_dpg/generated/typespec-ts/src/utils/deserializeUtil.ts
```diff
@@ -0,0 +1,512 @@
...
+}
+
+/** deserialize function for WidgetData1Output | WidgetData2Output[] */
+export function deserializeWidgetData1AndWidgetData2ArrayUnion(
+  obj: WidgetData1Output | WidgetData2Output[]
+): WidgetData1 | WidgetData2[] {
+  if (isWidgetData2Array(obj)) {
+    return deserializeWidgetData2Array(obj);
+  }
+  return obj;
```
MaryGao, line 445: One if here.

packages/typespec-test/test/unionwidget_dpg/generated/typespec-ts/src/models/models.ts
```diff
@@ -0,0 +1,133 @@
...
+  weight: number;
+  color: ColorType;
+}
+
+/** Type of ColorType */
+/** */
+export type ColorType = "red" | "blue";
+
+export interface Widget1 {
+  data: WidgetData0 | WidgetData1;
```
MaryGao, line 15: I notice the definition part it is extened from Widget, why is there no this info generated?

```
model Widget {
  @key id: string;
  weight: int32;
  color: "red" | "blue";
}

model Widget1 extends Widget {
  data: WidgetData0 | WidgetData1
}
```

packages/typespec-test/test/unionwidget_dpg/generated/typespec-ts/src/utils/deserializeUtil.ts
qiaozha: Because WidgetData2 is defined as  
```
model WidgetData2 {
  @projectedName("json", "foo_prop")
  fooProp: string;
}
```
and WidgetData3 is defined as 
```
model WidgetData3 {
  @projectedName("json", "bar_prop")
  barProp: string;
}
```
both of them belong to special union variant, each type needs a special deserialize function. 
Whereas WidgetData1 is defined as 
```
model WidgetData1 {
  barProp: string;
}
```
which is just a normal object we don't need special deserialization for it. 
How many if statements depends on how many special union variants we have in this union, not by the number of it.

packages/typespec-test/test/unionwidget_dpg/generated/typespec-ts/src/models/models.ts
qiaozha: will look into this tomorrow.

packages/typespec-test/test/unionwidget_dpg/spec/main.tsp
qiaozha: I can add a simple one, but I don't think it's necessary to have all kinds of combination around 3 union variants.

packages/typespec-ts/src/modular/buildOperationUtils.ts
```diff
@@ -0,0 +1,441 @@
...
+      kind: StructureKind.Function,
+      docs: [`type predict function fpr ${type.name} from ${typeUnionNames}`],
+      name: `is${toPascalCase(formalizeTypeUnionName(type.name))}`,
+      parameters: [{ name: "obj", type: typeUnionNames }],
+      returnType: `obj is ${type.name}Output`
+    };
+    if (type.properties) {
+      statements.push(
+        `return ${type.properties
+          .filter((p) => !p.optional)
```
MaryGao, line 371: Does that mean we will pick the first required property as the indicator?

packages/typespec-test/test/unionwidget_dpg/spec/main.tsp
qiaozha: done, see test cases 25 and 26.

packages/typespec-ts/src/modular/buildOperationUtils.ts
qiaozha: No, this is to filter out all the required properties. not sure where you get the first required property ?

packages/typespec-test/test/unionwidget_dpg/generated/typespec-ts/src/models/models.ts
qiaozha: fixed.

packages/typespec-ts/src/modular/buildOperationUtils.ts
MaryGao: Hah, got it. I may misunderstand with our generated code because there is only one property involved.

So we will include all required proerties as type indicator, right?

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -701,13 +706,13 @@ function deserializeResponseValue(
   required: boolean
 ): string {
   const coreUtilSet = importSet.get("@azure/core-util");
+  const deserializeUtils = importSet.get("../utils/deserializeUtil.js");
```
MaryGao, line 709: Not the priority, just wondering if this could work in multi-client case.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/utils/deserializeUtil.ts
qiaozha: This is how notification-hub does https://github.com/Azure/azure-sdk-for-js/tree/main/sdk/notificationhubs/notification-hubs/src/utils

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
qiaozha: As this deserilieUtils is highly related with models, I will consider this after #1892 is resolved.

packages/typespec-test/test/unionwidget_dpg/generated/typespec-ts/src/utils/deserializeUtil.ts
```diff
@@ -0,0 +1,596 @@
...
+}
+
+/** type predict function fpr WidgetData6Output array from WidgetData2Output | WidgetData4Output | WidgetData6Output[] */
+function isWidgetData6Array(
+  obj: WidgetData2Output | WidgetData4Output | WidgetData6Output[]
+): obj is WidgetData6Output[] {
+  if (Array.isArray(obj) && obj.length > 0) {
+    return (obj as WidgetData6Output[])[0].data !== undefined;
```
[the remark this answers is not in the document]
qiaozha: fixed

packages/typespec-ts/src/modular/buildOperationUtils.ts
qiaozha: we will include all required properties in this type and all its properties of any property which is a model type.  see https://github.com/Azure/autorest.typescript/pull/1962/files#diff-fee0329cc5e0bcbba7157ff4ae985fdbb1f41a2a389e764a084dbdf1487ff82eR533-R543 for details.

packages/typespec-ts/test/modularUnit/deserializeUtil.spec.ts
```diff
@@ -0,0 +1,1065 @@
+import { assert } from "chai";
+import { emitModularDeserializeUtilsFromTypeSpec, emitModularOperationsFromTypeSpec } from "../util/emitUtil.js";
+import { assertEqualContent } from "../util/testUtil.js";
+
+describe.only("modular special union deserialization", () => {
```
MaryGao, line 5: remove `only`?

packages/typespec-ts/test/modularUnit/deserializeUtil.spec.ts
```diff
@@ -0,0 +1,1065 @@
...
+      }
+      
+      /** deserialize function for WidgetData0Output | WidgetData1Output */
+      export function deserializeWidgetData0AndWidgetData1Union(
+        obj: WidgetData0Output | WidgetData1Output
+      ): WidgetData0 | WidgetData1 {
+        if (isWidgetData1(obj)) {
+          return deserializeWidgetData1(obj);
+        }
+        return obj;
```
MaryGao, line 91: In this case there is no special property in `WidgetData0` so we would directly assign that object in RLC to its parent objection in modular. I was wondering if we need to assign a new one to modular e.g `Object.assign`.

Another thinking is if we accept the above rule do we need to apply this rule to general models e.g body position. If both models between RLC and Modular are exactly the same we could simply copy the reference.

qiaozha: The difference between deep copy and reference copy is where there are cases these two objects might have different life cycles. Can you list any examples that can actual cause the object in Modular and the object in RLC in two differenct life cycle and cause potential issues here ? Also,, I am not quite familiar with JavaScript and TypeScript garbage collection, so I may not have any preference here. @xirzec Could you help provide some insights ? Thanks

qiaozha: If we are really going to deep copy the object, Object.assign may not work in this case, because at the step of `return obj` obj represents all the non-special union variants,, which are not necessarily required to be a tsp model type. it could be simple boolean/number/string or array type.

packages/typespec-ts/test/modularUnit/deserializeUtil.spec.ts
qiaozha: good catch!

qiaozha, line 5:
```suggestion
describe("modular special union deserialization", () => {
```

packages/typespec-ts/src/modular/buildOperationUtils.ts
MaryGao: Could we add a UT that is a special union A | B with their required properties are the same but with different optional parameters?

packages/typespec-ts/test/modularUnit/deserializeUtil.spec.ts
lirenhe: I would just only we face a real case that returning ref value would cause some problem, we should consider using deep copy. Otherwise, let's make the code simple and clean.

packages/typespec-ts/src/modular/buildOperationUtils.ts
qiaozha: I don't think this is a valid case, if there's an object that is missing the optional properties, we will never know whether it's type A or type B by all means.

qiaozha:
In general, we have two kinds of basic special union variants.

1. datetime type
2. binary array type

If we consider model type, we have the following three type.
3. model with property of datetime type.
4. model with property of binary array type.
5. model that has different property name between rest layer and modular layer.
6. nested model i.e. model with property that is a model with one of the above three conditions.

If we consider array type, with all the above 6 types as the element types: 

if we consider Record type, with all the above 6 types as the value types?   Still pending.

packages/typespec-ts/src/modular/buildOperationUtils.ts
MaryGao: I may take this case as negative testing, if tsp could define this.

packages/typespec-ts/test/modularUnit/deserializeUtil.spec.ts
xirzec: If we're talking about deserializing a JSON payload, all objects will initially be created by `JSON.parse`, so at least we don't have to worry about the consumer having a reference to these types. (There is a small exception where if they hook up an `onResponse` callback they could get a reference, but I think that is acceptable for our purposes.)

So I'm not too worried about deep copy for deserializing. The other direction (serialization) is trickier because we know that consumers sometimes re-use the same object many times in their inputs (e.g. defining a common part of their objects in one place) and we have to be careful to not mutate input parameters.

packages/typespec-ts/src/modular/buildOperationUtils.ts
```diff
@@ -0,0 +1,774 @@
...
+import {
+  FunctionDeclarationStructure,
+  SourceFile,
+  StructureKind
+} from "ts-morph";
+import { Imports as RuntimeImports } from "@azure-tools/rlc-common";
+import { UsageFlags } from "@typespec/compiler";
+
+/**
+ * This function creates a file under /api for each operation group.
```
lirenhe, line 18: Please ensure the comment is correct for this large file.

packages/typespec-ts/src/modular/buildOperationUtils.ts
```diff
@@ -0,0 +1,774 @@
...
+          getTypeDeserializeFunction(
+            utilsFile,
+            et,
+            serializeType,
+            importSet,
+            model.runtimeImports
+          );
+        }
+      });
+      const deserializeFUnctionName = getDeserializeFunctionName(
```
lirenhe, line 77:
```suggestion
      const deserializeFunctionName = getDeserializeFunctionName(
```

packages/typespec-ts/src/modular/buildOperationUtils.ts
lirenhe: It would be good that you could list the supported case that we could tell the difference between the models in unions.

packages/typespec-test/test/unionwidget_dpg/spec/main.tsp
```diff
@@ -0,0 +1,257 @@
...
+  data: utcDateTime | bytes;
+}
+
+model Widget32 extends Widget {
+  data: WidgetData0 | Record<string>;
+}
+
+model Widget33 extends Widget {
+  data: WidgetData2 | Record<string>;
+}
```
qiaozha, line 179: Added test cases that we can't handle here.
for cases 29-31, we can't handle it because both of them are serialized into string, we don't know if we should deserialize when we get a string. they will all generate as any.
for cases 32, there's no special union variant in this case, we don't need to generate the util for it.
for cases 33, there's special union variant and there's also a dictionary, we don't know how to serialize and deserialize without furthur information. so it will generate as any.

packages/typespec-ts/src/modular/buildOperationUtils.ts
```diff
@@ -0,0 +1,871 @@
...
+ */
+export function buildOperationUtils(model: ModularCodeModel) {
+  const serializeUtilFiles = [];
+  for (const serializeType of ["serialize", "deserialize"]) {
+    const usageCondition =
+      serializeType === "serialize" ? UsageFlags.Input : UsageFlags.Output;
+    const specialUnions = model.types.filter(
+      (t) =>
+        !isSpecialExcludeUnion(t) &&
+        isSpecialUnion(t) &&
```
MaryGao, line 29: Idealy i'd like to distinguish below cases:
- we don't need to serializae/deserialize between rlc and modular
- we can't do these utils so we exclude them like "Date | string"

For the former we don't do anything; for the latter it may introduce non-compile code wihtout further customizations.

packages/typespec-ts/src/modular/buildOperationUtils.ts
```diff
@@ -0,0 +1,871 @@
...
+ * 1. datetime type
+ * 2. bytes type
+ * If we consider model type, we have the following three type.
+ * 3. model with property of datetime type.
+ * 4. model with property of binary array type.
+ * 5. model that has different property name between rest layer and modular layer.
+ * 6. nested model i.e. model with property that is a model with one of the above three conditions.
+ * If we consider array type, with all the above 6 types as the element types.
+ */
+export function isSpecialUnionVariant(t: Type): boolean {
```
MaryGao, line 146: do we consider the named union?

packages/typespec-ts/src/modular/buildOperationUtils.ts
```diff
@@ -0,0 +1,871 @@
...
+  prefix: string,
+  serializeType: string
+): string {
+  const typeProperties = getAllProperties(type);
+  return typeProperties
+    .filter((p) => !p.optional)
+    .map((p) => {
+      const condition: string[] = [];
+      if (p.type.type === "model") {
+        const properties = [];
```
MaryGao, line 838: why do we need `properties` here?

packages/typespec-test/test/unionwidget_dpg/spec/main.tsp
```diff
@@ -0,0 +1,257 @@
...
+
+@error
+model Error {
+  code: int32;
+  message: string;
+}
+
+interface WidgetService {
+  @get @route("customGet/response") customGetResponse(): Widget;
```
MaryGao, line 188: will we have any difference if the union happens in body and in normal property e.g binary type?

packages/typespec-ts/src/modular/buildOperationUtils.ts
```diff
@@ -0,0 +1,871 @@
...
+ * This function creates serialize and deserialize utils for special unions and that are used in the operation.
+ */
+export function buildOperationUtils(model: ModularCodeModel) {
+  const serializeUtilFiles = [];
+  for (const serializeType of ["serialize", "deserialize"]) {
+    const usageCondition =
+      serializeType === "serialize" ? UsageFlags.Input : UsageFlags.Output;
+    const specialUnions = model.types.filter(
+      (t) =>
+        !isSpecialExcludeUnion(t) &&
```
MaryGao, line 28: Not sure we could enum all excludes cases, if not is it possible to set the white list explicitly? Any items not in that list would be taken as not supported.

PR #1976, opened 2023-08-16
Improve readme "Get started" and fix credential policy issue.

lukben2000:


lukben2000:
@microsoft-github-policy-service agree

packages/autorest.typescript/src/generators/clientFileGenerator.ts
```diff
@@ -710,7 +710,7 @@ function getTrack2DefaultContent(
         });
         this.pipeline.addPolicy(
           coreRestPipeline.bearerTokenAuthenticationPolicy({
-            credential: credentials,
+            credential: options.credential,
```
MaryGao, line 713: Could you add more descriptions for this fix?

My understanding is that we enable this policy only when there is a required parameter `credentials` so we should not go to optional one `options.credential`.

lukben2000: I had to do a search-and-replace temp fix for my project, when using this package. In the previous implementation `credentials` are not defined. So it might be a matter of exposing this variable then, but let me just have a second look when I get off work

lukben2000: I am running autorest with the following command:
`autorest --typescript  --package-name=api-client --input-file=api-specs.json --package-version=1.0.0 --add-credentials=true --generate-metadata=true --generate-client-as-impl=true --generate-client-interfaces --output-folder=api-client-package`

Maybe I am missing something in regards to the credentials?

MaryGao: Not sure what kind of issue you met, if you enable the option `add-credentials` it's better to define your credential details in OpenAPI([refer link](https://github.com/Azure/autorest/blob/main/docs/generate/authentication.md)), otherwise we will met [the known issue ](https://github.com/Azure/autorest.typescript/issues/1724) and I was trying to fix in [this pr](https://github.com/Azure/autorest.typescript/pull/1956).

Please file a new one if your case is different https://github.com/Azure/autorest.typescript/issues.

MaryGao:
@lukben2000 Could you try wiht our latest version https://www.npmjs.com/package/@autorest/typescript/v/6.0.9? I believe this should be fixed in our latest.

I will close this pr considering it's fixed and feel free to re-open it if still existing.

PR #1977, opened 2023-08-18
remove default value in Modular

qiaozha:
for comment https://github.com/Azure/autorest.typescript/pull/1975#discussion_r1297908714

### DIFF
--- a/packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts
+++ b/packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts
@@ -456,7 +456,7 @@ export function _beginAzureBatchImageGenerationSend(
       ...operationOptionsToRequestParameters(options),
       body: {
         prompt: prompt,
-        n: options.n ?? 1,
+        n: options?.n,
         size: options?.size,
         response_format: options?.responseFormat,
         user: options?.user,
--- a/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
+++ b/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
@@ -395,12 +395,9 @@ function getParameterMap(
     return getConstantValue(param);
   }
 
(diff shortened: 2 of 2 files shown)


packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -498,39 +497,6 @@ function getOptionalWithoutDefault(
...
-  if (param.type.type === "model") {
-    return `"${param.restApiName}": {${getRequestModelMapping(
-      param.type,
-      "options?." + param.clientName,
-      importSet
-    ).join(", ")}} ?? ${getQuotedValue(param)}`;
-  }
-  return `"${param.restApiName}": options.${
-    param.clientName
-  } ?? ${getQuotedValue(param)}`;
```
qiaozha, line 523: the question mark change is because of we now using getOptional which used to be getOptionalWithoutDefault and for OpenAI's case, they have a default value, they previously go into getOptionalWithDefault function.

qiaozha: Compared with all the places that is using options to build the request body, this is the only places that doesn't have a question mark. 

As options by literal means this parameter is optional, we should keep `options?. ` no matter if it has a default value or not.   Besides, even if it does have a default value. `options?.` won't report any errors.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -456,7 +456,7 @@ export function _beginAzureBatchImageGenerationSend(
       ...operationOptionsToRequestParameters(options),
       body: {
         prompt: prompt,
-        n: options.n ?? 1,
+        n: options?.n,
```
qiaozha, line 459: This is the change about question mark

MaryGao (approved):
LGTM

Issue #1856, opened 2023-05-18
[Modular] Implement Pagination


qiaozha:
Say we have below tsp definition
```
  @doc("List Widget resources")
  listWidgets is Operations.ResourceList<
    Widget,
    ListQueryParametersTrait<StandardListQueryParameters & SelectQueryParameter>
  >;
```

We would expect to generate modular operation

```typescript
  public listWidgets(
    options?: ListWidgetsOptionalParams
  ): PagedAsyncIterableIterator<Widget> 
```

### Option 1: One paging helper per operation

```typescript
public listWidgets(
    options?: ListWidgetsOptionalParams
  ): PagedAsyncIterableIterator<Widget> {
    const iter = this.__listWidgetsPagingAll(options);
    return {
      next() {
        return iter.next();
      },
      [Symbol.asyncIterator]() {
        return this;
      },
      byPage: (settings?: PageSettings) => {
        if (settings?.maxPageSize) {
          throw new Error("maxPageSize is not supported by this operation.");
        }
        return this.getEmptyNextLinkNamePagesPagingPage(options, settings);
      }
    };
  }


private async __listWidgetsPagingAll(
    options?: ListWidgetsOptionalParams
  ): AsyncIterableIterator<Widget> {
    ....
  }
```

### Option 2: Leverage pagination helper in RLC
```typescript
  public listWidgets(
    options?: ListWidgetsOptionalParams
  ): PagedAsyncIterableIterator<Widget> {
    const initialResponse = await _listWidgetsSend(
        context,
        options
      );
    return paginate(client, initialResponse, {
         customGetPage: () => { ... // deserilized logic}
   });
}
```

### Option 3: A modular generator
See details as below.

joheredi:
I think option 2 makes sense to start with. Assuming we can enhance it to support all paging scenarios. We can work afterwards to see if we want to optimize further by moving to option#3 but that would depend on bundle size analysis to decide if the gains are worth the effort.

MaryGao:
Take this as high priority considering the content safety would have paging operations. Previously we proposed two options and I have some concerns and I think we could have third one.

**Option 1: Similar to HLC's implementations**

This way we have to build everything by ourselves and I don't think that is necessary because we have [the getPagedAsyncIterator util](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-paging/src/getPagedAsyncIterator.ts#L14) in core-paging.

**Option 2: Leverage RLC's paginate helper**

[...]

PR #1978, opened 2023-08-18
Support modular paging

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1856


packages/typespec-ts/test/modularIntegration/generated/azure/core/src/api/operations.ts
```diff
@@ -260,12 +270,39 @@ export async function _listWithPageDeserialize(
...
 /** List with Azure.Core.Page<>. */
-export async function listWithPage(
+export function listWithPage(
   context: Client,
   options: ListWithPageOptions = { requestOptions: {} }
-): Promise<Page> {
-  const result = await _listWithPageSend(context, options);
-  return _listWithPageDeserialize(result);
+): PagedAsyncIterableIterator<Page> {
+  return listWithPageGenerator(context, options);
```
deyaaeldeen, line 277: I don't get why do we need to have the logic in a different function unless you plan on calling that function elsewhere too.

xirzec: my understanding is this is to make it simpler for customization -- trying to avoid having large amounts of code in an operation method to allow for simple wrappers and input/output transformations

MaryGao:
close and will reopen a new one

PR #1983, opened 2023-08-21
Support collection format in modular

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1984

When I implement the paging feature in modular I found the collect format is not supported and this would cause the compile error. So I need to fix it before [this pr](https://github.com/Azure/autorest.typescript/pull/1978).

- Enable the collection format test cases in cadl-ranch
- Handle the operation name conflicts


packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -1318,9 +1319,38 @@ function emitOperationGroups(
       operations: clientOperations
     });
   }
+  resolveConflictIfExist(operationGroups);
   return operationGroups;
 }
 
+function resolveConflictIfExist(operationGroups: OperationGroup[]) {
```
MaryGao, line 1326: Detect if there is name conflict cross operation groups, if yes, add operation group name prefix.

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -407,6 +417,22 @@ function getParameterMap(
   throw new Error(`Parameter ${param.clientName} is not supported`);
 }
 
+function getCollectionFormat(param: Parameter) {
```
MaryGao, line 420: Build the param with collection format helper function.

packages/typespec-ts/src/modular/buildCodeModel.ts
qiaozha: Can you add this check for RLC as well ? thx.

MaryGao: Done.

packages/typespec-ts/test/modularIntegration/generated/parameters/collection-format/src/models/models.ts
```diff
@@ -0,0 +1,2 @@
+// Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
MaryGao: Done

qiaozha (approved):

Issue #1957, opened 2023-08-07
Modular: Handle special header client-Request-Id


MaryGao:
The same issue https://github.com/Azure/autorest.typescript/issues/1906.

Below headers also need to be ingored, see issue https://github.com/Azure/typespec-azure/issues/3429

return-client-request-id
ocp-date
client-request-id

PR #1985, opened 2023-08-21
Support client request id modular and small fixes

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1957

- Ignore below headers in modular 
return-client-request-id
ocp-date
client-request-id

- Support `Array<Error>` and `Record<Error>`
- Add status check if no isUnexpected helper in RLC


packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -749,6 +750,9 @@ function emitBasicOperation(
   });
 
   for (const param of httpOperation.parameters.parameters) {
+    if (isCustomClientRequestIdParam(param)) {
```
MaryGao, line 753: In modular we just ignore this custom header parameter.

qiaozha: In the original issue, it also mentioned a return-client-request-id header, I don't see any logic is handling it in the response ?

packages/typespec-ts/src/transform/transformTelemetryInfo.ts
```diff
@@ -56,17 +60,23 @@ function getCustomRequestHeaderNameForClient(
   return undefined;
 }
 
+const CUSTOM_REQUEST_HEADER_NAME = "client-request-id";
```
qiaozha, line 63: Maybe put this inside a get function ?

packages/typespec-ts/test/modularIntegration/customClientRequestId.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+describe("RequestIdClient Classical Client", () => {
+  let client: RequestIdClient;
+
+  beforeEach(() => {
+    client = new RequestIdClient({
+      allowInsecureConnection: true
+    });
+  });
+
+  it("should add client-request-id in header transparently", async () => {
```
qiaozha, line 12: Can you add a test to pass the customizedHeaderName in the client constructor ?

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -477,16 +477,11 @@ type OptionalType = (Parameter | Property) & {
   type: { optional: true };
 };
 
-function isOptional(
-  param: Parameter | Property
-): param is OptionalType {
+function isOptional(param: Parameter | Property): param is OptionalType {
```
qiaozha, line 480: Is this because different prettier length options ? I've seen that kind of changes a lot.

packages/typespec-ts/src/modular/buildCodeModel.ts
qiaozha: Also, have we agreed to completely ignore this custom header in Modular ? what if customers set it via options.headers['client-request-id'] ?  And if customer has specify a diffrent header key than client-request-id, should we ignore that header key instead of this one ? 
Is it possible there are other headers that we want customers to ignore ? and client-request-id might be just one of them, are we going to have isXXXParam for all of them ?

packages/typespec-ts/test/modularIntegration/customClientRequestId.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+      assert.fail(err as string);
+    }
+  });
+
+  it("should override with x-test-client-request-id header", async () => {
+    try {
+      client = new RequestIdClient({
+        allowInsecureConnection: true,
+        telemetryOptions: {
+          clientRequestIdHeaderName: "x-test-request-id"
```
MaryGao, line 37: Blocked by this one https://github.com/Azure/autorest.typescript/issues/1987 and will fix it first.

packages/typespec-ts/test/modularIntegration/customClientRequestId.spec.ts
MaryGao: Added

packages/typespec-ts/test/modularIntegration/customClientRequestId.spec.ts
MaryGao: resoloved

packages/typespec-ts/src/modular/buildCodeModel.ts
MaryGao: I think the general guideline would be - the custom header would take the same role as original `x-ms-client-request-id`. If we ignore the latter we are supposed to ignore for custom one.

And I added test cases to demo on how to set other client header and how to override request id in modular. https://github.com/Azure/autorest.typescript/blob/bac92fb61ad9bd83938754139ada2935f3f881d3/packages/typespec-ts/test/modularIntegration/customClientRequestId.spec.ts

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -750,6 +751,9 @@ function emitBasicOperation(
   });
 
   for (const param of httpOperation.parameters.parameters) {
+    if (isIgnoredHeaderParam(param)) {
```
MaryGao, line 754: Three headers need to be ignored see https://github.com/Azure/typespec-azure/issues/3429

packages/typespec-ts/src/index.ts
```diff
@@ -166,9 +180,7 @@ export async function $onEmit(context: EmitContext) {
         buildModels(modularCodeModel, project, modularSourcesRoot, subfolder);
         buildModelsOptions(subClient, project, modularSourcesRoot, subfolder);
         const hasClientUnexpectedHelper =
-          needUnexpectedHelper.get(
-            subClient.rlcClientName.replace("Context", "Client")
-          ) ?? false;
+          needUnexpectedHelper.get(subClient.rlcClientName) ?? false;
```
MaryGao, line 183: Small fix: the set and get client name is inconsistancy in `needUnexpectedHelper`.

packages/typespec-ts/src/utils/operationUtil.ts
```diff
@@ -437,3 +449,133 @@ export function getCollectionFormatHelper(
...
+    await appendIfAbsent(route);
+  }
+}
+
+async function appendIfAbsent(route: HttpOperation) {
+  const errors = route.responses.filter((r) =>
+    isDefaultStatusCode(r.statusCode)
+  );
+  if (errors.length === 0) {
+    route.responses.push(FAKE_CORE_ERROR_RESPONSE as any);
```
MaryGao, line 491: There is no easy way to append the default error response so I  build a fake model to use if absent.

packages/typespec-ts/src/utils/operationUtil.ts
```diff
@@ -437,3 +449,133 @@ export function getCollectionFormatHelper(
...
+const _cache = new Map();
+export function getHttpOperationWithCache(
+  program: Program,
+  operation: Operation
+): [HttpOperation, readonly Diagnostic[]] {
+  const existing = _cache.get(operation);
+  if (existing) {
+    return existing;
+  }
+  const httpOperationRef = getHttpOperation(program, operation);
```
MaryGao, line 549: Every time `getHttpOperation` would build a new http operation according to original one so we need to cachec them otherwise the appending info would be lost.

.vscode/launch.json
```diff
@@ -98,7 +98,7 @@
       "runtimeArgs": [
         "ts-node",
         "./test/commands/gen-cadl-ranch.ts",
-        "--tag=modular",
+        "--tag=rlc",
```
MaryGao, line 101:
```suggestion
        "--tag=modular",
```

packages/typespec-ts/src/index.ts
```diff
@@ -68,15 +70,17 @@ export async function $onEmit(context: EmitContext) {
   >();
   const rlcCodeModels: RLCModel[] = [];
   let modularCodeModel: ModularCodeModel;
-  // 1. Enrich the dpg context with path detail and common options
+  // 1. Mutate context
```
MaryGao, line 73:
```suggestion
  // 1. Mutate context 
```

packages/typespec-test/test/healthinsight/generated/typespec-ts/src/outputModels.ts
```diff
@@ -18,33 +20,11 @@ export interface TrialMatcherResultOutput {
    */
   readonly status: string;
   /** An array of errors, if any errors occurred during the processing job. */
-  readonly errors?: Array<ErrorModelOutput>;
+  readonly errors?: Array<ErrorModel>;
```
MaryGao, line 23: Small fix: if the error type is from core we would directly re-use the one in js core.

packages/typespec-ts/src/utils/operationUtil.ts
```diff
@@ -437,3 +449,133 @@ export function getCollectionFormatHelper(
...
+    name: "ErrorResponseBase"
+  }
+};
+const FAKE_CORE_ERROR_RESPONSE = {
+  statusCode: "*",
+  type: FAKE_CORE_ERROR_RESPONSE_TYPE,
+  responses: [
+    {
+      body: {
+        contentTypes: ["application/json"],
```
qiaozha, line 533: Is it going to be problematic if this operation doesn't have "application/json" content type at all ?

packages/typespec-ts/src/utils/operationUtil.ts
```diff
@@ -437,3 +449,133 @@ export function getCollectionFormatHelper(
...
+async function appendIfAbsent(route: HttpOperation) {
+  const errors = route.responses.filter((r) =>
+    isDefaultStatusCode(r.statusCode)
+  );
+  if (errors.length === 0) {
+    route.responses.push(FAKE_CORE_ERROR_RESPONSE as any);
+  }
+}
+
+export async function createRLCEmitterTestHost() {
```
qiaozha, line 495: kind of don't get why have this in here ?

packages/typespec-ts/src/utils/operationUtil.ts
```diff
@@ -437,3 +449,133 @@ export function getCollectionFormatHelper(
...
+    libraries: [
+      HttpTestLibrary,
+      RestTestLibrary,
+      VersioningTestLibrary,
+      AzureCoreTestLibrary,
+      SdkTestLibrary
+    ]
+  });
+}
+const FAKE_CORE_ERROR_RESPONSE_TYPE = {
```
qiaozha, line 506: Is there any type from typespec compiler we can reuse ? Also, why can't we add the default response after the transform ? Feeling this way is depending too much on the typesspec compiler details, which may cause unnecessary breaking changes for us.

packages/typespec-ts/src/transform/transformTelemetryInfo.ts
```diff
@@ -37,7 +40,7 @@ function getCustomRequestHeaderNameForClient(
     );
     for (const op of operations) {
       const headerName = getCustomRequestHeaderNameForOperation(
-        ignoreDiagnostics(getHttpOperation(program, op))
+        ignoreDiagnostics(getHttpOperationWithCache(program, op))
```
qiaozha, line 43: Can you point to me where is getCustomRequestHeaderNameForClient being called ? thx.

packages/typespec-ts/src/transform/transformTelemetryInfo.ts
```diff
@@ -37,7 +40,7 @@ function getCustomRequestHeaderNameForClient(
...
       );
       if (headerName != undefined) {
```
qiaozha, line 45: !== ？

packages/typespec-ts/src/transform/transformTelemetryInfo.ts
```diff
@@ -47,26 +50,11 @@ function getCustomRequestHeaderNameForClient(
   const clientOperations = listOperationsInOperationGroup(dpgContext, client);
   for (const clientOp of clientOperations) {
     const headerName = getCustomRequestHeaderNameForOperation(
-      ignoreDiagnostics(getHttpOperation(program, clientOp))
+      ignoreDiagnostics(getHttpOperationWithCache(program, clientOp))
     );
     if (headerName != undefined) {
```
qiaozha, line 55: same here

packages/rlc-common/src/buildObjectTypes.ts
```diff
@@ -417,18 +421,31 @@ export function getPropertySignature(
...
-      ? property.outputTypeName
-      : property.typeName
-      ? property.typeName
-      : property.type;
-  if (property.typeName && property.fromCore) {
-    importedModels.add(property.typeName);
+  let type;
+  if (
+    property.type === "array" &&
+    (property as any).items &&
```
lirenhe, line 428: It would be good to leverage the actual type for the calculation to avoid using 'as any'.
I think this is another area we shall improve our codebase.

packages/typespec-ts/test/integration/encodeDatetime.spec.ts
```diff
@@ -98,7 +98,10 @@ describe("EncodeDatetimeClient Rest Client", () => {
             }
           });
         assert.strictEqual(result.status, "200");
-        assert.strictEqual(result.body.value, "2022-08-26T18:38:00.000Z");
+        assert.strictEqual(
+          (result as any).body.value,
```
qiaozha, line 102: why need to use as any here ?

packages/typespec-ts/test/modularIntegration/generated/client/structure/default/src/api/operations.ts
```diff
@@ -26,13 +33,19 @@ import {
...
   context: Client,
   options: OneOptions = { requestOptions: {} }
-): StreamableMethod<One204Response> {
+): StreamableMethod<One204Response | OneDefaultResponse> {
   return context
     .path("/one")
     .post({ ...operationOptionsToRequestParameters(options) });
 }
 
-export async function _oneDeserialize(_result: One204Response): Promise<void> {
```
qiaozha, line 35: I remeber there's a logic handing the rename from result to _result, can you remove this logic in the codegen if default response always exists.

packages/typespec-ts/test/modularIntegration/collectionFormat.spec.ts
```diff
@@ -33,6 +38,7 @@ describe("CollectionFormatClient Classical Client", () => {
       const result = await client.query.pipes(["blue", "red", "green"]);
       assert.strictEqual(result, undefined);
     } catch (err) {
+      console.log(err);
```
qiaozha, line 41: same, debug ?

packages/typespec-ts/test/modularIntegration/customClientRequestId.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  beforeEach(() => {
+    client = new RequestIdClient({
+      allowInsecureConnection: true
+    });
+  });
+
+  it("should add client-request-id in header transparently", async () => {
+    try {
+      const result = await client.get();
+      assert.isUndefined(result);
```
qiaozha, line 16: why this should be undefined ? is it a normal call to cadl-ranch ?

packages/typespec-ts/test/modularIntegration/customClientRequestId.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    } catch (err) {
+      console.log(err);
+      assert.fail(err as string);
+    }
```
qiaozha, line 20: 1

packages/typespec-ts/test/modularIntegration/customClientRequestId.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+      });
+
+      const result = await client.get({
+        requestOptions: {
+          headers: {
+            "client-request-id": overrideId
+          }
+        }
+      });
+      assert.isUndefined(result);
```
qiaozha, line 50: Same here, is it not a normal request that should return successfully ?

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -1255,3 +1247,20 @@ export function trimUsage(model: any) {
   }, {});
   return ordered;
 }
+
+export function isAzureCoreErrorType(t?: Type): boolean {
+  if (
+    t?.kind !== "Model" ||
+    !["Error", "ErrorResponse", "InnerError", "ErrorModel"].includes(t.name)
```
lirenhe, line 1254: shall we consider this comparison to be case-insensitive?  I double whether people would write sth. like 'Errormodel'

packages/rlc-common/src/buildObjectTypes.ts
```diff
@@ -419,16 +423,29 @@ export function getPropertySignature(
...
-      ? property.typeName
-      : property.type;
-  if (property.typeName && property.fromCore) {
-    importedModels.add(property.typeName);
+  let type;
+  if (
+    property.type === "array" &&
+    (property as any).items.fromCore &&
+    property.typeName
+  ) {
```
qiaozha, line 431: Consider the case of Record<Error>  ?

qiaozha (commented):
Please remove any consolde.log code if it's debugging. thx

packages/typespec-ts/test/modularIntegration/customClientRequestId.spec.ts
MaryGao: This is a normal call to cadl-ranch and its validation is for client request id in header. And the return type in modular would be void. If this call could be return directly without any exception and the validation passed.

packages/rlc-common/src/buildObjectTypes.ts
MaryGao: Yes, record is a valid case we could add it.

packages/typespec-ts/src/utils/operationUtil.ts
MaryGao: removed

packages/typespec-ts/src/transform/transformApiVersionInfo.ts
```diff
@@ -67,7 +67,7 @@ function getOperationQueryApiVersion(
       operationGroup
     );
     for (const op of operations) {
-      const route = ignoreDiagnostics(getHttpOperation(program, op));
+      const route = ignoreDiagnostics(getHttpOperationWithCache(program, op));
```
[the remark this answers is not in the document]
MaryGao: Every time `getHttpOperation` would build a new http operation according to original one so we need to cache them otherwise the appending info would be lost.

packages/typespec-ts/src/utils/operationUtil.ts
```diff
@@ -437,3 +443,122 @@ export function getCollectionFormatHelper(
...
+}
+
+export function getCustomRequestHeaderNameForOperation(
+  route: HttpOperation
+): string | undefined {
+  const params = route.parameters.parameters.filter(
+    isCustomClientRequestIdParam
+  );
+  if (params.length > 0) {
+    return "client-request-id";
```
lirenhe, line 544: Can we define the const value for it?

packages/typespec-ts/src/utils/operationUtil.ts
MaryGao: Please notice that we will not specify the content-type as `"application/json"` in generated code because this is our fallback content type in core. This content type is more like a bypass option to let rlc not generate any content type info for this operation.

But I do re-think that generating the `ErrorModel` from core is correct or not, if the server side throws any object with their own styles that would cause issues.

packages/typespec-ts/src/utils/operationUtil.ts
MaryGao: I don't think we have any in compiler side. If we do this after transformation we need to change both paths and responses, not sure somewhere also. I'd like to keep one place for this change.

packages/typespec-ts/src/utils/operationUtil.ts
MaryGao: aha, it seems conflict in comments https://github.com/Azure/autorest.typescript/pull/1985#discussion_r1300858206

packages/rlc-common/src/buildObjectTypes.ts
MaryGao: I created an issue as our MQ item https://github.com/Azure/autorest.typescript/issues/1993

packages/rlc-common/src/buildObjectTypes.ts
MaryGao: Added in UT.

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -128,6 +120,24 @@ export function getDeserializePrivateFunction(
...
+  } else {
+    const validStatus = [
+      ...new Set(
+        operation.responses
+          .flatMap((r) => r.statusCodes)
+          .filter((s) => s !== "default")
+      )
+    ];
+
+    if (validStatus.length > 0) {
```
MaryGao, line 132: Added status validation if no isUnexpected helper.

packages/typespec-ts/test/integration/encodeDatetime.spec.ts
```diff
@@ -129,6 +129,7 @@ describe("EncodeDatetimeClient Rest Client", () => {
               value: "Fri, 26 Aug 2022 14:38:00 GMT"
             }
           });
+
```
MaryGao, line 132:
```suggestion
```

packages/typespec-ts/test/unit/modelsGenerator.spec.ts
```diff
@@ -1801,4 +1806,186 @@ describe("Input/output model type", () => {
       });
     });
   });
+
+  describe("core error model", () => {
```
MaryGao, line 1810: Add test cases for error model

packages/typespec-ts/test/modularIntegration/collectionFormat.spec.ts
```diff
@@ -21,9 +21,14 @@ describe("CollectionFormatClient Classical Client", () => {
 
   it("should send multi format in query", async () => {
     try {
-      const result = await client.query.multi(["blue", "red", "green"]);
+      const result = await client.query.multi(["blue", "red", "green"], {
+        requestOptions: {
+          skipUrlEncoding: true
+        }
+      });
```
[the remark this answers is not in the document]
MaryGao: This case is not working before because we have no status check so the exception didn't happen.

packages/typespec-ts/test/integration/encodeDatetime.spec.ts
```diff
@@ -179,6 +179,7 @@ describe("EncodeDatetimeClient Rest Client", () => {
...
```
MaryGao, line 182:
```suggestion
```

packages/typespec-ts/test/integration/encodeDatetime.spec.ts
```diff
@@ -194,6 +194,7 @@ describe("EncodeDatetimeClient Rest Client", () => {
               value: "2022-08-26T18:38:00.000Z"
             }
           });
+
```
MaryGao, line 197:
```suggestion
```

packages/typespec-ts/test/integration/encodeDatetime.spec.ts
```diff
@@ -209,6 +209,7 @@ describe("EncodeDatetimeClient Rest Client", () => {
               value: "Fri, 26 Aug 2022 14:38:00 GMT"
             }
           });
+
```
MaryGao, line 212:
```suggestion
```

packages/typespec-ts/test/integration/encodeDatetime.spec.ts
```diff
@@ -224,6 +224,7 @@ describe("EncodeDatetimeClient Rest Client", () => {
               value: 1686566864
             }
           });
+
```
MaryGao, line 227:
```suggestion
```

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -128,6 +120,24 @@ export function getDeserializePrivateFunction(
...
+      statements.push(
+        `if(${validStatus
+          .map((s) => `"${s}" !== result.status`)
```
qiaozha, line 135:
```suggestion
          .map((s) => `result.status !== "${s}"`)
```
looks weird to put the status code in left side. but no strong opinions here.

qiaozha (approved):
LGTM

Issue #1939, opened 2023-07-26
Support final-state-schema in swagger


MaryGao:
Blocked by the issue: https://github.com/Azure/autorest/issues/4762.

Service swagger: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/networkcloud/resource-manager/Microsoft.NetworkCloud/preview/2023-05-01-preview/networkcloud.json#L881

After the autorest core fixes this issue we need to:
- Upgrade m4 and autorest core version
- Add test cases

MaryGao:
blocking the release request https://github.com/Azure/sdk-release-request/issues/4362

PR #1988, opened 2023-08-22
Upgrade the autorest core version in smoke and int test

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1939


packages/autorest.typescript/test/smoke/generated/networkcloud-resource-manager/src/models/index.ts
```diff
@@ -0,0 +1,6251 @@
...
+  /** The request body. */
+  bareMetalMachineCordonParameters?: BareMetalMachineCordonParameters;
+  /** Delay to wait until next poll, in milliseconds. */
+  updateIntervalInMs?: number;
+  /** A serialized poller which can be used to resume an existing paused Long-Running-Operation. */
+  resumeFrom?: string;
+}
+
+/** Contains response data for the cordon operation. */
+export type BareMetalMachinesCordonResponse = OperationStatusResult;
```
MaryGao, line 4953: With this upgrading we could generate the model in final-state-schema that is `OperationStatusResult`.

qiaozha (approved):

PR #1996, opened 2023-08-29
Support modular paging

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/1856

This is the initial version for modular paging. The whole discussion is under above issue.

- Basic paging support
- Disable paging & LRO for non-branded RLC/Modular

We build paging helpers and types in SDK side and keep the similar interface with HLC paging with slightly differences:
- Remove the `maxpagesize` in PageSetting
- Change the return type for byPage to `TPage & {
        continuationToken?: string;
    }`

Advanced paging cases will be covered in another issue https://github.com/Azure/autorest.typescript/issues/2134.


packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/review/ai-content-safety.api.md
```diff
@@ -129,6 +133,9 @@ export interface PagedTextBlocklist {
 export interface RemoveBlockItemsRequestOptions extends OperationOptions {
 }
 
+// @public (undocumented)
+export function setContinuationToken(page: unknown, continuationToken: string | undefined): void;
```
MaryGao, line 137: @xirzec I notice [your comment](https://github.com/Azure/autorest.typescript/issues/1199#issuecomment-1289735138) in original issue, do you think it's better to re-design below two helpers or just expose the same in modular?

Currently we just expose the get(set)ContinuationToken.

https://github.com/Azure/autorest.typescript/blob/9642e55f45349bb18b20fc2b026b62c1069b6663/packages/typespec-ts/test/modularIntegration/pageable.spec.ts#L60-L79

packages/typespec-ts/test/modularIntegration/pageable.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    // initiate another iterator starting with 2nd page
+    const continuationToken = getContinuationToken(firstPage.value);
+    const items: User[] = [];
+    for await (const pagedUsers of iter.byPage({ continuationToken })) {
+      items.push(...pagedUsers);
+    }
+    assert.strictEqual(items.length, 1);
+  });
+
+  it("maxPageSize is not allowed and should throw exceptions", async () => {
```
MaryGao, line 81: The `maxPageSize` setting in `byPage` is not supported and will throw exceptions directly.

packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/review/ai-content-safety.api.md
xirzec: I support us figuring out a better abstraction here and perhaps getting rid of the 'maxPageSize' param as well -- maybe we need to rethink the core-paging contract for modular?

@deyaaeldeen - what do you think? Is this borrowing too much trouble to fix now? I feel like if we don't tackle it for the new DPG codegen we're going to kick this can forever down the road.

MaryGao: @xizec regarding getting rid of this `maxPageSize` parameter in core, my understanding is if the relevant tsp defines the `maxPageSize` as query parameter, we should also respect this parameter in codegen side and so that customers could also set their own values, right?

```typescript
const iter = client.list({ 
       maxpagesize: 3 
});
```

deyaaeldeen: > @deyaaeldeen - what do you think? Is this borrowing too much trouble to fix now? I feel like if we don't tackle it for the new DPG codegen we're going to kick this can forever down the road.

That is a good point! I gave this some thought and I wonder if [this proposal](https://www.typescriptlang.org/play?target=99#code/JYOwLgpgTgZghgYwgAgApwOYQMoTGUDAZ2QG8AoZZBAe3FAFc4C6AVGgawhAH4AuZETBRCAbkrIAtnAAe6LNmAAvCP2QgGkgEbRxAX3LlQkWIhTzQzCABMAgkQCeIBAEkTcLQBsIAHgmsAUW9JbjByAD5kCBlIEGsSeydXdy9fQODQyLIJLQd5CAEfVnzkAF5kdIgQ8ABtAF0AGgr83HxCEnKWvAIQYnCACiJu9rVizBxh3qIASjLIxOc3aA9vH1JkAAdxgTGsJtp6DWZgNk5uASERXuQ9cP0gA) makes sense. The paging type variables get relegated to `byPage` itself emphasizing that it is not a core scenario to use `byPage` and assuming most scenarios don't require to know the paging details. It also updates the `byPage`'s return type to return both the page and the continuation token, making the relationship between both explicit and close by.

qiaozha: I think we can safely remove maxPageSize if it's not working currently and it should be considered as non-breaking. 

For continuationToken, from the HLC  
image
 It seems to me continuationToken = nextLink, and the setContinuationToken 
https://github.com/qiaozha/azure-sdk-for-js/blob/70507137e6626c4ed33d72bb59ba109ce0b9dc02/sdk/advisor/arm-advisor/src/pagingHelper.ts#L13 
just stores the continuationToken for each page data, I wonder if that's necessary? Is there a real scenario that customers want to getContinuationToken from where they have list those data before. If they want, they could just pass the nextLink to byPage as the continuationToken?

MaryGao: @qiaozha the requirement of `getContinuationToken` is coming from customers' ask and you could refer [their code here](https://github.com/Azure/autorest.typescript/issues/1199#issuecomment-1310817741). The intention is because we hide the nextLink detail in our public API so we provide two utilities to help them extract the underlying tokens in byPage iterator.

@deyaaeldeen I like the idea of the return type change for `byPage` and compared with current helper way and we could have two benefits:
- page items and its continuationToken are highly coupled so that we could encapsulate them in one place
- this could be easier to discover and use for customers

The drawback for relegating paging type variables is that we can't pass customized `PageSettings` when using getPagedAsyncIterator. These types could only be passed during calling byPage, which means they can't be used in codegen/SDK side but only for customers.

I think the `maxPageSize` issue still exists and directly throwing exception is not acceptable for me. Could you help me understand if getting ride of this in core would be chanllenging or not? Because I notice this param could be valuable for other downstreams. e.g [service bus](https://github.com/azure/azure-sdk-for-js/blob/61be11030937904bdc689fd4d0384e53ea250b33/sdk/servicebus/service-bus/src/serviceBusRuleManager.ts#L218-L219).

deyaaeldeen: > I think the maxPageSize issue still exists and directly throwing exception is not acceptable for me

I think we can take a code gen approach to supporting `maxPageSize` instead of always including it. If the API supports a `top` parameter (designated as such by a typespec decorator?), then generate a `maxPageSize` option and use it to set `top`. If the API doesn't, then don't generate such a parameter.

> The drawback for relegating paging type variables is that SDK would lose the ability to pass customized PageSettings when using getPagedAsyncIterator. These types could only be passed during calling byPage and only for customers.

If we're using the code gen approach to `PageSettings` described above, this shouldn't be an issue because `PageSettings` will no longer be a type variable. We can also use a code gen approach to the `TPage` type and get ride of the type variable as well. So for an API that doesn't have a `top` and has a page type of `Collection`, code gen can generate the following:
```ts
interface PageSettings {
  continuationToken?: string;
}

interface PaginatedAsyncIterable<
  TElement
> extends AsyncIterable<TElement>  {
  byPage: (settings?: PageSettings) => AsyncIterable<{ page: Collection, continuationToken: string }>;
}
```

MaryGao: Yeah I think we could take a code gen approach. Regarding the `maxPageSize` I would prefer to get rid of this in PageSettings with following reasons:

- We already expose the option `maxPageSize` in operation-level option which means we could 1) either duplicate it or 2) remove the operation one. The former would introduce confusion and the latter would cause real breakings for HLC.
- On the other hand just getting rid of this causes the API surface breaking but it would be safe considering the inner is directly throwing exception. And all operations could share the same `PageSettings` 
- There is no clear mapping from the real option to maxPageSize in tsp, it depends on the service team's implementation. I know some services use the `top`, some are `maxresults` and others use `maxpagesize`

packages/typespec-ts/test/commands/cadl-ranch-list.ts
```diff
@@ -225,11 +225,13 @@ export const modularTsps: TypeSpecRanchConfig[] = [
   },
   {
     outputPath: "azure/core",
-    inputPath: "azure/core/basic"
+    inputPath: "azure/core/basic",
+    debug: true
```
MaryGao, line 229:
```suggestion
```

packages/typespec-ts/test/commands/cadl-ranch-list.ts
```diff
@@ -225,11 +225,12 @@ export const modularTsps: TypeSpecRanchConfig[] = [
   },
   {
     outputPath: "azure/core",
-    inputPath: "azure/core/basic"
+    inputPath: "azure/core/basic",
```
MaryGao, line 228:
```suggestion
    inputPath: "azure/core/basic"
```

packages/typespec-ts/test/commands/cadl-ranch-list.ts
```diff
@@ -228,8 +228,9 @@ export const modularTsps: TypeSpecRanchConfig[] = [
     inputPath: "azure/core/basic"
   },
   {
-    outputPath: "encode/bytes",
-    inputPath: "encode/bytes"
+    outputPath: "payload/pageable",
+    inputPath: "payload/pageable",
+    debug: true
```
MaryGao, line 233:
```suggestion
```

packages/typespec-ts/test/commands/cadl-ranch-list.ts
```diff
@@ -228,8 +228,8 @@ export const modularTsps: TypeSpecRanchConfig[] = [
     inputPath: "azure/core/basic"
   },
   {
-    outputPath: "encode/bytes",
-    inputPath: "encode/bytes"
+    outputPath: "payload/pageable",
+    inputPath: "payload/pageable",
```
MaryGao, line 232:
```suggestion
    inputPath: "payload/pageable"
```

packages/typespec-test/test/batch_modular/generated/typespec-ts/review/batch.api.md
```diff
@@ -161,22 +161,22 @@ export class BatchClient {
...
+    listJobs(options?: ListJobsOptions): PagedAsyncIterableIterator<BatchJob>;
+    listJobSchedules(options?: ListJobSchedulesOptions): PagedAsyncIterableIterator<BatchJobSchedule>;
+    listJobsFromSchedule(jobScheduleId: string, options?: ListJobsFromScheduleOptions): PagedAsyncIterableIterator<BatchJob>;
+    listNodeExtensions(poolId: string, nodeId: string, options?: ListNodeExtensionsOptions): PagedAsyncIterableIterator<NodeVMExtension>;
+    listNodeFiles(poolId: string, nodeId: string, options?: ListNodeFilesOptions): PagedAsyncIterableIterator<NodeFile>;
+    listNodes(poolId: string, options?: ListNodesOptions): PagedAsyncIterableIterator<BatchNode>;
+    listPoolNodeCounts(options?: ListPoolNodeCountsOptions): PagedAsyncIterableIterator<PoolNodeCounts>;
+    listPools(options?: ListPoolsOptions): PagedAsyncIterableIterator<BatchPool>;
+    listPoolUsageMetrics(options?: ListPoolUsageMetricsOptions): PagedAsyncIterableIterator<PoolUsageMetrics>;
     listSubTasks(jobId: string, taskId: string, options?: ListSubTasksOptions): Promise<BatchTaskListSubtasksResult>;
```
lirenhe, line 176: Just curious why this is not updated?  Is it a paging operation?

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/models/pagingTypes.ts
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/**
+ * An interface that tracks the settings for paged iteration
+ */
+export interface PageSettings {
```
lirenhe, line 7: Is this file the same across all services?  How about moving this into the common library?

packages/typespec-test/test/batch_modular/generated/typespec-ts/review/batch.api.md
MaryGao: I checked this will not be devired as paging operation and I will confirm with service team any reason not do that.

lirenhe (commented):
I didn't find the change of unbranding projects, so does this change only apply for branding projects?

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/models/pagingTypes.ts
MaryGao: Currently we have same interface name in paging-core https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-paging/src/models.ts#L7. But we removed the `maxpagesize` considering it is not supported(see [dicusssion here](https://github.com/Azure/autorest.typescript/issues/1856#issuecomment-1711832806)).

We plan to add them to core when this version is stable in future.

MaryGao:
> I didn't find the change of unbranding projects, so does this change only apply for branding projects?

I think the ts-http-runtime lib is ready to support paging and I will disable the paging and lro support for non branding.  @xirzec Feel free to share your thoughts!

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -248,38 +249,69 @@ export function getOperationFunction(
...
+    returnType = buildType(type.name, type, type.format);
+  }
   const { name, fixme = [] } = getOperationName(operation);
   const functionStatement: OptionalKind<FunctionDeclarationStructure> = {
     docs: [
       ...getDocsFromDescription(operation.description),
       ...getFixmeForMultilineDocs(fixme)
     ],
-    isAsync: true,
+    isAsync: !isPaging,
```
lirenhe, line 270: Can the paging call be an LRO?

MaryGao: First there is no async/await pair, second the paging method is a sync method to get a async iter. So we don't need async label.

packages/typespec-test/test/batch_modular/generated/typespec-ts/review/batch.api.md
```diff
@@ -1645,6 +1645,20 @@ export interface OutputFileUploadOptions {
     uploadCondition: OutputFileUploadCondition;
 }
 
+// @public
+export interface PagedAsyncIterableIterator<TElement, TPage = TElement[], TPageSettings = PageSettings> {
```
deyaaeldeen, line 1649: Shouldn't this be referenced from core-paging?

deyaaeldeen: nvm, I see that the page settings definition is slightly different.

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/api/pagingHelpers.ts
```diff
@@ -0,0 +1,232 @@
...
+  : Array<unknown>;
+
+export function buildPagedAsyncIterator<
+  TElement,
+  TResponse extends PathUncheckedResponse = PathUncheckedResponse
+>(
+  client: Client,
+  initialSendFunction: (...args: any[]) => PromiseLike<TResponse>,
+  deserializeFunction: (result: TResponse) => Promise<unknown>,
+  sendFunctionArgs: any[] = []
```
deyaaeldeen, line 48: Instead of passing both `initialSendFunction` and `sendFunctionArgs`, we could pass a `getInitialResponse` of type `() => PromiseLike<TResponse>` and instantiate it as `() => initialSendFunction(...sendFunctionArgs)`.

Also, `deserializeFunction` doesn't seem to be named meaningfully, we know the client will already do basic deserialization on the response before this function is called. Perhaps we could name it `processResponseBody`? I am open to alternatives as well.

packages/typespec-test/test/batch_modular/generated/typespec-ts/review/batch.api.md
```diff
@@ -1645,6 +1645,20 @@ export interface OutputFileUploadOptions {
...
+    [Symbol.asyncIterator](): PagedAsyncIterableIterator<TElement>;
+    byPage: (settings?: TPageSettings) => AsyncIterableIterator<TPage & {
+        continuationToken?: string;
+    }>;
```
MaryGao, line 1653: I gave up to name this return type with a specific name like `ContinuablePage` because I found directly using `TPage & { continuationToken?: string; }` could be easier to discover the real type e.g User[] & { continuationToken?: string; } and they don't need to click into the definition.

```ts
export type ContinuablePage<TPage> = TPage & { continuationToken?: string };
```

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/api/pagingHelpers.ts
```diff
@@ -0,0 +1,210 @@
...
+      return {
+        page: values,
+        nextPageLink: nextLink,
+      };
+    },
+    byPage: (settings?: PageSettings) => {
+      const { continuationToken } = settings ?? {};
+      return getPageAsyncIterator(pagedResult, {
+        pageLink: continuationToken,
+      }) as any;
```
deyaaeldeen, line 57: why is this cast needed?

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/models/pagingTypes.ts
xirzec: Yeah, we could consider doing an update to remove `maxPageSize` and make any other needed breaking adjustments in the new core and align to that -- @deyaaeldeen were there other changes we had thought about? I dimly recall us having regrets over the generic parameters?

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/api/pagingHelpers.ts
```diff
@@ -0,0 +1,210 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import {
```
xirzec, line 4: if the plan is to diverge from core and re-align, maybe we should consider moving these helpers into the package temporarily so we don't have to cast between two similar types?

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/api/pagingHelpers.ts
```diff
@@ -0,0 +1,210 @@
...
+      return iter.next();
+    },
+    [Symbol.asyncIterator]() {
+      return this;
+    },
+    byPage: (settings?: PageSettings) => {
+      return iter.byPage(
+        settings as CorePageSettings
+      ) as unknown as AsyncIterableIterator<
+        TElement[] & { continuationToken?: string }
```
xirzec, line 74: I wonder if it would be better for us to standardize on the page always being an object to make it more coherent when there are other page-level properties (e.g. quota remaining and batch metadata)

Perhaps something like

```ts
{
  values: TElement[];
  continuationToken? string;
}
```

Alternatively, if we didn't want to pin the values to being an array, we could also consider making this page object itself be an iterator e.g.

```ts
Iterable<TElement> &
{
  continuationToken? string;
}
```

packages/typespec-ts/test/commands/cadl-ranch-list.ts
```diff
@@ -228,8 +228,8 @@ export const modularTsps: TypeSpecRanchConfig[] = [
     inputPath: "azure/core/basic"
   },
   {
-    outputPath: "encode/bytes",
-    inputPath: "encode/bytes"
+    outputPath: "payload/pageable",
```
qiaozha, line 231: why remove encode bytes test case ?

qiaozha: Also, we probably should clear the generated folder in the CI before we run the generate-tsp-only script as the cadl-ranch-list change is not being correctly reflected.

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/api/pagingHelpers.ts
MaryGao: I duplicated the references in core so the castcading is no need now.

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/api/pagingHelpers.ts
MaryGao: Both definitions work for me, but I updated the change to the following because it would not introduce any breakings for HLC.

```ts
export type ContinuablePage<TElement, TPage = TElement[]> = TPage & {
  /**
   * The token that keeps track of where to continue the iterator
   */
  continuationToken?: string;
};
```

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/api/pagingHelpers.ts
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  return value ?? [];
+}
+
+/**
+ * Checks if a request failed
+ */
+function checkPagingRequest(response: PathUncheckedResponse): void {
+  const Http2xxStatusCodes = [
```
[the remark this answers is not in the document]
MaryGao: I prefer to keep this https://restfulapi.net/http-status-codes/#2xx.

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/api/pagingHelpers.ts
```diff
@@ -0,0 +1,210 @@
...
+
+/**
+ * Extracts the itemName and nextLinkName from the initial response to use them for pagination
+ */
+function getPaginationProperties(initialResponse: PathUncheckedResponse) {
+  // Build a set with the passed custom nextLinkNames
+  const nextLinkNames = new Set(["nextLink", "odata.nextLink"]);
+
+  // Build a set with the passed custom set of itemNames
+  const itemNames = new Set(["value", "items"]);
```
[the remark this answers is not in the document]
MaryGao: I added one in widget_dpg, https://github.com/Azure/autorest.typescript/pull/1996/commits/e090a70ee4e095a9f23cf89087e7d7446f907a1c.

For multi-client we have case in loadtesting_modular.

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/api/pagingHelpers.ts
```diff
@@ -0,0 +1,210 @@
...
+  client: Client,
+  getInitialResponse: () => PromiseLike<TResponse>,
+  processResponseBody: (result: TResponse) => Promise<unknown>
+): PagedAsyncIterableIterator<TElement> {
+  let firstRun = true;
+  let itemName: string, nextLinkName: string | undefined;
+  const firstPageLinkPlaceholder = "";
+  const pagedResult: PagedResult<TElement[]> = {
+    firstPageLink: firstPageLinkPlaceholder,
+    getPage: async (pageLink: string) => {
```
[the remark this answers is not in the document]
MaryGao: It will be coverred in another issue https://github.com/Azure/autorest.typescript/issues/2134. Currently I am not clear if we could define them in typespec.

But for us it could be easy to implement by adding optional config to specify the calling.

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/api/pagingHelpers.ts
```diff
@@ -0,0 +1,210 @@
...
+      `Pagination failed with unexpected statusCode ${response.status}`,
+      response
+    );
+  }
+}
+
+/**
+ * Extracts the itemName and nextLinkName from the initial response to use them for pagination
+ */
+function getPaginationProperties(initialResponse: PathUncheckedResponse) {
```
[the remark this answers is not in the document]
MaryGao: I removed this helper and adding the options for `buildPagedAsyncIterator` to pass the itemName and nextLink pair and I think this way we would not cause any confusion if service side mismatched.

packages/typespec-ts/src/utils/operationUtil.ts
```diff
@@ -530,3 +530,18 @@ export function isIgnoredHeaderParam(param: HttpOperationParameter) {
...
   );
 }
+
+export function parseNextLinkName(
+  paged: PagedResultMetadata
+): string | undefined {
+  return paged.nextLinkProperty?.name;
+}
+
+export function parseItemName(paged: PagedResultMetadata): string | undefined {
```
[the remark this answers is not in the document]
MaryGao: The nested item name path case is not supported see issue https://github.com/Azure/autorest.typescript/issues/2134. So we have a common helper to extract the data temporarily.

packages/typespec-ts/test/commands/cadl-ranch-list.ts
MaryGao: > why remove encode bytes test case ?

I will revert this change and it is introduced wrongly because in latest cadl-ranch the cases are changed for encode/bytes.

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/api/pagingHelpers.ts
```diff
@@ -0,0 +1,210 @@
...
+  TElement,
+  TResponse extends PathUncheckedResponse = PathUncheckedResponse
+>(
+  client: Client,
+  getInitialResponse: () => PromiseLike<TResponse>,
+  processResponseBody: (result: TResponse) => Promise<unknown>
+): PagedAsyncIterableIterator<TElement> {
+  let firstRun = true;
+  let itemName: string, nextLinkName: string | undefined;
+  const firstPageLinkPlaceholder = "";
```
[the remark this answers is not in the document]
MaryGao: I refactored this logic a little because now we will build them by scratch not referenced the core-paging helpers.

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/api/pagingHelpers.ts
```diff
@@ -0,0 +1,210 @@
...
+ * Gets for the value of nextLink in the body
+ */
+function getNextLink(body: unknown, nextLinkName?: string): string | undefined {
+  if (!nextLinkName) {
+    return undefined;
+  }
+
+  const nextLink = (body as Record<string, unknown>)[nextLinkName];
+
+  if (typeof nextLink !== "string" && typeof nextLink !== "undefined") {
```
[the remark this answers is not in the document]
MaryGao: previously we used the core-paing helper so the firstPage is required with string. So I put a placeholder for that value, but I think this could be changed because Jeff suggested us to build our own helper to avoid cascadings.

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -248,38 +249,69 @@ export function getOperationFunction(
...
-    ? buildType(response.type.name, response.type, response.type.format)
-    : { name: "", type: "void" };
-
+  let returnType = { name: "", type: "void" };
+  if (response.type?.type) {
+    let type = response.type;
+    if (isPaging) {
+      type = extractPagingType(type, operation.itemName) ?? type;
+    }
+    returnType = buildType(type.name, type, type.format);
```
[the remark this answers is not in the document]
MaryGao: I may not get your point here, currently if there is no `@items` defined we would not generate paging for them. Because we don't know the detailed item type.

packages/typespec-ts/test/modularUnit/operations.spec.ts
```diff
@@ -522,4 +522,130 @@ describe("operations", () => {
...
+                    () => _testSend(context, options),
+                    _testDeserialize,
+                    {itemName: "lists"}
+                    );
+        }`,
+        true
+      );
+    });
+
+    it("should generate paging if no @items defined", async () => {
```
MaryGao, line 585: If there is no `@items` defined we would report warnings for specific operations and not generate paging for them. Because we don't know the detailed item type. For post paging I add a case in widget_dpg to cover that.

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/api/pagingHelpers.ts
```diff
@@ -0,0 +1,233 @@
...
+ */
+interface PagedResult<
+  TElement,
+  TPage = TElement[],
+  TPageSettings = PageSettings
+> {
+  /**
+   * Link to the first page of results.
+   */
+  firstPageLink?: string;
```
MaryGao, line 28: @deyaaeldeen I notice the property is required in core definition but here this would be optional because we can call from getInitialResponse in this first time.

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/models/pagingTypes.ts
deyaaeldeen: Yes we can make the generic page settings extend from the default page settings type, see your comment in https://github.com/Azure/azure-sdk-for-js/pull/23404#discussion_r990535750. Otherwise, I created this wish list back in the day which has one item for now which will simplify the implementation a bit: https://github.com/Azure/azure-sdk-for-js/issues/21836

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/api/pagingHelpers.ts
deyaaeldeen: Yeah, it is required in case the implementation will send the first call to get the first page, but if we can that first page outside the implementation, then yeah it doesn't have to be required.

qiaozha (approved):

qiaozha (approved):
LGTM

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/models/pagingTypes.ts
MaryGao: I created an issue in sdk repo to focus on the alignment https://github.com/Azure/azure-sdk-for-js/issues/27868.

MaryGao:
Jeff's approval:
image

packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/src/api/pagingHelpers.ts
```diff
@@ -0,0 +1,183 @@
...
+    return undefined;
+  }
+
+  const nextLink = (body as Record<string, unknown>)[nextLinkName];
+
+  if (typeof nextLink !== "string" && typeof nextLink !== "undefined") {
+    throw new RestError(
+      `Body Property ${nextLinkName} should be a string or undefined`
+    );
+  }
```
deyaaeldeen, line 154: What if it is set to null? I was just burnt by this 😅

packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/src/models/pagingTypes.ts
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/**
+ * An interface that tracks the settings for paged iteration
```
deyaaeldeen, line 5:
```suggestion
 * Options for the byPage method
```

packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/src/models/pagingTypes.ts
```diff
@@ -0,0 +1,89 @@
...
+ */
+export interface PageSettings {
+  /**
+   * The token that keeps track of where to continue the iterator
```
deyaaeldeen, line 9:
```suggestion
   * A reference to a specific page to start iteration from.
```

packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/src/api/pagingHelpers.ts
MaryGao: Goot catch! I would add the possible value also `null`.

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/models/pagingTypes.ts
```diff
@@ -0,0 +1,89 @@
...
+   * a function to implement the `byPage` method on the paged async iterator.
+   */
+  byPage?: (
+    settings?: TPageSettings
+  ) => AsyncIterableIterator<ContinuablePage<TElement, TPage>>;
+
+  /**
+   * A function to extract elements from a page.
+   */
+  toElements?: (page: TPage) => unknown[];
```
deyaaeldeen, line 80: We initially typed this as `unknown` to preserve backward compatibility but this is not a concern in this PR so we can do the right thing and return TElement[].
```suggestion
  toElements?: (page: TPage) => TElement[];
```

deyaaeldeen (approved):
LGTM, thanks for addressing all feedback!

PR #1997, opened 2023-08-30
Throw body error as rest error

### LINKED ISSUE #1972: Throw rest errors in the right shape
The modular code generator generates code that checks if the response is unexpected and throws the response body if it indeed is, see https://github.com/Azure/autorest.typescript/blob/28e2087088f15cce2a56f4c971948a1671d50d83/packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts#L57

However this is problematic because the response body in this case is of the following format `{ error: { message: ..., code: ... } }` and the body is not in the correct REST error shape. We need to throw `result.body.error` if the `error` field exists.

qiaozha:
fixes #1972


deyaaeldeen (commented):
Should we throw a `RestError`?

packages/typespec-ts/test/modularIntegration/generated/client/structure/default/src/api/operations.ts
```diff
@@ -34,7 +35,13 @@ export function _oneSend(
...
 export async function _oneDeserialize(result: One204Response): Promise<void> {
   if (result.status !== "204") {
-    throw result.body;
+    const internalError = (result.body as any).error || result.body || result;
+    const message = `Unexpected status code ${result.status}`;
+    throw new RestError(internalError.message ?? message, {
+      statusCode: result.status,
+      request: result.request,
+      response: result.body as PipelineResponse,
+    });
```
qiaozha, line 44: @deyaaeldeen as HttpResponse doesn't have a response property and I convert the body into the PipelineResponse here, could you help provide some insights ? Thank you.

packages/typespec-ts/test/modularIntegration/generated/client/structure/default/src/api/operations.ts
```diff
@@ -34,7 +35,13 @@ export function _oneSend(
 
 export async function _oneDeserialize(result: One204Response): Promise<void> {
   if (result.status !== "204") {
-    throw result.body;
+    const internalError = (result.body as any).error || result.body || result;
+    const message = `Unexpected status code ${result.status}`;
+    throw new RestError(internalError.message ?? message, {
+      statusCode: result.status,
```
deyaaeldeen, line 41: I think status code is the actual string code field in the body and not the status of the response.

deyaaeldeen: We probably need to try to set it to `result.body.error.code ?? result.body.code`.

packages/typespec-ts/test/modularIntegration/authApiKey.spec.ts
```diff
@@ -29,7 +29,7 @@ describe("ApiKeyClient Classical Client", () => {
       await client.invalid();
       assert.fail("Expected an exception to be thrown.");
     } catch (err: any) {
-      assert.strictEqual(err.error, "invalid-api-key");
+      assert.strictEqual(err.message, "Unexpected status code 403");
```
qiaozha, line 32: for this case, the return error is { body: { error: "invalid-api-key" }}, but we don't think the error should be just be a string in the following code.  I wonder if we should work around this case or leave it ? @deyaaeldeen 
```ts
    const internalError = (result.body as any).error || result.body || result;
    const message = `Unexpected status code ${result.status}`;
    throw new RestError(internalError.message ?? message, {
      statusCode: Number(result.status),
      code: internalError.code,
      request: result.request,
    });
```

deyaaeldeen: What service is this? I think we should have a discussion with them about how their error responses are not conforming to the Azure REST API design guidelines.

I don't think other languages are doing anything special for this case based on the email thread I sent to you so I vote to leave it as is. /cc @xirzec @lmazuel

packages/typespec-test/test/chatApi_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -40,7 +41,13 @@ export async function _createStreamingDeserialize(
...
 ): Promise<ChatCompletionChunk> {
   if (result.status !== "200") {
-    throw result.body;
+    const internalError = (result.body as any).error || result.body || result;
+    const message = `Unexpected status code ${result.status}`;
+    throw new RestError(internalError.message ?? message, {
+      statusCode: Number(result.status),
+      code: internalError.code,
+      request: result.request,
+    });
```
deyaaeldeen, line 50: I think we should pull out this logic into a helper function because it is exactly the same. Also, perhaps it should be exported by core.

qiaozha: Just found core-client-rest has exported an function called createRestError, which is similar like line 46 to line 49, except createRestError doesn't add the internalError.code into RestError, do you think if we change the above into something like 
```
    const internalError = (result.body as any).error || result.body || result;
    const message = `Unexpected status code ${result.status}`;
    throw createRestError(internalError.message ?? message, result);
```
is enough or we should improve the createRestError ?

packages/typespec-ts/test/modularIntegration/authApiKey.spec.ts
qiaozha: This is a cadl ranch test case, I agree to leave it as is.

packages/typespec-test/test/chatApi_modular/generated/typespec-ts/src/api/operations.ts
deyaaeldeen: I propose to update `createRestError` as follows:
- add an overload that takes just the response and take care of creating the message
- set the error code property

This way, the generated code would be just `throw createRestError(result)`. /cc @xirzec

xirzec: looks good

lirenhe (approved):

packages/typespec-ts/package.json
```diff
@@ -28,8 +28,8 @@
     "generate-tsp-only:rlc": "node --loader ts-node/esm ./test/commands/gen-cadl-ranch.ts --tag=rlc",
     "generate-tsp-only:modular": "node --loader ts-node/esm ./test/commands/gen-cadl-ranch.ts --tag=modular",
     "generate-tsp-only:non-branded": "npm run generate-tsp-only:non-branded-rlc && npm run generate-tsp-only:non-branded-modular",
-    "generate-tsp-only:non-branded-rlc": "ts-node ./test/commands/gen-cadl-ranch.ts --tag=non-branded-rlc",
-    "generate-tsp-only:non-branded-modular": "ts-node ./test/commands/gen-cadl-ranch.ts --tag=non-branded-modular",
+    "generate-tsp-only:non-branded-rlc": "node --loader ts-node/esm ./test/commands/gen-cadl-ranch.ts --tag=non-branded-rlc",
+    "generate-tsp-only:non-branded-modular": "node --loader ts-node/esm ./test/commands/gen-cadl-ranch.ts --tag=non-branded-modular",
```
MaryGao, line 32: just curious why we have this change?

packages/typespec-ts/test/modularIntegration/authOauth2.spec.ts
```diff
@@ -57,7 +57,8 @@ describe("OAuth2Context in API Layer", () => {
       await invalid(context);
       assert.fail("Expected an exception to be thrown.");
     } catch (err: any) {
-      assert.strictEqual(err.error, "invalid-grant");
+      console.log(err);
```
MaryGao, line 60: comment out?

packages/typespec-ts/test/modularIntegration/authOauth2.spec.ts
```diff
@@ -57,7 +57,8 @@ describe("OAuth2Context in API Layer", () => {
...
+      assert.strictEqual(err.message, "Unexpected status code: 403");
```
MaryGao, line 61: could we add some assertion for the whole erro object?

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -135,9 +135,16 @@ export function getDeserializePrivateFunction(
   if (needUnexpectedHelper) {
     statements.push(
       `if(${needSubClient ? "UnexpectedHelper." : ""}isUnexpected(result)){`,
-      "throw result.body",
+      `throw createRestError(result);`,
       "}"
     );
+    const specifier = getImportSpecifier("restClient", runtimeImports);
```
xirzec, line 141: nit: feels like we should have an "addImportSpecifier" helper that is smart enough to check if it exists and to populate the importSet

xirzec (approved):
Looks great! Left a minor nit suggestion that can be addressed in a later PR.

deyaaeldeen (approved):
Thanks for addressing the feedback!

packages/rlc-common/src/helpers/importsUtil.ts
```diff
@@ -95,3 +96,61 @@ export function getImportSpecifier(
     ""
   );
 }
+
+export function addImportToSpecifier(
```
qiaozha, line 100: add the addImportToSpecifier helper here. And also add helper functions about clear import set and addImportsToFiles. @xirzec

packages/typespec-ts/test/modularIntegration/authOauth2.spec.ts
qiaozha: iirc, a few years ago, there's some security concern that we should not include raw response in the error as of privacy related concerns ?

qiaozha: iirc, a few years ago, there's some security concern that we should not include raw response in the error as of privacy related concerns ?

xirzec: I think if we use `@azure/logger` and a level of `verbose` we can sidestep some of those concerns since it wouldn't show up in logs unless explicitly enabled

PR #2006, opened 2023-09-06
support encode for modular

qiaozha:
add support for encode for datetime, duration, bytes in Modular.

for datetime, Modular layer will always be a Date type, but RLC layer will be different.
1. In the case "rfc7231" or if the default in header, RLC layer will be a utc datetime string.
  serialize: someDate.toUTCString()
  deserialize: Date(someUTCString)
2. In the case of "unixTimestamp", RLC layer will be a number.
  serialize: someDate.getTime()
  deserialize: Date(someUnixTimestamp)
3. In the case of "rfc3339":or default not in header, RLC layer will be a ISO datetime string.
  serialize: someDate.toISOString()
  deserialize: Date(someISOString)

for duration
1. In the case of format "seconds", both Modular and RLC layer will use number.
5. In the case of "ISO8601" or default, both Modular and RLC layer will use string.
 
for bytes: 
1. Modular layer will be Uint8Array and RLC layer will be string no matter what format is using
  serialize: uint8ArrayToString(value, format)
  deserialize: stringToUint8Array(value, format)

fixes https://github.com/Azure/autorest.typescript/issues/2007


packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -437,10 +445,30 @@ function getCollectionFormat(param: Parameter) {
   }
   const isMulti = (param.format ?? "").toLowerCase() === "multi";
   const additionalParam = isMulti ? `, "${param.restApiName}"` : "";
-  if (!param.optional) {
-    return `"${param.restApiName}": ${collectionInfo}(${param.clientName}${additionalParam})`;
+  const serializeHelperImport = importSet.get("../rest/index.js");
+  if (serializeHelperImport) {
+    serializeHelperImport.add(collectionInfo);
+  } else {
+    importSet.set("../rest/index.js", new Set<string>().add(collectionInfo));
```
MaryGao, line 452: I don't think we need to add the importSet here because:
- the default fix missing import could help us
- the fixed path `../rest/index.js` may not work in multi client.

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/loadTestRun/api/operations.ts
```diff
@@ -1082,7 +1082,12 @@ export function _listMetricsSend(
         metricNamespace: options?.metricNamespace,
         timespan: options?.timespan,
       },
-      body: { filters: options?.filters },
+      body: {
+        filters: (options?.filters ?? []).map((p) => ({
+          name: p["name"],
+          values: p["values"],
+        })),
+      },
```
MaryGao, line 1090: Could you point to me where is the fix code?

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -63,8 +63,7 @@ export function _publishCloudEventSend(
               ? uint8ArrayToString(event["dataBase64"], "base64")
               : undefined,
           type: event["type"],
-          time:
-            event["time"] !== undefined ? new Date(event["time"]) : undefined,
+          time: event["time"]?.toISOString(),
```
MaryGao, line 66: I notice the typespec type here is `utcDateTime` but we generate to `toISOString`

https://github.com/Azure/autorest.typescript/blob/1235e0ba4762ac1171f5777b50eafb1b7e43a8d0/packages/typespec-test/test/eventgrid_modular/spec/main.tsp#L64-L65C24

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -437,10 +445,30 @@ function getCollectionFormat(param: Parameter) {
...
   }
-  return `"${param.restApiName}": options?.${param.clientName} !== undefined ? ${collectionInfo}(options?.${param.clientName}${additionalParam}): undefined`;
+  if (!param.optional) {
+    return `"${param.restApiName}": ${collectionInfo}(${serializeRequestValue(
+      param.type,
```
MaryGao, line 456: I think the change makes sense for me.

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
qiaozha: will remove this, previously it's not working because we fail to export it in the rest/index.js.

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -525,7 +557,17 @@ function getOptional(param: OptionalType, importSet: Map<string, Set<string>>) {
...
+    param.type,
+    `options?.${param.clientName}`,
+    importSet,
+    false,
+    param.format === undefined &&
+      (param as Parameter).location === "header" &&
+      param.type.type === "datetime"
+      ? "headerDefault"
+      : param.format
+  )}`;
```
qiaozha, line 570: It should be here fixing the options.filters

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
qiaozha: resolved.

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
qiaozha: I suspect typespec compiler does not provide any format info to these types, but let me check.

qiaozha: typescript date.toUTCString() is "Fri, 08 Sep 2023 08:50:03 GMT" and date.toISOString() is "2023-09-08T08:42:49.439Z" 
therefore use toISOString for utcDateTime makes more sense. and toUTCString for offsetDateTime makes more sense to me.

MaryGao: Yeah, it makes sense to me also. Just checking with Time here https://github.com/microsoft/typespec/pull/1899#issuecomment-1713185531

MaryGao (approved):
LGTM and just left small comments for checking and one ut.

packages/typespec-ts/test/modularUnit/modelsGenerator.spec.ts
```diff
@@ -13,6 +13,710 @@ describe("modular model type", () => {
   });
 });
 
+describe("modular encode test for property type datetime", () => {
```
MaryGao, line 16: nit: one more case with utcDateTime without encode for header property.

qiaozha: done

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
MaryGao: It seems both of them share the same default value.

https://microsoft.github.io/typespec/next/standard-library/http/encoding#utcdatetime-and-offsetdatetime

qiaozha: They are not the same 
image

Issue #2014, opened 2023-09-11
[JS] Missing Property From Extension in Deserialize Method


minhanh-phan:
There is a missing property in the deserialize method. The `AudioTranscriptionVerboseJson` interface extends from the `AudioTranscriptionSimpleJson` face. In the deserialize method for `AudioTranscriptionVerboseJson`, the property `text` from `AudioTranscriptionSimpleJson` is missing.

For reference, the type definition generated is [here](https://github.com/Azure/azure-sdk-for-js/blob/2ce87a283d531d2b309cd8953467c532ff2092c5/sdk/openai/openai/sources/generated/src/models/models.ts#L15C1-L26C2)
```ts
/** Transcription response containing only the transcribed text. */
export interface AudioTranscriptionSimpleJson {
  /** Transcribed text. */
  text: string;
}

/** Transcription response. */
export interface AudioTranscriptionVerboseJson
  extends AudioTranscriptionSimpleJson {
  /** Audio transcription task. */
  task: AudioTranscriptionTask;
  /** Language detected in the source audio file. */
  language: string;
  /** Duration. */
  duration: number;
  /** Segments. */
  segments: AudioTranscriptionSegment[];
}
```
The generated code for the deserialize method returns this object [here](https://github.com/Azure/azure-sdk-for-js/blob/2ce87a283d531d2b309cd8953467c532ff2092c5/sdk/openai/openai/sources/generated/src/api/operations.ts#L170C1-L186C5)

```ts
return {
    task: result.body["task"],
    language: result.body["language"],
    duration: result.body["duration"],
    segments: (result.body["segments"] ?? []).map((p) => ({
      id: p["id"],
      start: p["start"],
      end: p["end"],
      text: p["text"],
      temperature: p["temperature"],
      averageLogProb: p["avg_logprob"],
      compressionRatio: p["compression_ratio"],
      noSpeechProb: p["no_speech_prob"],
      tokens: p["tokens"],
      seek: p["seek"],
    })),
  };
```

Expect: Return the property `text` from the extended interface `AudioTranscriptionSimpleJson`. The model definition in TypeSpec is [here](https://github.com/Azure/azure-rest-api-specs/blob/fb4f558b15075826656a3061d332d1a198352dd4/specification/cognitiveservices/OpenAI.Inference/models/audio/commons.tsp#L90C1-L103C2)

Reproduce: 
[...]

PR #2017, opened 2023-09-12
fix-get-parent-properties

qiaozha:
fixes https://github.com/Azure/autorest.typescript/issues/2014

### DIFF
--- a/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
+++ b/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
@@ -156,11 +156,11 @@ export function getDeserializePrivateFunction(
         response.type.format
       )}`
     );
-  } else if (response?.type?.properties) {
+  } else if (getAllProperties(response?.type).length > 0) {
     statements.push(
       `return {`,
       getResponseMapping(
-        response.type.properties ?? [],
+        getAllProperties(response.type) ?? [],
         "result.body",
         importSet
       ).join(","),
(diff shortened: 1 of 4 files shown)


packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -943,3 +943,15 @@ export function hasPagingOperation(codeModel: ModularCodeModel) {
     )
   );
 }
+
+function getAllProperties(type: Type): Property[] {
+  const properties: Property[] = [];
+  if (!type) {
+    return properties;
+  }
+  type.parents?.forEach((p) => {
```
MaryGao, line 952: We should also consider the multiple hierarchy.

A -> B -> C.

packages/typespec-ts/test/modularUnit/modelsGenerator.spec.ts
```diff
@@ -804,4 +804,83 @@ describe("inheritance & polymorphism", () => {
       }`
     );
   });
+
+  it.only("should handle inheritance model in operations", async () => {
```
MaryGao, line 808: skip `only` & could you add test cases for A -> B -> C with different discriminator?

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -943,3 +943,15 @@ export function hasPagingOperation(codeModel: ModularCodeModel) {
     )
   );
 }
+
+function getAllProperties(type: Type): Property[] {
+  const properties: Property[] = [];
```
MaryGao, line 948: nit: not sure this is caused the duplicated properties issues when discriminator property exists both parent and child model. see below cadl-ranch cases:

https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/type/model/inheritance/single-discriminator/main.tsp#L12-L25C1

MaryGao (approved):
Left small comments.

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
qiaozha: Yeah, I found that, when I was trying to enable cadl ranch test cases for model inheritance in modular.

PR #2019, opened 2023-09-15
use-string-as-offset-datetime-type-in-modular

qiaozha:
fix https://github.com/Azure/azure-sdk-for-js/pull/27087#issuecomment-1720020323

### DIFF
--- a/packages/typespec-ts/src/modular/buildCodeModel.ts
+++ b/packages/typespec-ts/src/modular/buildCodeModel.ts
@@ -1080,7 +1080,7 @@ function emitStdScalar(
     case "plainTime":
       return { type: "datetime", format: newScalar.format ?? "time" };
     case "offsetDateTime":
-      return { type: "datetime", format: newScalar.format ?? "rfc7231" };
+      return { type: "string" };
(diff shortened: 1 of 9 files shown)


MaryGao (approved):
LGTM

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -1080,7 +1080,7 @@ function emitStdScalar(
     case "plainTime":
       return { type: "datetime", format: newScalar.format ?? "time" };
     case "offsetDateTime":
-      return { type: "datetime", format: newScalar.format ?? "rfc7231" };
+      return { type: "string" };
```
MaryGao, line 1083: Do you think we should apply this change to RLC also?

I have no strong preference but it seems reasonable to not generate Date for RLC if modular won't return date.

Core would use ISO format by default.
https://github.com/MaryGao/azure-sdk-for-js/blob/update-emitter-package-lock-file/sdk/core/core-client-rest/src/urlHelpers.ts#L50

PR #2020, opened 2023-09-15
Try not flatten payload

### LINKED ISSUE #2016: Should not flatten first layer of payload in Modular.
Let's say we have the following TypeSpec
```tsp
    model Foo {
      prop1: string;
      prop2: int64;
      prop3: utcDateTime;
      prop4: offsetDateTime;
      prop5: Bar;
    }
    model Bar {
      prop1: string;
      prop2: int64;
    }
    op read(@body body: Foo): { @body body: Foo };
```

we will get a flattened payload like as shown in the following operations.ts, but we should not flatten it, it maybe only useful when there's some RP migrate from swagger into TypeSpec.
```typescript
import { TestingContext as Client } from '../rest/index.js';
import { StreamableMethod, operationOptionsToRequestParameters } from '@azurerest/coreclient';

export function _readSend(
    context: Client,
    prop1: string,
    prop2: number,
    prop3: Date,
    prop4: Date,
    prop5: Bar,
    options: ReadOptions = { requestOptions: {} }
): StreamableMethod<Read200Response> {
    return context.path('/').post({
        body: {
            prop1: prop1,
            prop2: prop2,
            prop3: prop3.toISOString(),
            prop4: prop4.toUTCString(),
            prop5: { prop1: prop5['prop1'], prop2: prop5['prop2'] },
        },
    });
}
export async function _readDeserialize(result: Read200Response): Promise<Foo> {
    if (result.status !== '200') {
        throw result.body;
    }
    return {
        prop1: result.body['prop1'],
        prop2: result.body['prop2'],
        prop3: new Date(result.body['prop3']),
        prop4: new Date(result.body['prop4']),
        prop5: {
            prop1: result.body.prop5['prop1'],
            prop2: result.body.prop5['prop2'],
        },
    };
}
export async function read(
    context: Client,
    prop1: string,
    prop2: number,
    prop3: Date,
    prop4: Date,
    prop5: Bar,
    options: ReadOptions = { requestOptions: {} }
): Promise<Foo> {
    const result = await _readSend(context, prop1, prop2, prop3, prop4, prop5, options);
    return _readDeserialize(result);
}

```

qiaozha:
fixes #2016


packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -815,6 +815,21 @@ function emitBasicOperation(
...
+        ? bodyParameter.clientName + "Parameter"
+        : bodyParameter.clientName;
+  }
+  parameters
+    .filter((param) => {
+      return param.clientName === name && !param.isReadOnly && param.required;
+    })
+    .forEach((param) => {
+      param.clientName = param.clientName + "Parameter";
+    });
```
qiaozha, line 832: when there's parameter name is the same as the operation name, ts will report compile error.

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -186,28 +186,6 @@ function getOperationSignatureParameters(
...
-  } else if (operation.bodyParameter?.type.type === "byte-array") {
-    parameters.set(
-      operation.bodyParameter.clientName,
-      buildType(
-        operation.bodyParameter.clientName,
-        operation.bodyParameter.type,
-        operation.bodyParameter.type.format
-      )
-    );
-  }
```
qiaozha, line 210: This is the flatten part,

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -362,24 +350,13 @@ function buildBodyParameter(
...
-        );
-      } else {
-        bodyParts.push(getParameterMap(param, importSet));
-      }
-    }
+    const bodyParts: string[] = getRequestModelMapping(
+      bodyParameter.type,
+      bodyParameter.clientName,
+      importSet
+    );
```
qiaozha, line 357: As the model is unflattened, we can just getRequestModelMapping without considering if any of those properties are coming from required signature or option bag.

packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/review/ai-content-safety.api.md
```diff
@@ -9,6 +9,16 @@ import { KeyCredential } from '@azure/core-auth';
 import { OperationOptions } from '@azure-rest/core-client';
 import { TokenCredential } from '@azure/core-auth';
 
+// @public
+export interface AddOrUpdateBlockItemsOptions {
+    blockItems: TextBlockItemInfo[];
+}
+
+// @public (undocumented)
+export interface AddOrUpdateBlockItemsOptions {
```
lirenhe, line 18: Seems to have duplicate exports here for 'AddOrUpdateBlockItemsOptions', is this expected?

packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/review/ai-content-safety.api.md
```diff
@@ -125,6 +151,16 @@ export interface PagedTextBlocklist {
     value: TextBlocklist[];
 }
 
+// @public
+export interface RemoveBlockItemsOptions {
+    blockItemIds: string[];
+}
+
+// @public (undocumented)
+export interface RemoveBlockItemsOptions {
```
lirenhe, line 160: Same here.

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/review/eventgrid.api.md
```diff
@@ -42,12 +47,12 @@ export interface CloudEvent {
 // @public (undocumented)
 export class EventGridClient {
     constructor(endpoint: string, credential: KeyCredential, options?: EventGridClientOptions);
-    acknowledgeCloudEvents(lockTokens: string[], topicName: string, eventSubscriptionName: string, options?: AcknowledgeCloudEventsOptions): Promise<AcknowledgeResult>;
-    publishCloudEvent(event: CloudEvent, topicName: string, options?: PublishCloudEventOptions): Promise<Record<string, any>>;
-    publishCloudEvents(events: CloudEvent[], topicName: string, options?: PublishCloudEventsOptions): Promise<Record<string, any>>;
+    acknowledgeCloudEvents(topicName: string, eventSubscriptionName: string, lockTokens: AcknowledgeOptions, options?: AcknowledgeCloudEventsOptions): Promise<AcknowledgeResult>;
```
lirenhe, line 50: What are the root causes of those parameter order change?

packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/src/ContentSafetyClient.ts
```diff
@@ -57,18 +59,18 @@ export class ContentSafetyClient {
...
     options: AnalyzeTextRequestOptions = { requestOptions: {} }
   ): Promise<AnalyzeTextResult> {
-    return analyzeText(this._client, text, options);
+    return analyzeText(this._client, body, options);
   }
 
   /** A sync API for harmful content analysis for image. Currently, we support four categories: Hate, SelfHarm, Sexual, Violence. */
   analyzeImage(
-    image: ImageData,
+    body: AnalyzeImageOptions,
```
lirenhe, line 70: Can we have a better name instead of 'body' here?  I think that is consistent with HLC.
    analyzeTextOptions:  AnalyzeTextOptions

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/review/eventgrid.api.md
qiaozha: This is expected, because previously it will put the flattened parameters ahead of other required parameters.

packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/src/ContentSafetyClient.ts
qiaozha: This is their body name and AnalyzeImageOptions is their model name. I checked with them, they won't like to change it.

packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/review/ai-content-safety.api.md
qiaozha: Let me double check on this.

packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/src/ContentSafetyClient.ts
joheredi: Yeah, I was thinking the same as Renhe, but it looks like this is following what is defined in the spec. Python DPG has body as the parameter name as well

packages/typespec-test/test/openai_modular/generated/typespec-ts/review/openai_modular.api.md
```diff
@@ -264,12 +391,12 @@ export type ImageSize = string;
 // @public (undocumented)
 export class OpenAIClient {
     constructor(endpoint: string, credential: KeyCredential | TokenCredential, options?: OpenAIClientOptions);
-    beginAzureBatchImageGeneration(prompt: string, options?: BeginAzureBatchImageGenerationOptions): Promise<BatchImageGenerationOperationResponse>;
+    beginAzureBatchImageGeneration(body: ImageGenerationOptions, options?: BeginAzureBatchImageGenerationOptions): Promise<BatchImageGenerationOperationResponse>;
```
joheredi, line 394: @qiaozha I guess there will be libraries that actually wanted to flatten and were getting that behavior by luck. What changes do we need in the typespec here to keep the flattened properties? I think in this case we want it flattened

qiaozha: After confirming with other DPG, we will probably support by alias spread. I will add a ut for that, and for openai's case, we will probably need to convince them to change the typespec. 

Also, as previously, we put the flattened required parameters ahead of other required parameters, should we still keep this https://github.com/Azure/autorest.typescript/pull/2020#discussion_r1327042746 I feel like put those flattened parameters behind will make more sense.  Also, I found c# is putting them behind https://github.com/Azure/azure-sdk-for-net/blob/99cc57c41c7fa2a3df564ad0c6a745c001cd4517/sdk/openai/Azure.AI.OpenAI/api/Azure.AI.OpenAI.netstandard2.0.cs#L372 in their manual customization.

qiaozha:
@joheredi @xirzec I want to confirm about the parameter order in flattened case. 
Let's say we have tsp defined as 
```
alias Widget = {
  @visibility("read", "update")
  @path
  id: string;

  weight: int32;

  @query
  color: "red" | "blue";
};

interface Widgets {
  @patch update(@path id1: string, @query color1: string, ...Widget): Widget;
}
```
what should be update method signature should be ? 

Option1: 
we follow the parameter that is defined in the typespec, regardless of what their location should be. we will get `id1, color1, id, weight, color`

Option2: 
we follow pattern like path > query > body flattened param, and inside path or query or body, we follow the typespec defined order, we will get
`id1, id, color1, color, weight`

Another question about the path parameters order is what if the path is something like `/ids/{id}/id1s/{id1}` in the above case ?
should we have` id, id1, color1, color, weight` if we go to option2 ? 

As parameters order doesn't really matter to a API, so the way how OpenAPI3 emitter handles the parameter order may not be very helpful to us. see the typespec playground link [here](https://cadlplayground.z22.web.core.windows.net/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBUeXBlU3BlYy5IdHRwOwpAc2VydmljZSh7CiAgdGl0bGU6ICJXaWRnZXQgU8YbIiwKICB2ZXJzaW9uOiAiMS4wLjAiLAp9KQpuYW1lc3BhY2UgRGVtb8cvOwoKYWxpYXMgx0Y9IMRbQHZpc2liaWxpdHkoInJlYWQiLCAidXBkYXRlIinEIHBhdGgKICBpZDogc3RyaW5nOwoKICB3ZWlnaHQ6IGludDMyxRJAcXVlcnkKICBjb2xvcjogInJlZCIgfCAiYmx1ZSI7Cn07CgppbnRlcmbkAJjmAIRz5gCDcGF0Y2ggxnUoxXEgaWQxyHAsx1vGWcsXLi4uxkcpOsdQOwp9Cg%3D%3D&e=%40typespec%2Fopenapi3&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%7D)

xirzec:
@qiaozha I think my preference is option 1 where we follow the spec order, since then the spec author can control it by reordering the parameters. We'd still extract any optional values and put them in an options bag at the end, right?

qiaozha:
As much as I prefer the order that is defined in typespec, I have to point out sometimes service team could be too casual about the parameters order from version to version which is not an api level breaking but is a SDK breaking. and we have to done things either asking service team to change the order back or adding workaround manually to avoid unnecessary SDK breaking changes. I am hoping we could do something in modular to reduce the communicating and manual effort.

xirzec:
> As much as I prefer the order that is defined in typespec, I have to point out sometimes service team could be too casual about the parameters order from version to version which is not an api level breaking but is a SDK breaking. and we have to do things either asking service team to change the order back or adding workaround manually to avoid unnecessary SDK breaking changes. I am hoping we could do something in modular to reduce the communicating and manual effort.

Yeah, if you think it will reduce churn to follow the same convention as Python I'm okay with it, though they have some behavior we can't directly map like headers always being kwargs (we could pull them into options but that might mean we have required options)

qiaozha:
Confirmed with @joheredi  offline, we will keep the original order,  as Option2 can only resolve parameter order changes between different parameter location and if there's changes inside each parameter location, it will still be problematic.

packages/typespec-ts/test/modularUnit/modelsGenerator.spec.ts
```diff
@@ -1126,6 +1135,403 @@ describe("inheritance & polymorphism", () => {
...
+      prop3?: utcDateTime;
+      @query
+      prop4: offsetDateTime;
+      prop5?: Bar;
+    };
+    model Bar {
+      prop1: string;
+      prop2: int64;
+    }
+    op read(@path pathParam: string, ...Foo, @query queryParam: string): OkResponse;
```
qiaozha, line 1358: Here, I put the alias in the middle, and path and query related parameter can be order by typespec correctly, but as there's no way to know the order between bodyParameter and other parameters, I generate the body flattened parameter prop2 to the last.

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -40,8 +43,8 @@ import {
 
 export function _publishCloudEventSend(
   context: Client,
-  event: CloudEvent,
   topicName: string,
+  event: CloudEvent,
```
qiaozha, line 47: In this case, both the previous change and new change are flattened payload because their typespec is defined as 
```
  @post op PublishCloudEvent is Azure.Core.RpcOperation<{
    @doc("content type")
    @header("content-type")
    contentType: "application/cloudevents+json; charset=utf-8";

    @doc("Topic Name.")
    @path
    topicName: string;

    @body 
    event: {event: CloudEvent};
  }, PublishResult>;
```
where the body parameter is actually an anonymous model, and we are using model.name === "" to determine if there's an alias, anonymous model will also fit into the condition, so we also need to flatten the payload here, as of the order change is from this comment

qiaozha: Also, please note in their latest typespec from main branch, they already removed the anonymous model https://github.com/Azure/azure-rest-api-specs/blob/main/specification/eventgrid/Azure.Messaging.EventGrid/main.tsp#L197

xirzec:
It feels a bit odd to have the new body params models be suffixed with `Options` given many of those operation methods still have another `Options` type at the end. It's also strange to call it `Options` when it has a required property, especially when it's the only required property.

It makes it tempting to flatten when there is exactly one required property in the body... in theory the service can never change to add another required body property without a break, yes? If they added an optional body property we could pull that into the actual options bag.

qiaozha:
@xirzec I am not sure I totally get your point. In the below case, prop2, prop3 and prop5 all belong to a body parameter's properties. excapt the body parameter's own name is ""
```
    alias Foo = {
      @path
      prop1: string;
      prop2: int64;
      prop3?: utcDateTime;
      @query
      prop4: offsetDateTime;
      prop5?: Bar;
    };
    model Bar {
      prop1: string;
      prop2: int64;
    }
    op read(@path pathParam: string, ...Foo, @query queryParam: string): OkResponse;
```
As we agree to flatten the body payload in this scenario, we will flatten all of them to the top level and because prop2 is required, it will go into the operation signature, prop3 and prop5 are optional, they will go into the option bag the same as the other optional ordinary query/header parameters if there's any.
So our operation signature would loook like 
```
      export interface ReadOptions extends OperationOptions {
        prop3?: Date;
        prop5?: Bar;
      }
      export async function read(
        context: Client,
        pathParam: string,
        prop1: string,
        prop4: string,
        queryParam: string,
        prop2: number,
        options: ReadOptions = { requestOptions: {} }
      ):
```
And we build our payload like 
```
        return context
          .path("/{pathParam}/{prop1}", pathParam, prop1)
          .post({
            queryParameters: { prop4: prop4, queryParam: queryParam },
            body: {
              prop2: prop2,
              prop3: options?.prop3?.toISOString(),
              prop5: {
                prop1: options?.prop5?.["prop1"],
                prop2: options?.prop5?.["prop2"],
              },
            },
          });
```
If there're more required body properties in the above case, they will just go to the operation signature the same as other prop2. not in the option bag.  

Let me know if it makes sense to you. Thanks

qiaozha:
> .. in theory the service can never change to add another required body property without a break, yes? 

To answer this question, I think it's a breaking no matter we flatten it or not. if we don't flatten, the breaking will be a model layer breaking, if we flatten it, it will be an operation signature breaking.

xirzec:
Sorry, I think it will help if I give specific examples that I was looking at in the generated output:

```ts
export interface AddOrUpdateBlockItemsOptions {
  blockItems: TextBlockItemInfo[];
}


export class ContentSafetyClient {
  addOrUpdateBlockItems(blocklistName: string, body: AddOrUpdateBlockItemsOptions, options?: AddOrUpdateBlockItemsRequestOptions): Promise<AddOrUpdateBlockItemsResult>;
}
```

In the above it would feel better to name the first interface something like `AddOrUpdateBlockItemsBody`. At that point it makes better sense to me and works, especially in cases where we have many required body properties.

So now it would look like:

```ts

```ts
export interface AddOrUpdateBlockItemsBody {
  blockItems: TextBlockItemInfo[];
}


export class ContentSafetyClient {
  addOrUpdateBlockItems(blocklistName: string, body: AddOrUpdateBlockItemsBody, options?: AddOrUpdateBlockItemsRequestOptions): Promise<AddOrUpdateBlockItemsResult>;
}
```


In this specific case though, I was thinking it is a little silly to have an entire object just for one property, so as an optimization we could flatten bodies that have a single required property so that the above looked like:

```ts
export class ContentSafetyClient {
  addOrUpdateBlockItems(blocklistName: string, blockItems: TextBlockItemInfo[], options?: AddOrUpdateBlockItemsRequestOptions): Promise<AddOrUpdateBlockItemsResult>;
}
```

Does that make sense? So less of a global rule and more of a specific optimization for body objects with only one required property.

qiaozha:
`AddOrUpdateBlockItemsOptions` is the name from their typespec model, it's not generated by codegen. I specifically checked with them, they wouldn't like to change. see comment here https://github.com/Azure/azure-sdk-for-js/pull/26542#discussion_r1321082457 
In their latest typespec, they have changed it into AddBlockItemsOptions and the options still there https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/ContentSafety/models.tsp#L180

I have to say, I have mixed feelings about flatten one required property body payload. 
1. it indeed makes the user experience better, and it's one of the main reason we have x-ms-client-flatten in swagger.
2. from breaking change perspective.
    a. as state here https://github.com/Azure/autorest.typescript/pull/2020#issuecomment-1730756274 adding a required property is a breaking either way
    b. if it's making this required property optional, but keep the body parameter required, in the case of not flatten, it will not be a breaking but in the case of flatten, it will be a breaking. 
    c. In the case of body parameter is optional, current HLC implementation is to put the body parameter into the option bag, Modular is still keep the body in the operation signature. so changing both that required property and the body parameter itself from required to non-required 
         c1:  in the case of not flatten, it will be a breaking for hlc and will be a non breaking for Modular.
         c2: in the case of flatten, it will be a breaking for both hlc and Modular.
3. flatten will overall make the serialize and deserialize more complex. And as we are actually flatten the body payload not the model, it will cause inconsistency between the request and response if that body model is being used in a roundtrip case.  One of the reason I don't want to flatten body payload is that I want to have a model based serialize and deserialize in Modular as suggested here https://github.com/Azure/autorest.typescript/issues/2016#issuecomment-1715332779. 
4. As x-ms-client-flatten has been used in so many places in current rest api specs repo https://github.com/search?q=repo%3AAzure%2Fazure-rest-api-specs%20x-ms-client-flatten&type=code, it will be unavoidable to support model flatten unless we want a very large scale of breaking changes when they migrate to typespec. 

My preference would be we could support one required property flatten in model based flatten when we get to the x-ms-client-flatten thing ?

packages/typespec-ts/src/modular/buildCodeModel.ts
MaryGao: Could you point me to the testing code?

qiaozha: https://github.com/Azure/autorest.typescript/pull/2020/files#diff-3a40e76bf6b3c7b410b6fd9cc284f9de22165a47721a28d5b51d08b0f24c3563R34 in the model usage cadl ranch test case, there's a method called input and it has a parameter called input.

MaryGao:
> Does that make sense? So less of a global rule and more of a specific optimization for body objects with only one required property.

I agree with Qiaoqiao that not setting this rule for one required property. This may introduce un-necessary breakings and also introduce complexity for specific conditions.

I am okay with current implementation and as we discussed we still have the requirement of flattening especially for mgmt SDKs not and I prefer we continue monitoring this topic. 
- the only supported case in tsp is to flatten spread alias: https://github.com/Azure/cadl-ranch/pull/247#discussion_r1156503591
- how to interpret the nested alias with spread is not finalized in tsp https://github.com/Azure/cadl-ranch/pull/247#discussion_r1157859847
- we have no way to flatten some layer of property in tsp now

packages/typespec-ts/test/modularIntegration/generated/parameters/spread/src/SpreadClient.ts
```diff
@@ -0,0 +1,85 @@
...
+    spreadWithMultipleParameters: (
+      id: string,
+      xMsTestHeader: string,
+      prop1: string,
+      prop2: string,
+      prop3: string,
+      prop4: string,
+      prop5: string,
+      prop6: string,
+      options?: AliasSpreadWithMultipleParametersOptions
```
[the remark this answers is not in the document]
qiaozha: Not sure I get your point, are you suggesting to put some of the required parameters into option bag ? or are you suggesting to put the whole alias into option bag and not flatten it to the top level, in other words, determine flatten or not by how many parameters it has ? Also, is there any indicator says that alias itself is optional or not ?

packages/typespec-test/test/openai_modular/generated/typespec-ts/review/openai_modular.api.md
MaryGao: > What changes do we need in the typespec here to keep the flattened properties?

See my two cents: https://github.com/Azure/autorest.typescript/pull/2020#issuecomment-1733142926

packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/src/ContentSafetyClient.ts
qiaozha: as Jeff confirmed with Brian, we should not done anything tricky to manipulate the name in codegen side.

packages/typespec-test/test/openai_modular/generated/typespec-ts/review/openai_modular.api.md
qiaozha: unit test case added.

MaryGao (approved):
Left small comments, overall LGTM.

packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/review/ai-content-safety.api.md
```diff
@@ -18,13 +23,18 @@ export interface AddOrUpdateBlockItemsResult {
     value?: TextBlockItem[];
 }
 
+// @public
+export interface AnalyzeImageOptions {
+    categories?: ImageCategory[];
+    image: ImageData_2;
```
MaryGao, line 29: Could you help me understand it is `ImageData_2`?

packages/typespec-ts/src/modular/buildCodeModel.ts
MaryGao: Just curious it causes compiler error? I tested it's okay in [playground](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABFApgZygClRgXIjAJxjAHMBKRAbwChF7EIE04AbFAOlblO3SnI0AvjRpMwLdlx58MmAEQALFK27zy5IA).

packages/typespec-ts/test/modularIntegration/generated/parameters/spread/src/SpreadClient.ts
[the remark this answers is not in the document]
MaryGao: Confirmed with .Net they will not do any aggregation in their last design no matter how many parameters are. So I think we could skip this discussion.

packages/typespec-ts/test/modularUnit/modelsGenerator.spec.ts
MaryGao, line 1358: Do we have similar one with below?

```
// case 1
op read(@path pathParam: string, @query queryParam: string, ...Foo): OkResponse;

// case 2
op read(@query queryParam: string,@path pathParam: string, ...Foo): OkResponse;
```

packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/review/ai-content-safety.api.md
qiaozha: see https://github.com/Azure/azure-sdk-for-js/pull/26542#discussion_r1327118415 and https://github.com/Azure/azure-sdk-for-js/pull/26542#discussion_r1327869367

packages/typespec-ts/test/modularUnit/modelsGenerator.spec.ts
qiaozha: https://github.com/Azure/autorest.typescript/pull/2020/files/bc99bd8fb15251fb0759e1a69e8294153dd8e0f6#diff-6231ca8cedd024d1da6d77e9da66db4fc774f033279e977b4fcd55629ebe6e62R1252 case1 

I don't think we need a case 2. to prove that we don't make a difference between the path and query parameter as we get all those parameters from typespec, the only difference is body parameter.

packages/typespec-ts/src/modular/buildCodeModel.ts
qiaozha: It's reports compile error when classical client layer calls the api layer [link](https://www.typescriptlang.org/play?#code/FAYwNghgzlAEAqBTKAXWBvYtYH1wEtEA7FALlgiIE8BuLWfIgBwFcUAKe7R1s2VAE6MA5gBousAPZMU+SUSjl0sAYgCOLZCgDyMuQqUBfWIfoBKcgAUBkgLb4oiADwA3SfgAmAPgwTVKFgEiBmY2dhQACwcAOjwwQhJREN4k6Vl5KDM6bFNDIA)

lirenhe (approved):

xirzec (approved):

qiaozha:
@joheredi I will merge this PR first, let me know if you have other concerns. 

Also, please note, we will need to work with OpenAI to use spread of alias if they want a flattened operation sigature.

PR #2002, opened 2023-09-04
Refactor the sample generation into rlc-common & Generate samples with mock content

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2000

The pr is mainly focusing on generating samples if no real examples provided. Here is the high-level design:

### **rlc-common**
- transformSampleGroups.ts

In common layer we provide a transform function which could generate sample content based on path/schema/operation details in RLCModel. Currently we only provide **mock value** generation which could be improved if we have **suggested values** in future.

```typescript
export function transformSampleGroups(model: RLCModel, allowMockValue = true)
```

- valueGenerationUtil.ts

Helper functions to help generate values, the idea here is to try to suggest the ts-type from schema, if can't we will fallback to suggest ts-type from the generated type iteself, e.g `Record<string, "test" | 1>` would be a record type.

```typescript
export function generateParameterTypeValue(
  type: string,
  parameterName: string,
  schemaMap: Map<string, Schema>,
  path: Set<string> = new Set(),
  _allowMockValue = true
)
```

- typeUtil.ts

Helper utils to detect the detailed ts types.

### **typespec-ts**

The only thing in typespec project needs to do is apply the common transform logic and small improvements in schema for better detection.

```typescript
model.sampleGroups = transformSampleGroups(
    model,
    options?.generateSample ===
      true /* Enable mock sample content if generateSample === true */
  );
```

### **autorest.typescript**

- rlcSampleGenerator.ts

We prefer to generate sample from swagger examples first and if no swagger examples, we will generate mock sample.

### **Testing**

- `rlc-common/test/helpers/typeUtil.spec.ts` testing for different type convertion
- `typespec-ts/test/unit/transform/transformSchemas.spec.ts` testing for distinct schema data
- `typespec-ts/test/unit/sample/generateSampleContent.spec.ts` integration testing from typespec to generated values
- smoke testing both in swagger and typespec

PR #2022, opened 2023-09-17
Enable UT for rlc-common project

MaryGao:
see comment here: https://github.com/Azure/autorest.typescript/pull/2002#discussion_r1326778024

### DIFF
--- a/packages/rlc-common/package.json
+++ b/packages/rlc-common/package.json
@@ -22,7 +22,7 @@
     "build": "rimraf --glob dist/* dist-esm/* types/* && tsc -p tsconfig.json && tsc -p tsconfig-cjs.json && node publishPackage.js",
     "build:cjs": "rimraf --glob dist/* dist-esm/* types/* && tsc -p tsconfig-cjs.json && node publishPackage.js",
     "build:esm": "rimraf --glob dist/* dist-esm/* types/* && tsc -p tsconfig.json && node publishPackage.js",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "mocha"
   },
   "keywords": [],
   "author": "",
@@ -41,7 +41,21 @@
     "@types/lodash": "^4.14.182",
     "fs-extra": "^10.0.0",
     "@types/fs-extra": "^8.1.0",
-    "ts-node": "^10.7.0"
+    "ts-node": "^10.7.0",
+    "mocha": "^10.2.0",
+    "chai": "^4.3.7",
+    "@types/chai": "^4.3.4",
+    "@types/mocha": "^10.0.1"
+  },
+  "mocha": {
+    "extension": [
+      "ts",
+      "tsx"
+    ],
+    "loader": "ts-node/esm",
+    "spec": [
+      "test/**/*.spec.ts"
+    ]
   },
   "bugs": {
     "url": "https://github.com/Azure/autorest.typescript/issues"
--- a/packages/rlc-common/package.json
+++ b/packages/rlc-common/package.json
@@ -44,10 +44,10 @@
     "ts-node": "^10.7.0",
     "mocha": "^10.2.0",
     "chai": "^4.3.7",
+    "@types/chai": "^4.3.4",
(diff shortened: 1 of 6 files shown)


lirenhe:
Can we have some vscode config that enable the debugging of unit tests?

lirenhe (commented):
Can we have some vscode config that enable the debugging of unit tests?

packages/rlc-common/package.json
```diff
@@ -41,7 +41,21 @@
...
+  },
+  "mocha": {
+    "extension": [
+      "ts",
+      "tsx"
+    ],
+    "loader": "ts-node/esm",
+    "spec": [
+      "test/**/*.spec.ts"
+    ]
```
qiaozha, line 58: Is this the recommend way ? I feel like we should be consistent.

qiaozha (commented):
Maybe update the document here https://github.com/Azure/autorest.typescript/blob/main/packages/rlc-common/CONTRIBUTING.md#testing

packages/rlc-common/package.json
MaryGao: Update to align with typespec-ts

lirenhe (approved):

packages/rlc-common/package.json
```diff
@@ -22,7 +22,7 @@
     "build": "rimraf --glob dist/* dist-esm/* types/* && tsc -p tsconfig.json && tsc -p tsconfig-cjs.json && node publishPackage.js",
     "build:cjs": "rimraf --glob dist/* dist-esm/* types/* && tsc -p tsconfig-cjs.json && node publishPackage.js",
     "build:esm": "rimraf --glob dist/* dist-esm/* types/* && tsc -p tsconfig.json && node publishPackage.js",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "mocha test/**/*.spec.ts"
```
qiaozha, line 25:
```suggestion
    "test": "cross-env TS_NODE_PROJECT=tsconfig.json mocha -r ts-node/register --experimental-specifier-resolution=node --experimental-modules=true  --timeout 4000 './test/unit/**/*.spec.ts'"
```
something like this ? 

Also, I am thinking about when we generate modular, we import the module as esm, otherwise, we will use cjs name  https://github.com/qiaozha/autorest.typescript/blob/5023ca95291e7f8b8374a5afa222d56cacd34aac/packages/rlc-common/src/helpers/nameConstructors.ts#L119 

Can loader as "ts-node/esm" cover cases in cjs ?

MaryGao: I enabled the Unit Test for both cjs and esm in [commit](https://github.com/Azure/autorest.typescript/pull/2022/commits/7ec3a09020e824cfad643988fe101dac3ed36635). Please notice the suffix is not allowed in spec.ts file because cjs module can't find relevant dependencies. But it is okay when enabling option `experimental-specifier-resolution`.

image


Please also notice anothere detail for other packages in typespec-ts and autorest.typescript is we only enabled one kind of module resolution for their UT.

MaryGao: I revert this change because I notice that our sources use the .js extenstion to find the dependency which will cause the cjs can't find module. I believe we could do this only based on the built code not directly on the sources using `ts-node/register`. Let's enhance this when necessary.

image

qiaozha: Do you find this issue exist in the both esm and cjs loader or just cjs loader ? if it's the later, would you mind change this way ?
`cross-env TS_NODE_PROJECT=tsconfig.json mocha -r ts-node/register --loader=ts-node/esm --experimental-specifier-resolution=node  test/**/*.spec.ts`

MaryGao: `experimental-specifier-resolution` feature is for esm and you could refer the [link here](https://nodejs.cn/api/cli/experimental_specifier_resolution_mode.html).

qiaozha: Got it, Thanks for the pointer.

qiaozha (approved):

Issue #1955, opened 2023-08-03
support generating import statements with ".js" extension for relative modules


jeremymeng:
In ESM packages (`"type": "module"` in package.json) we need to add ".js" extension to relative module names in the import statements.  

for example: 
![image](https://github.com/Azure/autorest.typescript/assets/7583839/fc1f444b-0669-47b3-9dd7-2a891d453c22)

https://github.com/Azure/azure-sdk-for-js/pull/26162/commits/160d5a783fc32fa7702596b8d93f9c81917c5668#diff-5bf285a14916a38f4e6e1e5a1966436e11b64907a10cf3a1f17b604dc6348595

we need to add `.js` or `/index.js` depending on whether default is used.

davidyuk:
This seems to be fixed in @autorest/typescript@6.0.39
https://github.com/Azure/autorest.typescript/blob/main/packages/autorest.typescript/CHANGELOG.md#6039-2025-04-28
https://github.com/Azure/autorest.typescript/pull/3128

PR #2023, opened 2023-09-18
Update `mocha` to version `^10.0.0` in generated package

jeremymeng:
Also update commands as neccessary with similar changes that is already in azure-sdk-for-js repo (see https://github.com/Azure/azure-sdk-for-js/pull/27019)


jeremymeng:
I think currently tests in packages with `"type": "module"` would fail because we don't have ".js" extension in relative imports in the included test files. We need an option to https://github.com/Azure/autorest.typescript/issues/1955

jeremymeng:
@qiaozha @MaryGao please have a look when you have a chance.

packages/autorest.typescript/test/smoke/generated/compute-resource-manager/package.json
```diff
@@ -35,7 +35,8 @@
     "@azure/identity": "^2.0.1",
     "@azure-tools/test-recorder": "^3.0.0",
     "@azure-tools/test-credential": "^1.0.0",
-    "mocha": "^7.1.1",
+    "mocha": "^10.0.0",
+    "esm": "^3.2.18",
```
qiaozha, line 39: just curious, why not add @types/mocha here ?

packages/rlc-common/src/metadata/buildPackageFile.ts
```diff
@@ -289,6 +292,17 @@ function restLevelPackage(model: RLCModel, hasSamplesGenerated: boolean) {
...
+    if (isTypeSpecTest) {
+      packageInfo["mocha"] = {
+        extension: ["ts"],
+        timeout: "1200000",
+        loader: "ts-node/esm"
+      };
+      packageInfo.scripts["unit-test:node"] =
+        'mocha --reporter ../../../common/tools/mocha-multi-reporter.js --timeout 1200000 --full-trace "test/{,!(browser)/**/}*.spec.ts"';
+      packageInfo.scripts["integration-test:node"] =
+        'nyc mocha --require source-map-support/register --reporter ../../../common/tools/mocha-multi-reporter.js --timeout 5000000 --full-trace "dist-esm/test/{,!(browser)/**/}*.spec.js"';
```
qiaozha, line 304: I feel like this common/tools/mocha-multi-reporter.js is from the js repo, are you sure this only applies to isTypeSpecTest not azureSdkForJs ？

packages/typespec-ts/src/modular/buildProjectFiles.ts
```diff
@@ -242,12 +242,12 @@ export function emitPackage(
     packageInfo.scripts["unit-test"] =
       "npm run unit-test:node && npm run unit-test:browser";
     packageInfo.scripts["unit-test:node"] =
-      'mocha -r esm --require ts-node/register --reporter ../../../common/tools/mocha-multi-reporter.js --timeout 1200000 --full-trace "test/{,!(browser)/**/}*.spec.ts"';
+      'mocha --reporter ../../../common/tools/mocha-multi-reporter.js --timeout 1200000 --full-trace "test/{,!(browser)/**/}*.spec.ts"';
```
qiaozha, line 245: have the same question here and the below.

packages/autorest.typescript/src/generators/static/packageFileGenerator.ts
```diff
@@ -194,7 +194,8 @@ function regularAutorestPackage(
     packageInfo.devDependencies["@azure/identity"] = "^2.0.1";
     packageInfo.devDependencies["@azure-tools/test-recorder"] = "^3.0.0";
     packageInfo.devDependencies["@azure-tools/test-credential"] = "^1.0.0";
-    packageInfo.devDependencies["mocha"] = "^7.1.1";
+    packageInfo.devDependencies["mocha"] = "^10.0.0";
+    packageInfo.devDependencies["esm"] = "^3.2.18";
```
qiaozha, line 198: looks like this is missing @types/mocha ?

jeremymeng: The old version doesn't include @types/mocha but I probably should add it too.

jeremymeng: I was assuming that some packages might be fine without it (JS repo just upgraded to mocha v10 and some generated packages there may be missing @types/mocha.  But adding it shouldn't hurt.

packages/autorest.typescript/test/smoke/generated/compute-resource-manager/package.json
jeremymeng: Explained above. I can add it

packages/rlc-common/src/metadata/buildPackageFile.ts
jeremymeng: Good point! azureSDKForJs is using `dev-tool run` command which has these baked in already.  Would removing `--reporter ../../../common/tools/mocha-multi-reporter.js` good enough? I think it's for having a couple test result report in JS repo

packages/typespec-ts/src/modular/buildProjectFiles.ts
jeremymeng: is the question here about `../../../common/tools/mocha-multi-reporter.js`? It is in old code too.  should it be removed too?

The reason to remove esm and ts-node/register requires is that for modular we will use the ts-node/esm loader

qiaozha: I am a little curious what's the difference between the two ? what I might know ts-node/register is for commonjs, but I am kind of getting confused why put the esm and ts-node/register together before.
I would be really appreciate it if you can explain more about the background to me.

packages/autorest.typescript/src/generators/static/packageFileGenerator.ts
qiaozha: Thank you :)

packages/typespec-ts/src/modular/buildProjectFiles.ts
jeremymeng: 'ts-node/register' transform .ts to .js but they are still in ESM.  For packages without "type": "module", js files are treated as commonjs, so we need 'esm' to transform ESM test files into commonjs

packages/rlc-common/src/metadata/buildPackageFile.ts
jeremymeng: Mocha will still output the result to the console.  With common/tools/mocha-multi-reporter.js in js-repo, mocha also outputs test results to some files, our build pipelines can then upload them to their associated build so we can see the results in the Tests tab.

packages/autorest.typescript/src/generators/static/packageFileGenerator.ts
jeremymeng: fixed

packages/autorest.typescript/test/smoke/generated/compute-resource-manager/package.json
jeremymeng: fixed

packages/autorest.typescript/test/integration/generated/datasearch/package.json
```diff
@@ -85,7 +87,7 @@
     "unit-test:node": "cross-env TEST_MODE=playback npm run integration-test:node",
     "unit-test:browser": "echo skipped",
     "integration-test": "npm run integration-test:node && npm run integration-test:browser",
-    "integration-test:node": "mocha -r esm --require ts-node/register --timeout 1200000 --full-trace test/*.ts --reporter ../../../common/tools/mocha-multi-reporter.js",
+    "integration-test:node": "cross-env TS_NODE_COMPILER_OPTIONS=\"{\\\"module\\\":\\\"commonjs\\\"}\" mocha -r esm --require ts-node/register --timeout 1200000 --full-trace --reporter ../../../common/tools/mocha-multi-reporter.js test/*.ts",
```
MaryGao, line 90: Could you help me understand why we need below config? My understanding is it is commonjs module by default if without type:module in package.json.

```
cross-env TS_NODE_COMPILER_OPTIONS=\"{\\\"module\\\":\\\"commonjs\\\"}\
```

packages/rlc-common/src/metadata/buildPackageFile.ts
```diff
@@ -288,6 +291,17 @@ function restLevelPackage(model: RLCModel, hasSamplesGenerated: boolean) {
       packageInfo.scripts["integration-test:node"] =
         "dev-tool run test:node-js-input -- --timeout 5000000 'dist-esm/test/**/*.spec.js'";
     }
+    if (isTypeSpecTest) {
```
MaryGao, line 294: I think the condition should be `!azureSdkForJs` because we will get rid of dev-tool dependency not in sdk repo, but not only in typespec test env.

packages/typespec-ts/src/modular/buildProjectFiles.ts
```diff
@@ -265,6 +265,12 @@ export function emitPackage(
       "./dist-esm/test/public/utils/env.js":
         "./dist-esm/test/public/utils/env.browser.js"
     };
+
+    packageInfo["mocha"] = {
```
MaryGao, line 269: Is this config required for SDK repo also?

packages/typespec-ts/src/modular/buildProjectFiles.ts
```diff
@@ -208,8 +208,8 @@ export function emitPackage(
     packageInfo.devDependencies["@azure-tools/test-credential"] = "^1.0.0";
     packageInfo.devDependencies["@azure/identity"] = "^2.0.1";
     packageInfo.devDependencies["@azure-tools/test-recorder"] = "^3.0.0";
-    packageInfo.devDependencies["mocha"] = "^7.1.1";
-    packageInfo.devDependencies["@types/mocha"] = "^7.0.2";
+    packageInfo.devDependencies["mocha"] = "^10.0.0";
```
MaryGao, line 211: I'd like to confirm one thing which may be not relevant to this pr but impact the sample test file generations.

In current dev-tool we check there exists `type: module` in package.json to detect if this is a esm or cjs module, then if it is esm we should all follow the suffix `.js` within imports.

```
import { EventGridManagementClient } from "../src/eventGridManagementClient.js";
```

jeremymeng: > I'd like to confirm one thing which may be not relevant to this pr but impact the sample test file generations.

It would impact source code and test code, if we want to do it correctly in codegen when generate packages with `"type": "module"`.  We should generate relative imports with ".js" extension.

> dev-tool will not do any workaround to mitigate import errors if no js suffix. Is it correct or not?

we can add functionality to dev-tool to help add the .js extension for existing packages (generated or not).  I still think it's better to handled it in the codegen

packages/typespec-ts/src/modular/buildProjectFiles.ts
jeremymeng: openai has similar config in a mocha config file

packages/autorest.typescript/test/integration/generated/datasearch/package.json
jeremymeng: > dev-tool will not do any workaround to mitigate import errors if no js suffix. Is it correct or not?

That's right, ".js" file are consider in cjs format.  But our target in tsconfig is ES2017, which is ESM, so we override it to commonjs when using ts-node.  I don't remember about the exact mocha v10 change but without env variable it gives errors.

packages/autorest.typescript/src/generators/static/packageFileGenerator.ts
```diff
@@ -214,8 +216,7 @@ function regularAutorestPackage(
       packageInfo.scripts["integration-test:node"] =
         "dev-tool run test:node-ts-input -- --timeout 1200000 'test/*.ts'";
     } else {
-      packageInfo.scripts["integration-test:node"] =
-        "mocha -r esm --require ts-node/register --timeout 1200000 --full-trace test/*.ts --reporter ../../../common/tools/mocha-multi-reporter.js";
+      packageInfo.scripts["integration-test:node"] = `cross-env TS_NODE_COMPILER_OPTIONS="{\\\"module\\\":\\\"commonjs\\\"}" mocha -r esm --require ts-node/register --timeout 1200000 --full-trace --reporter ../../../common/tools/mocha-multi-reporter.js test/*.ts`;
```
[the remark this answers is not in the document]
jeremymeng: You are right. only js repo has this.  I probably kept it because the old code has it.  Will remove.

packages/typespec-ts/src/modular/buildProjectFiles.ts
jeremymeng: We could switch to a separate config file in another PR if desired.

packages/rlc-common/src/metadata/buildPackageFile.ts
jeremymeng: This is using the same condition above where we set `"type": "module"` for which we need to add the mocha `ts-node/esm` loader option.  There's no dev-tool or js repo specific code here. The difference is that `-r esm --require ts-node/register` is not needed.

packages/autorest.typescript/src/generators/static/packageFileGenerator.ts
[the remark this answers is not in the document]
jeremymeng: Removed

packages/typespec-ts/src/modular/buildProjectFiles.ts
MaryGao: > we can add functionality to dev-tool to help add the .js extension for existing packages (generated or not). I still think it's better to handled it in the codegen

Yeah sure I also think we should handle it in codegen.

packages/typespec-ts/src/modular/buildProjectFiles.ts
MaryGao: No strong preferences here and I think it's fine if the config is quite simple.

packages/autorest.typescript/test/rlcIntegration/generated/bodyStringRest/package.json
```diff
@@ -34,7 +34,7 @@
     "format": "prettier --write --config ../../../.prettierrc.json --ignore-path ../../../.prettierignore \"src/**/*.ts\" \"*.{js,json}\"  \"test/**/*.ts\" \"samples-dev/**/*.ts\"",
     "generate:client": "autorest --typescript swagger/README.md && npm run format",
     "integration-test:browser": "karma start --single-run",
-    "integration-test:node": "nyc mocha -r esm --require source-map-support/register --reporter ../../../common/tools/mocha-multi-reporter.js --timeout 5000000 --full-trace \"dist-esm/test/{,!(browser)/**/}*.spec.js\"",
+    "integration-test:node": "nyc mocha -r esm --require source-map-support/register --timeout 5000000 --full-trace \"dist-esm/test/{,!(browser)/**/}*.spec.js\"",
```
jeremymeng, line 37: I think this command will fail now without the esm-workaround.js in the js-repro.  However, in the old code it also uses `--reporter ../../../common/tools/mocha-multi-reporter.js` which is from js-repo, so it is broken already.

If we really have usage for this, then generating an esm-workaround.js should be simple enough.

jeremymeng: We would just need to add a generated file and add `-r esm-workaround` to mocha command line options

https://github.com/Azure/azure-sdk-for-js/blob/main/common/tools/esm-workaround.js

packages/typespec-ts/src/modular/buildProjectFiles.ts
jeremymeng: yes, tracked by #1955

packages/autorest.typescript/test/rlcIntegration/generated/bodyStringRest/package.json
MaryGao: Thanks for the info and I think we need to fix this in non-branded azure env and tracked here https://github.com/Azure/autorest.typescript/issues/2055

MaryGao (approved):
Approved.

packages/autorest.typescript/test/rlcIntegration/generated/bodyStringRest/package.json
jeremymeng: Hope we can migrate everything to "type": "module" then workarounds would be no longer needed.

jeremymeng:
@MaryGao @qiaozha thanks for the through review!

PR #2031, opened 2023-09-22
fix-offset-datetime-in-rlc

qiaozha:
fix https://github.com/Azure/azure-sdk-for-js/pull/27087#issuecomment-1720020323


MaryGao (approved):
LGTM

PR #2032, opened 2023-09-22
support-server-parameter-in-modular

qiaozha:
pendIng  https://github.com/Azure/cadl-ranch/pull/425 and https://github.com/Azure/cadl-ranch/pull/424 

In this PR, it contains the following changes:
1. upgrade the cadl ranch version
2. support parameterized host parameter in modular and update the test cases for client structure tests and resilience models, add tests for server path parameters.
3. add request body and response body test cases in encode bytes cases in both RLC and Modular. (Pending some issues)
4. add response headers test case for encode datetime in both RLC and Modular.

fixes https://github.com/Azure/autorest.typescript/issues/2040


packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/EventGridContext.ts
```diff
@@ -16,11 +16,6 @@ export function createEventGrid(
   credential: KeyCredential,
   options: EventGridClientOptions = {}
 ): EventGridContext {
-  const baseUrl = endpoint;
-  options.credentials = {
-    ...options.credentials,
-    apiKeyHeaderName: "SharedAccessKey",
```
qiaozha, line 22: This should be expected as we already set SharedAccessKey as default apiKeyHeaderName in the rest client factory here 
https://github.com/Azure/autorest.typescript/blob/main/packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/rest/eventGridClient.ts#L24-L26

common/config/rush/pnpm-lock.yaml
```diff
@@ -102,28 +102,28 @@ importers:
     devDependencies:
       '@azure-tools/test-recorder': 3.0.0
       '@azure/abort-controller': 1.1.0
-      '@azure/core-util': 1.4.0
+      '@azure/core-util': 1.5.0
       '@azure/core-xml': 1.3.4
-      '@microsoft.azure/autorest.testserver': 3.3.47
-      '@types/chai': 4.3.5
-      '@types/chai-as-promised': 7.1.5
+      '@microsoft.azure/autorest.testserver': 3.3.48
```
qiaozha, line 107: This is the reason why we the changes in `packages/autorest.typescript` folder.

qiaozha: the example key in testserver 3.3.47 version is string_putNull
```
        "x-ms-examples": {
          "string_putNull": {
            "$ref": "./examples/string_getBase64Encoded.json"
          }
        },
```
then changed into string_getBase64Encoded in version 3.3.48
```
       "x-ms-examples": {
          "string_getBase64Encoded": {
            "$ref": "./examples/string_getBase64Encoded.json"
          }
        },
```

qiaozha:
Pending on adding a few test cases for encoding.

packages/typespec-ts/test/modularIntegration/encodeBytes.spec.ts
```diff
@@ -167,4 +169,132 @@ describe("EncodeBytesClient Rest Client", () => {
...
+  describe.skip("request body", () => {
+    const pngFile = readFileSync(
+      resolve("../../packages/typespec-ts/temp/assets/image.png")
+    );
+    it(`should post bytes`, async () => {
+      try {
+        const result = await client.requestBody.default(
+          stringToUint8Array("dGVzdA==", "base64"),
+          {
+            requestOptions: { headers: { "content-type": "application/json" } }
```
qiaozha, line 182: we need to specify contentType as application/json if the body is string type. otherwise, cadl ranch will report parse error from raw body parser. in autorest testserver case, we will generate a contentType parameter like 
```
export interface StringPutNullMediaTypesParam {
  /** Request content type */
  contentType?: "application/json";
}
```
@MaryGao do you have any pointer on why we could generate the contentType in this case but not from the TypeSpec ? Thank you.

MaryGao: We don't have this kind of filter for tsp input. There is no contentType defined in tsp so we will not generate it for them.

I think the issue is introduced in core https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-client-rest/src/sendRequest.ts#L157-L174 when getting body content.

If we specify contentType the core would format as JSON.stringify(body) in our case it would be `'"dGVzdA=="'`; it not, it would be pure string that is `"dGVzdA=="`. 

The former would be a correct json but the latter would not, so cadl-ranch throws exceptions.

packages/autorest.typescript/test/rlcIntegration/generated/bodyStringRest/samples-dev/stringGetBase64EncodedSample.ts
```diff
@@ -14,14 +14,14 @@ dotenv.config();
  * @summary Get value that is base64 encoded
  * x-ms-original-file: /@microsoft.azure/autorest.testserver/swagger/examples/string_getBase64Encoded.json
  */
-async function stringPutNull() {
+async function stringGetBase64Encoded() {
```
MaryGao, line 17: Just wondering why we have these samples changed?

lirenhe (approved):

packages/autorest.typescript/test/rlcIntegration/generated/bodyStringRest/samples-dev/stringGetBase64EncodedSample.ts
qiaozha: https://github.com/Azure/autorest.typescript/pull/2032#discussion_r1337101537 see the comment here

packages/typespec-ts/test/integration/encodeBytes.spec.ts
```diff
@@ -186,4 +187,132 @@ describe("EncodeDatetimeClient Rest Client", () => {
     });
 
   });
-});
\ No newline at end of file
+
+  describe("request body", () => {
+    const pngFile = readFileSync(resolve("../../packages/typespec-ts/temp/assets/image.png"));
```
MaryGao, line 192: Could you help me understand why the case can be passed in ci? I didn't find we have a image.png file in temp folder.

MaryGao: I notice these cases are skipped?

qiaozha: the test cases are skipped but we can read the file because I have changed the string to copy the whole cadl-ranch-specs folder not just http folder https://github.com/Azure/autorest.typescript/pull/2032/files#diff-717e8cb22e747bf40ac8bb73b4111d6e4478be1e4e92f821531028d4e136e766R23 here.

packages/typespec-ts/test/integration/generated/client/structure/default/src/serviceClient.ts
```diff
@@ -7,16 +7,19 @@ import { ServiceClient } from "./clientDefinitions";
...
  * @param options - the parameter for all optional parameters
  */
 export default function createClient(
+  endpoint: string,
   clientParam: string,
   options: ClientOptions = {}
 ): ServiceClient {
   const baseUrl =
-    options.baseUrl ?? `http://localhost:3000/client/structure/${clientParam}`;
+    options.baseUrl ?? `${endpoint}/client/structure/${clientParam}`;
```
MaryGao, line 20: What is the difference for endpoint and baseUrl? 

If they are the same I think we should change to 

```const baseUrl = `${options.baseUrl ?? endpoint}/client/structure/${clientParam}`;
```

qiaozha: @joheredi previously we had an issue here https://github.com/Azure/autorest.typescript/issues/1337 which implies that baseUrl and endpoint are the same thing, and I have this PR https://github.com/Azure/azure-sdk-for-js/pull/21125 to mark the baseUrl as deprecated to encourage customer to use endpoint instead,  but from the core-client-rest https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-client-rest/src/common.ts#L319C1-L322C20, we are still using baseUrl in the RLC, are they not the same thing in the data plane context ? What's your suggestion here ?

qiaozha:
will handle the default value for parameterized host in another issue https://github.com/Azure/autorest.typescript/issues/2049

joheredi (approved):

packages/typespec-ts/test/integration/generated/client/structure/default/src/serviceClient.ts
joheredi: We should use endpoint, I'll have to dig a bit deeper in core-client-rest and see why we kept baseUrl. But in this case I think we should go with `endpoint`

xirzec: Prefer endpoint over baseUrl (it aligns better with our terminology in docs), AFAIK baseUrl is only for compat and can be deprecated/removed when possible

qiaozha: Thanks @xirzec and @joheredi for confirming about this. I want to get things a little more clearer.

First of all, when we say baseUrl and endpoint are the same thing, we are referring to the one in the [ClientOptions here](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-client-rest/src/common.ts#L305) not the endpoint parameter in this case, right ?

Secondly, If the answer for the first question is yes, based on the above comments, our action items would be, 
1. add an endpoint parameter, and mark baseUrl as deprecated in core-client-rest ClientOptions.
2. change the codegen to use the `options.endpoint`  like: 
    ```typescript
      const baseUrl =
        options.endpoint ?? options.baseUrl ??  `http://localhost:3000/client/structure/${clientParam}`;
    ```
    or like the below if we care about the variable name.
    ```typescript
      const endpoint =
        options.endpoint ?? options.baseUrl ??  `http://localhost:3000/client/structure/${clientParam}`;
      // other logic code
      const client = getClient(endpoint, options) as ServiceClient;
    ```

For this specific case, I prefer to interpret the endpoint is just a parameter which is part of the options.endpoint and happens to have the same name as endpoint, we can rename it to any other name if possible, so for @MaryGao's suggestion of changing something like below, it doesn't quite make sense to me.
```
const baseUrl = `${options.baseUrl ?? endpoint}/client/structure/${clientParam}`;
``` 

I prefer to do the codegen side change i.e. action item 2 after action item 1 is completed. @MaryGao let me know if you have other concerns ?

MaryGao: I am okay with endpoint over baseUrl, but still confused about what extract value for endpoint. Say we have a url `http://localhost:3000/client/structure/${clientParam}`  it looks to me the endpoint and baseUrl in options or the endpoint [in typespec](https://github.com/microsoft/typespec/blob/main/packages/http/README.md#examples-11) are the same, that is `http://localhost:3000`.

qiaozha: In typespec example, endpoint refers to the first parameter in the @ server decorator, and the typespec for this test case is like this way
```
@server(
  "{endpoint}/client/structure/{client}",
  "",
  {
    @doc("Need to be set as 'http://localhost:3000' in client.")
    endpoint: url = "http://localhost:3000",

    @doc("Need to be set as 'default', 'multi-client', 'renamed-operation', 'two-operation-group' in client.")
    client: ClientType,
  }
)
```

so using `http://localhost:3000/client/structure/${clientParam}` to represent `options.endpoint` makes more sense to me. 
To me, domain name and ip:port is more like a host instead of endpoint. 

Another point is, it may bring some implementation complexity to codegen if we use options.endpoint to represent the host instead of including the whole url prefix.

qiaozha: As confirmed by @joheredi offline, we should use `http://localhost:3000/client/structure/${clientParam}` to represent `options.endpoint`.

Also, create an issue https://github.com/Azure/autorest.typescript/issues/2050 to track the work items as we discussed.

MaryGao: Sure thing! Thanks for the confirmation!

MaryGao (approved):
Approved and thanks for identifying these details.

qiaozha:
I will merge this first, but @xirzec feel free to let me know if you have different thoughts about the above discussion, we can always revisit this.

PR #2052, opened 2023-10-10
Use node16 in tsconfig.json

timotheeguerin:
Use typespec recommended tsconfig setting https://microsoft.github.io/typespec/extending-typespec/basics#4-create-libts, this is needed to correctly support node exports.

Switch module and moduleResolution to use node16 to tell typescript that it should resolve modules following node 16+ exports

See this stackoverflow for details https://stackoverflow.com/questions/58990498/package-json-exports-field-not-working-with-typescript

Change in last release of typescript removed the `typesVersions` which was the legacy way of specifying the types for additional path. So this setting is required to work correctly with node exports


packages/typespec-ts/tsconfig.json
```diff
@@ -5,8 +5,8 @@
     "composite": true,
     "lib": ["ES2019"],
     "target": "es2019",
-    "module": "NodeNext",
-    "moduleResolution": "NodeNext",
+    "module": "node16",
+    "moduleResolution": "node16",
```
timotheeguerin, line 9: Change to this file shouldn't actually be needed if you prefer to stay to `nodenext`

packages/typespec-ts/tsconfig.test.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
-    "module": "ESNext",
-    "moduleResolution": "Node"
+    "module": "node16",
+    "moduleResolution": "node16"
```
timotheeguerin, line 5: This is now getting into some conflict where the generated tests are not compatible with node16 imports

packages/typespec-ts/test/integration/generated/azure/core/src/clientDefinitions.ts
timotheeguerin, line 1: Not sure how this is related to this change, does it update the typespec dependency automatically or something?

packages/typespec-ts/tsconfig.json
```diff
@@ -5,8 +5,8 @@
     "composite": true,
     "lib": ["ES2019"],
     "target": "es2019",
-    "module": "NodeNext",
-    "moduleResolution": "NodeNext",
+    "module": "node16",
```
MaryGao, line 8: The issue we found happens in dev version `"@azure-tools/typespec-azure-core": "0.35.0-dev.11"` and see pr https://github.com/Azure/autorest.typescript/pull/2053. And in our main we use the stable version not dev one.

I may suggest you to directly create a pr based on my branch. So we could verify it works or not.

timotheeguerin: That change should happen regardless of the version of typespec that you are using. You might get into the same issue with other packages in the future. So I would just merge that if it is not causing any issues anyway. If there is more problem to the dev version then we can investigate further.

MaryGao (approved):
LGTM

PR #2053, opened 2023-10-11
upgrade to the latest dev

MaryGao:


MaryGao:
tested and the vscode intellisense work as expected.
image

PR #2054, opened 2023-10-11
support additional properties in RLC

### LINKED ISSUE #2033: Models not generated in typescript SDK
Hi team,

I'm using `@azure-tools/typespec-ts` emitter to generate RLC SDK from typespec files directly but facing an issue where some of the models are not generated in the RLC SDK. This can be reproduced by using the below typespec. `Identifers` and `SearchFilter` were not generated even though they are referenced by another property. 
```
import "@typespec/openapi";
import "@typespec/http";
import "@typespec/json-schema";
using TypeSpec.Http;
using OpenAPI;
using TypeSpec.JsonSchema;

@service({
  title: "Sample API",  
  termsOfService: "http://swagger.io/terms/",
  version: "0.1",
})
@server("https://example.com", "Single server endpoint")

@route("/")
namespace SampleAPI {
    @route("/pets")
    interface PetOps {
        @get list(...ItemPerPage, @query search?: SearchFilter, ...FieldsFilter) : {@header Schema: string;...CustomResponse} | Errors;
    }

    @route("/cat")
    interface CatOps extends ResourceService<CatData,CustomResponse> {}
}

interface ResourceService <CatData, CustomResponse> {
    @doc("Read resource")
    @get read(...CatData): {@header Schema: string;...CustomResponse}| Errors;
}

@jsonSchema
model CatData is Resource <"CatData"> {}

model Resource <resourceType extends string> extends Identifiers{
    @visibility("read") type: resourceType;
    @visibility("read") id: string;
    summary: string;
    tags: string[];
}  

model CustomResponse {
    body: string;
    id: string;
    name: string;
}

model Identifiers {
    @visibility("read") type: string;
    id?: string;
    space?: string;
    name?: string;
}
model ItemPerPage {
    @query
    @minValue(1)
    @maxValue(100)
    itemPerPage: int32 = 50,
}
model SearchFilter{
    names: string[],
    codes: string[],
    tags: string[],
}
model FieldsFilter{
    @query fields?: string
}
alias Errors = {
@statusCode statusCode: 400 | 401 | 403 | 404 | 405 | 406 | 408 | 410 |  428 | 429 |
                        500 | 501 | 503 ; ...ServiceErrorResponse};

model ServiceErrorResponse {
    error: InnerErrorResponse
}

@error
model InnerErrorResponse {
    code: int32 | null;
    message: string | null;
    id: string | null,
    status: string | null
}

```
Dependencies (please note that I didn't use the latest version with @typespec/compiler v0.48 because it's throwing out error when compiling)
```
    "@typespec/compiler": "^0.47.0",
    "@typespec/json-schema": "^0.47.0",
    "@typespec/openapi3": "^0.47.0",
    "@typespec/rest": "^0.47.0",
    "@azure-tools/typespec-client-generator-core": "~0.33.0",
    "@azure-tools/typespec-azure-core": "~0.33.0",
    "@azure-tools/typespec-ts": "~0.15.0"
```
Would be appreciated if someone could take a look :) And please let me know if you need more information on this.

qiaozha:
fixes #2033 

Also fix the boolean/number/string literal type error in the modular layer.


qiaozha:
As, the VC interface in the below option1 and option2 has the same effect.
```typescript
interface C {
  prop1: string;
  prop2: boolean
}
// option 1
interface VC extends Record<string, C[]> {}

// option 2
interface VC {
  [prop: string]: C[]
}
```

And we are using type index like option 2 in current HLC. 

Also,the implementation of Record is actually defined to use type index in a way. 
```typescript
type Record<K extends keyof any, T> = {
    [P in K]: T;
};
```

I will  use option 2 to express additional properties in the generated code. 

@MaryGao @joheredi let me know if you have other thoughts.

joheredi:
I would prefer we go with option 1, the main reason is documentation as it is more readable and self documenting than the indexer syntax. They are equivalent and shouldn't cause a breaking change if we move to Record

MaryGao:
I have no concern if they can express the same functionality. And no strong preference here, both work for me.

packages/typespec-ts/src/modular/helpers/typeHelpers.ts
```diff
@@ -38,7 +38,7 @@ export function getType(type: Type, format?: string): TypeMetadata {
     case "boolean":
       return { name: getNullableType(type.type, type) };
     case "constant": {
-      let typeName: string = type.value ?? "undefined";
+      let typeName: string = type.value?.toString() ?? "undefined";
```
MaryGao, line 41: What is the difference with and without toString()?

packages/rlc-common/src/helpers/schemaHelpers.ts
```diff
@@ -3,9 +3,14 @@
 
 import { Schema } from "../interfaces.js";
 
-export function isDictionarySchema(schema: Schema) {
+export function isDictionarySchema(
+  schema: Schema,
+  filterEmpty: boolean = false
```
MaryGao, line 8: Could you help me understand what role the `filterEmpty` takes?

packages/rlc-common/src/buildObjectTypes.ts
```diff
@@ -363,24 +363,34 @@ function getImmediateParentsNames(
 
   const extendFrom: string[] = [];
 
-  // If an immediate parent is a DictionarySchema, that means that the object has been marked
+  // If an immediate parent is an empty DictionarySchema, that means that the object has been marked
   // with additional properties. We need to add Record<string, unknown> to the extend list and
-  if (objectSchema.parents.immediate.find(isDictionarySchema)) {
+  if (
```
MaryGao, line 368: > with additional properties. We need to add Record<string, unknown> to the extend list and
Just curious the comment says that the additional property is supported in modular?

MaryGao: Just notice this is rlc code.

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -617,27 +617,7 @@ function getSchemaForModel(
...
-  ) {
-    // Take the base model schema but carry across the documentation property
-    // that we set before
-    const baseSchema = getSchemaForType(dpgContext, model.baseModel, usage);
-    modelSchema = {
-      ...baseSchema,
-      description: modelSchema.description
-    };
-  } else if (model.baseModel) {
+  if (model.baseModel) {
```
MaryGao, line 620: Could you help me understand why we have this change? We don't need to handle template model by ourselves?

packages/typespec-ts/src/modular/helpers/typeHelpers.ts
qiaozha: If the type is literal type like true, what we get for typeName is not a string but a true value. though it's claimed to be a string, but it doesn't do convert by itself.

packages/typespec-ts/src/utils/modelUtils.ts
qiaozha: In the case of `model A extends Record<B> {}`, we will run into this condition, which will return a model without name, and we will skip the generation of the model A. As I really don't know what this is trying to do, and I found there's no such logic in OpenAPI3 emitter, so just removed it and found nowhere is reporting errors about it.

packages/typespec-ts/test/unit/modelsGenerator.spec.ts
```diff
@@ -847,6 +840,448 @@ describe("Input/output model type", () => {
...
+          id: string;
+        }`
+      );
+    });
+
+    it("should handle model additional properties is from record of object", async () => {
+      const schemaOutput = await emitModelsFromTypeSpec(`
+      model VegetableCarrot is Record<Carrots> {
+        testProp: Carrots
+      }
```
qiaozha, line 1156: not sure if it's appropriate to generate Vegetables as 
```ts
        export interface Vegetables {
          carrots: Record<string, Carrots>;
          beans: Record<string, Beans>;
        }
```

qiaozha: @joheredi in the case of is keyword being used in record to express additional properties,
we currently just drop the model `VegetableCarrot` generation, use the `Record<string, Carrots>` to replace it. which has information loss about VegetableCarrot's own properties. And according to Timothee's comment here https://github.com/Azure/cadl-ranch/pull/330#issuecomment-1555035443 we should not treat extends and is as the same thing, 
One possible solution here is to use indexer syntax here to express additional properties for is Record scenario. like we should generate model VegetableCarrot as 
```ts
export interface VegetableCarrot {
  [prop: string]: Carrots;
  testProp: Carrots;
}
```
How do you think ? feel free to come up with different solutions here :)

joheredi: I think we should still use Record when setting additional properties

```ts
export interface VegetableCarrot {
  [prop: string]: Carrots;
  testProp: Carrots;
}
```

is identical to

```ts
export interface VegetableCarrot extends Record<string, Carrots> {
  testProp: Carrots;
}
```

But using records is better for documenting. If I understand correctly Timothee's comment is in the context of TypeSpec's `is` vs `extends` which I think makes sense in that context and will have implications when emitting OpenAPI, but I also think modeling the TypeScript interface extending record is the best way we can represent it for our customers.



When additional properties are not in the picture then if we saw

```ts
model Foo is Record<Carrots>;
```

we  should generate

```ts
type Foo = Record<string, Carrots>;
```

and for 

```ts
model Foo extends Carrots {
   foo: string;
}
```

```ts
interface Foo extends Carrots {
   foo: string
}
```

qiaozha: it's a little strange to handle `is` differently based on whether there's extra properties in that model.
Another concern is, this is how we handle `is` in normal model context.
```
model Foo is Carrots {
  prop: string;
};
model Carrots {
  prop1: numeric;
};
```
and we will not generate model Carrots, just generate model Foo like 
```ts
interface Foo {
   prop: string;
   prop1: number;
}
```

packages/rlc-common/src/helpers/schemaHelpers.ts
joheredi, line 8: minor comment: I usually prefer to add an optional options bag to hold the single filterEmpty option as it helps with readability

for example

```ts
export interface IsDictionaryOptions {
   filterEmpty?: boolean;
}

export function isDictionarySchema(
  schema: Schema,
  options?: IsModelOptions = {filterEmpty: false}
)
``` 

This produces a call like this

```ts
isDictionary(scheman, {filterEmpty: true});

//vs
isDictionary(scheman, true);
```

It helps when reading the code later on, you don't have to go to the function definition to see what the boolean value means

qiaozha: Thanks for the suggestion, I updated.

MaryGao (approved):
approved.

PR #2056, opened 2023-10-13
fix-binary-in-multipart-form-data

### LINKED ISSUE #2045: Support @encode("binary")

qiaozha:
fixes #2045 
Currently, we will try to treat binary bytes the same as other encoding to use stringToUint8Array and unit8ArrayToString util to convert it, which is incorrect, 

For binary in RLC layer,
if it's in input model, we should generate it as `string | Uint8Array | ReadableStream<Uint8Array> | NodeJS.ReadableStream`, 
if it's in output model, we will generate it as `Uint8Array`

For binary in Moular layer, current implemtation is treat it as Uint8Array regardless input or output. 

Another change in this pr is that, previously, we will speculate if a bytes by default means binary data in the context of multipart/form-data request. As this is not yet decided by https://github.com/microsoft/typespec/issues/2370, we will not do such assumption.

One more thing here, in the past swagger generation, because there's no way to express all those parts together in the multipart/form-data request, we have to manual build the body parameter, but in typespec, we naturally support to use model to group all parts for multipart/form-data request, we don't need this manual build body parameter step now. that's why I delete the `transformMultiFormBody` function in parameters transform. 
If that model is defined anonymously, we will just generate a anonymously model.
If that model has a name, we will just give it a name and put it into the models.ts


packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -222,20 +222,74 @@ export function includeDerivedModel(model: Model): boolean {
...
   }
-  return applyIntrinsicDecorators(program, scalar, result);
+  const withDecorators = applyEncoding(
+    dpgContext,
+    scalar,
+    result
+      ? applyIntrinsicDecorators(dpgContext.program, scalar, result)
+      : undefined
+  );
+  if (withDecorators.type === "string" && withDecorators.format === "binary") {
```
lirenhe, line 287: Can this be the generic handling for format equals binary?  Any other types that we need special handling?

qiaozha: First of all, bytes with any other encodings like base64 or base64url are just pure string in RLC, we only need to special handle binary bytes in both standard scalar and non-standard scalar. this is the place where we handle non-standard scalar and I also add similar logic in standard scalar here https://github.com/Azure/autorest.typescript/pull/2056/files#diff-b9d6429c3d7eab57015f06928f591a73117fa57b586c4eb7f1bdd7c1ec2e16d9R1047-R1057 
Secondly, the type we get from typespec for `@encode("binary")` is string no matter what the original type is. 
if we have something like 
```
  @encode("binary")
  weight: int32;
```
we also get in OpenAPI3 emitter.
```yaml
        weight:
          type: string
          format: binary
```

packages/typespec-ts/test/unit/parametersGenerator.spec.ts
```diff
@@ -318,11 +317,14 @@ describe("Parameters.ts", () => {
     it("contentTypes has array data defined in form body", async () => {
       const parameters = await emitParameterFromTypeSpec(
         `
+        @encode("binary")
```
MaryGao, line 320: What if there is no `@encode("binary")` for form-data?

qiaozha: Then it would just be a normal request.

lirenhe (approved):

joheredi (approved):

Issue #2059, opened 2023-10-16
Use `dev-tool run bundle` in build scripts


deyaaeldeen:
See https://github.com/Azure/azure-sdk-for-js/commit/72d460d5ba9afd37526e4e44d3ef5ed4f24e2cfd

We upgraded rollup to v3 and we now rely on dev-tool to bundle all rush-managed libraries. For packages generated to live inside the mono repo, they should use the `dev-tool run bundle` command in the build scripts and don't include rollup in their dev dependencies.

deyaaeldeen:
Addressed

Issue #1703, opened 2022-12-15
CI failed randomly with error `Error: read ECONNRESET at TLSWrap.onStreamRead`


MaryGao:
CI failed randomly with error `Error: read ECONNRESET at TLSWrap.onStreamRead` and every time re-run would fix it.

https://dev.azure.com/azure-sdk/public/_build/results?buildId=2034829&view=logs&j=91420f01-9daf-53d3-aec1-75f797de9149&t=3aed5a45-624c-53bb-998d-671906525409&l=274

MaryGao:
see https://github.com/Azure/autorest/issues/4322

tadelesh:
Got same issue for Go. Seems throttling issue.

PR #2060, opened 2023-10-16
Use `dev-tool run bundle` in generated azure-sdk-for-js packages

jeremymeng:
for bunding.

- update test baselines.

#2059


deyaaeldeen (approved):
Thanks! we should also update the `files` list to not include the `rollup.config.js` file.

packages/autorest.typescript/test/integration/generated/appconfigurationexport/package.json
```diff
@@ -18,17 +18,12 @@
...
-    "@rollup/plugin-node-resolve": "^13.1.3",
     "mkdirp": "^2.1.2",
-    "rollup": "^2.66.1",
-    "rollup-plugin-sourcemaps": "^0.6.3",
     "typescript": "~5.0.0",
     "uglify-js": "^3.4.9",
     "rimraf": "^5.0.0",
-    "dotenv": "^16.0.0"
+    "dotenv": "^16.0.0",
+    "@azure/dev-tool": "^1.0.0"
```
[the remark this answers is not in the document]
jeremymeng: yes it's only added when `azureSdkForJs` is true https://github.com/Azure/autorest.typescript/pull/2060/files/d9cba5b1c8aa5729ae34f84c7b8115d9018150ae#diff-c0268d76bd99f6fb8f8cdb15c036f8df312c03e7adfb022f56bad59beb5d68f0R186

jeremymeng:
The smoke tests are failing because they do npm install and npm build, however, installation fails to get `@azure/dev-tool`.  Any suggestion to fix these? These tests have the AzureSdkForJs option.

jeremymeng:
I changed smoke tests to not test for azure-sdk-for-js since dev-tool is not available outside of js-repo. Yeah it's unfortunate that we cannot test the scenario, but hopefully we can fix things quickly whenever generated packages failed to build in the js repo.

packages/autorest.typescript/src/generators/static/packageFileGenerator.ts
```diff
@@ -189,6 +182,19 @@ function regularAutorestPackage(
...
+    packageInfo.scripts["build"] =
+      "npm run clean && tsc && dev-tool run bundle && npm run minify && mkdirp ./review && npm run extract-api";
+  } else {
+    packageInfo.devDependencies["@rollup/plugin-commonjs"] = "^24.0.0";
+    packageInfo.devDependencies["@rollup/plugin-json"] = "^6.0.0";
+    packageInfo.devDependencies["@rollup/plugin-multi-entry"] = "^6.0.0";
+    packageInfo.devDependencies["@rollup/plugin-node-resolve"] = "^13.1.3";
+    packageInfo.devDependencies["rollup"] = "^2.66.1";
+    packageInfo.devDependencies["rollup-plugin-sourcemaps"] = "^0.6.3";
+  }
```
MaryGao, line 196: Thanks for contributing this! Could you help understand below questions?

- Does this apply for both RLC, HLC and Modular?
- Doest this change apply for both swagger and typespec path?

If yes I think we still have changes in rlc-common and typespec-ts.

packages/autorest.typescript/test/integration/generated/appconfiguration/package.json
```diff
@@ -18,17 +18,12 @@
...
-    "@rollup/plugin-node-resolve": "^13.1.3",
     "mkdirp": "^2.1.2",
-    "rollup": "^2.66.1",
-    "rollup-plugin-sourcemaps": "^0.6.3",
     "typescript": "~5.0.0",
     "uglify-js": "^3.4.9",
     "rimraf": "^5.0.0",
-    "dotenv": "^16.0.0"
+    "dotenv": "^16.0.0",
+    "@azure/dev-tool": "^1.0.0"
```
MaryGao, line 26: Could you help understand why this change happens in `not azureSDKForJs` env?

https://github.com/Azure/autorest.typescript/blob/main/packages/autorest.typescript/test/commands/test-swagger-gen.ts#L35-L45

packages/autorest.typescript/test/smoke/generated/arm-package-deploymentscripts-2019-10-preview/package.json
```diff
@@ -21,17 +21,17 @@
...
     "@rollup/plugin-multi-entry": "^6.0.0",
     "@rollup/plugin-node-resolve": "^13.1.3",
-    "mkdirp": "^2.1.2",
     "rollup": "^2.66.1",
-    "rollup-plugin-sourcemaps": "^0.6.3",
-    "typescript": "~5.0.0",
-    "uglify-js": "^3.4.9",
-    "rimraf": "^5.0.0",
-    "dotenv": "^16.0.0"
+    "rollup-plugin-sourcemaps": "^0.6.3"
```
MaryGao, line 34: > The smoke tests are failing because they do npm install and npm build, however, installation fails to get @azure/dev-tool. Any suggestion to fix these? These tests have the AzureSdkForJs option.

Is the issue existing? I assume in non js-repo there will have no real change just order change.

packages/autorest.typescript/test/smoke/generated/arm-package-deploymentscripts-2019-10-preview/package.json
```diff
@@ -50,7 +50,6 @@
     "src/**/*.ts",
     "README.md",
     "LICENSE",
-    "rollup.config.js",
```
MaryGao, line 53: Is the deletion of this line expected in non js-repo?

jeremymeng: Yes, there's no value in packing rollup config file.  it's for dev environment only.

packages/autorest.typescript/src/generators/static/packageFileGenerator.ts
jeremymeng: Yes you are right,  any packages in the js repo should use dev-tool to bundle.

packages/autorest.typescript/test/smoke/generated/arm-package-deploymentscripts-2019-10-preview/package.json
jeremymeng: Prior to this change it doesn't matter because all generated use `rollup` so works fine with npm install.  Now package generated for js repo would failed to install. the old `AzureSdkForJs` option (`--AzureSdkForJs=false`) really is NotAzureSdkForJs so I renamed it and added it for all smoke tests.

packages/autorest.typescript/src/generators/static/packageFileGenerator.ts
jeremymeng: Turns out that you and Qiaoqiao already use `dev-tool run bundle` for js repo in rlc-common and typespec-ts :)

jeremymeng:
@MaryGao @qiaozha any other feedback?

packages/autorest.typescript/test/integration/generated/appconfiguration/package.json
```diff
@@ -47,14 +42,13 @@
...
     "LICENSE",
-    "rollup.config.js",
     "tsconfig.json",
     "review/*",
     "CHANGELOG.md",
     "types/*"
   ],
   "scripts": {
-    "build": "npm run clean && tsc && rollup -c 2>&1 && npm run minify && mkdirp ./review && npm run extract-api",
+    "build": "npm run clean && tsc && dev-tool run bundle && npm run minify && mkdirp ./review && npm run extract-api",
```
qiaozha, line 51: I am a little confused, this change is for azure-sdk-for-js: true  and here we are changing the integration test. right ? Has the integration test enabled azure-sdk-for-js: true ?

jeremymeng: From what I can find, this test doesn't specify azureSdkForJs option

https://github.com/Azure/autorest.typescript/blob/main/packages/autorest.typescript/test/commands/test-swagger-gen.ts#L628

By default we generate for JS repo

https://github.com/Azure/autorest.typescript/blob/c3d9f9bf3c5d45ff9fea680fe73918428e7721ed/packages/autorest.typescript/src/utils/autorestOptions.ts#L123

jeremymeng:
What command do I need to run to update the failing RLC smoke tests? I did `smoke-test` locally but don't see any change afterwards

qiaozha:
I think it's a known issue https://github.com/Azure/autorest.typescript/issues/1703, unfortunately, we don't have time to investigate this for now.

Issue #2067, opened 2023-10-17
Dpg - Generate anonymous models inline


joheredi:
TypeSpec allows defining anonymous models:

```ts
model Foo {
   bar: {
       baz: string;
    }
}
```

TypeScript allows generating this so we should generate similar to the TypeSpec definition

```ts
interface Foo {
   bar: {
       baz: string;
    }
}
```

This will solve some issues identified:

- Problem with TypeScript Emitter not generating anonymous or inline objects
- Issue with inline models not serializing. Workaround: Handwrite model. 
- output model for Error (=> ErrorModel) is not being generated. Workaround: Handwrite model which also solves the isUnexpected issue. 
- Issue with isUnexpected not narrowing down types in functions like _createFineTuningJobDeserialize.

PR #2072, opened 2023-10-19
Anonymous models inline support

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2067


packages/typespec-ts/test/modularUnit/anonymousModel.spec.ts
```diff
@@ -0,0 +1,671 @@
...
+        assertEqualContent(
+          modelFile?.getFullText()!,
+          `
+        export interface ReturnBody {
+          emptyAnomyous: {};
+          emptyAnomyousArray: {}[];
+          emptyAnomyousDict: Record<string, {}>;
+          emptyModel: EmptyModel;
+          emptyModelArray: EmptyModel[];
+          emptyModelDict: Record<string, EmptyModel>;
```
MaryGao, line 536: This is the generated model for empty anonymous/named model now we don't treat them as `Record<string, any>`.

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -792,18 +792,22 @@ function emitBasicOperation(
   }
 
   let bodyParameter: any | undefined;
+  // FIXME handle the empty body case
```
joheredi, line 795: Are we planning to cover this in the current PR or later work? If later, let's file an issue to track

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -792,18 +792,22 @@ function emitBasicOperation(
...
-      if (bodyParameter.type.properties.length > 0) {
-        for (const param of bodyParameter.type.properties) {
-          // const emittedParam = emitParameter(context, param.type, "Method");
-          param.implementation = "Method";
-          param.location = param.location ?? "body";
-          parameters.push(param);
-        }
+    // Flatten the body parameter if it is an anonymous model
+    if (
+      bodyParameter.type.type === "model" &&
```
joheredi, line 802: Can we use isEmptyModel here?

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -792,18 +792,22 @@ function emitBasicOperation(
...
+      bodyParameter.type.name === "" &&
+      bodyParameter.type.properties.length > 0
+    ) {
+      for (const param of bodyParameter.type.properties) {
+        // const emittedParam = emitParameter(context, param.type, "Method");
```
joheredi, line 807: if not needed, lets remove the commented out code

joheredi (approved):
Looks great, left a few minor comments. I'm really excited to see this working now :smile:

packages/typespec-ts/src/modular/buildCodeModel.ts
MaryGao: No this is anonymous but with properties

packages/typespec-ts/src/modular/buildCodeModel.ts
MaryGao: I don't think we don't need this comment more

packages/typespec-ts/test/modularUnit/anonymousModel.spec.ts
```diff
@@ -0,0 +1,973 @@
...
+            options
+          );
+          return _readDeserialize(result);
+        }`,
+          true
+        );
+      });
+    });
+    describe("happens as a property in response body", async () => {
+      it("should map empty anonymous  model({}) => Record<string, any> & empty named model(A {}) => Record<string, any>", async () => {
```
MaryGao, line 815: I would call out for the test case where 
- We treat the anonymous empty model `{}` as `Record<string, any>` the benefit is the customers could assign objects with properties but will take pontential breaking if the contract is changed to add optional properties in e.g `{ a?: string }`

packages/typespec-ts/test/modularUnit/anonymousModel.spec.ts
```diff
@@ -0,0 +1,973 @@
...
+        // Model name referred in operations.ts
+        verifyReturnTypeAsEmpty(
+          operationFiles?.[0]?.getFullText()!,
+          "PublishResult"
+        );
+      });
+
+      it("should return non-empty anonymous  model({ ... })", async () => {
+        const tspContent = `
+        op read(): { foo?: {bar: string | null}};
```
MaryGao, line 765: Please notice if the return type is anonymous the inline type would be duplicated in different method.

```typescript


export async function _readDeserialize(
     result: Read200Response
): Promise<{ foo?: { bar: string | null } }>

export async function read(
    context: Client,
    options: ReadOptions = { requestOptions: {} }
): Promise<{ foo?: { bar: string | null } }>
```

packages/typespec-ts/test/modularUnit/anonymousModel.spec.ts
```diff
@@ -0,0 +1,973 @@
...
+          context: Client,
+          options: ReadOptions = { requestOptions: {} }
+        ): Promise<ReturnBody> {
+          const result = await _readSend(context, options);
+          return _readDeserialize(result);
+        }`
+        );
+      });
+
+      it("should map non-empty anonymous  model({ ... }) => { ... }", async () => {
```
qiaozha, line 895: I wonder if we should add some ut for non-empty anonymous model like above empty model's case ?
          nonemptyAnomyous: {...};
          nonemptyAnomyousArray: {...}[];
          nonemptyAnomyousDict: Record<{...}>;

qiaozha (commented):
wonder if you have added test for the sample generation with anonymous model ?

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/rest/models.ts
```diff
@@ -186,7 +186,7 @@ export interface ChatCompletionRequestMessage {
    */
   name?: string;
   /** The name and arguments of a function that should be called, as generated by the model. */
-  function_call?: object;
```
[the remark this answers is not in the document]
MaryGao: this is because we have logic to fall back to schema.type and then it would be `object`. Generally this only works for fixed enum schema or some constants schema and its type would be relevant values.

MaryGao:
> wonder if you have added test for the sample generation with anonymous model ?

Added here https://github.com/Azure/autorest.typescript/pull/2072/commits/abec42aae63673470772f708abe30c6ca09b7af0

packages/typespec-ts/test/modularUnit/anonymousModel.spec.ts
MaryGao: updated

Issue #2066, opened 2023-10-17
Dpg - Generate namespace hierarchy


joheredi:
This involves generating a hierarchy structure for DPG - Classical and Modular layers.

Here's the discussion on this topic: https://github.com/Azure/autorest.typescript/issues/2035#issuecomment-1759879287

PR #2080, opened 2023-10-24
Support hierarchy subpath

### LINKED ISSUE #2035: Ensure all operations (non-overloads) have unique names
Consider the following definitions for two APIs:

```
@route("/chat")
namespace Chat {
  @route("/completions")
  interface Completions {
    @tag("OpenAI")
    @post
    @operationId("createChatCompletion")
    create(
      @body body: CreateChatCompletionRequest,
    ): CreateChatCompletionResponse | ErrorResponse;
  }
}
@route("/completions")
interface Completions {
  @tag("OpenAI")
  @post
  @operationId("createCompletion")
  create(
    @body body: CreateCompletionRequest,
  ): CreateCompletionResponse | ErrorResponse;
}
```

Both operations are named `create` and the generated code names corresponding methods `create` which is not valid:
```ts
  create(
    model:
      | string
      | "gpt4"
      | "gpt-4-0314"
      | "gpt-4-0613"
      | "gpt-4-32k"
      | "gpt-4-32k-0314"
      | "gpt-4-32k-0613"
      | "gpt-3.5-turbo"
      | "gpt-3.5-turbo-16k"
      | "gpt-3.5-turbo-0301"
      | "gpt-3.5-turbo-0613"
      | "gpt-3.5-turbo-16k-0613",
    messages: ChatCompletionRequestMessage[],
    options: CreateOptions = { requestOptions: {} }
  ): Promise<CreateChatCompletionResponse> {
    return create(this._client, model, messages, options);
  }

create(
    model:
      | string
      | "babbage-002"
      | "davinci-002"
      | "text-davinci-003"
      | "text-davinci-002"
      | "text-davinci-001"
      | "code-davinci-002"
      | "text-curie-001"
      | "text-babbage-001"
      | "text-ada-001",
    prompt: string | string[] | number[] | number[],
    bestOf: number | null,
    options: CreateOptions = { requestOptions: {} }
  ): Promise<CreateCompletionResponse> {
    return create(this._client, model, prompt, bestOf, options);
  }
```

Causing this TS error:
```
src/OpenAIClient.ts(62,3): error TS2393: Duplicate function implementation.
```

qiaozha:
fixes #2035 #2066

This PR is for hierarchy client support, document some agreements we have:
1. `hierarchyClient` will be enabled by default, unless it's specified as false.
1. If `hierarchyClient` is true, we will generate hierarchy client in the classic layer. every namespace and interface will be considered as a layer.
2. if `hierarchyClient` is true, generate hierarchy client subpath in the api layer. customer will no longer be able to access the deep layered apis from the top ./api subpath export.
3. we will generate models with full namespace as prefix if we detect there's model name conflict or `enableModelNamespace` is true.
4. we will generate hierarchy folder structure in the api layer, and generate classic folder to wrapper the classical client calls to the api layer. which has the same hierarchy folder structure as the api layer.  which is also true if `hierarchyClient` is false and `enableOperationGroup` is true, we will just not export the api subpath in that case.
5. we will generate all the operation options in the models/options.ts and if there're any name conflicts, we will add full namespace as prefix to avoid them. Though I tried to put them in the api hierarchy subpath to avoid resolve the conflicts with namespace prefix, but the api extractor doesn't look happy and we have to export all the operation options to the top level index.ts.
6. If `hierarchyClient` is false, and `enableOperationGroup` is false, `enableModelNamespace` is also false, there should be no code change, which can be proved in those `hierarchyClient` false cases.

About testing,
I have enabled `hierarchyClient` true for three smoke test, `hierarchy_generic`, `openai_generic` and `widget_dpg`, 
And disabled `hierarchyClient` for all the rest smoke tests and modular integration tests. The change for modularIntegration tests are code reorganize as point out in the above point 4. Their API doesn't change which can be proved from no integration tests change.


packages/typespec-test/test/anomalyDetector/generated/typespec-ts/src/models.ts
```diff
@@ -2,14 +2,14 @@
 // Licensed under the MIT license.
 
 /** The request of entire or last anomaly detection. */
-export interface UnivariateDetectionOptions {
+export interface UnivariateUnivariateDetectionOptions {
```
qiaozha, line 5: This is the change we get if we by default add the namespace as prefix for models.

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/audio/transcriptions/index.ts
```diff
@@ -0,0 +1,62 @@
...
+import {
+  StreamableMethod,
+  operationOptionsToRequestParameters,
+} from "@azure-rest/core-client";
+import { CreateTranscriptionOptions } from "../../../models/options.js";
+
+export function _createTranscriptionSend(
+  context: Client,
+  audio: CreateTranscriptionRequest,
+  options: CreateTranscriptionOptions = { requestOptions: {} }
```
qiaozha, line 23: We need to figure out a way to rename the operation options if there's operation name conflict even in the hierarchy subpath because currently we generate all the options in the `src/models/options.ts` file. Currently we name the options type as `OperationName + Options`.

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/audio/translations/index.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (isUnexpected(result)) {
+    throw result.body;
+  }
+
+  return {
+    text: result.body["text"],
+  };
+}
+
+export async function createTranslation(
```
qiaozha, line 54: I noticed that in the https://github.com/joheredi/unbranded_lib/blob/main/src/api/audio/translations/index.ts#L51 manual code, the function here is called `create`, I wonder if this is from previous typespec or we should generate something to replace the duplicated part into a empty string if it's the same as the interface name ?

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/index.ts
```diff
@@ -6,33 +6,3 @@ export {
...
-  listFineTuneEvents,
-  cancelFineTune,
-  listModels,
-  retrieve,
-  deleteOperation,
-  createImage,
-  createImageEdit,
-  createImageVariation,
-  createModeration,
-} from "./operations.js";
```
qiaozha, line 38: In this example, we don't have client level operation i.e. operations belongs to the global namespace directly, I wonder where should we put the client level operations in the context of hierarchy subpath ?

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/models/index.ts
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) Microsoft Corporation.
```
qiaozha, line 1: For this file name, this operation are indeed inside an interface called `Models`. I feel like it's kind of confusing given models means where we put model types. Wonder if codegen should do something about it ?

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/models/index.ts
```diff
@@ -0,0 +1,135 @@
...
+    object: result.body["object"],
+    deleted: result.body["deleted"],
+  };
+}
+
+/**
+ *  @fixme delete is a reserved word that cannot be used as an operation name. Please add @projectedName(
+ *       "javascript", "<JS-Specific-Name>") to the operation to override the generated name.
+ */
+export async function deleteOperation(
```
qiaozha, line 128: this is renamed as the original operation name is `delete` which is a reserve word.

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/OpenAIClient.ts
```diff
@@ -115,207 +121,185 @@ export class OpenAIClient {
...
-    return createImageVariation(this._client, image, options);
-  }
-
-  createModeration(
-    content: CreateModerationRequest,
-    options: CreateModerationOptions = { requestOptions: {} }
-  ): Promise<CreateModerationResponse> {
-    return createModeration(this._client, content, options);
-  }
+  audioTranscriptions = {
```
qiaozha, line 124: This is not correctly implemented yet, it should be `audio { transcriptions : {}}`. 
But I notice that we generate a file for each operation group and put that file into src folder directly, I wonder if we should put them into some folder ? Also, do we only generate file for first layer operation group ? what if there's more deeper layer than 3 or 4 ?

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/audio/transcriptions/index.ts
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Microsoft Corporation.
```
qiaozha, line 1: Is it really neccessary to have a folder for each interface ? And it doesn't look like will have another file other than the index.ts right ? Could we just generate a file like `src/api/audio/transcriptions.ts` for this interface but still export this file as an independent subpath ?

qiaozha (commented):
Initial review for some implementation details to be confirmed.

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/index.ts
qiaozha: Another question is, in the context of client level operation, which subpath should we export those operations in the api layer ?

packages/typespec-test/test/anomalyDetector/generated/typespec-ts/src/models.ts
joheredi: Do you know how expensive it would be to implement this only when there are conflicts? If it is not too expensive we can go that route and have a flag in case we want to turn it on/off

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/audio/translations/index.ts
joheredi: We should use the name in TypeSpec. The manual code I had was assuming typespec called `create`. Operations don't have to be globally unique, just unique in the namespace.

joheredi: I'm fine appending something to the interface name to avoid duplicates, such as operation hierarchy

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/index.ts
joheredi: I was thinking client level operations remain as we have them today. Would that be possible, so how this gets generated is driven by how the typespec is structured

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/models/index.ts
joheredi: Unfortunately I don't think codegen needs to do anything about this unless there is a conflict with reserved words or words we use internally and cause a problem

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/models/index.ts
joheredi: Yeah, I need to close on this with Brian. Some ideas were `deleteOp` or `$delete`. fyi - @bterlson

joheredi: Lets choose one for now and we can modify later. Maybe $delete will be the one with least potential to cause troubles to the generator in the short term

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/audio/transcriptions/index.ts
joheredi: Yes that works, lets make sure we make it easy in the furute if we need to change this structure

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/OpenAIClient.ts
joheredi: I wouldn't be opposed to put them in folders, as long as the folder tells clearly that it contains top level operations. Will having this in folders help with the multilayer group? I'm not opposed to have multiple layers deep as long as it can be generated reliably. Do you have any thoughts here?

qiaozha: https://github.com/joheredi/unbranded_lib/pull/1#pullrequestreview-1696282481 I have this proposal to generate them with hierarchy layers otherwise, codegen would have to resolve the conflicts. let me know how you think ? 
Another consideration is, should we put those interfaces `AudioOperations` and `AudioTranslationOperations` and functions `getAudios` and `getAudioTranslations` into the api layer ? given the file structures are so much alike ?

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/audio/transcriptions/index.ts
joheredi: Is the the one you mention in your proposal where we;d be using the folder names?

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/OpenAIClient.ts
```diff
@@ -113,209 +52,39 @@ export class OpenAIClient {
...
+    this.embeddings = getEmbeddingsOperations(this._client);
+    this.files = getFilesOperations(this._client);
+    this.fineTunes = getFineTunesOperations(this._client);
+    this.models = getModelsOperations(this._client);
+    this.images = getImagesOperations(this._client);
+    this.moderations = getModerationsOperations(this._client);
+  }
+
+  /** The operation groups for AudioTranscriptions */
+  public readonly audio: AudioOperations;
```
qiaozha, line 69: readonly ?

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/OpenAIClient.ts
```diff
@@ -113,209 +52,39 @@ export class OpenAIClient {
   constructor(credential: KeyCredential, options: OpenAIClientOptions = {}) {
     this._client = createOpenAI(credential, options);
     this.pipeline = this._client.pipeline;
```
qiaozha, line 54: how are we going to handle pipeline ?

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/OpenAIClient.ts
```diff
@@ -113,209 +52,39 @@ export class OpenAIClient {
...
-    return createImageVariation(this._client, image, options);
-  }
-
-  createModeration(
-    content: CreateModerationRequest,
-    options: CreateModerationOptions = { requestOptions: {} }
-  ): Promise<CreateModerationResponse> {
-    return createModeration(this._client, content, options);
-  }
+    this.audio = getAudioOperations(this._client);
```
qiaozha, line 55: This should not be a breaking for mgmt plane.

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/models/index.ts
qiaozha: to use $delete and other reserve word for operations.

qiaozha (commented):
Still missing the subpath export info in the package.json

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/models/index.ts
qiaozha: issue created https://github.com/Azure/autorest.typescript/issues/2084

packages/typespec-test/test/anomalyDetector/generated/typespec-ts/src/models.ts
qiaozha: Add an option  enableModelNamespace to allow customer to append the namespace information in the model name.  By default, we will detect if there's model name conflict, if yes we will set this option as true when this option is unset. But if customer has specifically set a true or false value to it, we will just respect their choice. 
Also, this option applies to both RLC and Modular.

qiaozha: In the anomalyDetector's case, we have enabled this enableModelNamespace: true, it will use the namespace in the prefix, even if it doesn't have conflicts.
In the hierarchy generic's case, we don't set this option, but it detect there's a model name conflict, so it will enable it to add the full namespace prefix in the model name.

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/index.ts
qiaozha: Yes, see from the hierarchy generic test case here https://github.com/Azure/autorest.typescript/pull/2080/files#diff-da9b0137d726c15c00ff45b81aaec51cf260b9831ab85dbc025efc1650c4e890R25-R27

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/audio/transcriptions/index.ts
```diff
@@ -0,0 +1,65 @@
...
+  if (isUnexpected(result)) {
+    throw result.body;
+  }
+
+  return {
+    text: result.body["text"],
+  };
+}
+
+export async function audioTranscriptionsCreate(
```
qiaozha, line 58: This is incorrectly renamed, we shouldn't rename the api layer operations if hierarchy clients is enable which should by default be enabled.

packages/typespec-test/test/openai_generic/generated/typespec-ts/package.json
```diff
@@ -23,6 +23,54 @@
...
+      "types": "./types/src/models/index.d.ts",
+      "import": "./dist-esm/src/models/index.js"
+    },
+    "./api/images": {
+      "types": "./types/src/images/index.d.ts",
+      "import": "./dist-esm/src/images/index.js"
+    },
+    "./api/moderations": {
+      "types": "./types/src/moderations/index.d.ts",
+      "import": "./dist-esm/src/moderations/index.js"
```
qiaozha, line 73: api subpath exports added here.

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/OpenAIClient.ts
joheredi: Yeah lets keep this readonly

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/OpenAIClient.ts
joheredi: lets leave this for now and will revisit in a later PR

packages/typespec-test/test/hierarchy_generic/generated/typespec-ts/src/FooClient.ts
```diff
@@ -0,0 +1,33 @@
...
+    this.b = getBOperations(this._client);
+    this.d = getDOperations(this._client);
+  }
+
+  op1(body: A, options: Op1Options = { requestOptions: {} }): Promise<void> {
+    return op1(this._client, body, options);
+  }
+
+  /** The operation groups for BB */
+  public readonly b: BOperations;
```
joheredi, line 35: I think we'll have to define this like

```ts
public readonly b: {
  b: {
    op1: (body: BA, options?: BBOp1Options) => Promise<void>;
  };
}
```

To avoid the need to export the Operations interface which is really junk for our customers

packages/typespec-test/test/hierarchy_generic/generated/typespec-ts/src/rest/models.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+export interface A {
+  prop1: string;
+}
+
+export interface Ba {
```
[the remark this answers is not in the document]
joheredi: Yeah we should unify, can you please file an issue to track this. Good catch

packages/typespec-test/test/openai_generic/generated/typespec-ts/review/openai-generic.api.md
```diff
@@ -601,64 +601,48 @@ export interface Model {
...
-    listFineTuneEvents(fineTuneId: string, options?: ListFineTuneEventsOptions): Promise<ListFineTuneEventsResponse>;
-    // (undocumented)
-    listFineTunes(options?: ListFineTunesOptions): Promise<ListFineTunesResponse>;
-    // (undocumented)
-    listFineTuningEvents(fineTuningJobId: string, options?: ListFineTuningEventsOptions): Promise<ListFineTuningJobEventsResponse>;
-    // (undocumented)
-    listModels(options?: ListModelsOptions): Promise<ListModelsResponse>;
-    // (undocumented)
-    listPaginatedFineTuningJobs(options?: ListPaginatedFineTuningJobsOptions): Promise<ListPaginatedFineTuningJobsResponse>;
+    // Warning: (ae-forgotten-export) The symbol "AudioOperations" needs to be exported by the entry point index.d.ts
```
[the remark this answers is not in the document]
joheredi: I think we'll have to generate the type inline in the class property. See my comment below

packages/typespec-test/test/hierarchy_generic/generated/typespec-ts/src/FooClient.ts
qiaozha: As discussed offline, we should export the operation interface for now, otherwise we will have ugly classical client in the docs site. /cc  @bterlson

packages/typespec-test/test/openai_generic/generated/typespec-ts/review/openai-generic.api.md
```diff
@@ -417,13 +503,41 @@ export interface FineTuningJobEvent {
...
     b64Json?: Uint8Array;
     url?: string;
 }
 export { Image_2 as Image }
 
+// @public (undocumented)
+export interface ImagesOperations {
+    // (undocumented)
+    images: {
+        create: (image: CreateImageRequest, options?: CreateOptions_10) => Promise<ImagesResponse>;
```
qiaozha, line 535: I wonder if it's okay to have this CreateOptions_10 here ?

packages/typespec-test/test/openai_generic/generated/typespec-ts/review/openai-generic.api.md
```diff
@@ -534,6 +655,37 @@ export interface OpenAIFile {
...
+// src/classic/fineTuning/jobs/index.ts:41:5 - (ae-forgotten-export) The symbol "CancelOptions" needs to be exported by the entry point index.d.ts
+// src/classic/images/index.ts:22:5 - (ae-forgotten-export) The symbol "CreateOptions_10" needs to be exported by the entry point index.d.ts
+// src/classic/images/index.ts:26:5 - (ae-forgotten-export) The symbol "CreateEditOptions" needs to be exported by the entry point index.d.ts
+// src/classic/images/index.ts:30:5 - (ae-forgotten-export) The symbol "CreateVariationOptions" needs to be exported by the entry point index.d.ts
+// src/classic/models/index.ts:21:5 - (ae-forgotten-export) The symbol "ListOptions_4" needs to be exported by the entry point index.d.ts
+// src/classic/models/index.ts:22:5 - (ae-forgotten-export) The symbol "RetrieveOptions_4" needs to be exported by the entry point index.d.ts
+// src/classic/models/index.ts:23:5 - (ae-forgotten-export) The symbol "DeleteOptions_2" needs to be exported by the entry point index.d.ts
+// src/classic/moderations/index.ts:13:5 - (ae-forgotten-export) The symbol "CreateOptions_11" needs to be exported by the entry point index.d.ts
```
qiaozha, line 687: Are we going to do something about these warnings ? If yes, we will have to resolve the operation options conflict.

qiaozha (commented):
manual change for the operation interface export proposal.

packages/typespec-test/test/hierarchy_generic/generated/typespec-ts/src/rest/models.ts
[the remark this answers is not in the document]
qiaozha: Issue created https://github.com/Azure/autorest.typescript/issues/2086

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/OpenAIClient.ts
qiaozha: issue created  https://github.com/Azure/autorest.typescript/issues/2087

packages/typespec-test/test/openai_generic/generated/typespec-ts/review/openai-generic.api.md
joheredi: I think this is just how API extractor works in order to be able to show everything in the same file (*.api.md) with no conflicts. The actual API that is exported won't have _10

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/loadTestAdministration/LoadTestAdministrationClient.ts
```diff
@@ -57,119 +27,11 @@ export class LoadTestAdministrationClient {
...
-    testId: string,
-    body: TestServerMetricConfig,
-    options: CreateOrUpdateServerMetricsConfigOptions = { requestOptions: {} }
-  ): Promise<TestServerMetricConfig> {
-    return createOrUpdateServerMetricsConfig(
-      this._client,
-      testId,
-      body,
-      options
+    this.administrationOperations = getAdministrationOperationsOperations(
```
qiaozha, line 30: In this case, there's no `@operationGroup` decorator in the the client.tsp, it should not be treated as   operation group here.

packages/typespec-ts/test/modularIntegration/generated/client/structure/renamed-operation/src/RenamedOperationClient.ts
```diff
@@ -39,17 +39,23 @@ export class RenamedOperationClient {
     this.pipeline = this._client.pipeline;
   }
 
-  group = {
```
qiaozha, line 42: In this case, there is a decorator in the group field and we have set hierarchyClient as false. we should not remove it.

packages/typespec-ts/test/modularIntegration/generated/encode/bytes/src/BytesClient.ts
```diff
@@ -71,139 +32,21 @@ export class BytesClient {
   constructor(options: BytesClientOptions = {}) {
     this._client = createBytes(options);
     this.pipeline = this._client.pipeline;
+    this.query = getQueryOperations(this._client);
+    this.property = getPropertyOperations(this._client);
+    this.header = getHeaderOperations(this._client);
+    this.requestBody = getRequestBodyOperations(this._client);
+    this.responseBody = getResponseBodyOperations(this._client);
```
qiaozha, line 39: In the case of hierarchyClient is false, and operationGroup is true, we will just organize the code in the hierarchy way for better documentation purpose, and we will not export the api layer as subpath. @joheredi  how do you think ? it should have no api interface change in both classic layer and api layer expect for the change of exporting the operation group operations interface which we have discussed earlier that have to be exported as of the api extractor and the change of operation operations change as I memtioned earlier.

qiaozha (commented):
left some reviews after disable hierarchy client in most of the other test cases.

packages/typespec-ts/test/modularIntegration/generated/parameters/collection-format/src/api/query/index.ts
```diff
@@ -60,8 +60,8 @@ export async function queryMulti(
 export function _querySsvSend(
   context: Client,
   colors: string[],
-  options: QuerySsvOptions = { requestOptions: {} }
-): StreamableMethod<QuerySsv204Response> {
+  options: QueryQuerySsvOptions = { requestOptions: {} }
```
qiaozha, line 63: Doesn't seems correct.

packages/typespec-ts/test/modularIntegration/generated/parameters/collection-format/src/api/query/index.ts
```diff
@@ -7,29 +7,29 @@ import {
   buildSsvCollection,
   buildTsvCollection,
   CollectionFormatContext as Client,
-  QueryCsv204Response,
-  QueryMulti204Response,
-  QueryPipes204Response,
-  QuerySsv204Response,
-  QueryTsv204Response,
-} from "../rest/index.js";
+  Csv204Response,
```
qiaozha, line 10: need to figure out why this is missing Query

packages/typespec-test/test/chatApi_modular/generated/typespec-ts/package.json
```diff
@@ -23,6 +23,10 @@
     "./models": {
       "types": "./types/src/models/index.d.ts",
       "import": "./dist-esm/src/models/index.js"
+    },
+    "./api/genericChatClient": {
+      "types": "./types/src/api/genericChatClient/index.d.ts",
+      "import": "./dist-esm/src/api/genericChatClient/index.js"
```
qiaozha, line 29:
```suggestion
```

packages/typespec-test/test/loadtesting_modular/tspconfig.yaml
```diff
@@ -9,8 +9,8 @@ options:
     title: Azure Load Testing
     description: Azure Load Testing Client
     azureSdkForJs: false
```
qiaozha, line 11:
```suggestion
    azureSdkForJs: false
    enableOperationGroup: true
```

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/loadTestAdministration/LoadTestAdministrationClient.ts
qiaozha: Correction, as there's operation name conflict in this rlc layer, enableOperationGroup is set as true, therefore we will use the interface name as the operation group name. which should take effect in both RLC and modular layer. So this change is expected.

joheredi (approved):

PR #2088, opened 2023-10-27
fix-default-byte-as-binary-payload-in-modular

### LINKED ISSUE #2082: Content-type of application/octet-stream is being base64 encoded...
We have TypeSpec that looks like:
```
op analyzeFromStream is Azure.Core.RpcOperation<
  {
    @doc("The format of the HTTP payload.")
    @header
    contentType: "application/octet-stream";

    @doc("The image to be analyzed")
    @body
    imageContent: bytes;
...
```

And it's producing TypeScript:
```
    .post({
      contentType: (options.contentType as any) ?? "application/octet-stream",
      queryParameters: { mode: mode },
      body: uint8ArrayToString(imageContent, "base64"),
    });
```

qiaozha:
fixes #2082

### DIFF
--- a/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
+++ b/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
@@ -144,7 +144,7 @@ export function getDeserializePrivateFunction(
     }
   }
 
-  if (response?.type?.type === "any") {
+  if (response?.type?.type === "any" || response.isBinaryPayload) {
     statements.push(`return result.body`);
   } else if (getAllProperties(response?.type).length > 0) {
     statements.push(
@@ -386,7 +386,6 @@ function buildBodyParameter(
 
   if (
     bodyParameter.type.type === "byte-array" &&
-    bodyParameter.type.format !== "binary" &&
     !bodyParameter.isBinaryPayload
   ) {
     const coreUtilSet = importSet.get("@azure/core-util");
--- a/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
+++ b/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
@@ -386,7 +386,8 @@ function buildBodyParameter(
 
   if (
     bodyParameter.type.type === "byte-array" &&
-    bodyParameter.type.format !== "binary"
+    bodyParameter.type.format !== "binary" &&
+    !bodyParameter.isBinaryPayload
   ) {
     const coreUtilSet = importSet.get("@azure/core-util");
     if (!coreUtilSet) {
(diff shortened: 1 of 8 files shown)


packages/typespec-ts/test/modularUnit/apiOperations.spec.ts
```diff
@@ -0,0 +1,224 @@
+import { assert } from "chai";
+import { emitModularOperationsFromTypeSpec } from "../util/emitUtil.js";
+import { assertEqualContent } from "../util/testUtil.js";
+
+describe.only("api operations in Modular", () => {
```
MaryGao, line 5: skip only

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -407,8 +408,9 @@ function buildBodyParameter(
           bodyParameter.clientName
         }, "${getEncodingFormat(bodyParameter.type)}")`;
   } else if (
-    bodyParameter.type.type === "byte-array" &&
-    bodyParameter.type.format === "binary"
+    (bodyParameter.type.type === "byte-array" &&
+      bodyParameter.type.format === "binary") ||
+    bodyParameter.isBinaryPayload
```
MaryGao, line 413: It seems that when `bodyParameter.type.type === "byte-array" &&
      bodyParameter.type.format === "binary"` is also binary, why not enhance the isBinaryPayload logic to cover that case?

MaryGao (approved):
Approved and left minor comments

packages/typespec-ts/test/modularUnit/apiOperations.spec.ts
```diff
@@ -0,0 +1,224 @@
+import { assert } from "chai";
+import { emitModularOperationsFromTypeSpec } from "../util/emitUtil.js";
+import { assertEqualContent } from "../util/testUtil.js";
+
+describe("api operations in Modular", () => {
+  it("should handle contentTypes has binary data", async () => {
+    const tspContent = `
+    @route("/uploadFileViaBody")
+    @post op uploadFileViaBody(
+      @header contentType: "application/octet-stream",
```
lirenhe, line 10: From Swagger, the payload of octet-stream would look like the below, so could you help to ensure the typespec is correct for this case?

requestBody:
  content:
    application/octet-stream:
      schema:
        type: string
        format: binary # binary data
    image/png:
      schema:
        type: string
        format: base64 # base64-encoded data

lirenhe (approved):

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -83,7 +83,10 @@ export function isByteOrByteUnion(dpgContext: SdkContext, type: Type) {
 }
 
 function isByteType(schema: any) {
-  return schema.type === "string" && schema.format === "byte";
+  return (
+    schema.type === "string" &&
+    (schema.format === "byte" || schema.format === "binary")
```
qiaozha, line 88: If it's self defined scalar, we will get the format of binary as special handled.

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
qiaozha: Changed both to use isBinaryPayload and fixed a small bug in isByteType.

packages/typespec-ts/test/modularUnit/apiOperations.spec.ts
qiaozha: Issue created https://github.com/microsoft/typespec/issues/2615

Issue #2098, opened 2023-11-06
Special character can not be generated as a parameter name from sample generation


kazrael2119:
https://github.com/Azure/azure-rest-api-specs/blob/b672a0b301338a570af2e5430b4b7691f909a094/specification/azureintegrationspaces/resource-manager/Microsoft.IntegrationSpaces/preview/2023-11-14-preview/examples/Applications_ValidateBusinessProcessDevelopmentArtifact.json#L19

this parameter name can't be generated.

kazrael2119:
https://github.com/Azure/autorest.typescript/pull/2099

PR #2099, opened 2023-11-06
fix sample generation

kazrael2119:
fix https://github.com/Azure/autorest.typescript/issues/2098
Special character can not be generated as a parameter name from sample generation, this pr is to fix this

### DIFF
--- a/packages/autorest.typescript/src/utils/valueHelpers.ts
+++ b/packages/autorest.typescript/src/utils/valueHelpers.ts
@@ -179,13 +179,8 @@ export function getParameterAssignment(
           propName = normalizeName(initPropName, NameType.Property, true);
         }
         let propRetValue: string;
-        if (propName.indexOf("/") > -1 || propName.match(/^\d/)) {
-          propRetValue =
+        propRetValue =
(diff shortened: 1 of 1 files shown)


kazrael2119:
related sample: https://github.com/Azure/azure-sdk-for-js/pull/27655/files#diff-57333f3ec9d7cf6ea2a6f5ff75e1cb39ed9ade679f7fdeb23d759bc185ca5189R87

packages/autorest.typescript/src/utils/valueHelpers.ts
```diff
@@ -179,13 +179,8 @@ export function getParameterAssignment(
           propName = normalizeName(initPropName, NameType.Property, true);
         }
         let propRetValue: string;
-        if (propName.indexOf("/") > -1 || propName.match(/^\d/)) {
-          propRetValue =
+        propRetValue =
```
MaryGao, line 182: Please update to below:

```
const propRetValue = `"${propName}": ` + getParameterAssignment(property, isRLCSample);
```

MaryGao (approved):
LGTM

PR #2103, opened 2023-11-08
Add coverage cases for arrays-item-types in cadl-ranch

zzhxiaofeng:
When generate test SDK for arrays-item-types, it failed and returns an error which is shown as following:
image

### DIFF
--- a/packages/typespec-ts/test/commands/cadl-ranch-list.ts
+++ b/packages/typespec-ts/test/commands/cadl-ranch-list.ts
@@ -259,5 +259,9 @@ export const modularTsps: TypeSpecRanchConfig[] = [
   {
     outputPath: "authentication/union",
     inputPath: "authentication/union"
+  },
+  {
+    outputPath: "arrays/itemTypes",
+    inputPath: "type/array"
   }
 ];
--- a/packages/typespec-ts/test/modularIntegration/generated/arrays/items/src/logger.ts
+++ b/packages/typespec-ts/test/modularIntegration/generated/arrays/items/src/logger.ts
@@ -1,0 +1,5 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { createClientLogger } from "@azure/logger";
+export const logger = createClientLogger("arrays-item-types");
--- a/packages/typespec-ts/test/modularIntegration/generated/arrays/items/src/rest/arrayClient.ts
+++ b/packages/typespec-ts/test/modularIntegration/generated/arrays/items/src/rest/arrayClient.ts
@@ -1,0 +1,35 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { getClient, ClientOptions } from "@azure-rest/core-client";
+import { logger } from "../logger.js";
+import { ArrayContext } from "./clientDefinitions.js";
+
+/**
+ * Initialize a new instance of `ArrayContext`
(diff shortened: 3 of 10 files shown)


v-jiaodi:
@qiaozha When running integration-test, get error as follow:
image

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -944,7 +944,7 @@ function deserializeResponseValue(
           type.elementType!,
           "p",
           runtimeImports,
-          required,
+          type.elementType?.nullable === false ? false : true,
```
qiaozha, line 947:
```suggestion
          true,
```

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -944,7 +944,7 @@ function deserializeResponseValue(
           type.elementType!,
           "p",
           runtimeImports,
-          required,
+          true,
```
qiaozha, line 947: Just found typespec can't not define this way [playground link](https://cadlplayground.z22.web.core.windows.net/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdCI7CgpAc2VydmljZSh7CiAgdGl0bGU6ICJXaWRnZXQgU8YbIiwKfSkKbmFtZXNwYWNlIERlbW%2FHGzsKCnVzaW5nIFR5cGVTcGVjLkh0dHA70BVSZXN0OwoKbW9kZWwgVGVzdCDEbmNvbnRlbnQ6IChzdHLEMHwgdW5kZWZpbmVkKVtdOwp9Cm9wIHJlYWQoQGJvZHkgxAU6xUApOiB2b2lkOwo%3D&e=%40typespec%2Fopenapi3&options=%7B%7D)

qiaozha: I would assume elementTypes are all required, for nullable, since the current current nullable logic has problem, I would prefer to have another independent pr to fix it. @MaryGao  let me know if you have different thoughts.

v-jiaodi:
Add test case , encountered the following error:
image

packages/typespec-ts/test/modularIntegration/arrayItemTypes.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+    type: "string",
+    defaultValue: ["hello", ""]
+  },
+  {
+    type: "float32",
+    defaultValue: [42.42]
+  },
+  {
+    type: "datetime",
+    defaultValue: [new Date("2022-08-26T18:38:00Z")]
```
qiaozha, line 33: Datetime type is a Date typescript type in modular.

packages/typespec-ts/test/modularIntegration/arrayItemTypes.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+  },
+  {
+    type: "unknown",
+    defaultValue: [1, "hello", null]
+  },
+  {
+    type: "model",
+    defaultValue: [
+      { property: "hello", children: undefined },
+      { property: "world", children: undefined }
```
qiaozha, line 47: the children: undefined are added by the current serialization and deserialization logic in modular.

qiaozha (approved):

Issue #2108, opened 2023-11-09
Generator defaults to an empty array for optional array properties


timovv:
Encountered when generating the unbranded OpenAI client:

### Context

`CreateChatCompletionRequest` has the following array property declared in TypeSpec with `@minItems(1)`:

https://github.com/Azure/autorest.typescript/blob/8fdc016a9080e6ac9decfb85319503791390c6ab/packages/typespec-test/test/openai_generic/spec/completions/models.tsp#L159-L162

### Expected behavior

When `functions` is not specified in the input, the property should not be present in the request.

### Actual behavior

`functions` has value `[]` in the request. While this makes sense for required array properties, it does not make sense in this scenario, and causes issues, since `[]` does not satisfy the `@minItems(1)` constraint. OpenAI returns an error when passing `[]` here.

MaryGao:
@timovv It should be fixed in the latest codegen, feel free to have a check! Thanks!

PR #2111, opened 2023-11-13
Fix the empty default value issue for array in body model

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2108

### DIFF
--- a/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
+++ b/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
@@ -904,7 +904,7 @@ function deserializeResponseValue(
         : `${restValue} !== undefined? new Date(${restValue}): undefined`;
     case "combined":
       return `${restValue} as any`;
-    case "list":
+    case "list": {
       const prefix =
         required && !type.nullable
           ? `${restValue}`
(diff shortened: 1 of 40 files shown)


packages/typespec-ts/test/modularUnit/operations.spec.ts
```diff
@@ -4,159 +4,522 @@ import { assertEqualContent } from "../util/testUtil.js";
...
+      }
+      `
+      );
+    });
   });
 
-  it("required nullable header would report diagnostic", async () => {
-    try {
+  // TODO: need to fix the tests
+  describe.skip("array as body", () => {
```
MaryGao, line 168: Optionaly and nullable body check will be in another pr.

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -975,16 +980,20 @@ function serializeRequestValue(
         default:
           return `${clientValue}${required ? "" : "?"}.toISOString()`;
       }
-    case "list":
+    case "list": {
+      const prefix =
+        required && !type.nullable
```
qiaozha, line 985: if it's requried but nullable, are you sure this works?

qiaozha (approved):

PR #2122, opened 2023-11-15
Add modular coverage cases for azure-core-traits in cadl-ranch

zzhxiaofeng:
fixes https://github.com/Azure/autorest.typescript/issues/2102

Added List:
1.azure-core-traits


packages/typespec-ts/test/modularIntegration/azureCoreTraits.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+  beforeEach(() => {
+    client = new TraitsClient({
+      allowInsecureConnection: true
+    });
+  });
+
+  it("client smoke test", async () => {
+    try {
+      const result = await client.smokeTest(1, "123", {
+        ifNoneMatch: '"invalid"',
```
zzhxiaofeng, line 16: The value of `ifNoneMatch` is set to `'"invalid"'` and `ifMatch` is set to `'"valid"'` in mock file. They have an extra pair of single quotes than normal `string` type. Please have a check if there are any issues here.

MaryGao: I think this is by design and etag value is quoted string https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag#examples.

packages/typespec-ts/test/modularIntegration/projection.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+      const result = await client.property.client({ defaultName: true });
+      assert.isUndefined(result);
+    } catch (err) {
+      assert.fail(err as string);
+    }
+  });
+
+  it("should use language property", async () => {
+    try {
+      const result = await client.property.language({ defaultName: true });
```
MaryGao, line 33: We need to adopt the client and javascript projected name.

MaryGao (approved):
LGTM

MaryGao:
Merged this pr. @qiaozha feel free to add any comments!

PR #2124, opened 2023-11-16
Add rlc coverage cases for scalar, client-generator-core-access, client-generator-core-usage and payload-pageable in cadl-ranch

zzhxiaofeng:
Fixes https://github.com/Azure/autorest.typescript/issues/2102

Added List:
1.scalar
2.client-generator-core-access
3.client-generator-core-usage
4.payload-pageable


packages/typespec-ts/test/integration/azureClientGeneratorCoreAccess.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  beforeEach(() => {
+    client = AccessClientFactory({
+      allowInsecureConnection: true,
+      retryOptions: {
+        maxRetries: 0
+      }
+    });
+  });
+
+  matrix([testedUrl], async (params: UrlDetail) => {
```
MaryGao, line 52: In our practice the matrix is not recommanded because it would introduce casting like as any, so we can't verify type correctly.

Could you remove the matrix and sperate them one by one?

packages/typespec-ts/test/integration/azureClientGeneratorCoreAccess.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    it(`should access ${params.value} in ${params.type}`, async () => {
+      try {
+        const result = await client
+          .path(
+            `/azure/client-generator-core/access/${params.type}/${params.value}` as any
+          )
+          .get({ queryParameters: { name: "myname" } });
+        assert.strictEqual(result.status, "200");
+        assert.strictEqual(
+          JSON.stringify(result.body),
```
MaryGao, line 62: Why do we need the stringify here? I think the correct way is directly using `.` for any property e.g `result.body.name`.

packages/typespec-ts/test/integration/azureClientGeneratorCoreAccess.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+        .path(
+          "/azure/client-generator-core/access/relativeModelInOperation/operation"
+        )
+        .get({
+          queryParameters: { name: "Madge", inner: { name: "Madge" } }
+        });
+      assert.strictEqual(result.status, "200");
+      assert.strictEqual(
+        JSON.stringify(result.body),
+        JSON.stringify({ name: "Madge", inner: { name: "Madge" } })
```
MaryGao, line 83: +1 here.

I can understand this would be easier, but may not catch any type mismatch errors. Could you use `.` for any property instead?

packages/typespec-ts/test/integration/azureClientGeneratorCoreAccess.spec.ts
zzhxiaofeng: I have removed the matrix and sperate them one by one

packages/typespec-ts/test/integration/azureClientGeneratorCoreAccess.spec.ts
zzhxiaofeng: I have updated the code as you said and will do the same in the future.

packages/typespec-ts/test/integration/azureClientGeneratorCoreAccess.spec.ts
zzhxiaofeng: I have updated the code as you said and will do the same in the future.

packages/typespec-ts/test/commands/cadl-ranch-list.ts
```diff
@@ -184,6 +184,14 @@ export const rlcTsps: TypeSpecRanchConfig[] = [
...
     outputPath: "client/structure/two-operation-group",
     inputPath: "client/structure/two-operation-group"
+  },
+  {
+    outputPath: "scalar",
+    inputPath: "type/scalar"
+  },
+  {
+    outputPath: "azure/clientGeneratorCore/access",
+    inputPath: "azure/client-generator-core/access"
```
MaryGao, line 194: Could you update the `/azure/client-generator-core/usage` cases also?

zzhxiaofeng: I have updated this case.

MaryGao (approved):
LGTM

packages/typespec-ts/test/integration/azureClientGeneratorCoreUsage.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  beforeEach(() => {
+    client = UsageClientFactory({
+      allowInsecureConnection: true,
+      retryOptions: {
+        maxRetries: 0
+      }
+    });
+  });
+
+  it("should input usage model in Operation", async () => {
```
MaryGao, line 18:
```suggestion
  it("should post input usage model in Operation", async () => {
```

packages/typespec-ts/test/integration/azureClientGeneratorCoreAccess.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+  beforeEach(() => {
+    client = AccessClientFactory({
+      allowInsecureConnection: true,
+      retryOptions: {
+        maxRetries: 0
+      }
+    });
+  });
+
+  it("no decorator in public operation", async () => {
```
MaryGao, line 18:
```suggestion
  it("should get no decorator in public operation", async () => {
```

packages/typespec-ts/test/integration/payloadPageable.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+      retryOptions: {
+        maxRetries: 0
+      }
+    });
+  });
+
+  it("should get pagable list without skiptoken", async () => {
+    try {
+      const result = await client
+        .path("/payload/pageable")
```
MaryGao, line 21: Please refer this for paging cases

https://github.com/Azure/autorest.typescript/blob/main/packages/autorest.typescript/test/rlcIntegration/pagingRest.spec.ts#L26-L40

zzhxiaofeng: I have updated this test file as you said.

packages/typespec-ts/test/integration/payloadPageable.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  beforeEach(() => {
+    client = PageableClientFactory({
+      allowInsecureConnection: true,
+      retryOptions: {
+        maxRetries: 0
+      }
+    });
+  });
+
+  it("should get pagable list", async () => {
```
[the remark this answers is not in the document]
zzhxiaofeng: I have updated this test file as you said

PR #2135, opened 2023-11-20
Fix unix timestamps

timovv:
JS timestamps are in milliseconds from the epoch by default whereas standard Unix timestamps are in seconds. This PR changes the code to divide/multiply by 1000 for the conversion


packages/typespec-test/test/ai/generated/typespec-ts/src/models/models.ts
```diff
@@ -1675,16 +1675,16 @@ export function threadMessageSerializer(item: ThreadMessage): any {
...
     status: item["status"],
     incomplete_details: item["incompleteDetails"],
     completed_at: !item["completedAt"]
       ? item["completedAt"]
-      : item["completedAt"].getTime(),
+      : (item["completedAt"].getTime() / 1000) | 0,
     incomplete_at: !item["incompleteAt"]
       ? item["incompleteAt"]
-      : item["incompleteAt"].getTime(),
+      : (item["incompleteAt"].getTime() / 1000) | 0,
```
maorleger, line 1687:
```suggestion
      : (item["incompleteAt"].getTime() / 1000) || 0,
```

Should this be || ?

maorleger: Also Do we want to default to 0 or should we fail-fast?

timovv: The bitwise or with 0 removes the decimal part of the number

maorleger (approved):
LGTM. One nit about the bitwise or usage but totally up to you!

packages/typespec-test/test/ai/generated/typespec-ts/src/models/models.ts
timovv: Probably clearer, yeah, but FWIW I've seen bitwise used in this way a few times in the wild so wasn't under the impression that this is particularly uncommon pattern. Since this is generated code that folks generally aren't going to be editing by hand I don't think it's worth the hassle of regenerating the tests. I appreciate the suggestion though.

PR #2150, opened 2023-11-29
Update tcgc to 0.36.1

timovv:


qiaozha (approved):
Just a little curious if the 0.36.1 is trying to resolve some kind of issues ? which is not reflected in our code ?

timovv:
0.36.1 added an extra parameter to a decorator so Laurent asked we update it so that we don't get failures when trying to emit if the new extra parameter is present. At least that's my understanding.

Issue #2141, opened 2023-11-22
Generate code error for unions when add rls coverages in cadl-ranch


zzhxiaofeng:
**Issue description:**
There is an error in the code generated for unions. The error appears in file `parameters.ts`. The detailed error code is shown as following:
![image](https://github.com/Azure/autorest.typescript/assets/20970631/f941c025-4053-4a29-97a8-78828ba02922)

@MaryGao for notification.

PR #2154, opened 2023-12-04
Fix named union issue and integrate the test case for unions

MaryGao:
Fix named union issue and integrate the test case for unions

fixes https://github.com/Azure/autorest.typescript/issues/2141


packages/typespec-ts/src/transform/transformSchemas.ts
```diff
@@ -130,8 +130,9 @@ export function transformSchemas(
         getGeneratedModels(indexer.value, context);
       }
       for (const prop of model.properties) {
+        const [, propType] = prop;
         if (
-          prop[1].type.kind === "Model" &&
+          propType.type.kind === "Model" &&
```
qiaozha, line 135: why not use propType to replace all the places that has refered to prop[1] ?

qiaozha (approved):

packages/typespec-ts/src/transform/transformSchemas.ts
MaryGao: it should be replaced and will update that

PR #2155, opened 2023-12-04
Upgrade the cadl-ranch and integrate the scalar test cases

MaryGao:
Upgrade the cadl-ranch & integrate the scalar test cases & support to generate named union type for header.

### DIFF
--- a/packages/typespec-ts/src/modular/buildCodeModel.ts
+++ b/packages/typespec-ts/src/modular/buildCodeModel.ts
@@ -1469,12 +1469,13 @@ function emitOperationGroups(
     string,
     OperationGroup
   >();
-  for (const operationGroup of listOperationGroups(context, client)) {
+  for (const operationGroup of listOperationGroups(context, client, true)) {
     const operations: HrlcOperation[] = [];
     const name = operationGroup.type.name;
     for (const operation of listOperationsInOperationGroup(
       context,
-      operationGroup
+      operationGroup,
+      true
     )) {
       operations.push(emitOperation(context, operation, name, rlcModels));
     }
@@ -1483,7 +1484,11 @@ function emitOperationGroups(
     }
   }
   const clientOperations: HrlcOperation[] = [];
-  for (const operation of listOperationsInOperationGroup(context, client)) {
+  for (const operation of listOperationsInOperationGroup(
+    context,
+    client,
+    true
+  )) {
     clientOperations.push(emitOperation(context, operation, "", rlcModels));
   }
   if (clientOperations.length > 0) {


packages/typespec-ts/test/integration/generated/scalar/src/clientDefinitions.ts
```diff
@@ -8,6 +8,16 @@ import {
   BooleanModelPutParameters,
   UnknownGetParameters,
   UnknownPutParameters,
+  DecimalTypeResponseBodyParameters,
+  DecimalTypeRequestBodyParameters,
```
MaryGao, line 12: these cases should be supported in https://github.com/Azure/autorest.typescript/issues/2157

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -1073,18 +1073,28 @@ function getSchemaForStdScalar(
...
+        format: "float32"
       });
     case "float":
       return applyIntrinsicDecorators(program, type, {
         type: "number",
         format: "float"
       });
+    case "decimal":
+      return applyIntrinsicDecorators(program, type, {
+        type: "number",
```
MaryGao, line 1090: the temp workaround to ci failures and detailed fix would be covered https://github.com/Azure/autorest.typescript/issues/2157

packages/typespec-ts/test/integration/generated/azure/core-traits/src/responses.ts
```diff
@@ -33,7 +33,7 @@ export interface SmokeTestDefaultResponse extends HttpResponse {
 
 export interface RepeatableAction200Headers {
   /** Indicates whether the repeatable request was accepted or rejected. */
-  "repeatability-result"?: "accepted" | "rejected";
+  "repeatability-result"?: RepeatabilityResultOutput;
```
MaryGao, line 36: This is introduced by typespec-azure-core upgrade in which the definition is changed to named union.

packages/typespec-ts/test/unit/modelsGenerator.spec.ts
```diff
@@ -346,6 +346,166 @@ describe("Input/output model type", () => {
     });
   });
 
+  describe("decimal generation", () => {
+    it("should handle decimal -> number", async () => {
```
MaryGao, line 350: Map the decimal to `number` in JS directly, and will 
- add below comments 
```
Please note the field was supposed to be a decimal but JavaScript does not have a native 'BigDecimal' data type.
So it was converted to a number instead. It is recommended to use a third-party library like 'decimal.js' to handle 
any calculations.
```
- report a warning 
```
Please note the decimal type will be converted to number. If you strongly care about precision you can use @encode to encode it as a string for the property - {your prop name}.
```

packages/typespec-ts/test/unit/modelsGenerator.spec.ts
```diff
@@ -346,6 +346,166 @@ describe("Input/output model type", () => {
...
+         * any calculations.
+        */
+        prop: number;
+      }
+      `
+      );
+    });
+
+    // TODO: pending with typespec definition https://github.com/microsoft/typespec/issues/2762
+    it.skip("should handle decimal/decimal128 with encode `string`", async () => {
```
MaryGao, line 470: Encoding as string is still pending typespec discussion https://github.com/microsoft/typespec/issues/2762.

packages/typespec-ts/test/integration/scalar.spec.ts
```diff
@@ -77,4 +78,118 @@ describe("Scalar Client", () => {
...
+  });
+
+  it("should post decimal verify", async () => {
+    try {
+      // prepare the verification
+      const getResult = await client
+        .path("/type/scalar/decimal/prepare_verify")
+        .get();
+      // do any calculation based on third party library
+      let total = new Decimal(0);
```
MaryGao, line 157: Import `decimal.js` for calculation.

MaryGao:
closed as no need.

Issue #2149, opened 2023-11-29
[Codegen/TypeSpec] missing error response import in source code


kazrael2119:
ci check failure in rush build https://github.com/Azure/azure-rest-api-specs/pull/26702/checks?check_run_id=18990593006

it's caused by missing import in the generated source code https://github.com/azure-sdk/azure-sdk-for-js/pull/7787/files#diff-23670a4188ba7a22ff45baffecb05db777e683c05a5090b1b0edcb83309ac32bR47

another question: 
don't know why codegen generated another [Error response](https://github.com/azure-sdk/azure-sdk-for-js/pull/7787/files#diff-23670a4188ba7a22ff45baffecb05db777e683c05a5090b1b0edcb83309ac32bR353) only for the interface
related code in [typespec](https://github.com/Azure/azure-rest-api-specs/blob/8450f55a6cbd2b67e73d1a5710d6a242cd0498ec/specification/purview/Azure.Analytics.Purview.DataMap/routes.tsp#L717-L725) 
related Error response in [model.tsp](https://github.com/Azure/azure-rest-api-specs/blob/8450f55a6cbd2b67e73d1a5710d6a242cd0498ec/specification/purview/Azure.Analytics.Purview.DataMap/models.tsp#L450-L462)
related Error response in [routes.tsp](https://github.com/Azure/azure-rest-api-specs/blob/8450f55a6cbd2b67e73d1a5710d6a242cd0498ec/specification/purview/Azure.Analytics.Purview.DataMap/routes.tsp#L46-L92)

qiaozha:
close it as won't fix

PR #2156, opened 2023-12-04
Fix the error response model issues in purview data map

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2149


packages/typespec-ts/test/unit/responsesGenerator.spec.ts
```diff
@@ -286,6 +286,76 @@ describe("Responses.ts", () => {
       `
       );
     });
+
+    it("mixed custom error response and core error response", async () => {
```
qiaozha, line 290: Is this what happened in purview ? I doubt why not asking them to be consistent ? And Modular will never work for cases like this, because we don't split input model and output model.

MaryGao: yes, because purview mixed rpc operation and azure core template operations which introduced duplicate definitions.

MaryGao: For purview case they would change to consistent error response I think and here is comment https://github.com/Azure/azure-rest-api-specs/pull/26702#discussion_r1414911972.

But from our side I prefer to fix this to generate as what the customer defines otherwise this will fail to build.

This would not be an issue for Modular because the model marked as @error won't be generated. Any body shape is acceptable when calling createRestError.

This would be an issue if these models happen in normal 2xx response body. But I haven't met this case yet.

PR #2162, opened 2023-12-08
using-effective-payload-type

qiaozha:
fix easm ci error https://github.com/Azure/azure-rest-api-specs/pull/26980/checks?check_run_id=19398286042


packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -1293,7 +1295,17 @@ export function getBodyType(
       }
     }
     if (resourceType && bodyModel.name === "") {
-      bodyModel = resourceType;
+      const metadataInfo = createMetadataInfo(program);
```
lirenhe, line 1298: Can we have a simple test case e.g. unit test to verify this change?

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -1293,7 +1295,17 @@ export function getBodyType(
...
+      const visibility = resolveRequestVisibility(
```
MaryGao, line 1299: Maybe some test cases to help understand the visibility impact?

Another point is if the type name would be duplicated if happening in different visibility scopes?

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -1293,7 +1295,17 @@ export function getBodyType(
...
+        program,
+        route.operation,
+        route.verb
+      );
+      const payloadType = metadataInfo.getEffectivePayloadType(
```
MaryGao, line 1304: Does this only happen in body type? Is it possible in general property if there is a anonymous model?

lirenhe (approved):

packages/typespec-ts/src/utils/modelUtils.ts
MaryGao: yes, the visibility is not supported yet, so I am wondering if this could cause some issues or maybe it's better to hard code visibility?

qiaozha: I don't understand what's the value to get a wrong model out here ?  @lirenhe  thoughts ?

.scripts/ci.yml
```diff
@@ -129,18 +129,18 @@ stages:
...
+            Linux_Node18:
               Pool: ${{ parameters.LinuxPool }}
               OSVmImage: "ubuntu-20.04"
-              NodeTestVersion: "16.x"
+              NodeTestVersion: "18.x"
             Windows_Latest:
               Pool:
               OSVmImage: "Windows-latest"
-              NodeTestVersion: "16.x"
+              NodeTestVersion: "18.x"
```
jeremymeng, line 143: maybe use 20.x for Latest?

qiaozha: updated. Thanks for the advice.

PR #2163, opened 2023-12-12
Remove unused npm `audit` script from generated packages

jeremymeng:
- as `audit` is not used any more in the azsdk js repo
- and remove unused rush commands. They were copied from azure-sdk-for-js repo but not used any more.
- update test baselines


jeremymeng:
interesting. Nothing changes from last week but now CI is failing

common/config/rush/command-line.json
```diff
@@ -164,65 +158,36 @@
...
         "summary": "Run typedoc on projects and generate documentation",
         "enableParallelism": true,
         "ignoreMissingScript": true
       },
-  
+
       // Global Commands
-      {
-        "commandKind": "global",
-        "name": "list-files",
```
qiaozha, line 179: I guess these are copied from somewhere but actually doesn't work at all ?

qiaozha (approved):
LGTM

common/config/rush/command-line.json
jeremymeng: yeah looks copied from azsdkjs repo. It's not used anywhere as fare as I know.

PR #2165, opened 2023-12-13
Add support for a TypeSpec/sdk regeneration preview pipeline

hallipr:


packages/typespec-ts/package.json
```diff
@@ -71,7 +71,13 @@
...
     "@azure/core-util": "^1.4.0",
     "eslint-plugin-require-extensions": "0.1.3",
-    "@typespec/ts-http-runtime": "1.0.0-alpha.20231129.4"
+    "@typespec/ts-http-runtime": "1.0.0-alpha.20231129.4",
+    "@azure-tools/typespec-azure-core": "0.37.2",
+    "@azure-tools/typespec-client-generator-core": "0.37.0",
+    "@typespec/compiler": "0.51.0",
+    "@typespec/http": "0.51.0",
+    "@typespec/rest": "0.51.0",
+    "@typespec/versioning": "0.51.0"
```
MaryGao, line 80: Why we need this in devdependencies? I believe they are in peerDependencies already.

hallipr: We need to ensure that the sdk repo uses the same version of the libraries that the emitter was built against. To do this, we use 2 related patterns:
- **devDependencies pinned peerDependencies range:** We declare in our peerDependencies that we require and are compatible with any version within the range `>=0.51.0 <1.0.0`.  In devDependencies, we pick a specific version to build and test against.  The version in devDependencies becomes an informal way of declaring to consumers which version within the range was used by the library developers for build and test.
- **forwarded dependencies:** To ensure that the sdk repo uses the same version of the typespec modules that the emitter used, we need to pin the typespec modules as devDependencies in emitter-package.json.  This follows the normal npm convention of having the emitter declare that it depends on the peerDependency within a range and have the consumer (emitter-package.json) provide the dependency using a specific version within that range.

We don't want to hand edit emitter-package.json to keep it in sync with the emitter's package.json, so we just infer the packages to be pinned by comparing the dependencies in package.json.  If a package is mentioned in both peer and dev dependencies, we assume it's doing the peer/dev range pinning and we forward the dev dependency spec along to [emitter-package.json](https://github.com/Azure/azure-sdk-for-python/blob/7160d94f344f346b159ebe266b9b06b420a1a977/eng/emitter-package.json):
```json
  "devDependencies": {
    "@azure-tools/typespec-azure-core": "~0.37.0",
    "@typespec/versioning": "~0.51.0",
    "@typespec/http": "~0.51.0",
    "@azure-tools/typespec-client-generator-core": "~0.37.0",
    "@typespec/compiler": "~0.51.0",
    "@typespec/rest": "~0.51.0"
  },
```

joheredi (approved):

PR #2169, opened 2023-12-19
Ser deser for discriminated union and polymorphic base

### LINKED ISSUE #1982: RangeError: Maximum call stack size exceeded at getResponseMapping
Full error:

> ExternalError: Emitter "@azure-tools/typespec-ts" failed!
File issue at https://github.com/Azure/autorest.typescript/issues

> RangeError: Maximum call stack size exceeded
    at getResponseMapping (file:///home/codespace/workspace/sdk/openai/openai/TempTypeSpecFiles/OpenAI.Inference/node_modules/@azure-tools/typespec-ts/dist/src/modular/helpers/operationHelpers.js:400:35)
    at deserializeResponseValue (file:///home/codespace/workspace/sdk/openai/openai/TempTypeSpecFiles/OpenAI.Inference/node_modules/@azure-tools/typespec-ts/dist/src/modular/helpers/operationHelpers.js:441:59)
    at getResponseMapping (file:///home/codespace/workspace/sdk/openai/openai/TempTypeSpecFiles/OpenAI.Inference/node_modules/@azure-tools/typespec-ts/dist/src/modular/helpers/operationHelpers.js:419:53)
    at getResponseMapping (file:///home/codespace/workspace/sdk/openai/openai/TempTypeSpecFiles/OpenAI.Inference/node_modules/@azure-tools/typespec-ts/dist/src/modular/helpers/operationHelpers.js:412:179)
    at deserializeResponseValue (file:///home/codespace/workspace/sdk/openai/openai/TempTypeSpecFiles/OpenAI.Inference/node_modules/@azure-tools/typespec-ts/dist/src/modular/helpers/operationHelpers.js:441:59)
    at getResponseMapping (file:///home/codespace/workspace/sdk/openai/openai/TempTypeSpecFiles/OpenAI.Inference/node_modules/@azure-tools/typespec-ts/dist/src/modular/helpers/operationHelpers.js:419:53)
    at getResponseMapping (file:///home/codespace/workspace/sdk/openai/openai/TempTypeSpecFiles/OpenAI.Inference/node_modules/@azure-tools/typespec-ts/dist/src/modular/helpers/operationHelpers.js:412:179)
    at deserializeResponseValue (file:///home/codespace/workspace/sdk/openai/openai/TempTypeSpecFiles/OpenAI.Inference/node_modules/@azure-tools/typespec-ts/dist/src/modular/helpers/operationHelpers.js:441:59)
    at getResponseMapping (file:///home/codespace/workspace/sdk/openai/openai/TempTypeSpecFiles/OpenAI.Inference/node_modules/@azure-tools/typespec-ts/dist/src/modular/helpers/operationHelpers.js:419:53)
    at getResponseMapping (file:///home/codespace/workspace/sdk/openai/openai/TempTypeSpecFiles/OpenAI.Inference/node_modules/@azure-tools/typespec-ts/dist/src/modular/helpers/operationHelpers.js:412:179)
    at runEmitter (file:///home/codespace/workspace/sdk/openai/openai/TempTypeSpecFiles/OpenAI.Inference/node_modules/@typespec/compiler/dist/src/core/program.js:583:19)
    at async compile (file:///home/codespace/workspace/sdk/openai/openai/TempTypeSpecFiles/OpenAI.Inference/node_modules/@typespec/compiler/dist/src/core/program.js:257:9)
    at async compileOnce (file:///home/codespace/workspace/sdk/openai/openai/TempTypeSpecFiles/OpenAI.Inference/node_modules/@typespec/compiler/dist/src/core/cli/actions/compile/compile.js:37:25)
    at async compileAction (file:///home/codespace/workspace/sdk/openai/openai/TempTypeSpecFiles/OpenAI.Inference/node_modules/@typespec/compiler/dist/src/core/cli/actions/compile/compile.js:21:9)

## Repro steps:
- edit `sdk/openai/openai/tsp-location.json` and set the commit to `0d3d201639b9e7f3f448bbd08b1fe2fd461ecf49`
- run `rushx generate`

You can find the typespec in https://github.com/Azure/azure-rest-api-specs/pull/25271

This is high priority to unblock the openai release scheduled later this month.

### LINKED ISSUE #2139: Serialization Incomplete For Polymorphic Input
In OpenAI, we have `ChatRequestMessage` that is an union of 5 different types. The deserialization logic for the property only parses the shared properties between the different types.

The Typespec definition for this type is [here](https://github.com/Azure/azure-rest-api-specs/blob/c8be3dc86f4dcfd3a2bd4f8a56233ff6212e679e/specification/cognitiveservices/OpenAI.Inference/models/completions/chat_messages.tsp#L80C1-L176C2)
```ts
@discriminator("role")
@added(ServiceApiVersions.v2023_05_15)
@doc("An abstract representation of a chat message as provided in a request.")
model ChatRequestMessage {
  @doc("The chat role associated with this message.")
  role: ChatRole;
}

// An example of a type that inherits from ChatRequestMessage 
@added(ServiceApiVersions.v2023_05_15)
@doc("""
A request chat message containing system instructions that influence how the model will generate a chat completions
response.
""")
model ChatRequestSystemMessage extends ChatRequestMessage {
  @doc("The chat role associated with this message, which is always 'system' for system messages.")
  role: ChatRole.system;

  @doc("The contents of the system message.")
  content: string;

  @doc("An optional name for the participant.")
  name?: string;
}
```

This will generate the code [here](https://github.com/Azure/azure-sdk-for-js/blob/5ad1e5dd5afe164eb79669a44810859f7d408dcb/sdk/openai/openai/sources/generated/src/api/operations.ts#L546C1-L559C70)

```ts
export function _getChatCompletionsSend(
  context: Client,
  deploymentId: string,
  body: ChatCompletionsOptions,
  options: GetChatCompletionsOptions = { requestOptions: {} }
): StreamableMethod<
  GetChatCompletions200Response | GetChatCompletionsDefaultResponse
> {
  return context
    .path("/deployments/{deploymentId}/chat/completions", deploymentId)
    .post({
      body: {
        messages: body["messages"].map((p) => ({ role: p["role"] })),
```

And we will receive this error
```
Type '{ role: string; }[]' is not assignable to type 'ChatRequestMessage[]'.
  Type '{ role: string; }' is not assignable to type 'ChatRequestMessage'.
    Type '{ role: string; }' is missing the following properties from type 'ChatRequestFunctionMessage': name, contentts(2322)
```

To reproduce, generate from the latest commit from [this PR](https://github.com/Azure/azure-rest-api-specs/pull/26737), commit id `c8be3dc86f4dcfd3a2bd4f8a56233ff6212e679e`

qiaozha:
fixes #2139 
fixes #1982

### Summary

In this PR, we will support serialize/deserialize for discriminated union and polymorphic base, as we previously agree that we should not handle other unions because it's hard to predict the type as of the additional properties. 

As such we will not generate predict function for unions, but we still need to distinguish which union variants are special, we only generate special handling logic for those special variants.

We also address the circular reference issue in this PR, as circular reference in inheritance and in combination could be handled differently

### Special Union Variants / Subtypes

In general, we have two kinds of basic special union variants.
1. datetime type
2. binary array type

For model type, we have the following basic three types of special union variants.
3. model with property of datetime type.
4. model with property of binary array type.
5. model that has different property name between rest layer and modular layer.

To extend the combination, we get:
6. nested model i.e. model with property that is a model with one of the above 4-6 conditions.
7. model with property of special union.
8. An array type, with all the above 7 types as the element types.
9. A record type, with all the above 7 types as the value type. (TBD...)

### Example for Discriminated Union

The typespec would be like 
```
    @discriminator("kind")
    union WidgetData {
      kind0: WidgetData0;
      kind1: WidgetData1;
    }

    model WidgetData0 {
      kind: "kind0";
      fooProp: string;
    }
    
    model WidgetData1 {
      kind: "kind1";
      data: bytes;
    }
```
In this case, WidgetData0 is not a special union, but WidgetData1 is, we will not generate special serialize or deserialize logic for it. Our deserialize util would just be like 
```ts
      /** deserialize function for WidgetData1 */
      function deserializeWidgetData1(obj: WidgetData1Output): WidgetData1 {
        return {
          kind: obj["kind"],
          data:
            typeof obj["data"] === "string"
              ? stringToUint8Array(obj["data"], "base64")
              : obj["data"],
        };
      }
      
      /** deserialize function for WidgetDataOutput */
      export function deserializeWidgetData(obj: WidgetDataOutput): WidgetData {
        switch (obj.kind) {
          case "kind1":
            return deserializeWidgetData1(obj);
          default:
            return obj;
        }
      }
```

### Example for Polymorphic Base 

For simplicity, in OpenAI's case, we have typespec like 

```
@doc("An abstract representation of a structured content item within a chat message.")
@discriminator("type")
model ChatMessageContentItem {
  type: string;
}

model ChatMessageTextContentItem extends ChatMessageContentItem {
  type: "text";
  text: string;
}

model ChatMessageImageContentItem extends ChatMessageContentItem {
  type: "image_url";

  @projectedName("json", "image_url")
  imageUrl: ChatMessageImageUrl;
}

model ChatMessageImageUrl {
  @doc("The URL of the image.")
  url: url;

  detail?: ChatMessageImageDetailLevel; // ChatMessageImageDetailLevel is an normal enum
}
```
In this case, ChatMessageTextContentItem is not a special union variant, but ChatMessageImageContentItem is, our serialize util would be like 
```ts
/** serialize function for ChatMessageImageContentItem */
function serializeChatMessageImageContentItem(
  obj: ChatMessageImageContentItem
): ChatMessageImageContentItemRest {
  return {
    type: obj["type"],
    image_url: { url: obj.imageUrl["url"], detail: obj.imageUrl["detail"] },
  };
}

/** serialize function for ChatMessageContentItemUnion */
export function serializeChatMessageContentItemUnion(
  obj: ChatMessageContentItemUnion
): ChatMessageContentItemRest {
  switch (obj.type) {
    case "image_url":
      return serializeChatMessageImageContentItem(
        obj as ChatMessageImageContentItem
      );
    default:
      return obj;
  }
}
```

### Circular Reference Issue 

1. Circular reference from inheritance

This is for typespec like 
```
    @discriminator("kind")
    model Pet {
      kind: string;
      name: string;
      weight?: float32;
    }
    model Cat extends Pet {
      kind: "cat";
      meow: int32;
    }
    @discriminator("type")
    model Dog extends Pet {
      kind: "dog";
      type: string;
      bark: string;
    }
    model Gold extends Dog {
      type: "gold";
      friends: Pet[];
    }
```
In this case, Gold has a property which is type of his ancestors, but the overall model doesn't have anything special subtype, we will not generate the serialize and deserialize utils, if Gold has a birthday property of utcDateTime type, then we should generate the serialize utils for it.

**It is possible that a discriminated union can also have circular reference, the logic should be the same, but ut TBD...**

2. Circular reference from combination

This is for typespec like 
```
      export interface Foo {
        name: string;
        weight?: number;
        bar: Bar;
      }
      
      export interface Bar {
        foo: Foo;
      }
```
In this case, Foo has a property point to Bar and Bar also has a property point to Foo, as this is out of the scope of discriminated union and polymorphic base, the current logic is, we will just detect if these models has anything special, if it's not, we will just give whatever we get to the Modular layer if it's for deserialize and to the Rest layer if it's for serialize. otherwse, we will generate as any. a case in point is *ChatMessage* in OpenAI's case.

### LRO and Paging dependency issue

While integrate the latest OpenAI's typespec, I found they have a LRO operation that is defined in their typespec, but is removed in their typespec, which cause us need to generate LRO helper in RLC, but doesn't need to generate the LRO in Modular, as we currently don't really split RLC from Modular, we need to take both into consideration about whether we have a core-lro or core-paging dependency.


packages/typespec-test/test/openai_modular/generated/typespec-ts/src/utils/serializeUtil.ts
```diff
@@ -0,0 +1,101 @@
...
+  ChatMessageContentItem,
+} from "../models/models.js";
+
+/** serialize function for ChatRequestUserMessage */
+function serializeChatRequestUserMessage(
+  obj: ChatRequestUserMessage
+): ChatRequestUserMessageRest {
+  return {
+    role: obj["role"],
+    content: obj["content"] as any,
```
qiaozha, line 27: In this case, the type of content is defined as `alias ChatMessageContent = string | ChatMessageContentItem[];`  and `ChatMessageContentItem` is another polymohic base with special subtype. which we could handle here https://github.com/Azure/autorest.typescript/pull/2169/files#diff-1a35b9806b61290726e7a05f336d1849f89253f8e4c1591a991919a90b578733R90-R100, but because of the extra variant string, we just generate it as any now. Not sure if you have other thoughts here ? @joheredi @xirzec @MaryGao

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -246,7 +547,9 @@ export function _getChatCompletionsSend(
     .post({
       ...operationOptionsToRequestParameters(options),
       body: {
-        messages: body.messages as any,
+        messages: body["messages"].map((p) =>
+          serializeChatRequestMessageUnion(p)
```
qiaozha, line 551: This is the place we call the serialization for `ChatRequestMessage[]` which element type is a polymorphic base.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/utils/serializeUtil.ts
```diff
@@ -0,0 +1,101 @@
...
+    case "assistant":
+      return serializeChatRequestAssistantMessage(
+        obj as ChatRequestAssistantMessage
+      );
+    case "tool":
+      return serializeChatRequestToolMessage(obj as ChatRequestToolMessage);
+    default:
+      return obj;
+  }
+}
```
qiaozha, line 77: The current serialize and deserialize logic of discriminated union or polymorphic base is we only handle those subtypes that are special union variant. other will fall into the default logic.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/utils/serializeUtil.ts
```diff
@@ -0,0 +1,101 @@
...
+  };
+}
+
+/** serialize function for ChatRequestMessage */
+export function serializeChatRequestMessageUnion(
+  obj: ChatRequestMessage
+): ChatRequestMessageRest {
+  switch (obj.role) {
+    case "user":
+      return serializeChatRequestUserMessage(obj as ChatRequestUserMessage);
```
qiaozha, line 67: because our current implementation adds the polymorphic base into the union, we can't tell if it's `ChatRequestUserMessage` or `ChatRequestMessageParent`  type if the discriminator value is `user`, we have to add a type casting here.

packages/typespec-ts/test/modularUnit/serializeUtil.spec.ts
```diff
@@ -0,0 +1,1410 @@
+import { assert } from "chai";
+import {
+  emitModularSerializeUtilsFromTypeSpec,
+  emitModularOperationsFromTypeSpec
+} from "../util/emitUtil.js";
+import { assertEqualContent } from "../util/testUtil.js";
+
+describe.only("modular special union serialization", () => {
```
qiaozha, line 8:
```suggestion
describe("modular special union serialization", () => {
```

packages/typespec-ts/test/modularUnit/serializeUtil.spec.ts
```diff
@@ -0,0 +1,1410 @@
...
+      color: "red" | "blue";
+    }
+    
+    model Widget1 extends Widget {
+      data: WidgetData
+    }
+
+    interface WidgetService {
+      @get @route("customGet1") customGet1(@body body: Widget1): void;
+    }
```
qiaozha, line 653: add ut for discriminated union here.

packages/typespec-ts/test/modularUnit/serializeUtil.spec.ts
```diff
@@ -0,0 +1,1410 @@
...
+        return { kind: obj["kind"], data: uint8ArrayToString(obj["data"], "base64") };
+      }
+      
+      /** serialize function for WidgetData */
+      export function serializeWidgetDataUnion(obj: WidgetData): WidgetDataRest {
+        switch (obj.kind) {
+          case "kind1":
+            return serializeWidgetData1(obj);
+          default:
+            return obj;
```
qiaozha, line 682: in this case, kind0 is not a special union variant, we don't need to do anything here, could just return whatever we get.

packages/typespec-ts/test/modularUnit/serializeUtil.spec.ts
```diff
@@ -0,0 +1,1410 @@
...
+        const result = await _customGet1Send(context, body, options);
+        return _customGet1Deserialize(result);
+      }
+      `,
+      true
+    );
+  });
+});
+
+describe.only("modular special union deserialization", () => {
```
qiaozha, line 743:
```suggestion
describe("modular special union deserialization", () => {
```

packages/typespec-test/test/openai_generic/generated/typespec-ts/review/openai-generic.api.md
```diff
@@ -118,7 +118,7 @@ export interface CreateChatCompletionRequest {
     model: string | "gpt4" | "gpt-4-0314" | "gpt-4-0613" | "gpt-4-32k" | "gpt-4-32k-0314" | "gpt-4-32k-0613" | "gpt-3.5-turbo" | "gpt-3.5-turbo-16k" | "gpt-3.5-turbo-0301" | "gpt-3.5-turbo-0613" | "gpt-3.5-turbo-16k-0613";
     n?: number | null;
     presencePenalty?: number | null;
-    stop?: string | string[];
+    stop?: Stop;
```
qiaozha, line 121: A small fix for named union.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/rest/responses.ts
```diff
@@ -4,27 +4,89 @@
...
   BatchImageGenerationOperationResponseOutput,
 } from "./outputModels.js";
 
 /** The request has succeeded. */
-export interface GetEmbeddings200Response extends HttpResponse {
+export interface GetAudioTranscriptionAsPlainText200Response
+  extends HttpResponse {
   status: "200";
-  body: EmbeddingsOutput;
+  body: string;
```
qiaozha, line 20: need to check what makes the change here.

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -869,7 +898,9 @@ export function getResponseMapping(
         (property.restApiName === "message" ||
           property.restApiName === "messages") &&
         (property.type.name === "ChatMessage" ||
-          property.type.elementType?.name === "ChatMessage")
+          property.type.elementType?.name === "ChatMessage" ||
+          property.type.name === "ChatResponseMessage" ||
+          property.type.elementType?.name === "ChatResponseMessage")
```
qiaozha, line 903: There is a circular reference in `ChatResponseMessage` and we haven't fix that issue yet, have to add the temporary work around here.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/models/models.ts
```diff
@@ -532,101 +1316,334 @@ export interface ChatCompletions {
...
+}
+
+/** Alias for ChatRequestMessage */
+export type ChatRequestMessage =
+  | ChatRequestSystemMessage
+  | ChatRequestUserMessage
+  | ChatRequestAssistantMessage
+  | ChatRequestToolMessage
+  | ChatRequestFunctionMessage
+  | ChatRequestMessageParent;
```
qiaozha, line 1600: Put a pin here, we have to use Union as the suffix so that we are consistent with hlc.

qiaozha (commented):
initial reviews

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -1105,6 +1102,7 @@ function needsDeserialize(type?: Type) {
 
 export function hasLROOperation(codeModel: ModularCodeModel) {
   return (codeModel.clients ?? []).some((c) =>
+    c.rlcHelperDetails.hasLongRunning ||
```
qiaozha, line 1105: This is to resolve the LRO operation exists in RLC but is removed in the client decorator, and we have to use the core-lro dependency in package.json anyway.
@MaryGao, maybe you need to pay attention to other lro related logic in the new LRO implementation.

.vscode/launch.json
```diff
@@ -154,6 +154,26 @@
...
+        "--timeout",
+        "999999",
+        "--colors",
+        "./test/modularUnit/**/*.spec.ts"
+      ],
+      "runtimeExecutable": "npx",
+      "skipFiles": ["<node_internals>/**"],
+      "type": "pwa-node",
+      "outFiles": ["${workspaceFolder}/packages/*/dist/**/*.js"]
+    },
```
qiaozha, line 176:
```suggestion
```

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/review/load-testing.api.md
```diff
@@ -83,7 +83,7 @@ export interface DimensionValue {
 // @public (undocumented)
 export interface DimensionValueList {
     // (undocumented)
-    readonly value: string[];
+    value: string[];
```
joheredi, line 86: why is `readonly` being removed?

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -294,10 +304,15 @@ function processModelProperties(
     if (!hasDiscriminator) {
       newValue.properties.push({ ...discriminatorInfo });
     }
-    newValue.alias = `${newValue.name}Parent`;
-    newValue.isPolyBaseModel = true;
-    discriminatorInfo?.aliases.push(`${newValue.alias}`);
-    newValue.aliasType = discriminatorInfo?.aliases.join(" | ");
+    if (newValue.name && newValue.name !== "") {
```
joheredi, line 307: `if(newValue.name)`  or `if(Boolean(newValue.name))`  should be enough to check for undefined or empty string

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -294,10 +304,15 @@ function processModelProperties(
...
+      discriminatorInfo?.aliases.push(`${newValue.name}`);
+      newValue.alias = `${newValue.name}`;
+      newValue.name = `${newValue.name}Union`;
```
joheredi, line 310: Do we need to suffix all of these with `Union`?

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -294,10 +304,15 @@ function processModelProperties(
...
+      newValue.aliasType = discriminatorInfo?.aliases.join(" | ");
+      newValue.types = discriminatorInfo?.discriminatedSubtypes;
+      newValue.isPolyBaseModel = true;
```
joheredi, line 313: can we name this property `isPolymorphicBaseModel`?

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -267,10 +276,11 @@ function getEffectiveSchemaType(program: Program, type: Model | Union): Model {
 function processModelProperties(
   context: SdkContext,
   newValue: any,
-  model: Model
+  model: Model,
+  usage: UsageFlags
```
joheredi, line 280: I think we should add usage here and in the other places where it is being added for the first time, as part of an options bag

packages/typespec-ts/src/modular/buildCodeModel.ts
joheredi: Can we also add some docs describing what this check means?

packages/typespec-test/test/openai_modular/generated/typespec-ts/review/openai_modular.api.md
```diff
@@ -39,74 +164,280 @@ export interface AzureCognitiveSearchIndexFieldMappingOptions {
...
+// @public
+export type ChatCompletionsToolSelectionPreset = string;
+
+// @public
+export interface ChatFinishDetails {
+    type: string;
+}
+
+// @public
+export type ChatFinishDetailsUnion = StopFinishDetails | MaxTokensFinishDetails | ChatFinishDetails;
```
joheredi, line 350: Why do we have to include ChatFinishDetails as part of the union? IMO it should only be extended by StopFinishDetails and MaxTokensFinishDetails but not be part of the union

packages/typespec-ts/src/modular/buildCodeModel.ts
joheredi: But if there is no obvious dfault value for this and has to be required maybe it makes sense to keep it as positional

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -833,29 +861,40 @@ export function getResponseMapping(
...
+          "deserialize"
+        );
+        definition = `"${property.clientName}": ${deserializeFunctionName}(${propertyFullName})`;
       } else {
+        if (
+          property.type.name === "ChatMessage" ||
+          property.type.name === "ChatResponseMessage"
+        ) {
+          property;
+        }
```
qiaozha, line 885:
```suggestion
```

joheredi (approved):

packages/typespec-ts/test/modularUnit/serializeUtil.spec.ts
```diff
@@ -0,0 +1,1521 @@
...
+
+  it("should generate serialize util if there's a special discriminated union variant of model with bytes property", async () => {
+    const tspContent = `
+    @discriminator("kind")
+    union WidgetData {
+      kind0: WidgetData0;
+      kind1: WidgetData1;
+    }
+
+    model WidgetData0 {
```
lirenhe, line 636: It would be good to include more test cases like:

WidgetData0 & WidgetData1 all contains the data of binary.
WidgetData0 contains the data of datetime and WidgetData1 contains the data of binary.

lirenhe (approved):

packages/typespec-ts/test/modularUnit/serializeUtil.spec.ts
qiaozha: For case 1, probably not, for case 2, I think I have a case. let me find the link.

qiaozha:
I will merge this PR and use https://github.com/Azure/autorest.typescript/issues/2253 to track other non-problematic union serialize and deserialize work.  Let me know if anyone has any concerns.

Issue #2157, opened 2023-12-05
Support decimal type from typespec


MaryGao:
https://github.com/Azure/azure-rest-api-specs-pr/pull/13119#discussion_r1376998182

In typespec, we have `decimal` and `decimal128`. And in health insight we have [decimal type used](https://github.com/Azure/azure-rest-api-specs-pr/blob/d0af087407c0f4ee2ce49a9f51d26ad580d8e616/specification/cognitiveservices/HealthInsights/HealthInsights.Common/model.common.fhir.elements.tsp#L178C1-L179C19). 

```
  @doc("Numerical value (with implicit precision)")
  value?: decimal;
```

https://microsoft.github.io/typespec/language-basics/built-in-types#numeric-types

TypeSpec definition for decimal:

```
/**
 * A decimal number with any length and precision. This represent any `decimal` value possible.
 * It is commonly represented as `BigDecimal` in some languages.
 */
scalar decimal extends numeric;

/**
 * A 128-bit decimal number.
 */
scalar decimal128 extends decimal;
```

We need to decide on how to support this type in both RLC and Modular.

xirzec:
How is this being represented over the wire? Presumably they are not sending it as a JSON number since that would potentially lose precision. If it's a string my preference is to keep it as a string and let the consumer use whatever userland library they prefer to handle the number manipulation.

MaryGao:
Using @encode as string to preserve decimal's precision could be our recommandation. But I may have two concerns:
- Java/.Net/Python have the native support for decimal, not sure it is a good idea to suggest that for JS
- Service team may follow an external specification and we have no choice to encode them as string e.g health insight follows the [FHIR standard quantity model](https://www.hl7.org/fhir/R4/datatypes.html#Quantity) and [discussion here](https://github.com/Azure/azure-rest-api-specs-pr/pull/13119#discussion_r1377135364). 

I agree that for calculations, users could pick their own lib then converting to a number or string for us(aka client SDK) and within the safe range it should be SAFE also.

However I am worrying about this has potential risks but it is a silent process. 

Not sure we need to involve more communications with service team to clarify this because eventually they will take the cost.

On the other hand how to educate end users that the contact with server side is based on `decimal` so you may pick your own decimal lib for any calculations. And we expect `0.2 + 0.1 = 0.3` not `0.30000000000000004`.

MaryGao:
pls ignore the wrong linked to lro epic :)

bterlson:
[...]

PR #2170, opened 2023-12-19
Support decimal type in RLC

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2157 
Support decimal in RLC - map the decimal to `number` in JS directly, and will 
- add below comments in generated code
```
Please note the field was supposed to be a decimal but JavaScript does not have a native 'BigDecimal' data type.
So it was converted to a number instead. It is recommended to use a third-party library like 'decimal.js' to handle 
any calculations.
```
- report a warning 
```
Please note the decimal type will be converted to number. If you strongly care about precision you can use @encode to encode it as a string for the property - {your prop name}.
```
Encoding as string is still pending typespec discussion https://github.com/microsoft/typespec/issues/2762.

### DIFF
--- a/packages/typespec-ts/src/utils/modelUtils.ts
+++ b/packages/typespec-ts/src/utils/modelUtils.ts
@@ -1052,7 +1052,7 @@ function getSchemaForStdScalar(
     case "safeint":
       return applyIntrinsicDecorators(program, type, {
         type: "number",
-        format: "int64"
+        format: "safeint"
       });
     case "uint8":
       return applyIntrinsicDecorators(program, type, {
@@ -1077,17 +1077,43 @@ function getSchemaForStdScalar(
     case "float64":
       return applyIntrinsicDecorators(program, type, {
         type: "number",
-        format: "double"
+        format: "float64"
       });
     case "float32":
       return applyIntrinsicDecorators(program, type, {
         type: "number",
-        format: "float"
+        format: "float32"
       });
     case "float":
       return applyIntrinsicDecorators(program, type, {
         type: "number",
         format: "float"
+      });
+    case "decimal":
+      reportDiagnostic(program, {
+        code: "decimal-to-number",
+        format: {
+          propertyName: relevantProperty?.name ?? ""
+        },
+        target: relevantProperty ?? type
+      });
+      return applyIntrinsicDecorators(program, type, {
+        type: "number",
+        format: "decimal",
+        description: "decimal"
+      });
+    case "decimal128":
+      reportDiagnostic(program, {
+        code: "decimal-to-number",
+        format: {
+          propertyName: relevantProperty?.name ?? ""
+        },
+        target: relevantProperty ?? type
+      });
+      return applyIntrinsicDecorators(program, type, {
+        type: "number",
+        format: "decimal128",
+        description: "decimal128"
       });
     case "string":
       if (format === "binary") {
@@ -1243,6 +1269,16 @@ function getEnumStringDescription(type: any) {
   return undefined;
 }
 
+function getDecimalDescription(type: any) {
+  if (
+    (type.format === "decimal" || type.format === "decimal128") &&
+    type.type === "number"
+  ) {
+    return `Please note the field was supposed to be a ${type.format} but JavaScript does not have a native 'BigDecimal' data type.\nSo it was converted to a number instead. It is recommended to use a third-party library like 'decimal.js' to handle\nany calculations.`;
+  }
+  return undefined;
+}
+
 export function getFormattedPropertyDoc(
   program: Program,
   type: ModelProperty | Type,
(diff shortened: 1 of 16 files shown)


packages/typespec-ts/test/integration/scalar.spec.ts
```diff
@@ -77,4 +78,118 @@ describe("Scalar Client", () => {
...
+  });
+
+  it("should post decimal verify", async () => {
+    try {
+      // prepare the verification
+      const getResult = await client
+        .path("/type/scalar/decimal/prepare_verify")
+        .get();
+      // do any calculation based on third party library
+      let total = new Decimal(0);
```
MaryGao, line 157: Include `decimal.js` for calculation.

qiaozha: Is this meant to be a test case for round trip ? and because we have precision loss, we have to leverage the third party library ? I thought we already have precison loss, when we get the data from json.parse ?

MaryGao: There are two kinds of risks - 1) calculation and 2) serialization/deserialization. You could see detailed [discussion here](https://github.com/Azure/autorest.typescript/issues/2157#issuecomment-1850671951).

Here is mainly for cadl-ranch test case for calculation, in that case we need three decimals to add together and then the sum would be used to verify in server side.

qiaozha: Thanks for the detailed explaination, I am not quite sure if it's necessary to import the decimal.js here? Just wondering if there's any case that toFixed could not work but decimal.js can, giving the result and the added numbers are all primitive number type ? 
@joheredi @xirzec any thoughts here ?

MaryGao: I may not get your point of toFixed. Let me explain the case more, in cadl-ranch we would get three decimals they are 0.1, 0.1 and 0.1. Then we need to add these three decimals and then send the total `0.3` back to server side.

In JavaScript directly adding them would lose the precision and we got `0.30000000000000004`.

```js
const a = 0.1 +0.1 +0.1;
console.log(a, a.toFixed());
// [LOG]: 0.30000000000000004,  "0" 
```

Including the `decimal.js` as a dev dependency is trying to simulate customers' behavior who really cares about precision for any calculation.

joheredi: In the first example you are missing the digit parameter of toFixed

```ts
const total = 0.1 + 0.1 + 0.1;
console.log(total, total.toFixed(1));
// [LOG]: 0.30000000000000004,  "0.3" 
```

I think we should try not to add the external library unless completly necessary

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -1243,14 +1269,25 @@ function getEnumStringDescription(type: any) {
   return undefined;
 }
 
+function getDecimalDescription(type: any) {
+  if (
+    (type.format === "decimal" || type.format === "decimal128") &&
+    type.type === "number"
+  ) {
+    return `Please note the field was supposed to be a ${type.format} but JavaScript does not have a native 'BigDecimal' data type.\nSo it was converted to a number instead. It is recommended to use a third-party library like 'decimal.js' to handle\nany calculations.`;
```
joheredi, line 1277: if we are going verbose maybe we can do something like

```ts
/**
 * NOTE: This property is represented as a 'number' in JavaScript, but it corresponds to a 'decimal' type in other languages.
 * Due to the inherent limitations of floating-point arithmetic in JavaScript, precision issues may arise when performing arithmetic operations.
 * If your application requires high precision for arithmetic operations or when round-tripping data back to other languages, consider using a library like decimal.js, which provides an arbitrary-precision Decimal type.
 * For simpler cases, where you need to control the number of decimal places for display purposes, you can use the 'toFixed()' method. However, be aware that 'toFixed()' returns a string and may not be suitable for all arithmetic precision requirements.
 * Always be cautious with direct arithmetic operations and consider implementing appropriate rounding strategies to maintain accuracy.
 */
```

joheredi (approved):
Left a couple of comments but look fine otherwise

packages/typespec-ts/test/integration/scalar.spec.ts
MaryGao: There are two concerns, 1) toFixed would return a string; do we need to castcade to number? 2) the test case is trying to simulate the items preparation from server side during runtime then how do we know which digits we sould use in toFixed?

qiaozha: @joheredi is right about my previous point, my another point is that even if we need to use decimal.js to handle the calculation as of the result out of number range, because we can only accept number result, the decimal.js toNumber will also have precision loss. how is that precision loss different than using toFixed ?

lirenhe: One suggestion:  Can we just send 0.30000000000000004 to server which is the actual result and let service side to return error code. 
We could still verify the 'failure' error code or exception and add a comment that this case failed because we will lose the precision when doing the calculation and this is a known issue in JS.

qiaozha (approved):

Issue #1971, opened 2023-08-14
Better modularization of request/response logic


deyaaeldeen:
The modular code generator generates the following function to handle processing the response: https://github.com/Azure/autorest.typescript/blob/b9413192a7ed0b7694e8205db33b578665f5e557/packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts#L53 and it mainly checks if the response is unexpected: https://github.com/Azure/autorest.typescript/blob/b9413192a7ed0b7694e8205db33b578665f5e557/packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts#L56 and create the output object: https://github.com/Azure/autorest.typescript/blob/b9413192a7ed0b7694e8205db33b578665f5e557/packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts#L60

The two parts can be refactored into two separate functions. I need the function that just creates the output object to call it to deserialize server-sent events. That will save me from doing it manually, see https://github.com/Azure/azure-sdk-for-js/blob/cbbcc187abee610b7b11c1be7858a3bbee1d6fe2/sdk/openai/openai/src/api/operations.ts#L572 for an example.

PR #2180, opened 2023-12-26
Use reshape to simplify deserialization

joheredi:
Integrate the new reshape function from core-util to simplify serialization and deserialization.

reshape is WIP in https://github.com/Azure/azure-sdk-for-js/pull/28095


xirzec:
Very nice! I love that we're removing more lines than we're adding in this PR. 

I assume we could do a similar change for serializing input shapes from the customer?

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -283,105 +265,63 @@ export async function _getChatCompletionsDeserialize(
...
+    deserializedResponse,
+    "usage.prompt_tokens",
+    "promptTokens"
+  );
+  deserializedResponse = reshape(
+    deserializedResponse,
+    "usage.total_tokens",
+    "totalTokens"
+  );
+  return deserializedResponse as ChatCompletions;
```
qiaozha, line 324: I wonder if we should just pass a function to reshape a specific type in a more general way ? feel like if every property has been renamed, we will still need to traverse the object in the operatoins.ts ? https://github.com/Azure/autorest.typescript/issues/1971 maybe just leverage reshape in this better modularization ? 

/cc @dgetu

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -460,105 +400,63 @@ export async function _getChatCompletionsWithAzureExtensionsDeserialize(
     throw createRestError(result);
   }
 
-  return {
-    id: result.body["id"],
-    created: new Date(result.body["created"]),
-    choices: result.body["choices"].map((p) => ({
-      message: !p.message ? undefined : (p.message as any),
```
qiaozha, line 467: I am not sure if removing this would work, if I remember corrrectly, there's a recursive reference here. which has not yet been resolved. https://github.com/Azure/autorest.typescript/issues/1982 

/cc @dgetu

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -192,6 +185,77 @@ export function getDeserializePrivateFunction(
...
+  const response = operation.responses[0]!;
+
+  const responseProperties = getAllProperties(response.type);
+  const statements: string[] = [];
+  const visitedTypes = new Set<string>(); // Set to track visited types
+
+  statements.push(`let deserializedResponse: unknown = result.body;`);
+
+  deserialize(responseProperties, statements, visitedTypes);
+  statements.push(`return deserializedResponse as ${resultType};`);
```
qiaozha, line 198: just a little curious, why we need as here ?

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -752,17 +816,6 @@ function getRequestModelMapping(
...
-        (property.restApiName === "message" ||
-          property.restApiName === "messages") &&
-        (property.type.name === "ChatMessage" ||
-          property.type.elementType?.name === "ChatMessage")
-      ) {
-        definition = `"${property.restApiName}": ${
-          !property.optional
-            ? `${propertyFullName} as any`
-            : `!${propertyFullName} ? undefined : ${propertyFullName} as any`
-        }`;
```
qiaozha, line 765: This is the workaround for recursive reference in codegen.

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -895,6 +957,15 @@ export function getResponseMapping(
   return props;
 }
 
+function reshapeValue(type: Type) {
+  switch (type.type) {
+    case "datetime":
+      return "(value) => new Date(value as string)";
+    default:
```
qiaozha, line 964: missing bytes ? and datetime serialization is another case

packages/typespec-test/test/contentsafety_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -86,19 +86,8 @@ export async function _analyzeTextDeserialize(
...
-          blockItemId: p["blockItemId"],
-          blockItemText: p["blockItemText"],
-        })),
-    analyzeResults: result.body["analyzeResults"].map((p) => ({
-      category: p["category"],
-      severity: p["severity"],
-    })),
-  };
+  let deserializedResponse: unknown = result.body;
+  return deserializedResponse as AnalyzeTextResult;
```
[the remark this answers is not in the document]
joheredi: Yup I'm pushing that change next.

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/src/api/operations.ts
```diff
@@ -196,30 +196,18 @@ export async function _receiveCloudEventsDeserialize(
...
-        subject: p.event["subject"],
-      },
-    })),
-  };
+  let deserializedResponse: unknown = result.body;
+  deserializedResponse = reshape(
+    deserializedResponse,
+    "value[].event.data_base64",
+    "dataBase64"
+  );
```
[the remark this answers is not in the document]
joheredi: Yes, still need to handle the conversion. I let it out temporarily b/c I wanted to double check if we want to keep the current approach or need to update

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/api/operations.ts
joheredi: If the recursive properties don't need to be reshaped it should be fine, as we'll be returning the same as we got from the service. True, this doesn't handle recursive properties that need to be transformed.

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
joheredi: Because the reshape function returns `Unknown` I was thinking to suggest adding some type magic to the function so that it could actually returns the type we need, however as I started trying it out it became too complex and likely not worth investing in it and maintaining for the internal added type-safety

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
joheredi: Yeah I wanted to try this out without the workaround, since the properties in message don't need to be reshaped I think we should be fine without it

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
joheredi: yes, I'll add both of them

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
xirzec: I'm fine changing `reshape` to just deal in `any` if that saves you effort :)

PR #2184, opened 2024-01-02
Move to prettier v3

jeremymeng:
- use `dev-tool run vendored ` prefix for prettier when generating for Azure SDK for JS repo. Dependency on prettier is no longer used thus removed.  Related PR: https://github.com/Azure/azure-sdk-for-js/pull/28127
- updated prettier to v3 generated package.json in non-azsdkjs packages.
- upgrade to prettier v3 in this repo as codegen uses prettier to format generated code and we want generated code to be consistent and morden.


joheredi (approved):

packages/autorest.typescript/src/typescriptGenerator.ts
```diff
@@ -144,7 +144,7 @@ export async function generateTypeScriptLibrary(
 
     // Format the contents if necessary
     if (isJson || isSourceCode) {
-      fileContents = prettier.format(
+      fileContents = await prettier.format(
```
MaryGao, line 147: Thank you so much for the fix! Does that mean all generated codes are not format successfully?

MaryGao (approved):
LGTM.

packages/autorest.typescript/src/typescriptGenerator.ts
jeremymeng: yes, v3 makes the method async.

jeremymeng:
> just notice that there are a lot of file format changes and wondering why upgrading would cause that because our prettierTypeScriptOptions didn't change.

Prettier v3 has some changes in options' default values, the most impactful one to us is the `trailingComma` option now defaults to `"all"`.  There are formatting changes around tenery operator as well.

PR #2191, opened 2024-01-08
fix-extra-api-version-in-query-issue

### LINKED ISSUE #2148: [Codegen/TypeSpec] Handle apiVersion as Path Parameter
Microsoft Fabric's API guidelines require the API version to be a path parameter rather than a query parameter. We would like to use this tool for generating our TypeScript SDK from our TypeSpec definition but incorporating the API version as a path parameter cannot be accomplished as cleanly as we would like, at least from what we know of the tool. There are two things we would like to be able to do.

First, we would like to set the API version as part of the client creation or route definition rather than having to specify the API version as a parameter in every request. It seems like this is already a feature in the [autorest.python project](https://github.com/Azure/autorest.python/issues/1778).

Second, we would like `api-version` to not be automatically added as a query parameter to all requests. I believe there is already an [issue](https://github.com/Azure/autorest.typescript/issues/2094) that has been created related to this along with a [pull request](https://github.com/Azure/autorest.csharp/pull/3931) to resolve it.

Are there already features to support this use case that we are simply not aware of? If not, would it be possible to add this as a feature? 

Here is a quick example of what our TypeSpec looks like.

```
model ApiVersionPathParameter {
  apiVersion: string;
}

@resource("exampleResources")
model ExampleResource {
  @Key
  id: string;
}

@tag("ExampleResources")
interface ExampleResourceInterface {
  @autoRoute
  @readsResource(ExampleResource)
  get(
    ...ApiVersionPathParameter
    ...Resource.ResourceParameters<ExampleResource>
  ): ExampleResource;
}
```

The generated `clientDefinition.ts` file would then have a Routes interface like the following.

```
export interface Routes {
  (
    path: "/{apiVersion}/exampleResources/{id}",
    apiVersion: string,
    id: string
): ExampleResourcesGet;
```

Is there a way to either set the path to `"/v1/exampleResource/{id}"` with no `apiVersion` parameter or set the client's `baseUrl` to include the api version at the end and then have all paths look like `"/exampleResource/{id}"`?

qiaozha:
fixes #2148


packages/rlc-common/src/buildClient.ts
```diff
@@ -361,6 +361,11 @@ export function getClientFactoryBody(
         }
       });`;
   }
+
+  let apiVersionPolicyStatement = "";
+  if (model.apiVersionInfo?.definedPosition === "path") {
```
qiaozha, line 366: we should remove this policy regardless if there is default value for the path api version.

packages/typespec-test/test/chatApi_modular/generated/typespec-ts/src/rest/chatProtocolClient.ts
```diff
@@ -18,7 +18,7 @@ export default function createClient(
   options: ClientOptions = {}
 ): ChatProtocolContext {
   const baseUrl = options.baseUrl ?? `${endpoint}`;
-  options.apiVersion = options.apiVersion ?? "2023-10-01-preview";
+
```
qiaozha, line 21: In this case, there's no api version defined in the operations.

packages/typespec-test/test/anomalyDetector/generated/typespec-ts/src/anomalyDetectorClient.ts
```diff
@@ -51,5 +51,7 @@ export default function createClient(
     options
   ) as AnomalyDetectorClient;
 
+  client.pipeline.removePolicy({ name: "ApiVersionPolicy" });
```
qiaozha, line 54: there's is a path level api version, we should not add this policy as it's for query api version parameter.

packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/bodyComplexRestClient.ts
```diff
@@ -31,5 +30,6 @@ export default function createClient(
 
   const client = getClient(baseUrl, options) as BodyComplexRestClient;
 
+  client.pipeline.removePolicy({ name: "ApiVersionPolicy" });
```
MaryGao, line 33: A little bit suprised with this change, does that mean there is a path-level apiVersion in client?

qiaozha: No, this means, there's no query api version in client, there're three cases, we don't need this policy.
1. client level path parameter, in url.
2. operation level path parameter.
3. not defined.

packages/typespec-test/test/faceai/generated/typespec-ts/README.md
```diff
@@ -0,0 +1,57 @@
+# Azure Face REST client library for JavaScript
```
qiaozha, line 1: I put face API into smoke-test as there's multipart-form data related change. so I guess it's okay.

packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/bodyComplexRestClient.ts
MaryGao: Okay we have an option in ClientOptions for apiVersion, so I think 
1.for case 1 this option would take effective in path parameter
2.but i am not sure with case 2 and 3, personally i would prefer to deco the option if we plan to remove relevant core policy. not sure we have a way to do that.

packages/typespec-ts/src/transform/transformApiVersionInfo.ts
```diff
@@ -26,23 +27,20 @@ export function transformApiVersionInfo(
...
-    pathVersionDetail?.defaultValue ??
-    queryVersionDetail?.defaultValue;
-
-  // Clear the default value if there exists cross version
-  if (isCrossedVersion) {
-    defaultValue = undefined;
-  }
+  const defaultValue =
+    (pathVersionDetail || queryVersionDetail) && !isCrossedVersion
+      ? getEnrichedDefaultApiVersion(program, dpgContext) ??
```
MaryGao, line 40: I found there is inconsistency cross languages with default value and i am checking wiht tsp team [here](https://teams.microsoft.com/l/message/19:906c1efbbec54dc8949ac736633e6bdf@thread.skype/1704709079749?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1676977636162&teamName=Azure%20SDK&channelName=TypeSpec%20Discussion%20%F0%9F%90%AE&createdTime=1704709079749).

joheredi (approved):
It looks good to me

packages/rlc-common/src/interfaces.ts
```diff
@@ -106,7 +106,7 @@ export interface ApiVersionInfo {
   isCrossedVersion?: boolean;
 }
 
-export type ApiVersionPosition = "path" | "query" | "both" | "none";
+export type ApiVersionPosition = "path" | "query" | "url" | "duplicate" | "none";
```
MaryGao, line 109: not sure we could have a better name to distingush the `path` and `url`?

qiaozha: only the default value for api version defined in url and query can actually has effect right now, for the path parameter, currently we don't have a way to set it, at least in RLC. 

For the name, I don't have a strong opinion, maybe @joheredi could help to decide ? Thanks

packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/bodyComplexRestClient.ts
qiaozha: In case 2 and case 3, if there's any case that the previous policy added on query api version parameter work, this would either mean the service side has been using the api version twice or means they should have a query parameter for api version, but they don't.  Both ways don't see like a standard, I probably don't think we should support. not sure if @joheredi  or @xirzec has any different opinions on this ?

packages/typespec-ts/src/transform/transformApiVersionInfo.ts
qiaozha: Giving what Mark and Timothee mentioned in the channel, I think it's align with what we have in our implementation, also if there're cases that we should revisit our decision on the default value logic, we could file another PR to fix it, as this PR is trying to remove the extra api version in query parameter, what do you think ?

packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/bodyComplexRestClient.ts
qiaozha: I should point out, in the issue #2148, fabric has make it clear that they have a path parameter api version and they don't want api version version appear in their query parameter, which is the same case for face ai.

xirzec: it doesn't appear that this policy is publicly accessible to customer since core-client-rest doesn't export it, but is there a way we could indicate to the client (perhaps through a client option) to not include the policy in the first place?

packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/bodyComplexRestClient.ts
```diff
@@ -13,7 +13,6 @@ export default function createClient(
   options: ClientOptions = {}
 ): BodyComplexRestClient {
   const baseUrl = options.baseUrl ?? `http://localhost:3000`;
-  options.apiVersion = options.apiVersion ?? "2016-02-29";
```
qiaozha, line 16: don't need this.

packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/bodyComplexRestClient.ts
qiaozha: create an issue to track this https://github.com/Azure/azure-sdk-for-js/issues/28204

packages/rlc-common/src/interfaces.ts
qiaozha: @joheredi do you have any comments on this ? feel free to let me know. Thanks

joheredi: I wouldn't want to take a breaking change to rename path, also I think the path is a standard term. For url I tink I'd be fine to name it baseUrl, only if we don't need to do a breaking change for that.

MaryGao (approved):
LGTM

packages/rlc-common/src/interfaces.ts
qiaozha: I am not sure about the breaking here, I think ApiVersionPosition should be used internally? but I am fine to go with path and baseurl.

Issue #2186, opened 2024-01-04
TS2339 compile error when using both BearerAuth and OAuth2Auth for authentication


yufeih:
`@azure-tools/typespec-ts` version: 0.19.0

When using both `BearerAuth` and `OAuth2Auth`, the generated code results in `TS2339` error.

TypeSpec file:

```tsp
@useAuth([BearerAuth, OAuth2Auth<[{
    type: OAuth2FlowType.clientCredentials;
    tokenUrl: "https://login.microsoftonline.com/common/oauth2/v2.0/token";
    refreshUrl: "https://login.microsoftonline.com/common/oauth2/v2.0/token";
    scopes: ["foo/.default"];
}]>])
namespace MyService;
```

Generated code:

```ts
export default function createClient(
  credentials: TokenCredential | KeyCredential,
  options: ClientOptions = {}
): MyServiceClient {
   //...

  const client = getClient(baseUrl, credentials, options) as MyServiceClient;

  client.pipeline.addPolicy({
    name: "customKeyCredentialPolicy",
    async sendRequest(request, next) {
      request.headers.set("Authorization", "bearer " + credentials.key);  // <--- TS2339
      return next(request);
    },
  });
  return client;
}
```

Error:

```
Property 'key' does not exist on type 'TokenCredential | KeyCredential'.
  Property 'key' does not exist on type 'TokenCredential'.ts(2339)
```

MaryGao:
Thanks for reporting and this would be a codegen bug and the `customKeyCredentialPolicy` should only be added when the credential is `KeyCredential`. I will work on the fix.

MaryGao:
@yufeih I'd like to confirm with you one detail, for BearerAuth for the word `bearer ` in `Authorization` header do we care about it is capital case or lower case?

yufeih:
> @yufeih I'd like to confirm with you one detail, for BearerAuth regarding the word `bearer ` in `Authorization` header do we(server side) care about it is capital case or lower case?

Our server should treat is as case insensitive, it seems `Bearer` is more appropriate. Here is the codegen for C#:

```csharp
        private const string AuthorizationApiKeyPrefix = "Bearer";
```

MaryGao:
Thanks for the info! Yes, generally `Bearer` is the convention and server side should treat as case insensitive.

PR #2195, opened 2024-01-09
Fix the union credential issue

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2186


packages/rlc-common/src/buildPaginateHelper.ts
```diff
@@ -7,6 +7,7 @@ import { paginateContent } from "./static/paginateContent.js";
 
 export function buildPaginateHelper(model: RLCModel) {
   const pagingInfo = model.helperDetails;
+  const isBranded = model.options?.branded ?? true;
```
MaryGao, line 10:
```suggestion
```

packages/typespec-test/test/mslearn/generated/typespec-ts/src/learnClient.ts
```diff
@@ -0,0 +1,62 @@
...
+    credentials: {
+      scopes: options.credentials?.scopes ?? [
+        "https://learn.microsoft.com/.default",
+      ],
+    },
+  };
+
+  const client = getClient(baseUrl, credentials, options) as LearnClient;
+
+  if (isKeyCredential(credentials)) {
```
MaryGao, line 42: The helper is supposed to be added in core-auth.

packages/rlc-common/src/buildClient.ts
```diff
@@ -228,7 +240,8 @@ function isSecurityInfoDefined(
 
 export function getClientFactoryBody(
   model: RLCModel,
-  clientTypeName: string
+  clientTypeName: string,
+  isMultipleCredential: boolean
```
joheredi, line 244: I would suggest to instead of adding this as a positional parameter to add an options bag. This will help us prevent having a lot of positional parameters if we need to keep adding

```ts
interface GetClientFactoryOptions {
   isMultipleCredential: boolean;
}
export function getClientFactoryBody(
  model: RLCModel,
  clientTypeName: string,
  options?: GetClientFactoryBodyOptions
)

```

joheredi (approved):

packages/typespec-test/test/mslearn/generated/typespec-ts/.eslintrc.json
```diff
@@ -0,0 +1,11 @@
+{
+  "plugins": ["@azure/azure-sdk"],
+  "extends": ["plugin:@azure/azure-sdk/azure-sdk-base"],
```
lirenhe, line 3: As we already have unit tests for this feature, not sure whether it is necessary add ms-learn as a test project for now.

lirenhe (approved):

packages/rlc-common/src/buildClient.ts
qiaozha: updated.

PR #2205, opened 2024-01-16
fix-function-parameter-name-issue

### LINKED ISSUE #1871: Generation fails if an endpoint parameter named `function`
It happened that in our API we have a parameter named `function`, if I use autorest to generate API wrapper it prints the below error, and the process gets stuck. It would be good if autorest would modify the name somehow if it is a JS keyword.

spec file (reproduction-openapi.yaml)
```yaml
info:
  title: Aeternity Middleware
  version: 1.47.0
openapi: 3.0.0
paths:
  /contracts/logs:
    get:
      description: Get contract logs.
      operationId: get_contracts_logs
      parameters:
      - description: Prefix of the contract called function
        in: query
        name: function
        required: false
        schema:
          example: mint
          type: string
      responses:
        '200':
          content:
            application/json:
              schema:
                example: response
                type: string
          description: <description>
```

autorest configuration (reproduction.yaml)
```yaml
version: ^3.6.3
use-extension:
  '@autorest/typescript': ^6.0.2
  '@autorest/modelerfour': ^4.26.2
input-file: reproduction-openapi.yaml
generator: typescript
```

```bash
$ npx autorest ./reproduction.yaml
AutoRest code generation utility [cli version: 3.6.3; node: v18.16.0]
(C) 2018 Microsoft Corporation.
https://aka.ms/autorest
info    |    Loading AutoRest core      '/Users/<user>/.autorest/@autorestcore@3.9.6/nodemodules/@autorest/core/dist' (3.9.6)
info    |    Loading AutoRest extension '@autorest/typescript' (^6.0.2->6.0.2)
info    |    Loading AutoRest extension '@autorest/modelerfour' (^4.26.2->4.26.2)
warning | PreCheck/CheckDuplicateSchemas | Checking for duplicate schemas, this could take a (long) while.  Run with --verbose for more detail.
fatal   | SyntaxError: Variable declaration expected. (27:14)
  25 |   };
  26 | 
   27 | export const function: OperationQueryParameter = {
     |              ^
  28 |     parameterPath: ["options","function"],mapper: {
  29 |       serializedName: "function",
  30 |       type: {
```

qiaozha:
fixes #1871
fixes https://github.com/Azure/autorest.typescript/issues/1342


MaryGao (approved):
LGTM

packages/autorest.typescript/test/integration/generated/azureParameterGrouping/src/models/index.ts
```diff
@@ -33,7 +33,7 @@ export interface ParameterGroupingPostOptionalParameters {
 /** Parameter group */
 export interface ParameterGroupingPostReservedWordsParameters {
   /** 'from' is a reserved word. Pass in 'bob' to pass. */
-  from?: string;
+  fromParam?: string;
```
joheredi, line 36: Can we adjust the reserved words to make from not reserved when it is a property name?

joheredi: Also function, should be fine as a property name but not as an interface identifier

MaryGao: Yeah, it seems introducing un-necessary breakings here. Can we mitigate this?

packages/autorest.typescript/test/unit/transforms/parameterTransforms.spec.ts
```diff
@@ -422,7 +422,7 @@ describe("parameterTransforms", () => {
       assert.equal(p1.parameter, param1);
     });
 
-    it("should not guard non-required query parameters name", () => {
+    it("should guard non-required query parameters name", () => {
```
MaryGao, line 425: Is the mismatch between mapper and property name from required parameter?

Do we have a case for that?

joheredi (approved):

Issue #2220, opened 2024-01-17
optional property with all types


qiaozha:
TypeSpec link: https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/type/property/optionality/main.tsp

Hint: think of a case where parameter is optional property with special word as the name.

v-jiaodi:
When running the test case, the following error was encountered：
![image](https://github.com/Azure/autorest.typescript/assets/80496810/e4db250b-86c0-452f-b2f8-26e1f1a74677)

joheredi:
Blocked by https://github.com/Azure/autorest.typescript/issues/1971

qiaozha:
close it as completed via #2613

PR #2229, opened 2024-01-22
Add modular coverage cases for optional property in cadl-ranch

v-jiaodi:
Fixes https://github.com/Azure/autorest.typescript/issues/2220
Added List:
1.optional property.
run test case error:
image


joheredi:
This is being addressed in #2613 closing in favor of that PR

Issue #2207, opened 2024-01-17
TypeSpec Azure Core basic scenarios


qiaozha:
Typespec link 
https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/azure/core/basic/main.tsp 

we have several operations un-supported in Modular 
<img width="827" alt="image" src="https://github.com/Azure/autorest.typescript/assets/55688292/2ebc0936-a4d3-4ab2-afd3-b784d57e3f30">

v-jiaodi:
Azure Core BasicClient createOrReplace method request restriction, When add create or replace scenarios for azure core, get error as follow:

![image](https://github.com/Azure/autorest.typescript/assets/80496810/a6323853-8892-4189-b9ae-59169e10b2b6)

qiaozha:
pending core-client-rest 1.2.1 release.

PR #2236, opened 2024-01-23
Supplement modular azure core basic test case

v-jiaodi:
Fixes  https://github.com/Azure/autorest.typescript/issues/2207


packages/typespec-ts/package.json
```diff
@@ -62,7 +62,7 @@
     "@azure-tools/cadl-ranch-expect": "^0.9.0",
     "@azure-tools/cadl-ranch": "^0.10.0",
     "chalk": "^4.0.0",
-    "@azure-rest/core-client": "^1.1.6",
+    "@azure-rest/core-client": "/Users/zhangqiaoqiao/work/code/azure-sdk-for-js/sdk/core/core-client-rest/azure-rest-core-client-1.2.0.tgz",
```
qiaozha, line 65: we will need a new release of `@azure-rest/core-client` once https://github.com/Azure/azure-sdk-for-js/pull/27343 is merged.

packages/typespec-ts/test/modularIntegration/azureCore.spec.ts
```diff
@@ -159,4 +159,58 @@ describe("BasicClient Classical Client", () => {
...
+
+  it.only("should create or replace a user", async () => {
+    try {
+      const user = await client.createOrReplace(1, {
+        name: "Madge",
+        id: 1,
+        etag: "11bdc430-65e8-45ad-81d9-8ffa60d55b59"
+      }, {
+        contentType: "application/json"
+      } as any);
```
qiaozha, line 183: once core is fixed, we can pass the content type like this way or in the requestOptions.headers['content-type']

packages/typespec-ts/test/modularIntegration/azureCore.spec.ts
```diff
@@ -159,4 +159,58 @@ describe("BasicClient Classical Client", () => {
...
+      console.log(user);
+      assert.strictEqual(user?.id, 1);
+      assert.strictEqual(user?.name, "Madge");
+      assert.strictEqual(user?.etag, "11bdc430-65e8-45ad-81d9-8ffa60d55b59");
+    } catch (err) {
+      assert.fail(err as string);
+    }
+  });
+
+  it.only("should create or replace a user", async () => {
```
qiaozha, line 175:
```suggestion
  it("should create or replace a user", async () => {
```

qiaozha:
pending 1.2.1 release

packages/typespec-ts/test/modularIntegration/scalar.spec.ts
```diff
@@ -23,7 +29,13 @@ describe("Scalar Client", () => {
...
   it("should put string value", async () => {
     try {
-      const result = await client.string.put(JSON.stringify("test"));
+      const result = await client.string.put("test", {
+        requestOptions: {
+          headers: {
+            "content-type": "text/plain"
+          }
+        }
+      });
```
qiaozha, line 38: the previous way was working because core client rest will set the requested content type as application/json [here](https://github.com/Azure/azure-sdk-for-js/pull/27343/files#diff-c7a7b43a730d4a3de597de20c6de199f8468c0d29e27ccc52b6b9a26a5afbd52L65-L72) but will pass undefined when serialize the request body [here](https://github.com/Azure/azure-sdk-for-js/pull/27343/files#diff-c7a7b43a730d4a3de597de20c6de199f8468c0d29e27ccc52b6b9a26a5afbd52L83), and we will fall into the  situation [here](https://github.com/Azure/azure-sdk-for-js/pull/27343/files#diff-c7a7b43a730d4a3de597de20c6de199f8468c0d29e27ccc52b6b9a26a5afbd52L130-L133) which will pass the '"test"' to the cadl ranch server end, 
but with the core-client-rest 1.3.0, 
1. if we keep passing '"test"' without additional content type information, we will get the into the condition [here](https://github.com/Azure/azure-sdk-for-js/pull/27343/files#diff-c7a7b43a730d4a3de597de20c6de199f8468c0d29e27ccc52b6b9a26a5afbd52R86) and get the content type "application/json; charset=UTF-8" and when we serialize the request body with the content type, we will try to JSON.stringify it again. server side will get a ""\"test\""".
2. if we pass the raw string "test" without the content type, as we have decided, we should just leave the content-type as undefined [here](https://github.com/Azure/azure-sdk-for-js/pull/27343/files#diff-c7a7b43a730d4a3de597de20c6de199f8468c0d29e27ccc52b6b9a26a5afbd52R89) we will just send out the raw string to cadl ranch, which is wrong too. 
So we have to specify the original string, and the request content type here. and it also works if we set the content type as "application/json".

@joheredi  do you think if the core logic is somewhat broken ? or the cureent logic is the right?

qiaozha (approved):

PR #2238, opened 2024-01-24
[modular] Add the optional body case in cadl-ranch

MaryGao:
[modular] Add the optional body case in cadl-ranch

### DIFF
--- a/packages/typespec-ts/test/commands/cadl-ranch-list.ts
+++ b/packages/typespec-ts/test/commands/cadl-ranch-list.ts
@@ -216,6 +216,10 @@ export const nonBrandedRlcTsps: TypeSpecRanchConfig[] = [
 
 export const modularTsps: TypeSpecRanchConfig[] = [
   {
+    outputPath: "parameters/body-optionality",
+    inputPath: "parameters/body-optionality"
+  },
+  {
     outputPath: "models/usage",
     inputPath: "type/model/usage"
   },


.vscode/launch.json
```diff
@@ -147,7 +147,7 @@
         "--timeout",
         "999999",
         "--colors",
-        "./test/unit/**/*.spec.ts"
+        "./test/modularUnit/**/*.spec.ts"
```
MaryGao, line 150:
```suggestion
        "./test/unit/**/*.spec.ts"

```

.vscode/launch.json
```diff
@@ -148,6 +148,7 @@
         "999999",
         "--colors",
         "./test/unit/**/*.spec.ts"
+
```
MaryGao, line 151:
```suggestion
```

qiaozha (approved):
LGTM, but we really don't have ut that could reflect the change?

MaryGao:
> LGTM, but we really don't have ut that could reflect the change?

yeah, I enabled the one I used for debug [here](https://github.com/Azure/autorest.typescript/pull/2238/files#diff-0ea2042ac344d1ec3a3e679731353be8a3bb8cecc443e06030280d0a457af50cR166). But skipped other nullable cases and it is not supported yet.

I want to support them once the modular refactor is done.

Issue #2243, opened 2024-01-26
Integrate with advanced cadl-ranch cases for multipart/form data cases


MaryGao:

v-jiaodi:
Integration test error:
![image](https://github.com/Azure/autorest.typescript/assets/80496810/dde661a0-6db4-4b2f-b4ac-01a682af9abe)
Server error log:
![image](https://github.com/Azure/autorest.typescript/assets/80496810/0a11aaf0-037d-4fed-b300-93eb37fa66ff)

MaryGao:
I think we can close this as when supporting MFD we could adopt the case automatically.

PR #2247, opened 2024-01-26
Add rlc coverage cases for payload multipart/form-data in cadl-ranch

v-jiaodi:
Fixes https://github.com/Azure/autorest.typescript/issues/2243


qiaozha:
close it as already integrated

PR #2260, opened 2024-01-30
Add rlc coverage cases for server versions in cadl-ranch

v-jiaodi:
Fixes https://github.com/Azure/autorest.typescript/issues/2212


qiaozha:
close it as added in #2261

PR #2261, opened 2024-01-30
Support server versions in modular and add cadl-ranch test cases

v-jiaodi:
Fixes https://github.com/Azure/autorest.typescript/issues/2212

we only generate client side api version default value and remove it from operation level when all operations have an api version parameter. 

We didn't consider a special case in this PR, where RLC layer not all operations have an api version parameter but all Modular layer operations have an api version parameter.


packages/typespec-test/test/batch_modular/generated/typespec-ts/src/BatchClient.ts
```diff
@@ -1098,31 +1234,42 @@ export class BatchClient {
...
     poolId: string,
     options: ListNodesOptions = { requestOptions: {} },
   ): PagedAsyncIterableIterator<BatchNode> {
-    return listNodes(this._client, poolId, options);
+    return listNodes(this._client, apiVersion, poolId, options);
   }
 
   /** Gets information about the specified Compute Node Extension. */
   getNodeExtension(
+    apiVersion: string,
```
MaryGao, line 1264: I notice that this is the case RLC has default value so skiped to generate query param. 
Do we need to following same rule in RLC to promot apiVersion to client level?

Another thing is I notice this parameter is not passed to RLC even though it is required one now.

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/rest/batchClient.ts
```diff
@@ -18,7 +18,7 @@ export default function createClient(
   options: ClientOptions = {},
 ): BatchContext {
   const baseUrl = options.baseUrl ?? `${endpoint}`;
-  options.apiVersion = options.apiVersion ?? "2023-05-01.17.0";
```
qiaozha, line 21: this is expected, not all operations in batch spec contains an api-version parameter. we should not set the default value for it.

packages/typespec-ts/test/modularUnit/apiOperations.spec.ts
```diff
@@ -456,4 +451,280 @@ describe("api operations in Modular", () => {
       );
     });
   });
+
+  describe.only("apiVersion in query", () => {
```
qiaozha, line 455:
```suggestion
  describe("apiVersion in query", () => {
```

packages/typespec-ts/test/modularUnit/apiOperations.spec.ts
```diff
@@ -456,4 +451,280 @@ describe("api operations in Modular", () => {
       );
     });
   });
+
+  describe("apiVersion in query", () => {
```
qiaozha, line 455: added ut here.

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -153,6 +153,7 @@ const typesMap = new Map<EmitterType, HrlcType>();
 const simpleTypesMap = new Map<string, HrlcType>();
 const endpointPathParameters: Record<string, any>[] = [];
 let apiVersionParam: Parameter | undefined = undefined;
+let hasApiVersionInClient = true;
```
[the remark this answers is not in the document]
joheredi: Can we make this global variable scoped to the functions where it is needed?

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -843,17 +847,24 @@ function emitBasicOperation(
     namespaceHierarchies.push(operationGroupName);
   }
 
+  let hasApiVersionInOperation = false;
```
joheredi, line 850: can we join this and hasApiVersionInClient in a single variable like apiVersionLocation: "client" | "operation"?

packages/typespec-ts/src/transform/transformApiVersionInfo.ts
```diff
@@ -52,16 +48,20 @@ export function transformApiVersionInfo(
...
   dpgContext: SdkContext
 ): ApiVersionInfo | undefined {
   const apiVersionTypes = new Set<string>();
   const locations = new Set<ApiVersionPosition>();
   const clientOperations = listOperationsInOperationGroup(dpgContext, client);
+  let hasApiVersionInClient = true;
+  let hasApiVersionInOperation = true;
   for (const clientOp of clientOperations) {
-    const route = ignoreDiagnostics(getHttpOperation(program, clientOp));
+    hasApiVersionInOperation = false;
```
joheredi, line 61: can we remove this and set hasApiVersionInOperation = false in line 59? Will it be the same or do you need this here?

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -383,6 +383,9 @@ function getRequestParameters(
       param.location === "query" ||
       param.location === "body"
     ) {
+      if (param.clientName === "apiVersion") {
+        param;
+      }
```
qiaozha, line 388:
```suggestion
```

qiaozha, line 388:
```suggestion
```

packages/typespec-ts/test/modularIntegration/server.spec.ts
```diff
@@ -53,3 +55,83 @@ describe("Multiple Server Path Client", () => {
...
+      const result = await client.withoutApiVersion();
+      assert.isUndefined(result);
+    } catch (err) {
+      assert.fail(err as string);
+    }
+  });
+
+  it("should work with param", async () => {
+    try {
+      const result = await client.withQueryApiVersion();
```
qiaozha, line 122: in this case, query api-version parameter has become optional in the modular layer, we can skip passing it or pass it through options `{ apiVersion: "2022-12-01-preview" }`

qiaozha (approved):

packages/typespec-ts/src/transform/transformApiVersionInfo.ts
```diff
@@ -82,9 +82,15 @@ function getOperationApiVersion(
...
-    if (apiVersionTypes.size > 1) {
+    if (apiVersionTypes.size > 1 || !hasApiVersionInClient) {
       break;
     }
+    if (params.length === 1) {
+      hasApiVersionInOperation = true;
+    }
+    if (!hasApiVersionInOperation) {
+      hasApiVersionInClient = false;
+    }
```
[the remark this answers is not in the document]
qiaozha: this is trying to determine if there's a client level api version parameter.

packages/typespec-ts/src/transform/transformApiVersionInfo.ts
qiaozha: I think we will need to refresh it before we handle each operation.

packages/typespec-ts/src/modular/buildCodeModel.ts
qiaozha: I have moved hasApiVersionInClient into emitter options as we have agreed that we skip the consideration of modular and RLC are inconsistent for now and keep hasApiVersionInOperation as the local variable.

packages/typespec-ts/src/modular/buildCodeModel.ts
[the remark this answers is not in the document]
qiaozha: replied here https://github.com/Azure/autorest.typescript/pull/2261#discussion_r1530057265

packages/typespec-ts/test/modularIntegration/server.spec.ts
MaryGao: could we have a case for that? i believe it would failed with 400 in server side?

joheredi (approved):

packages/typespec-ts/test/modularIntegration/server.spec.ts
qiaozha: I added a test case, but I don't really think that matters, we have set that value in modular layer, so it doesn't matter customer pass it or not if it's the same value.

Issue #2210, opened 2024-01-17
media type payload


qiaozha:
TypeSpec link https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/payload/media-type/main.tsp 

we need to make sure our generated code can have a way to support customize content type for both request and response

v-jiaodi:
Add modular case,  test error as follow: 

<img width="625" alt="image" src="https://github.com/Azure/autorest.typescript/assets/80496810/56d5a3fb-8606-4eb5-9015-b4d2f337d645">

qiaozha:
it's not done yet, still pending on modular layer integration.

PR #2266, opened 2024-01-31
Add modular coverage cases for payload media type in cadl-ranch

v-jiaodi:
Fixes https://github.com/Azure/autorest.typescript/issues/2210


packages/typespec-ts/package.json
```diff
@@ -65,7 +65,7 @@
     "@azure-tools/cadl-ranch-expect": "^0.13.2",
     "@azure-tools/cadl-ranch": "^0.12.4",
     "chalk": "^4.0.0",
-    "@azure-rest/core-client": "^1.3.0",
+    "@azure-rest/core-client": "^1.4.0",
```
MaryGao, line 68: do we need to bump client version in generated code?

qiaozha: @v-jiaodi can you help send a fix to update that version ?

Issue #2222, opened 2024-01-17
all kinds of scalars


qiaozha:
TypeSpec link: https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/type/scalar/main.tsp

v-jiaodi:
Add modular case error: 
<img width="628" alt="image" src="https://github.com/Azure/autorest.typescript/assets/80496810/a6a4d35a-d3cf-46ad-a3c8-5264cf6e3e4f">

PR #2274, opened 2024-02-04
Update rlc type scalar  test case

v-jiaodi:
Fixes https://github.com/Azure/autorest.typescript/issues/2222


qiaozha:
we shouldn't change this.

MaryGao (commented):
I don't think this is necessary and the failure is expected. pls close this pr.

PR #2285, opened 2024-02-06
use 0.39.0-dev.8 tcgc

iscai-msft:


qiaozha:
@iscai-msft do you still need this ? Thanks

iscai-msft:
this was to update and make sure there were no breaking changes. I can close it bc this should be taken up as part of your migration to tcgc. thanks!

PR #2258, opened 2024-01-29
Support `File` for multipart/form data in RLC

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2176 and https://github.com/Azure/autorest.typescript/issues/2196 and https://github.com/Azure/autorest.typescript/issues/2200

The pr would cover the followings:

- Multipart/form-data support
  - Translate `bytes` to `| string  | Uint8Array | ReadableStream<Uint8Array> | NodeJS.ReadableStream | File`
  - Export `createFile, createFileFromStream` and relevant options from @azure/core-rest-pipeline
  - Integrate cadl-ranch cases
- Refactor media types code
- Handle bytes according to content types 

Please note below issues are NOT covered in this pr:
- https://github.com/Azure/autorest.typescript/issues/2279
- https://github.com/Azure/autorest.typescript/issues/2278
- https://github.com/Azure/autorest.typescript/issues/2277
- https://github.com/Azure/autorest.typescript/issues/2276
- https://github.com/Azure/autorest.typescript/issues/2273

PR #2286, opened 2024-02-07
upgrade the version of typespec

kazrael2119:
update the typespec compiler version to 0.53


packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/audio/translations/index.ts
```diff
@@ -31,7 +32,7 @@ export function _createSend(
       ...operationOptionsToRequestParameters(options),
       contentType: (options.contentType as any) ?? "multipart/form-data",
       body: {
-        file: audio["file"],
+        file: uint8ArrayToString(audio["file"], "base64"),
```
qiaozha, line 35: This is because we have to remove the encode in the typespec as the tcgc linter rule. And previously we have agreed that the default encoding is base64 for bytes. As we will support multipart/form-data later, I would prefer we don't do extra effort here. @joheredi what do you think?

packages/typespec-test/test/openai_modular/generated/typespec-ts/review/openai_modular.api.md
```diff
@@ -651,7 +665,7 @@ export interface Embeddings {
 // @public
 export interface EmbeddingsOptions {
     input: string[];
-    inputType: string;
+    inputType?: string;
```
qiaozha, line 668: This is because of the typespec change.

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/rest/models.ts
```diff
@@ -6,11 +6,7 @@ export interface CreateTranscriptionRequest {
    * The audio file object (not file name) to transcribe, in one of these formats: flac, mp3, mp4,
    * mpeg, mpga, m4a, ogg, wav, or webm.
    */
-  file:
-    | string
-    | Uint8Array
-    | ReadableStream<Uint8Array>
-    | NodeJS.ReadableStream;
+  file: string;
```
lirenhe, line 9: Could you help to explain why we change this?

qiaozha: For RLC layer, we only generate 
  ```
 string
    | Uint8Array
    | ReadableStream<Uint8Array>
    | NodeJS.ReadableStream;
``` for binary bytes, if it's other encoded bytes, we will just treat it as normal string, because of the tcgc lint rule for bytes in multipart/form-data request, we have to remove the encode decorator.

qiaozha: Maybe the right solution for RLC here is to figure out a way to suppress the linter rule in TCGC, instead of changing the typespec ?

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/api/audio/translations/index.ts
joheredi: Yes! we should have customers do the encoding explicitly in RLC

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/rest/models.ts
joheredi: This will come back to be a File with Mary's PR, is that correct?

joheredi (approved):

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/rest/models.ts
qiaozha: yes, @MaryGao can you help verify if your calculation will not rely on any encode information on bytes?

lirenhe (approved):

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/rest/models.ts
MaryGao: yes, the change is back now: https://github.com/Azure/autorest.typescript/pull/2258/files#diff-a83560ba5fcefbeffbde9968ac475fc716fdadef3cf6702aa15c166960e87170

PR #2295, opened 2024-02-17
[RLC] Support Versioning

joheredi:
This Pr introduces initial support for versioning in RLC and puts it under the experimental feature flag.

For now this generates N times where N is the number of versions in the spec, each RLC is put under a folder named after the version

```
src/
   v1/
   v2/
index.ts
```

It also generates a root level index.ts to export each version as a sub module, so customers can import it as:

```ts

import { v1 } from "@azure-rest/my-lib"

const client = v1.getClient();
```

Followup PRs will include exporting a "default" version in the root, and a later PR expand this support to Modular


qiaozha:
Should we have a small review for versioning in rlc support, I remember our previous discussion about this is we should have each RLC package per each endpoint each version ?  Personally, I am in favor of this approach to support versioning in RLC. :)

joheredi:
> Should we have a small review for versioning in rlc support, I remember our previous discussion about this is we should have each RLC package per each endpoint each version ? Personally, I am in favor of this approach to support versioning in RLC. :)

At some point before taking it out of the experimental flag, we definitely need a review. I was working closely with @bterlson on this and he is on board with the overall approach. I think we can go ahead implementing it and have a GA review once the implementation is completed for both RLC and Modular

packages/autorest.typescript/package.json
```diff
@@ -75,7 +75,7 @@
     "source-map-support": "^0.5.16",
     "ts-morph": "^15.1.0",
     "@azure/core-auth": "^1.6.0",
-    "@azure-tools/rlc-common": "workspace:^0.22.0"
+    "@azure-tools/rlc-common": "workspace:^0.23.0"
```
qiaozha, line 78: We should probably upgrade the version in `packages/autorest.typescript/package.json` too. and add the change log here to keep consistent.

packages/typespec-test/test/versioning/spec/tspconfig.yaml
```diff
@@ -0,0 +1,14 @@
+emit:
+  - "@azure-tools/typespec-ts"
+options:
+  "@azure-tools/typespec-ts":
+    azureSdkForJs: false
+    isModularLibrary: false
+    experimental: true
```
qiaozha, line 7: what should we generate if we don't pass experimental and the typespec is multi api version ?

packages/typespec-test/test/versioning/generated/typespec-ts/src/index.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import * as v1 from "./v1";
+import * as v2 from "./v2";
```
qiaozha, line 5: I wonder if we have considered use subpath export to support multiapi instead of using the sub namespaces?

packages/typespec-test/test/versioning/generated/typespec-ts/src/v1/demoServiceClient.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @param endpoint - The parameter endpoint
+ * @param options - the parameter for all optional parameters
+ */
+export default function createClient(
+  endpoint: string,
+  options: ClientOptions = {},
+): DemoServiceClient {
+  const baseUrl = options.baseUrl ?? `${endpoint}`;
+
+  const userAgentInfo = `azsdk-js-versioning-rest/1.0.0-beta.1`;
```
qiaozha, line 19: I think we are missing a logic to set default version for each single api version bits ?

packages/typespec-test/test/versioning/spec/tspconfig.yaml
qiaozha: I wonder if we should generate api version as client constructor parameter in that case?

qiaozha: I think we should generate the latest version if it's not experimental and the typespec is multi api version ? I remember previously, we had an issue that if an operation is removed in the latest version, we still generate this operation https://github.com/Azure/typespec-azure/issues/129

packages/typespec-test/test/versioning/generated/typespec-ts/src/v2/clientDefinitions.ts
```diff
@@ -0,0 +1,73 @@
...
+export interface Routes {
+  /** Resource for '/\{id\}' has methods for the following verbs: get, patch, delete */
+  (path: "/{id}", id: string): Get;
+  /** Resource for '/' has methods for the following verbs: post, get */
+  (path: "/"): Create;
+  /** Resource for '/customGet' has methods for the following verbs: get */
+  (path: "/customGet"): CustomGet;
+}
+
+export type DemoServiceClient = Client & {
```
qiaozha, line 71: If we don't add suffix v2 to the client name, will it be confusing to our customer ?

qiaozha (commented):
I am thinking in the case that service team is deprecating an older api version, how do we express it in typespec ? will it be generated correctly to remove the older version code ?

qiaozha:
Also, I remember that Mike has pointed out, in the case of deprecating, service team still need to maintain the old version for about two years until it's retired,  but client side like SDKs should stop releasing the SDK to include the apis that are deprecated.  And typespec is representing both service side and client side if I understand correctly ?

packages/typespec-test/test/versioning/spec/tspconfig.yaml
joheredi: For the scope of this PR we'll keep generating what we generate today. The superset of TypeSpec, with no regard to versioning. We should revisit this as we approach the completion of this feature

packages/typespec-test/test/versioning/generated/typespec-ts/src/index.ts
joheredi: Yeah, we should definetly do that for ESM. Maybe this for CJS and sub exports for ESM. I'll handle ESM in a follow up PR

packages/typespec-test/test/versioning/generated/typespec-ts/src/v1/demoServiceClient.ts
joheredi: This is a good question, not sure if we need to as this is client version not service version though

qiaozha: do you mean typespec api version is for client side not for service side ? Also, I am not quite sure what do you mean of service version ? currently, I think there's an ongoing task for deprecating @service.version here https://github.com/microsoft/typespec/issues/2821 And our current implementation for api version has issues in both rlc and modular, I have a few fix in https://github.com/Azure/autorest.typescript/pull/2261 but would like to hold until the deprecation from typespec is finished.

joheredi: What I mean is that user-agent is used to track the client version, not the service or API version. And, the client library version will be the same within the package.

packages/autorest.typescript/package.json
joheredi: you mean bump the version of autorest.typescript?

joheredi: Pushed a change for this

packages/typespec-test/test/versioning/generated/typespec-ts/src/v2/clientDefinitions.ts
joheredi: I wouldn't like to alter the generated names. If a user just needs a single version which may be the most common scenario they won't really care to know if they are on DemoServiceClientv1 or v2 since they have already imported it.

```ts
import {v2} from "@azure-rest/foo"

const client: v2.DemoServiceClient = ...
```

joheredi:
> Also, I remember that Mike has pointed out, in the case of deprecating, service team still need to maintain the old version for about two years until it's retired, but client side like SDKs should stop releasing the SDK to include the apis that are deprecated. And typespec is representing both service side and client side if I understand correctly ?

I think at that point the service team would remove the deprecated version from the spec. That doesn't mean they stop servicing, they can still maintain the service version, just that any new code gen won't include the deprecated version(s). @bterlson to confirm.

packages/typespec-test/test/versioning/generated/typespec-ts/src/v1/demoServiceClient.ts
qiaozha: Oh, I mean we missed a line like `  options.apiVersion = options.apiVersion ?? "v1";` above the userAgentInfo default prefix logic here. I just put the comment here, sorry for the misleading info.

joheredi: Gotcha, okay yeah that makes sense now. Let me look into this

PR #2300, opened 2024-02-21
Verify MFD fix in non branding

MaryGao:
merged my change and qiaoqiao's change and re-generate non-branded test, see commit https://github.com/Azure/autorest.typescript/commit/d10ecbc75a1c2ff7529b16a3917796ca23f0ef25


packages/typespec-test/test/todo_non_branded/generated/typespec-ts/src/parameters.ts
```diff
@@ -0,0 +1,142 @@
...
+export interface TodoItemsCreateFormBodyParam {
+  body?: {
+    item: TodoItem;
+    attachments?: (
+      | TodoUrlAttachment
+      | string
+      | Uint8Array
+      | ReadableStream<Uint8Array>
+      | NodeJS.ReadableStream
+      | File
```
qiaozha, line 83: has the non-branded core already included this File type ?

qiaozha: I didn't see where it's imported.

MaryGao: File is not provided by core but by the environment.

packages/typespec-test/test/todo_non_branded/generated/typespec-ts/src/parameters.ts
```diff
@@ -0,0 +1,142 @@
...
+  TodoItemsCreateJsonBodyParam &
+  RequestParameters;
+
+export interface TodoItemsCreateFormBodyParam {
+  body?: {
+    item: TodoItem;
+    attachments?: (
+      | TodoUrlAttachment
+      | string
+      | Uint8Array
```
MaryGao, line 80: With the [fix](https://github.com/Azure/autorest.typescript/pull/2258) we could generate the union of following for MFD. And @timovv would work on the sync between branding and non-branding.

@joheredi Do you think we could merge the codegen pr?
https://github.com/Azure/autorest.typescript/pull/2300/commits/d10ecbc75a1c2ff7529b16a3917796ca23f0ef25

PR #2305, opened 2024-02-27
enable lint check in smoke test

kazrael2119:
fix https://github.com/Azure/autorest.typescript/issues/2304


MaryGao:
close as it is not ready

PR #2307, opened 2024-02-27
upgrade-emitter-version-to-0.23.0

qiaozha:


packages/autorest.typescript/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@autorest/typescript",
-  "version": "6.0.15",
+  "version": "6.0.17",
```
qiaozha, line 3: It's caused by failing to bump the version in an earlier release https://github.com/Azure/autorest.typescript/commit/71273b38e5d262719e8d807c952ae3df925f4d29

MaryGao (approved):
Approved and only small comment.

packages/autorest.typescript/package.json
MaryGao: I may not get the point, why not 6.0.16?

qiaozha: In the commit id I comment here, it added the change log but didn't bump the version in package.json.  cause them to be inconsistent, we don't realize this issue in our last release.

PR #2338, opened 2024-03-08
[RLC] Hybrid codegen

joheredi:
Scoped to TypeSpec generation. Autorest generation should remain the same


packages/rlc-common/src/metadata/buildPackageFile.ts
```diff
@@ -174,10 +181,19 @@ function restLevelPackage(model: RLCModel) {
...
+      ...(model.options?.moduleKind === "esm" && { tshy: "1.11.1" })
     }
   };
 
+  if (model.options?.moduleKind === "esm") {
+    packageInfo.scripts["build"] =
+      "npm run clean && tshy && npm run extract-api";
+  } else {
+    packageInfo.scripts["build"] =
+      "npm run clean && tsc && rollup -c 2>&1 && npm run minify && mkdirp ./review && npm run extract-api";
```
qiaozha, line 194: is it for non-branded only ? I wonder what's the difference between line 194 and line 220-221 ?

packages/rlc-common/src/metadata/buildPackageFile.ts
```diff
@@ -233,63 +252,113 @@ function restLevelPackage(model: RLCModel) {
...
+
+      if (model.options?.moduleKind === "cjs") {
+        packageInfo.scripts["integration-test:browser"] =
+          "dev-tool run test:browser";
+        packageInfo.scripts["unit-test:browser"] = "dev-tool run test:browser";
+        packageInfo.scripts["unit-test:node"] =
+          "dev-tool run test:node-ts-input -- --timeout 1200000 --exclude 'test/**/browser/*.spec.ts' 'test/**/*.spec.ts'";
+        packageInfo.scripts["integration-test:node"] =
+          "dev-tool run test:node-js-input -- --timeout 5000000 'dist-esm/test/**/*.spec.js'";
+      } else if (model.options?.moduleKind === "esm") {
```
qiaozha, line 344: I wonder if it's possible to put the package.json metadata for cjs in branded/non-branded and esm in branded/non-branded cases separately instead of using so many conditions inside?

packages/rlc-common/src/test/template.ts
```diff
@@ -168,20 +168,18 @@ export async function createRecorder(context: Context): Promise<Recorder> {
 `;
 
 export const sampleTestContent = `
-import { Recorder } from "@azure-tools/test-recorder";
-import { assert } from "chai";
-import { createRecorder } from "./utils/recordedClient{{#if isModularLibrary}}.js{{/if}}";
-import { Context } from "mocha";
+// import { createRecorder } from "./utils/recordedClient{{#if isEsm}}.js{{/if}}";
+import { assert, beforeEach, afterEach, it, describe } from "vitest";
```
qiaozha, line 172: are we going to use vitest to replace mocha and test recorder now?

packages/typespec-test/test/openai/generated/typespec-ts/.tshy/esm.json
```diff
@@ -0,0 +1,12 @@
+{
```
qiaozha, line 1: are those files under .tshy folder auto generated when you run build with tshy ? and they are not supposed to be modified ? I am just curious as I don't see any code is generating those file?

qiaozha (commented):
left some small and probably naive comments :)

packages/autorest.typescript/src/restLevelClient/transforms/transformOptions.ts
```diff
@@ -17,9 +17,9 @@ export function transformOptions(model: CodeModel): RLCOptions {
     dependencyInfo,
     productDocLink,
     azureArm,
-    flavor,
+    flavor
   } = getAutorestOptions();
-  const options: RLCOptions = {};
+  const options: RLCOptions = { moduleKind: "cjs" };
```
mpodwysocki, line 22: There should be no reason to emit CJS since we can emit for that using tshy as well as for `react-native`

packages/rlc-common/src/metadata/buildPackageFile.ts
```diff
@@ -94,23 +94,30 @@ function restLevelPackage(model: RLCModel) {
...
-    files: [
-      "dist/",
-      generateTest ? "dist-esm/src/" : "dist-esm/",
-      `types/${packageDetails.nameWithoutScope ?? packageDetails.name}.d.ts`,
-      "README.md",
-      "LICENSE",
-      "review/*"
-    ],
+    files:
+      model.options?.moduleKind === "esm"
```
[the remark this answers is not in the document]
mpodwysocki: Same, there should be no reason we would ever need to use CJS or ESM determinations if we just use tshy to generate all of the builds for us.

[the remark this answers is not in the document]
mpodwysocki: Unless this is for legacy code and we're not switching over yet

packages/rlc-common/src/metadata/buildPackageFile.ts
```diff
@@ -233,63 +252,113 @@ function restLevelPackage(model: RLCModel) {
...
+      packageInfo.scripts["build:test"] = "tsc -p . && rollup -c 2>&1";
+    } else if (model.options?.moduleKind === "esm") {
+      packageInfo.devDependencies["@vitest/browser"] = "^1.3.1";
+      packageInfo.devDependencies["@vitest/coverage-istanbul"] = "^1.3.1";
+      packageInfo.devDependencies["playwright"] = "^1.41.2";
+      packageInfo.devDependencies["vitest"] = "^1.3.1";
+    }
     packageInfo.devDependencies["c8"] = "^8.0.0";
     packageInfo.devDependencies["source-map-support"] = "^0.5.9";
     packageInfo.devDependencies["ts-node"] = "^10.0.0";
```
mpodwysocki, line 290: We should probably fix this to use `tsx` and that work versus `ts-node`

packages/rlc-common/src/metadata/buildPackageFile.ts
```diff
@@ -233,63 +252,113 @@ function restLevelPackage(model: RLCModel) {
...
+      packageInfo.scripts["test:node"] =
+        "npm run clean && npm run build:test && npm run unit-test:node";
+      packageInfo.scripts["test:browser"] =
+        "npm run clean && npm run build:test && npm run unit-test:browser";
+      packageInfo.scripts["unit-test"] =
+        "npm run unit-test:node && npm run unit-test:browser";
       packageInfo.scripts["unit-test:node"] =
-        "dev-tool run test:node-ts-input -- --timeout 1200000 --exclude 'test/**/browser/*.spec.ts' 'test/**/*.spec.ts'";
+        // eslint-disable-next-line no-useless-escape
+        `cross-env TS_NODE_COMPILER_OPTIONS="{\\\"module\\\":\\\"commonjs\\\"}" mocha -r esm --require ts-node/register --timeout 1200000 --full-trace "test/{,!(browser)/**/}*.spec.ts"`;
```
mpodwysocki, line 302: We should leverage the work with `tsx` so that we can just call

`mocha --require tsx --timeout 1200000 --full-trace "test/{,!(browser)/**/}*.spec.ts"`

packages/rlc-common/src/metadata/buildPackageFile.ts
mpodwysocki: Agreed, it's a little confusing to follow when we could have a module for generating tshy builds and one for legacy.

packages/rlc-common/src/metadata/buildTsConfig.ts
```diff
@@ -4,39 +4,65 @@
...
-  extends: "../../../tsconfig.package",
-  compilerOptions: {
-    outDir: "./dist-esm",
-    declarationDir: "./types"
-  },
-  include: ["src/**/*.ts"]
+const restLevelTsConfigInAzureSdkForJs: (
+  model: RLCModel
+) => Record<string, any> = function (model: RLCModel) {
+  if (model.options?.moduleKind === "esm") {
```
mpodwysocki, line 10: These get confusing when having early returns and I think should be separated into different files to make it easier to reason about.

packages/rlc-common/src/metadata/buildVitestConfig.ts
```diff
@@ -0,0 +1,115 @@
...
+    "define": {
+      "process.env": process.env
+    },
+    ${options.isAzureSdkForJs ? `plugins: [browserMap()],` : ""}
+    "test": {
+      "reporters": ["basic", "junit"],
+      "outputFile": {
+        "junit": "test-results.browser.xml"
+      },
+      "browser": {
```
mpodwysocki, line 43: Did you want to disable security or some other thing here across the board or just for the recorder?

packages/rlc-common/src/test/template.ts
mpodwysocki: Yes, that is the plan to remove karma, mocha and we are fixing the test recorder to use either chai or vitest.

packages/typespec-test/test/openai/generated/typespec-ts/.tshy/esm.json
mpodwysocki: `tshy` auto generates these files and are safe to check in.

packages/typespec-test/test/openai/generated/typespec-ts/package.json
```diff
@@ -103,8 +87,30 @@
...
+    "productSlugs": [
+      "azure"
+    ],
     "disableDocsMs": true,
     "apiRefLink": "https://docs.microsoft.com/javascript/api/@msinternal/openai?view=azure-node-preview"
-  }
+  },
+  "tshy": {
+    "exports": {
+      "./package.json": "./package.json",
```
mpodwysocki, line 98: Did we want modular exports here such as /api as well?

packages/typespec-test/test/openai/generated/typespec-ts/package.json
```diff
@@ -103,8 +87,30 @@
...
     "productName": "Azure OpenAI API",
-    "productSlugs": ["azure"],
+    "productSlugs": [
+      "azure"
+    ],
     "disableDocsMs": true,
     "apiRefLink": "https://docs.microsoft.com/javascript/api/@msinternal/openai?view=azure-node-preview"
-  }
+  },
+  "tshy": {
```
mpodwysocki, line 96: Is there a reason why we're not doing browser builds as part of RLC generation?

packages/rlc-common/src/metadata/buildPackageFile.ts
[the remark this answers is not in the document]
joheredi: Yeah, I don't want to change the way autorest generates RLC, at least for now. There may also be external customers interested in this, for example, someone can generate an RLC without metadata inside their project instead of making it an external package. In this scenario, if they are not using ESM they'd be forced to migrate their project which may not be feasible.

packages/typespec-test/test/openai/generated/typespec-ts/package.json
joheredi: Yeah that's the next iteration. So far I've been working on RLC migration, once I get to modular I'll have those exports

packages/rlc-common/src/metadata/buildPackageFile.ts
joheredi: Yup! I'll try to refactor this

packages/rlc-common/src/metadata/packageJson/packageCommon.ts
```diff
@@ -0,0 +1,128 @@
...
+function getCjsEntrypointInformation({
+  name,
+  nameWithoutScope,
+  moduleKind
+}: PackageCommonInfoConfig) {
+  if (moduleKind !== "cjs") {
+    return;
+  }
+
+  const types = `./types/${nameWithoutScope ?? name}.d.ts`;
```
qiaozha, line 77: is it possible that we could generateTest or generateSample in cjs ? in which case, the `${nameWithoutScope ?? name}.d.ts` is probably not in the types folder top level ? also, in the case of multi-client, it might have the subfolder?

packages/typespec-test/test/openai/generated/typespec-ts/package.json
```diff
@@ -1,7 +1,5 @@
 {
   "name": "@msinternal/openai",
-  "sdk-type": "client",
```
qiaozha, line 3: I think we probably should keep the sdk-type here? I remember this is the configuration for the pipeline ?

packages/typespec-test/test/openai/generated/typespec-ts/package.json
qiaozha: I think it's because RLC types is meaningless as they are just generated for better user experience ?

packages/rlc-common/src/metadata/packageJson/azurePackageCommon.ts
```diff
@@ -0,0 +1,167 @@
...
+  moduleKind
+}: AzurePackageInfoConfig) {
+  if (moduleKind !== "cjs") {
+    return {};
+  }
+
+  return {
+    files: [
+      "dist/",
+      withTests ? "dist-esm/src/" : "dist-esm/",
```
qiaozha, line 74: I think we should consider a case where generateSample: true and generateTest: false?

packages/rlc-common/src/metadata/buildPackageFile.ts
```diff
@@ -36,391 +63,33 @@ export function buildPackageFile(model: RLCModel) {
...
+    Boolean(model.options?.azureSdkForJs) &&
+    Boolean(model.options?.flavor === "azure")
+  );
 }
 
-function appendPathWhenLint(generateTest?: boolean) {
-  return generateTest ? "test" : "";
+function isAzureStandalonePackage(model: RLCModel): boolean {
+  return (
+    Boolean(model.options?.flavor === "azure") && !model.options?.azureSdkForJs
```
qiaozha, line 91: will we remove dev-tool when we generating azure standalone package? there's customer issue about this https://github.com/Azure/autorest.typescript/issues/2343

packages/rlc-common/src/metadata/buildPackageFile.ts
```diff
@@ -36,391 +63,33 @@ export function buildPackageFile(model: RLCModel) {
...
-  }
-  if (generateSample) {
-    path = path + ` "samples-dev/**/*.ts"`;
-  }
-  return path;
+function isAzureMonorepoPackage(model: RLCModel): boolean {
+  return (
+    Boolean(model.options?.azureSdkForJs) &&
+    Boolean(model.options?.flavor === "azure")
+  );
```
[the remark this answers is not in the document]
joheredi: The difference is on the dev depencencies. azureSdkForJs uses dev-tool and azureSdkForJs === false doesn't. In practice flavor === "azure" && azureSdkForJs  === false only would happen in our tests. It may be interesting to explore how we can remove the need to have both. Maybe taking a git submodule dependency on azure-sdk-for-js dev-tool ? Anyway, I'll explore this in a follow up effort

packages/typespec-test/test/openai/generated/typespec-ts/package.json
joheredi: This is only relevant for azureSdkForJs === true

packages/rlc-common/src/metadata/packageJson/azurePackageCommon.ts
joheredi: you mean (withTests || withSamples) ?   "dist-esm/src/" : "dist-esm/",

packages/rlc-common/src/metadata/buildPackageFile.ts
joheredi: Yes no devtool for standalone azure or flavorless

packages/rlc-common/src/metadata/packageJson/azurePackageCommon.ts
qiaozha: yeah, I found this bug in my pr https://github.com/Azure/autorest.typescript/pull/2342

packages/rlc-common/src/metadata/buildPackageFile.ts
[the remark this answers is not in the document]
qiaozha: makes sense to me, thanks for the detailed explaination :)

packages/rlc-common/src/metadata/buildPackageFile.ts
joheredi: We'll have this as a follow up effort

packages/rlc-common/src/metadata/buildVitestConfig.ts
joheredi: This is needed, karma did it for all browser tests

packages/rlc-common/src/metadata/buildTsConfig.ts
joheredi: I'll refactor this in a follow up effort, I went ahead and refactored package generation in this PR

packages/rlc-common/src/metadata/buildApiExtractorConfig.ts
```diff
@@ -5,15 +5,22 @@ import { Project } from "ts-morph";
...
-  const { packageDetails, isModularLibrary, generateTest, generateSample } =
+  const { packageDetails, isModularLibrary, generateTest } =
     model.options || {};
   const project = new Project();
+
+  let mainEntryPointFilePath = "./dist/esm/index.d.ts";
+
+  if (model.options?.moduleKind === "cjs") {
+    mainEntryPointFilePath = `./types${
+      generateTest || isModularLibrary ? "/src" : ""
```
qiaozha, line 16: we don't need to consider the generateSample here?

packages/rlc-common/src/metadata/buildApiExtractorConfig.ts
```diff
@@ -5,15 +5,22 @@ import { Project } from "ts-morph";
 import { RLCModel } from "../interfaces.js";
 
 export function buildApiExtractorConfig(model: RLCModel) {
-  const { packageDetails, isModularLibrary, generateTest, generateSample } =
+  const { packageDetails, isModularLibrary, generateTest } =
     model.options || {};
   const project = new Project();
+
+  let mainEntryPointFilePath = "./dist/esm/index.d.ts";
```
qiaozha, line 12: I am trying to understand if it's esm, we don't really need to compile the samples and tests? so does the cjs compile result?

qiaozha (approved):
LGTM, just some minor comments

packages/typespec-test/test/anomalyDetector/generated/typespec-ts/package.json
```diff
@@ -1,109 +1,111 @@
...
+    }
   },
-  "//sampleConfiguration": {
-    "productName": "Anomaly Detector",
-    "productSlugs": ["azure"],
-    "disableDocsMs": true,
-    "apiRefLink": "https://docs.microsoft.com/javascript/api/@msinternal/ai-anomaly-detector?view=azure-node-preview"
-  }
+  "main": "./dist/commonjs/index.js",
+  "types": "./dist/commonjs/index.d.ts"
```
MaryGao, line 110: Does that mean we will build types in both cjs and esm? Both of them are the extract same definitions?

packages/typespec-test/test/authoring/generated/typespec-ts/package.json
```diff
@@ -1,111 +1,113 @@
...
+        "types": "./dist/browser/index.d.ts",
+        "default": "./dist/browser/index.js"
+      },
+      "react-native": {
+        "types": "./dist/react-native/index.d.ts",
+        "default": "./dist/react-native/index.js"
+      },
+      "import": {
+        "types": "./dist/esm/index.d.ts",
+        "default": "./dist/esm/index.js"
```
MaryGao, line 103: Could you help me understand the dist structure here? Will it be like:

```
dist/
├─ esm/
│  ├─ test/
│  ├─ samples-dev/
│  ├─ src?/
│  ├─ index.d.ts/ (whoes index file? src/sample/test?)
│  ├─ index.js/ 
.....
```

packages/rlc-common/src/metadata/buildTsConfig.ts
```diff
@@ -4,39 +4,65 @@
...
+      compilerOptions: {
+        module: "NodeNext",
+        moduleResolution: "NodeNext",
+        rootDir: "."
+      },
+      include: [
+        "./src/**/*.ts",
+        "./src/**/*.mts",
+        "./src/**/*.cts",
+        "test/**/*.ts"
```
MaryGao, line 22: the test folder may not be always there? e.g if customers disable the test gen or in non-branded we disable test gen by default?

packages/rlc-common/src/metadata/buildVitestConfig.ts
```diff
@@ -0,0 +1,118 @@
...
+        "provider": "playwright"
+      },
+      "fakeTimers": {
+        "toFake": ["setTimeout", "Date"]
+      },
+      "watch": false,
+      "include": ${
+        options.isAzureSdkForJs
+          ? `["dist-test/browser/**/*.spec.js"]`
+          : `["test/**/*.spec.ts"]`
```
MaryGao, line 59: i may not get the point that we need to distugsh with `isAzureSdkForJs` option with diff path?

packages/typespec-test/test/batch_modular/generated/typespec-ts/test/public/sampleTest.spec.ts
```diff
@@ -1,20 +1,18 @@
...
+import { assert, beforeEach, afterEach, it, describe } from "vitest";
 
 describe("My test", () => {
-  let recorder: Recorder;
+  // let recorder: Recorder;
 
-  beforeEach(async function (this: Context) {
-    recorder = await createRecorder(this);
+  beforeEach(async function () {
+    // recorder = await createRecorder(this);
```
MaryGao, line 11: Does that mean the recorder is not working yet? if yes can we release this change?

packages/autorest.typescript/src/restLevelClient/transforms/transformOptions.ts
MaryGao, line 22: I'd like to clarify some details, there are two main changes
- changed to esm 
- upgrade our build/test dependencies and steps

Are the two changes coupled with each other? Will the HLC need to adopt this? I mean if HLC won't can old code run correctly without changes?

packages/typespec-test/test/anomalyDetector/generated/typespec-ts/package.json
```diff
@@ -1,109 +1,111 @@
...
+  "exports": {
+    "./package.json": "./package.json",
+    ".": {
+      "browser": {
+        "types": "./dist/browser/index.d.ts",
+        "default": "./dist/browser/index.js"
+      },
+      "react-native": {
+        "types": "./dist/react-native/index.d.ts",
+        "default": "./dist/react-native/index.js"
```
MaryGao, line 97: interesting, should the cjs code be the same as browser one after building?

another detail to confirm is that the replacement logic will be still working, right?

https://github.com/Azure/autorest.typescript/blob/main/packages/autorest.typescript/test/smoke/generated/anomaly-detector-rest/package.json#L103

PR #2342, opened 2024-03-09
fix issues in todo: anonymous array model import and sample mock value for credential

### LINKED ISSUE #2326: fail to import array value type if it's inside a anonymous body type

### LINKED ISSUE #2330: mock value for key credential is missing in sample generation

qiaozha:
fixes #2326 
fixes #2330


packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -998,8 +998,14 @@ function getSchemaForArrayModel(
       if (schema.items.typeName) {
         if (schema.items.type === "dictionary") {
           schema.typeName = `${schema.items.typeName}[]`;
+          if (usage && usage.includes(SchemaContext.Output)) {
+            schema.outputTypeName = `(${schema.items.outputTypeName})[]`;
```
MaryGao, line 1002: nit: if this is model, `Array<Model>` would be more proper than `Model[]`.

MaryGao (approved):
LGTM and just small comments.

packages/typespec-ts/src/utils/modelUtils.ts
qiaozha: if we need to change from Model[] to Array<Model>, we need to change both the input type name and output type name, we should have an independent pr to fix it. I prefer to keep this pr simple.

packages/rlc-common/src/transformSampleGroups.ts
```diff
@@ -257,6 +272,11 @@ function convertClientLevelParameters(
         "DefaultAzureCredential",
         importedDict
       );
+    } else {
+      clientParams.push({
+        name: "credential",
+        assignment: `const credential = {getToken: () => Promise.resolve({ token: "{Your token}", expiresOnTimestamp: 0 })};`
```
qiaozha, line 278: @joheredi I wonder if you have any concerns for this part for generating the mock value for token credential as we don't have `@azure/identity` in non-branded scope.

packages/typespec-ts/src/utils/modelUtils.ts
MaryGao: sure

qiaozha:
I will merge it for now, @joheredi let me know if you have any concerns.

Issue #1925, opened 2023-07-13
Add support for @usage &  @access


MaryGao:

qiaozha:
we will have a feature flag generateAllModels for generating all orphan models.  instead of supporting @access and @usage directly in modular.

Issue #2214, opened 2024-01-17
special words in model properties, models, operations, parameters,


qiaozha:
TypeSpec link: https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/special-words/main.tsp 

Should consideration this issue https://github.com/Azure/autorest.typescript/issues/2084

kazrael2119:
![image](https://github.com/Azure/autorest.typescript/assets/98569699/1cca870a-c51e-4264-bd70-75fe6cc94334)
can't generate code for special-words

qiaozha:
Design Discussion Notes:
1. parameter
    ```ts
    declare function op(await: string): void;
    ```
    we will rename await => awaitParam

2. operation
      ```
      declare function delete(param: string): void;
      ```
      we will rename await => $await, but keep the original name whereever possible.
      
      ```ts
      // for operation options
      DeleteOptions
      FooDeleteOptions
      
      // for classcial client layer
      client.await() {
        $await();
      }
      ```


3. property // new in Modular
      ```ts
      interface Some {
        constructor: string;
      }
      type SomeConstructor = Some & Record<string, unknown>
      function foo(some: SomeConstructor) {
        return some.constructor;
      }
      foo({ constructor: "hello" })
      
      ```
      For normal property proposal: constructor => constructorProperty
      for parameter being used in query parameter: we use as any
      ```ts
      {
         constructor: constructorParam as any
      }
      ```

// newable reserved words
```ts
declare interface Boolean { };
const bool = new Boolean();
```

4. interface or class
     we will rename Boolean => BooleanModel

5. operation groups
     we will rename Boolean => BooleanOperations 


6. operation options

      // content-safety and ClientOptions reserved 
      ```ts
      model BarOptions {
      }
      op bar(@body options: BarOptions);
      ```
      we will rename our options as requestOptions: BarOptions => BarRequestOptions 
      ```ts
      declare interface BarOptions { 
          prop: string;
       };
       declare interface BarRequestOptions { 
          optional1: string;
       };
      declare function bar(options: BarOptions, requestOptions?: BarRequestOptions): void; 
      ```

qiaozha:
For point 6, since I found we have been using OptionalParams in HLC https://github.com/Azure/autorest.typescript/issues/2352 and it can resolved the Options conflicts issue, I just changed the Options to OptionalParams to avoid breaking and extra effort of resolve conflicts.

qiaozha:
[...]

PR #2356, opened 2024-03-14
Support usage and access

qiaozha:
fixes https://github.com/Azure/autorest.typescript/issues/1925


packages/typespec-test/test/widget_dpg/spec/client.tsp
```diff
@@ -0,0 +1,8 @@
+import "@azure-tools/typespec-client-generator-core";
+import "./main.tsp";
+
+using Azure.ClientGenerator.Core;
+
+// Some models from routes with suppressed visibility are still desired for custom public surface.
+@@access(WidgetService.NonReferencedModel, Access.public);
+@@usage(WidgetService.NonReferencedModel, Usage.input | Usage.output);
```
qiaozha, line 8: We still need this kind of configuration as tcgc getAllModels will not return the orphan models if we don't have this. I think it's fine as they will be required to other languages as well ?

packages/typespec-ts/src/lib.ts
```diff
@@ -11,6 +11,7 @@ import { Options } from "prettier";
 
 export interface EmitterOptions extends RLCOptions {
   branded?: boolean;
+  generateOrphanModels?: boolean;
```
qiaozha, line 14: Feel like it's better to put it as an emitter layer option instead of a rlc option.

packages/rlc-common/src/helpers/nameUtils.ts
```diff
@@ -62,7 +62,7 @@ export const ReservedModelNames: ReservedName[] = [
   { name: "package", reservedFor: [NameType.Parameter] },
   { name: "private", reservedFor: [NameType.Parameter] },
   { name: "protected", reservedFor: [NameType.Parameter] },
-  { name: "public", reservedFor: [NameType.Parameter] },
+  { name: "public", reservedFor: [NameType.Parameter, NameType.Operation] },
```
qiaozha, line 65: we will revisit all kinds of renaming when we handle https://github.com/Azure/autorest.typescript/issues/2214

packages/typespec-test/test/openai_modular/spec/tspconfig.yaml
```diff
@@ -5,6 +5,7 @@ options:
     azureSdkForJs: false
     isModularLibrary: true
     hierarchyClient: false
+    generateOrphanModels: true
```
lirenhe, line 8: Is there any documentation we need to update for this new setting?

lirenhe (approved):

joheredi (approved):

MaryGao (approved):
LGTM

packages/typespec-test/test/openai_modular/spec/tspconfig.yaml
qiaozha: added here https://github.com/Azure/autorest.typescript/pull/2356/files#diff-0dc6d1b55dc9a9970d589d1beb757622461950f62cd6107358e29fc205f04f8dR148

PR #2369, opened 2024-03-25
Add coverage cases for Azure Core Scalar in cadl-ranch

v-jiaodi:
Added List:
1. Azure core scalar: modular and rlc.

### DIFF
--- a/packages/typespec-ts/test/integration/azurecorescalar.spec.ts
+++ b/packages/typespec-ts/test/integration/azurecorescalar.spec.ts
@@ -7,10 +7,7 @@ describe("Azure Core Saclar Rest Client", () => {
 
   beforeEach(() => {
     client = AzureCoreScalarClientFactory({
-      allowInsecureConnection: true,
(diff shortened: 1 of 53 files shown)


packages/typespec-ts/test/integration/azurecorescalar.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+      assert.fail(err as string);
+    }
+  });
+
+  it("should put an Azure Location value", async () => {
+    try {
+      const result = await client.path("/azure/core/scalar/azureLocation").put({
+        body: "eastus",
+        headers: {
+          "content-type": "text/plain"
```
qiaozha, line 31: I think this is expected, as cadl-ranch is expecting string here https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/azure/core/scalar/mockapi.ts#L15

qiaozha (approved):

PR #2387, opened 2024-03-27
Support special words

### LINKED ISSUE #2084: use $ as the operation rename for reserved word.
If we have an operation name delete, currently, we will rename it as deleteOperation because of the reserved word. 
But we should rename it as $delete instead. see comment https://github.com/Azure/autorest.typescript/pull/2080#discussion_r1369876176

### LINKED ISSUE #2214: special words in model properties, models, operations, parameters,

### LINKED ISSUE #2352: Operation option interface is changed from DataProductsGetOptionalParams to DataProductsGetOptions

qiaozha:
fix #2214
fix #2352
fix #2084


packages/typespec-ts/test/modularIntegration/generated/encode/bytes/src/classic/header/index.ts
```diff
@@ -9,37 +9,32 @@ import {
   headerBase64urlArray,
 } from "../../api/header/index.js";
 import {
-  HeaderDefaultOptions,
-  HeaderBase64Options,
-  HeaderBase64urlOptions,
-  HeaderBase64urlArrayOptions,
+  DefaultOptions,
```
qiaozha, line 12: need to figure out why this changed.

packages/typespec-ts/test/modularIntegration/generated/scalar/src/ScalarClient.ts
```diff
@@ -3,13 +3,13 @@
 
 import { Pipeline } from "@azure/core-rest-pipeline";
 import {
-  getStringOperations,
-  StringOperations,
-} from "./classic/string/index.js";
+  getStringModelOperations,
```
qiaozha, line 6: why is this change here ?

packages/typespec-ts/test/modularIntegration/generated/server/path/multiple/src/MultipleClient.ts
```diff
@@ -21,8 +21,8 @@ export class MultipleClient {
   /** The pipeline used by this client to make requests */
   public readonly pipeline: Pipeline;
 
-  constructor(endpoint: string, options: MultipleClientOptions = {}) {
-    this._client = createMultiple(endpoint, options);
+  constructor(endpointParam: string, options: MultipleClientOptions = {}) {
```
qiaozha, line 24: feel like this change shouldn't be here.

packages/typespec-test/test/batch_modular/generated/typespec-ts/review/batch.api.md
```diff
@@ -14,7 +14,7 @@ export type AccessScope = string;
 
 // @public
 export interface AccountListSupportedImagesResult {
-    "odata.nextLink"?: string;
+    odataNextLink?: string;
```
qiaozha, line 17: we should not normalize the name here.

MaryGao (approved):
LGTM

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -1160,6 +1154,9 @@ export function serializeRequestValue(
         return `${clientValue} as any`;
       }
     default:
+      if (clientValue === "constructorParam") {
```
MaryGao, line 1157: is this the only case? do we need to do this for other xxxParam?

qiaozha: as offline discussed, this is the only case for now.

PR #2421, opened 2024-04-06
Emit test files for early feedback

dgetu:


packages/typespec-test/test/batch_modular/generated/typespec-ts/src/util/serializeUtil.ts
```diff
@@ -0,0 +1,3372 @@
...
+  BatchErrorOutput as RestBatchError,
+} from "../rest/index.js";
+
+export function serializeNodeRemoveOptions(
+  o: NodeRemoveOptions,
+): RestNodeRemoveOptions {
+  return {
+    nodeDeallocationOption: o["nodeDeallocationOption"],
+    resizeTimeout:
+      o["resizeTimeout"] === undefined ? o["resizeTimeout"] : FIXME,
```
qiaozha, line 324: I am curious where is the FIXME come from ? and why?

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/util/serializeUtil.ts
```diff
@@ -0,0 +1,3372 @@
...
+    nodeList: o["nodeList"],
+  };
+}
+
+export function serializeBatchPoolReplaceOptions(
+  o: BatchPoolReplaceOptions,
+): RestBatchPoolReplaceOptions {
+  return {
+    targetNodeCommunicationMode: o["targetNodeCommunicationMode"],
+    metadata: o["metadata"].map((e) => MISSING_SERIALIZER(e)),
```
qiaozha, line 334: Is this because we haven't started to handle array element type yet? Also, when we are handling the serialization inside one model, we could assume other model's serialization models has been build somewhere else, so what we need here is just that serializer's name. for example, if metadata's element type is called MetadataElement, we can just generate 
```suggestion
    metadata: o["metadata"].map((e) => serializeMetadataElement(e)),
```

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/util/serializeUtil.ts
```diff
@@ -0,0 +1,3372 @@
...
+  };
+}
+
+export function serializeBatchPoolResizeOptions(
+  o: BatchPoolResizeOptions,
+): RestBatchPoolResizeOptions {
+  return {
+    nodeDeallocationOption: o["nodeDeallocationOption"],
+    resizeTimeout:
+      o["resizeTimeout"] === undefined ? o["resizeTimeout"] : FIXME,
```
qiaozha, line 354: I rmemeber there's some improvement  about how to handle undefined and null more properly, so that we don't assign extra undefineds if that property doesn't exist.

packages/typespec-test/test/batch_modular/generated/typespec-ts/src/util/serializeUtil.ts
```diff
@@ -0,0 +1,3372 @@
...
+    vmSize: o["vmSize"],
+    displayName: o["displayName"],
+    id: o["id"],
+  };
+}
+
+export function serializeBatchJobCreateOptions(
+  o: BatchJobCreateOptions,
+): RestBatchJobCreateOptions {
+  return {
```
qiaozha, line 471: I don't know if you remember when we try to support dictionary type, we found that this kind of inline assignment way of serialization and deserialization, wouldn't work if the dictionary value type is complex data type? I can't find the link now.

packages/typespec-test/test/openai_non_branded/generated/typespec-ts/src/classic/fineTuning/index.ts
```diff
@@ -1,19 +0,0 @@
...
-import {
-  FineTuningJobsOperations,
-  getFineTuningJobsOperations,
-} from "./jobs/index.js";
-
-export interface FineTuningOperations {
-  jobs: FineTuningJobsOperations;
-}
-
-export function getFineTuningOperations(
```
qiaozha, line 13: why is this getting deleted ?

packages/typespec-ts/test/modularIntegration/generated/authentication/api-key/src/index.ts
```diff
@@ -2,4 +2,4 @@
 // Licensed under the MIT license.
 
 export { ApiKeyClient, ApiKeyClientOptions } from "./ApiKeyClient.js";
-export { ValidOptions, InvalidOptions } from "./models/index.js";
+export { InvalidAuth, ValidOptions, InvalidOptions } from "./models/index.js";
```
qiaozha, line 5: Is it a bug fix ?

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/.eslintrc.json
```diff
@@ -1,11 +0,0 @@
-{
```
qiaozha, line 1: this is a case for multi clients, we need to generate the serialize/deserialize utils for each client separately.

qiaozha (commented):
I can't find the source code change for this, is it because you forget to check in ?

dgetu:
> I can't find the source code change for this, is it because you forget to check in ?

I'm hoping to validate that the design I have can be integrated with the current `send/deserialize` functions so I don't have folks review something that is going to drastically change. That should be ready tomorrow.

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/.eslintrc.json
qiaozha: I would say, the client is divided conceptually, normally it's because architects think some of the operations should be grouped into a relatively independent client. it's very likely we will have models being used in more than one client. Since we are generating them in each separate subclient folder and we will not export the serialize util,  I am not too worried about the serializer implementation, just a small reminder of the file path will be different.

PR #2429, opened 2024-04-09
Update specialWords.spec.ts

qiaozha:


qiaozha:
close it in favor of #2431

PR #2430, opened 2024-04-10
Emit serialize utils

dgetu:


packages/typespec-ts/src/modular/serialization/serializers.ts
```diff
@@ -0,0 +1,245 @@
...
+  };
+}
+
+export function serializeProperty(
+  functionType: UsageFlags,
+  serializerMap: SerializerMap,
+  p: Property,
+  valueExpr: string
+): SerializerOutput {
+  const child = serializeType(functionType, serializerMap, p.type, valueExpr);
```
qiaozha, line 228: I think we need to consider recursive models so that we don't get stack overflow issues, but I found we have only leveraged special union serialize function ? another point is for the performance consideration?

packages/typespec-ts/src/modular/serialization/serializers.ts
```diff
@@ -0,0 +1,245 @@
...
+      }
+      return {
+        expr: isNormalUnion(type) ? valueExpr : `(${valueExpr}) as any`
+      };
+    case "enum":
+      // TODO: encodedName
+      return { expr: valueExpr };
+  }
+  return {
+    expr: "FIXME"
```
qiaozha, line 82: I am thinking if the fall back logic should be `valueExpr` ? Another point is for query/header/path parameters serialization, should we leverage the serializeType function helper too? I tend to.

packages/typespec-ts/src/index.ts
```diff
@@ -195,6 +196,7 @@ export async function $onEmit(context: EmitContext) {
         buildModelsOptions(modularCodeModel, subClient);
         const hasClientUnexpectedHelper =
           needUnexpectedHelper.get(subClient.rlcClientName) ?? false;
+        buildSerializers(modularCodeModel);
```
qiaozha, line 199: I remember in `serializeUtil.ts` we also generate serialize function for each union variant, just curious if it's possible to leverage the `buildSerializers` logic instead of generating it there?

packages/typespec-ts/src/modular/serialization/serializers.ts
```diff
@@ -0,0 +1,245 @@
...
+      case "headerDefault":
+        return `(${valueExpr}).toUTCString()`;
+      case "unixTimestamp":
+        return `(${valueExpr}).getTime()`;
+      case "rfc3339":
+      default:
+        return `(${valueExpr}).toISOString()`;
+    }
+  } else {
+    return `new Date(${valueExpr})`;
```
qiaozha, line 132: I am not sure why we need create a Date for Datetime serialization ? 
```suggestion
    return `${valueExpr}`;
```

packages/typespec-ts/src/modular/serialization/util.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (type.name && !map.has(type.name)) {
+      map.set(type.name, type);
+      stack.push(...(t.parents ?? []));
+    }
+  }
+  return Array.from(map.values());
+}
+
+export function getEncodingFormat(type: { format?: string }) {
+  const supportedFormats = ["base64url", "base64", "byte"];
```
qiaozha, line 39: https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-util/src/bytesEncoding.ts#L5 I think that's core supported format ?

packages/typespec-ts/src/modular/serialization/serializers.ts
```diff
@@ -0,0 +1,245 @@
...
+  return {
+    expr: "FIXME"
+  };
+}
+
+export function serializeByteArray(
+  functionType: UsageFlags,
+  type: Type,
+  valueExpr: string
+): SerializerOutput {
```
qiaozha, line 90: I feel like this part might be problematic for streaming type, but since we haven't discussed the support of mfd in modular layer yet, I am not sure what should we do here.

packages/typespec-ts/src/modular/serialization/util.ts
dgetu: This function has already been checked in elsewhere in the codebase as-is. I think UTF-8 is missing because tsp doesn't support encoding string data as base64. I don't find the docs to be clear about this though.

qiaozha: tsp `@encode` decorators accepts open string, so if we have something like 
```typespec
model Widget {
  @key id: string;
  weight: int32;
  color: "red" | "blue";
  @encode("utf-8")
  prop: string;
}
```
it won't report any errors. Though I do know typespec has a BytesKnownEncoding only specifies base64 and base64url https://github.com/microsoft/typespec/blob/6f68753b191fd5171eb5153fb7b8a547b9942b7f/packages/compiler/src/lib/decorators.ts#L693 

And I did ask them to add some common used type into the BytesKnownEncoding, but they refused to https://github.com/microsoft/typespec/issues/2204  😢

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -391,6 +392,19 @@ function getType(
     }
   }
 
+  if (type.kind === "Model") {
```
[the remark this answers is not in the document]
qiaozha: I mean body parameter's type could be ModelProperty type.  And it seems a fairly common usage in tsp.

packages/typespec-ts/src/modular/serialization/serializers.ts
qiaozha: I need to take some time to look into this, but maybe you can submit your local change.

joheredi:
I believe we can close this PR in favor of #2613

PR #2432, opened 2024-04-10
try to add usage and access to generate networkanalytics

kazrael2119:
use command:
npx tsp compile client.tsp


qiaozha:
pending on unused type alias is not getting removed fix.

qiaozha:
I think we can close this now, the orphan models is generating without additional configs now.

Issue #2254, opened 2024-01-29
use parellel mode to run the codegen test


qiaozha:
the current test is taking too long to finish, we should figure out a way to run the test parellelly.

qiaozha:
close it as supported.

PR #2439, opened 2024-04-11
Parallelize test

joheredi:
Parallelizing smoke-test and cadl-ranch test generation to speed up dev-workflow and pipeline runtime.

In local benchmark these are the results:

Smoke Test Baseline: 10 minutes
Smoke Test New: 1:57 minutes

RLC CadlRanch Baseline: 1:56 min
RLC CadlRanch New: 10sec

Will update with pipeline run benchmark once they are ready


packages/typespec-test/eng/smoke-test.js
```diff
@@ -1,129 +1,90 @@
...
-    const contents = await readdir(path);
-    console.log(`================Start ${folder}===============`);
-    if (contents.includes("skip")) {
-      console.log(`          ##### Skipped #####          `);
-    } else {
-      generate(path);
-      copyFile(path);
-      build(path);
+    // Wait for an available worker
+    if (activeWorkers >= maxConcurrentWorkers) {
```
dgetu, line 29:
```suggestion
    while (activeWorkers >= maxConcurrentWorkers) {
```

packages/typespec-ts/test/commands/gen-cadl-ranch.js
```diff
@@ -0,0 +1,102 @@
...
+
+  console.log(`Max concurrent workers: ${maxConcurrentWorkers}`);
+
+  for (const tsp of list) {
+    if (isDebugging === true && tsp.debug !== true) {
+      continue;
+    }
+
+    // Wait for an available worker
+    if (activeWorkers >= maxConcurrentWorkers) {
```
dgetu, line 46:
```suggestion
    while (activeWorkers >= maxConcurrentWorkers) {
```

dgetu (commented):
That's a huge improvement! Thanks for looking into this.

packages/typespec-ts/test/commands/gen-cadl-ranch.js
joheredi: Good catch!

packages/typespec-test/eng/smoke-test-worker.js
```diff
@@ -0,0 +1,108 @@
...
+}
+
+async function createSmokeTest(folder) {
+  const __filename = fileURLToPath(import.meta.url);
+  const __dirname = dirname(__filename);
+  const root = join(__dirname, "..");
+
+  const path = join(root, "test", folder);
+  const contents = await readdir(path);
+  console.log(`================Start ${folder}===============`);
```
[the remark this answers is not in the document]
joheredi: We were not, commited a change to synchronize them so logs are output as a unit

joheredi:
Closes #2254

timovv (commented):
Random thought -- not sure we need workers for this? From what I can tell, each worker is just waiting asynchronously for the real work to be done in a separate process, so they won't be actually doing anything blocking.

If that's true (and I might be missing something!) I think we'd be able to get away with just doing `Promise.all` or similar. If we're concerned about spawning everything up at the same time (a valid concern) then we can probably write something to ensure only a certain number of "worker promises" are actually being run at a time.

packages/typespec-test/eng/runCommand.js
```diff
@@ -0,0 +1,21 @@
...
+    const process = spawn(command, args, { cwd, shell: true });
+
+    let stdout = "";
+    let stderr = "";
+
+    process.stdout.on("data", (data) => (stdout += data.toString()));
+    process.stderr.on("data", (data) => (stderr += data.toString()));
+
+    process.on("close", () => {
+      resolve(stdout ?? stderr);
```
dgetu, line 14:
```suggestion
      resolve(stdout || stderr);
```

qiaozha (approved):
thanks for addressing it, feel free to merge if all the comments are resolved.

jeremymeng:
I'd love to have this!  Related: have we looked into running other tests in parallel too? Assuming most integration tests are independent, that would save a lot of inner-loop dev time.

joheredi:
> Random thought -- not sure we need workers for this? From what I can tell, each worker is just waiting asynchronously for the real work to be done in a separate process, so they won't be actually doing anything blocking.
> 
> If that's true (and I might be missing something!) I think we'd be able to get away with just doing `Promise.all` or similar. If we're concerned about spawning everything up at the same time (a valid concern) then we can probably write something to ensure only a certain number of "worker promises" are actually being run at a time.

Good point, I think we might be able to get away without workers. Since now we'll be using spawn instead of execSync I think that would hand it off to the OS to decide if any of those should be run in separate cores, overlapping a bit with what workers do.

I ran 5 times using workers and 5 times with `promise.allSettled`, it seems like workers are still faster, but may not be faster enough to justify the added complexity of workers

Workers: 1:43, 1:56, 2:11, 1:47, 1:42
promise.allSettled: 2:12, 2:13, 2:04, 2:18

Will run the promise.allSettled version in CI to see how it looks there vs workers

packages/typespec-ts/package.json
```diff
@@ -28,11 +28,11 @@
     "generate-and-run:rlc": "npm run generate-tsp-only:rlc && npm run integration-test:alone:rlc && npm run stop-test-server",
     "generate-and-run:modular": "npm run generate-tsp-only:modular && npm run integration-test:alone:modular && npm run stop-test-server",
     "generate-tsp-only": "npm run generate-tsp-only:rlc && npm run generate-tsp-only:modular && npm run generate-tsp-only:non-branded-rlc && npm run generate-tsp-only:non-branded-modular",
-    "generate-tsp-only:rlc": "node --loader ts-node/esm ./test/commands/gen-cadl-ranch.ts --tag=rlc",
-    "generate-tsp-only:modular": "node --loader ts-node/esm ./test/commands/gen-cadl-ranch.ts --tag=modular",
+    "generate-tsp-only:rlc": "node ./test/commands/gen-cadl-ranch.js --tag=rlc",
```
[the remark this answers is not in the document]
joheredi: Yes, I prefer we use node and js instead of ts-node and ts as it will run faster b/c ts-node needs to compile on the fly.

packages/typespec-test/eng/smoke-test.js
```diff
@@ -1,95 +1,136 @@
...
+  const contents = await readdir(path);
+  logger.log(`================Start ${path}===============`);
+  if (contents.includes("skip")) {
+    logger.log(`          ##### Skipped #####          `);
+  } else {
+    await generate(path, logger);
+    await build(path, logger);
+  }
+  logger.log(`================End ${path}===============`);
+  logger.flush();
```
MaryGao, line 133: Will the all logging mix together with disordered messages?

joheredi: no they are now synchronized so all logs from the same task are going to together

packages/typespec-test/test/openai_generic/tspconfig.yaml
```diff
@@ -4,7 +4,7 @@ options:
   "@azure-tools/typespec-ts":
     azureSdkForJs: false
     isModularLibrary: true
-    "emitter-output-dir": "{project-root}/../generated/typespec-ts"
+    "emitter-output-dir": "{project-root}/generated/typespec-ts"
```
qiaozha, line 7: I vaguely remember this was trying to test against some output folder issue, thought I am not sure if it's still value to keep it.

packages/typespec-test/test/widget_dpg/generated/typespec-ts/sources/generated/src/api/index.ts
```diff
@@ -1,8 +0,0 @@
-// Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
joheredi: The customization process is changing and this will no longer apply

PR #2440, opened 2024-04-11
Upgrade generated packages to have `typescript@~5.4.5`

jeremymeng:
related commit in the JS repo: https://github.com/Azure/azure-sdk-for-js/commit/398494acd8275f9d8d80c89663a5e72218180a0b


qiaozha (approved):

qiaozha:
I am curious if we should also update the typescript dependency version for codegen itself in each `packages/*/package.json` file?

jeremymeng:
> I am curious if we should also update the typescript dependency version for codegen itself in each `packages/*/package.json` file?

good point. yes we should stay on latest as much as possible.  I can open another PR.

Issue #2230, opened 2024-01-22
Implement RLC for new design


MaryGao:

MaryGao:
Given [data link](https://microsoft-my.sharepoint.com/:x:/p/v-ziweichen/Ed38NkSV-xNFmPCtqps_cNsB6iHSS-K_5DYueqkPLBCKhA?e=xmixZ9) we have 3 GAed LRO in RLCs - ![image](https://github.com/Azure/autorest.typescript/assets/9943211/b6225bb5-c56e-4f4f-8b04-548a56e77437)

qiaozha:
close it as completed.

Issue #2392, opened 2024-02-28
[modular] should throw exception if the initial response failed in poller helper


MaryGao:
In V3 LRO `submitted` should throw exception if poller's initialization failed

qiaozha:
close it as pr merged.

Issue #2316, opened 2024-02-28
Support `abortSignal` for LRO both for modular and RLC


MaryGao:
Support abortSignal for LRO both for modular and RLC

PR #2443, opened 2024-04-15
Support V3 SimplePollerLike in RLC LRO

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2230 https://github.com/Azure/autorest.typescript/issues/2392 https://github.com/Azure/autorest.typescript/issues/2316

- Implement the RLC interface `SimplePollerLike` with V3 lro
   - Generate V3 LRO both TypeSpec and Swagger RLC
   - Enable integration/browser testing to both work for V2 and V3 LRO in swagger
   - No breakings migrated to V3 but with `stopPolling` and `toString` are marked as deprecated
    
- Fix small issue - throw exceptions in modular poller helper
- Add LRO RPC test case for RLC
- Support abortSignal for LRO both for modular and RLC


packages/autorest.typescript/test/rlcIntegration/generated/pagingRest/package.json
```diff
@@ -27,8 +27,8 @@
     "@azure/core-rest-pipeline": "^1.5.0",
     "@azure/logger": "^1.0.0",
     "tslib": "^2.6.2",
-    "@azure/core-lro": "^2.5.4",
-    "@azure/abort-controller": "^1.0.0",
+    "@azure/core-lro": "3.0.0-beta.1",
```
qiaozha, line 30: is this a case for both paging and lro ? I thought we haven't supported it yet ?

packages/autorest.typescript/test/rlcIntegration/lroRest.spec.ts
```diff
@@ -93,7 +93,7 @@ describe("LRO Rest Client", () => {
       if (isUnexpected(response)) {
         assert.fail(`Unexpected status code ${response.status}`);
       }
-      assert.equal(poller.getOperationState().status, "canceled");
+      assert.equal(poller?.operationState?.status, "canceled");
```
qiaozha, line 96: https://github.com/Azure/autorest.typescript/pull/2443/files#diff-37bd95d53e94cd07f0b9e3056ff9ef0be5f33c55d49f1a3ddc62207d6f444751R115-R117 why we need to change the test since we have the interface here ?

packages/rlc-common/src/buildIndexFile.ts
```diff
@@ -299,9 +297,7 @@ function generateRLCIndex(file: SourceFile, model: RLCModel) {
     ]);
   }
 
-  // TODO: Disable LRO poller tentatively
-  // https://github.com/Azure/autorest.typescript/issues/2230
-  if (hasPollingOperations(model) && !model.options?.isModularLibrary) {
```
qiaozha, line 304: will rlc code be different based on if it's modular or pure rlc?

packages/rlc-common/src/interfaces.ts
```diff
@@ -172,7 +172,7 @@ export type PathParameter = {
 
 export interface OperationHelperDetail {
   lroDetails?: OperationLroDetail;
-  isPageable?: boolean;
+  isPaging?: boolean;
```
qiaozha, line 175: why we need this change ?

packages/autorest.typescript/package.json
```diff
@@ -3,7 +3,8 @@
   "version": "6.0.18",
   "scripts": {
     "build": "tsc -p . && npm run copyFiles",
-    "build:test:browser": "tsc -p tsconfig.browser-test.json && webpack --config webpack.config.test.js",
+    "build:test:browser:rlc": "tsc -p tsconfig.browser-test.json && webpack --config webpack.config.test.js --env mode=rlc",
+    "build:test:browser": "tsc -p tsconfig.browser-test.json && webpack --config webpack.config.test.js --env mode=hlc",
```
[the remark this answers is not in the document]
MaryGao: the mode is different

packages/autorest.typescript/test/commands/prepare-deps.ts
```diff
@@ -0,0 +1,28 @@
+import { spawn } from "child_process";
```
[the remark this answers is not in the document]
MaryGao: yes

packages/rlc-common/src/buildIndexFile.ts
MaryGao: no

packages/rlc-common/src/interfaces.ts
MaryGao: it's the property name suggested by cspell.

packages/autorest.typescript/test/rlcIntegration/lroRest.spec.ts
joheredi: right, can we keep getOperationState() maybe we can just wrap opearationState. Also why are we nullchecking, does getLongRunningPoller return undefined?

MaryGao: >  does getLongRunningPoller return undefined?
yes, because it is promise like now so the value would be undefined if the promise is not settled down.

MaryGao: After adopting to V3 in RLC, we have two deprecated APIs which may be a breaking for customers. But considering the impact is limited I would prefer not taking them as breakings. The fact is 1) only 2 GAed RLC with LRO and 2) the toString is mainly used for rehydration which would be uncommon cases. The stopPolling is actually not supported in V2 version that's why we remove this in V3.

@qiaozha @joheredi How do you think?


No change:
- isDone, isStopped, getOperationState, getResult, poll, pollUntilDone, onProgress

Deprecated:
- toString, use `serialize` instead
- stopPolling, deprecated due to actually no support

Newly added:
- submitted
- serialize

MaryGao: As discussed in the meeting, we would accept the `toString` breaking and remove this in `SimplePollerLike`. And we would investigate if this is feasible to support `stopPolling` in RLC.

packages/typespec-test/test/authoring/generated/typespec-ts/src/pollingHelper.ts
```diff
@@ -38,6 +39,74 @@ import {
...
+
+  /**
+   * Returns a string representation of the poller's operation. Similar to serialize but returns a string.
+   */
+  toString(): string;
+
+  /**
+   * Stops the poller from continuing to poll. Please note this will only stop the client-side polling
+   */
+  stopPolling(): void;
```
joheredi, line 107: we need to mark these as @deprecated

packages/typespec-ts/test/modularIntegration/lroStardard.spec.ts
```diff
@@ -73,15 +72,73 @@ describe("LROStandardClient Classical Client", () => {
...
+        } as any,
+        {
+          abortSignal: abortController.signal
+        }
+      );
+      try {
+        await poller.submitted();
+        abortController.abort();
+        await poller.poll();
+        assert.fail("Should throw an AbortError");
```
qiaozha, line 99: is the case here trying to testify abort submit? what would happen if the initial request has submitted successfully ? Also, why would it impact the polling? Also it reminds me of a question, is the abort about LRO itself or about some specific async operation?

qiaozha (approved):

packages/typespec-ts/test/integration/lroRpc.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+describe("RpcClient Rest Client", () => {
+  let client: RpcClient;
+
+  beforeEach(() => {
+    client = SpecsAzureCoreLroStandardClientFactory({
+      allowInsecureConnection: true
+    });
+  });
+
+  it("should put LRO response", async () => {
```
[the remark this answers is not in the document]
MaryGao: this is rpc lro case and testify we could generate and call rpc cases.

packages/typespec-ts/test/modularIntegration/lroStardard.spec.ts
MaryGao: there are 3 cases, one is to verify the initial request; one is to verify the polling; the other is for the both the initial and polling set and both are set.

joheredi (approved):

PR #2446, opened 2024-04-15
Replace `c8` with `nyc` for coverage in generated packages

jeremymeng:
Switching back to nyc as done in the js mono repo: https://github.com/Azure/azure-sdk-for-js/commit/e9561c1a523a57d169ccc5d3800dacce25aadfcb


packages/autorest.typescript/test/rlcIntegration/generated/bodyStringRest/package.json
```diff
@@ -42,7 +42,7 @@
     "@azure/identity": "^4.0.1",
     "@azure-tools/test-recorder": "^3.0.0",
     "autorest": "latest",
-    "c8": "^8.0.0",
+    "c8": "^9.0.0",
```
qiaozha, line 45:
```suggestion
    "c8": "^9.1.0",
```
I thought it's 9.1.0 that we are trying to upgrade to ?

qiaozha (commented):
Might not be related, but what is this c8 being used for in our sdk repo ? I can't find anywhere that refers to this package.

packages/autorest.typescript/test/rlcIntegration/generated/bodyStringRest/package.json
jeremymeng: You are right. It should be 9.1.0

jeremymeng:
> Might not be related, but what is this c8 being used for in our sdk repo ? I can't find anywhere that refers to this package.

It is used for code coverage (previously we used `nyc` but that has problem with `esm`)

https://github.com/Azure/azure-sdk-for-js/blob/83ef149d60001cb53ca412208e265e327598c1b9/common/tools/dev-tool/src/commands/run/testNodeJSInput.ts#L34

qiaozha (approved):

jeremymeng:
Thinking about it more. I actually want to switch back to `nyc` now that we removed `esm` dev dependency.

PR #2447, opened 2024-04-15
[WIP] New design for multipart (type helper implementation)

timovv:
For discussion -- possible implementation of new `multipart/form-data`. This implementation uses a type helper, `FormDataPayload<T>`, to construct the expected body shape. The type helper is [currently implemented in the Core client package](https://github.com/Azure/azure-sdk-for-js/blob/3800693d8d63ce6cd9efdacb391161330f3eb7df/sdk/core/core-client-rest/src/multipart.ts#L53-L115) next to the actual multipart serialization logic, although we could also export it from somewhere else or emit it directly as needed. I don't feel strongly any way.

### Why a type helper?

The new multipart body shape is a complex type which requires creating a union consisting of all the possible parts. For example, the following

```typespec
model ExampleMultipartBody {
  propertyA: string;
  propertyB: bytes;
  propertyC: string[];
}
```

Needs to be modelled as

```ts
type ExampleMultipartBody = Array<
  | { name: "propertyA", body: string }
  | {
      name: "propertyB",
      body: Uint8Array | /* ...etc, other binary types... */,
      filename?: string,
      type?: string,
    }
  | {
      name: "propertyC",
      body: string,
    }
>;
```

This presents a few potential issues:

- It is technically quite difficult to model this given the structure of the generator (limited support for type aliases in the rlc-common Schema model and in ts-morph)
- It is difficult to model certain concepts such as inheritance (if a model extends another we need to pull in all the inherited properties into the union)
- Information is lost when generating the type i.e. whether the property is optional or multiple values are permitted

If we use a helper type alias, generating the original model and then using the helper where required, we circumvent all of these issues. We also get the benefit of preserving the original model which could provide context to the user on what the expected shape is. Using the type helper, the body parameter looks something like this:

```ts
export interface ExampleMultipartBodyParam {
  body: FormDataPayload<ExampleMultipartBody>;
}
```

### Alternatives

The obvious alternative to this design is to straight up generate the type alias. This is probably achievable but will require more work and consideration of the different edge cases. However, it might be easier for users to understand when they look at the types. If we think that this is important enough then we could go this route instead.

### Other changes made in this PR

- We don't need to re-export the `File` polyfill anymore, since the file name and MIME type can be set using the part descriptor directly, so I have removed that and the associated comments. We still include `File` in the union of accepted types for file uploads; it's just not necessary to provide a fill since there are now other ways to provide a filename.

### To do

- [ ] Merge and release Core PR first
- [ ] Create and regenerate e2e tests

### DIFF
--- a/packages/rlc-common/src/buildIndexFile.ts
+++ b/packages/rlc-common/src/buildIndexFile.ts
@@ -5,7 +5,6 @@ import { Project, SourceFile } from "ts-morph";
 import { NameType, normalizeName } from "./helpers/nameUtils.js";
 import {
   hasCsvCollection,
-  needsFilePolyfil,
   hasInputModels,
   hasMultiCollection,
   hasOutputModels,
@@ -19,7 +18,6 @@ import {
 import { RLCModel } from "./interfaces.js";
 import * as path from "path";
 import { getImportModuleName } from "./helpers/nameConstructors.js";
-import { getImportSpecifier } from "./helpers/importsUtil.js";
 
 export function buildIndexFile(model: RLCModel) {
   const multiClient = Boolean(model.options?.multiClient),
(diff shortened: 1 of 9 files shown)


timovv:
Closing in favor of #2455

PR #2453, opened 2024-04-18
Upgrade the tcgc to v0.41.3

v-jiaodi:


packages/typespec-test/test/openai/generated/typespec-ts/src/outputModels.ts
```diff
@@ -9,8 +9,6 @@ export interface DeploymentOutput {
   readonly deploymentId: string;
 }
 
-export interface AzureCoreTraitsOutput {}
```
qiaozha, line 12: what is causing the change here?

MaryGao: this is somehow introduced by our template logic and i remove that.

packages/typespec-ts/test/modularIntegration/generated/azure/core-traits/src/rest/outputModels.ts
```diff
@@ -1,15 +1,6 @@
...
 // Licensed under the MIT license.
 
-/** Provides the 'If-*' headers to enable conditional (cached) responses */
-export interface ConditionalRequestHeadersOutput {}
-
-/** Provides the 'ETag' header to enable conditional (cached) requests */
-export interface EtagResponseEnvelopeOutput {}
-
-/** Provides the 'x-ms-client-request-id' header to enable request correlation in requests and responses. */
-export interface ClientRequestIdHeaderOutput {}
```
MaryGao, line 11: these are just empty models and should not be generated and it is introduced by templateArgs logic.

packages/typespec-test/test/openai_modular/generated/typespec-ts/src/rest/outputModels.ts
```diff
@@ -3,12 +3,6 @@
 
 import { ErrorModel } from "@azure-rest/core-client";
 
-/** A specific deployment */
-export interface DeploymentOutput {
-  /** Specifies either the model deployment name (when using Azure OpenAI) or model name (when using non-Azure OpenAI) to use for this request. */
-  readonly deploymentId: string;
```
MaryGao, line 9: bugfix: this is only input model and should not be generated as output.

PR #2455, opened 2024-04-18
[WIP] New design for multipart - no type helper

timovv:
See https://github.com/Azure/autorest.typescript/pull/2447 for context. This version generates without a type helper. Hopefully this will improve understanding.

The scope of this first round of work is to create the API surface correctly. In follow up PRs, we should look at
- Generating good docs which highlight how to construct the body and show which parts are optional, required, etc
- Generating samples

These follow up items will improve the user experience.

### DIFF
--- a/packages/rlc-common/src/buildObjectTypes.ts
+++ b/packages/rlc-common/src/buildObjectTypes.ts
@@ -14,6 +14,7 @@ import {
   Parameter,
   Property,
   RLCModel,
+  Schema,
   SchemaContext
 } from "./interfaces.js";
 
@@ -36,6 +37,7 @@ export function buildObjectInterfaces(
     if (
       objectSchema.alias ||
       objectSchema.outputAlias ||
+      objectSchema.isMultipartBody ||
       objectSchema.fromCore
     ) {
       continue;
@@ -67,6 +69,37 @@ export function buildObjectAliases(
   const objectAliases: TypeAliasDeclarationStructure[] = [];
 
   for (const objectSchema of objectSchemas) {
+    if (objectSchema.isMultipartBody) {
+      const propertySignatures = getPropertySignatures(
+        objectSchema.properties ?? {},
+        schemaUsage,
+        importedModels,
+        { flattenArrays: true }
+      );
+
+      const objectTypeDefinitions = propertySignatures.map((sig) => {
+        if (sig.type?.toString().includes("File")) {
+          return `{name: ${sig.name}, body: ${
+            sig.type ?? "unknown"
+          }, filename?: string, type?: string }`;
+        } else {
+          return `{name: ${sig.name}, body: ${sig.type ?? "unknown"} }`;
+        }
+      });
+
+      objectAliases.push({
+        kind: StructureKind.TypeAlias,
+        ...(objectSchema.description && {
+          docs: [{ description: objectSchema.description }]
+        }),
+        name: objectSchema.typeName!,
+        type: `FormData | Array<${
+          objectTypeDefinitions.join("|") || "unknown"
+        }>`,
+        isExported: true
+      });
+    }
+
     if (objectSchema.alias || objectSchema.outputAlias) {
       const description = objectSchema.description;
       const modelName = schemaUsage.includes(SchemaContext.Input)
@@ -424,10 +457,15 @@ export function getImmediateParentsNames(
   return [...parents, ...extendFrom];
 }
 
+interface GetPropertySignatureOptions {
+  flattenArrays?: boolean;
+}
+
 function getPropertySignatures(
   properties: { [key: string]: Property },
   schemaUsage: SchemaContext[],
-  importedModels: Set<string>
+  importedModels: Set<string>,
+  options: GetPropertySignatureOptions = {}
 ) {
   let validProperties = Object.keys(properties);
   const readOnlyFilter = (name: string) =>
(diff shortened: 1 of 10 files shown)


packages/typespec-ts/test/integration/generated/payload/multipart/src/index.ts
```diff
@@ -8,11 +8,5 @@ export * from "./parameters.js";
 export * from "./responses.js";
 export * from "./clientDefinitions.js";
 export * from "./models.js";
-export {
-  createFile,
-  createFileFromStream,
-  type CreateFileOptions,
-  type CreateFileFromStreamOptions,
-} from "@azure/core-rest-pipeline";
```
MaryGao, line 11: is it because of filename and content type attributes so we could remove these helpers?

MaryGao:
Thanks for making this! The approach looks good to me:)~

packages/typespec-ts/test/integration/multipart.spec.ts
```diff
@@ -82,79 +53,64 @@ describe("MultiPartClient Rest Client", () => {
...
-          }
-        });
-      assert.strictEqual(result.status, "204");
-    });
-    it("createFileFromStream could return file objects", async () => {
-      const content = await readFile(imgPath);
-      const file = createFileFromStream(() => content, "hello.jpg", {
-        type: "image/jpg"
-      });
+    it("allows specifying MIME type and filename", async () => {
```
MaryGao, line 56: just curious without the helper `createFileFromStream` how to feed the big file in a streaming way?

timovv: You can just pass the stream directly:

```ts
const result = await client
        .path("/multipart/form-data/check-filename-and-content-type")
        .post({
          contentType: "multipart/form-data",
          body: [
            { name: "profileImage", body: fs.createReadStream("/path/to/file"), filename, contentType }
          ]
        });
```

packages/typespec-ts/test/integration/generated/payload/multipart/src/index.ts
timovv: Yeah, that, and the fact that we now accept streams as part bodies, so no need for createFileFromStream.

packages/rlc-common/src/metadata/packageJson/azurePackageCommon.ts
```diff
@@ -31,10 +31,13 @@ export function getAzureCommonPackageInfo(config: AzurePackageInfoConfig) {
...
   hasPaging,
-  specSource
+  specSource,
+  isModularLibrary
 }: AzurePackageInfoConfig) {
   let dependencies: Record<string, string> = {
-    "@azure-rest/core-client": specSource === "Swagger" ? "^1.4.0" : "^2.0.0",
+    // Swagger and modular libraries currently only support the old multipart/form-data implementation.
+    "@azure-rest/core-client":
+      specSource === "Swagger" || isModularLibrary ? "^1.4.0" : "^2.0.0",
```
MaryGao, line 40: why modular is depend on v1?

timovv: For modular to depend on v2, we need to update the modular API models as well, and probably also figure out how we want to handle serialization, which is a whole other piece of work which is I think out of scope for this PR. So I kept modular on the old version for now. Obviously we aren't going to keep it like that.

MaryGao: > For modular to depend on v2, we need to update the modular API models as well, and probably also figure out how we want to handle serialization

Personally I prefer to always use the latest version in modular. We didn't support mfd feature in core rest v1 either, it's fine considering there is no real need to modular mfd yet except openai(but i heard that we plan to depreate azure openai ) AND modular is still in preview stage.  The version mismatch may introduce un-necessary maintainance effort.

joheredi: I think it depends, if making modular take a dependency on v2 means there is additional effort in this PR I would favor keeping it in v1. If it is just a matter fo switching the string, go for it. However I think it is most likely the former, if so I'm fine keeping the 2 versions in the meantime.

joheredi (approved):
Looks good! Great improvement :)

packages/rlc-common/src/metadata/packageJson/azurePackageCommon.ts
timovv: ah, I think I see what you mean. I was under the impression that MFD was somewhat working in modular v1 given OpenAI is using it. But that is because they are doing customization, right? In that case I will change this to use 2.0, since there won't be a regression (it was already broken).

At the moment I have modular just generating normal models for MFD so as not to break the compilation.

MaryGao: yes, openai has customization here.

qiaozha (approved):
LGTM, just curious what would happen if a customer passes inconsistent content type and body in multipart form data operation ? how would the core serialize the body and send the request?

MaryGao (approved):
LGTM

packages/rlc-common/src/buildObjectTypes.ts
```diff
@@ -40,6 +48,19 @@ export function buildObjectInterfaces(
     ) {
       continue;
     }
+
+    // FIXME: disabling new multipart generation for modular while we figure out the story
+    if (objectSchema.isMultipartBody && !model.options?.isModularLibrary) {
```
MaryGao, line 53: nit: what RLC version will be generated in modular SDK? I image that new design would align with v2, old design would align with v1.

packages/rlc-common/src/buildObjectTypes.ts
```diff
@@ -54,6 +75,66 @@ export function buildObjectInterfaces(
...
+      type: "string"
+    },
+    {
+      name: "contentType",
+      hasQuestionToken: true,
+      type: "string"
+    }
+  ];
+
+function buildMultipartPartDefinitions(
```
MaryGao, line 92: Do we consider the case for  `A extends B`?

timovv: Not currently but we can revisit if this becomes a requirement

packages/rlc-common/src/buildObjectTypes.ts
timovv: We just generate a normal RLC model in the modular SDK (as if it was a plain old JSON operation) because of this check. This is more in line with v1 of the Core package. We can't generate the v2 models here yet as that results in a compile error, since the serializers don't serialize into the array shape.

timovv:
> LGTM, just curious what would happen if a customer passes inconsistent content type and body in multipart form data operation ? how would the core serialize the body and send the request?

If you set the content type to multipart/form-data and pass something unexpected it currently just JSON.stringifies it, which is what we were also doing in the old MFD implementation. I opened an issue for Core tracking this earlier: https://github.com/Azure/azure-sdk-for-js/issues/29424

PR #2462, opened 2024-04-19
use `dev-tool run extract-api` in packages generated for JS mono repo

jeremymeng:
This follows the commit in JS repo: https://github.com/Azure/azure-sdk-for-js/commit/224f7c1c1493ad737b22fd58ae634ccad8cdd71b

### DIFF
--- a/packages/autorest.typescript/src/generators/static/packageFileGenerator.ts
+++ b/packages/autorest.typescript/src/generators/static/packageFileGenerator.ts
@@ -185,6 +185,7 @@ function regularAutorestPackage(
     packageInfo.devDependencies["@azure/dev-tool"] = "^1.0.0";
     packageInfo.scripts["build"] =
       "npm run clean && tsc && dev-tool run bundle && npm run minify && mkdirp ./review && npm run extract-api";
+    packageInfo.scripts["extract-api"] = "dev-tool run extract-api";
   } else {
     packageInfo.devDependencies["@rollup/plugin-commonjs"] = "^24.0.0";
     packageInfo.devDependencies["@rollup/plugin-json"] = "^6.0.0";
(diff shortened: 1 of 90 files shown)


packages/autorest.typescript/src/generators/static/packageFileGenerator.ts
```diff
@@ -185,6 +185,7 @@ function regularAutorestPackage(
     packageInfo.devDependencies["@azure/dev-tool"] = "^1.0.0";
     packageInfo.scripts["build"] =
       "npm run clean && tsc && dev-tool run bundle && npm run minify && mkdirp ./review && npm run extract-api";
+    packageInfo.scripts["extract-api"] = "dev-tool run extract-api";
```
MaryGao, line 188: May I know if we need to switch all actions to dev-tool?

jeremymeng: For the JS monorepo we would prefer the consistent and centralized dev-tool command.  It adds support for the subpath exports.

packages/rlc-common/src/metadata/packageJson/buildAzureMonorepoPackage.ts
```diff
@@ -223,7 +225,7 @@ function getEsmScripts({ moduleKind }: AzureMonorepoInfoConfig) {
   return {
     "build:test": "npm run clean && tshy && dev-tool run build-test",
     build:
-      "npm run clean && tshy && mkdirp ./review && api-extractor run --local",
+      "npm run clean && tshy && mkdirp ./review && dev-tool run extract-api",
```
MaryGao, line 228: maybe no need to `mkdirp ./review` twice?

MaryGao (approved):
LGTM

PR #2469, opened 2024-04-23
Additional properties for modular legacy client support

qiaozha:
fixes https://github.com/Azure/autorest.typescript/issues/2470


packages/typespec-ts/src/modular/buildOperations.ts
```diff
@@ -49,7 +49,9 @@ export function buildOperationFiles(
         subfolder && subfolder !== "" ? subfolder + "/" : ""
       }api/${operationFileName}.ts`
     );
-
+    if (operationFileName.startsWith("extendsDifferentSpreadFloat")) {
+      operationGroup;
+    }
```
qiaozha, line 54:
```suggestion
```

packages/rlc-common/src/interfaces.ts
```diff
@@ -240,6 +240,7 @@ export interface RLCOptions {
   flavor?: PackageFlavor;
   enableModelNamespace?: boolean;
   hierarchyClient?: boolean;
+  compatibilityMode?: boolean;
```
MaryGao, line 243: Just confirm - does the `legacy` title mean HLC's behavior? Is our current pr for HLC?

Because I notice the HLC use property bag instead?
- https://github.com/Azure/autorest.testserver/blob/main/swagger/additionalProperties.json#L269-L271
- https://github.com/Azure/autorest.typescript/blob/main/packages/autorest.typescript/test/integration/generated/additionalProperties/src/models/index.ts#L36

qiaozha: For you information 
```ts
export interface PetAPString {
  /** Describes unknown properties. The value of an unknown property can be of "any" type. */
  [property: string]: any;
  id: number;
  name?: string;
  /** NOTE: This property will not be serialized. It can only be populated by the server. */
  readonly status?: boolean;
}
```
 is equivalent to 
```ts
export interface PetAPString extends Record<string, any> {
  id: number;
  name?: string;
  /** NOTE: This property will not be serialized. It can only be populated by the server. */
  readonly status?: boolean;
}
``` 
which is what I am implementing in this PR right now. and the difference between the indexer one and the extends record one has already been decided in previous PR https://github.com/Azure/autorest.typescript/pull/2054#issuecomment-1758730274. 

if we are using the additionalProperties bag, i.e. the non-legacy client description for this would be:
```ts
export interface PetAPString {
   /** Describes unknown properties. The value of an unknown property can be of "any" type. */
  additionalProperties: Record<string, string>; // there's no any in this case.
  id: number;
  name?: string;
  /** NOTE: This property will not be serialized. It can only be populated by the server. */
  readonly status?: boolean;
}
```

MaryGao: I was wondering how we could handle following cases?

```
model Base {
  foo: int32;
}
model A extends Base{
  ...Record<int32>
  prop: int32
}
```

qiaozha:
```ts
    interface Base {
      foo: number;
    }
    interface A extends Base, Record<string, number> {
      prop: number;
    }
```

qiaozha: I don't see any problem with the above ts expression, if there is,I don't think it's in the scope of this PR. 
And also FYI, even if the above syntax is valid in typespec, but if we write it as 
```
model Base {
  foo: int32;
}
model A extends Base, Record<int32>{
  prop: int32
}
```
which is hypothetically equivalent, but in typespec, it's a syntax error. which means we should not recommend the original extends Base tsp. 

And according to my previous conversation with tsp team, If we really want to express model A has a Base class and additional properties, spread of Base is the recommend typespec expression. 

```
model Base {
  foo: int32;
}

model A {
  ...Base;
  ...Record<int32>;
  prop: int32;
}
```
which I don't see we have any problem to express it.

packages/typespec-ts/test/modularIntegration/generated/models/propertyAdditional/src/rest/outputModels.ts
```diff
@@ -0,0 +1,269 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/** The model extends from Record<unknown> type. */
+export interface ExtendsUnknownAdditionalPropertiesOutput
+  extends Record<string, any> {
```
MaryGao, line 6: Considering the type generated in rlc and modular is different, not sure there would happen for compile error between them because of type inconsistancy?

qiaozha: Can you point out what is the difference here? the legacy client should be the same in both Modular and RLC.

qiaozha: or you mean unknown specifically? which we have to use any for both input and output as we don't split the input models and output models in Modular.

qiaozha: Not sure if @joheredi has any different thoughts.

packages/typespec-ts/src/modular/emitModels.ts
```diff
@@ -48,10 +48,11 @@ function getCoreLroType(name: string, coreLroTypes: Set<string>) {
...
     (t) =>
-      (t.type === "model" || t.type === "enum") &&
-      !isAzureCoreErrorSdkType(t) &&
-      !isAzureCoreLroSdkType(t) &&
-      !(t.type == "model" && t.name === "")
+      ((t.type === "model" || t.type === "enum") &&
+        !isAzureCoreErrorSdkType(t) &&
+        !isAzureCoreLroSdkType(t) &&
+        !(t.type == "model" && t.name === "")) ||
+      (t.type === "dict" && t.properties?.length && t.properties.length > 0)
```
joheredi, line 55: We should consider making these more readable

joheredi (approved):

packages/typespec-ts/src/modular/emitModels.ts
qiaozha: thanks for the advice, I have updated to have an independent function to make it work.

packages/typespec-ts/test/modularIntegration/generated/models/propertyAdditional/src/rest/outputModels.ts
MaryGao: Thanks for the explainantion! `any` makes sense for me. I just was thinking because there exist type difference between rlc and modular, for example tsp datetime would be translated to Date in modular but string in RLC, would it be a problem here?

Another question is if there exist ser/derser logic e.g properties renaming, how to handle additional properties? is it in current pr scope?

packages/typespec-ts/test/modularUnit/enumUnion.spec.ts
```diff
@@ -621,7 +621,10 @@ describe("header parameters", () => {
       const schemaOutput = await emitModularModelsFromTypeSpec(
         tspDefinition,
         false,
-        true
+        true,
+        false,
+        false,
+        false
```
MaryGao, line 627: I awalys want to do refactoring these options to option bag, otherwise we should be carefully with the order.

packages/rlc-common/src/interfaces.ts
MaryGao: Maybe not this pr, i just think out a case in non-legacy mode in tsp service defined an `additionalProperty` but also `...Record<int32>` for that model, how could we resolve the conflict?

packages/typespec-ts/src/lib.ts
```diff
@@ -221,6 +223,12 @@ const libDef = {
       messages: {
         default: paramMessage`Please note the header ${"type"} is not serializable.`
       }
+    },
+    "compatible-additional-properties": {
```
MaryGao, line 227: Is this temp warning because our non-legacy code is not ready?

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -601,8 +609,10 @@ function buildBodyParameter(
       return `\nbody: ${bodyParameter.clientName},`;
     }
   } else if (
-    bodyParameter.type.type === "model" &&
-    bodyParameter.type.aliasType
+    (bodyParameter.type.type === "model" &&
+      (bodyParameter.type.aliasType ||
+        allParents.some((p) => p.type === "dict"))) ||
+    bodyParameter.type.type === "dict"
```
MaryGao, line 615: so we will directly return the body?

MaryGao (approved):
Approved and left some confirmation!

qiaozha:
serialize and deserialize can't be support for additional properties now, create an issue to track unsupported cases in compatible mode https://github.com/Azure/autorest.typescript/issues/2506

PR #2491, opened 2024-05-03
Fix TypeSpec compiler options API version override

### LINKED ISSUE #2490: [Bug] API Version Not Respected in `createClient`
**Description:**
When generating a client that targets a specific API version, the generated `createClient` function does not respect the specified `api-version` flag in `tspconfig.yaml`. Although models and operations for the specific version are generated correctly, `createClient` defaults to the latest version instead of the specified one.

**Relevant Details:**
- TCGC was updated to support targeting specific API versions through the `api-version` flag in `tspconfig.yaml`, as introduced by [this change](https://github.com/Azure/typespec-azure/pull/700/files#diff-3599739b76751d3c27620bc862fbe98c2ba6e0cb2fedfe25b264176b10dfb86e).
- The TypeScript emitter is already utilizing TCGC with this feature support.
- The issue is specifically observed in the generated `createClient` function which continues to use the default (latest) API version.

**Suspected Culprit:**
- The issue might be stemming from how the API version is handled or defaulted in the `createClient` function within the TypeScript Emitter. Specifically, the logic in this [line of code](https://github.com/Azure/autorest.typescript/blob/3b06cf501ad088d58378c15e772a9282a36a1605/packages/rlc-common/src/buildClient.ts#L310) might be ignoring the specified API version.

**Steps to Reproduce:**
1. Configure `tspconfig.yaml` with a specific `api-version`.
2. Generate a client using the Typescript Emitter.
3. Observe that the generated client's `createClient` function defaults to the latest API version instead of using the specified version.

**Expected Behavior:**
The `createClient` function should respect the API version specified in `tspconfig.yaml` and use it instead of defaulting to the latest version.

**Actual Behavior:**
The `createClient` function ignores the specified API version and defaults to the latest version.

**Screenshots/Logs:**
Issue Screenshot

**Environment:**
- Typescript Emitter version: 0.28.0
- Typespec Compiler version: 0.55.0

dgetu:
Fixes #2490

### DIFF
--- a/packages/typespec-ts/src/transform/transformApiVersionInfo.ts
+++ b/packages/typespec-ts/src/transform/transformApiVersionInfo.ts
@@ -28,26 +28,24 @@ export function transformApiVersionInfo(
 ): ApiVersionInfo | undefined {
   const queryVersionDetail = getOperationApiVersion(client, dpgContext);
   const pathVersionDetail = extractPathApiVersion(urlInfo);
-  const isCrossedVersion =
-    pathVersionDetail?.isCrossedVersion || queryVersionDetail?.isCrossedVersion;
-  const defaultValue =
-    (pathVersionDetail || queryVersionDetail) && !isCrossedVersion
-      ? getDefaultApiVersionString(dpgContext) ??
-        pathVersionDetail?.defaultValue ??
-        queryVersionDetail?.defaultValue
-      : undefined;
 
-  if (pathVersionDetail && queryVersionDetail) {
-    return pathVersionDetail;
-  }
   if (pathVersionDetail && !queryVersionDetail) {
     dpgContext.hasApiVersionInClient = true;
   }
+
+  const definedPosition =
+    pathVersionDetail && queryVersionDetail
+      ? pathVersionDetail.definedPosition
+      : extractDefinedPosition(queryVersionDetail, pathVersionDetail);
+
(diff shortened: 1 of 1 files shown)


packages/typespec-ts/src/transform/transformApiVersionInfo.ts
```diff
@@ -28,31 +28,61 @@ export function transformApiVersionInfo(
...
     defaultValue
   };
 }
 
+function getDefaultValue(
+  dpgContext: SdkContext,
+  queryVersionDetail?: ApiVersionInfo,
+  pathVersionDetail?: ApiVersionInfo
+): string | undefined {
+  if (dpgContext.apiVersion) {
```
qiaozha, line 63: I am a little doubt about we add this logic by ourself, I think we have used tcgc getDefaultApiVersion https://github.com/qiaozha/autorest.typescript/blob/d69af33e2327c9017c0aee17566d9c5e60759b53/packages/typespec-ts/src/utils/modelUtils.ts#L70. @dgetu can you check if that api has returned the correct default value? if it's not, is it that api's responsibility to return the correct default api version, if we have specified api-version in tspconfig.yaml ?

joheredi:
@dgetu I beleive we can close this PR as this was fixed upstream in TCGC. Feel free to reopen if needed

Issue #2492, opened 2024-05-07
Support `void` type in RLC and modular


MaryGao:
This is issue found when generating fleet ([tsp here](https://github.com/Azure/azure-rest-api-specs/blob/bce261b592674cd74a6603ea65b8561fb20ed154/specification/containerservice/Fleet.Management/fleet.tsp#L237)) and also in sphere tsp: https://github.com/Azure/autorest.typescript/pull/2498.

![image](https://github.com/Azure/autorest.typescript/assets/9943211/de3b6e45-c92d-4a78-8ba3-86fc16d4fb55)

See the discussion in typespec side https://github.com/microsoft/typespec/issues/3299.

MaryGao:
Case 1: void as request body & void as response

[tsp](https://cadlplayground.z22.web.core.windows.net/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBUeXBlU3BlYy5IdHRwOwoKb3AgdGVzdChAYm9keSDEBTogdm9pZCnGBzs%3D&e=%40typespec%2Fopenapi3&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%7D) 

```
op test(@body body: void): void;
```

RLC generated

```ts
// omit the parameter body
export type TestParameters = RequestParameters;

// omit the response body
export interface Test200Response extends HttpResponse {
  status: "200";
}
```

Modular generated

```ts
export async function test(
  context: Client,
  // omit the body parameter
  options: TestOptionalParams = { requestOptions: {} },
): Promise<void> 
```

Case 2: void as response body

[tsp](https://cadlplayground.z22.web.core.windows.net/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBUeXBlU3BlYy5IdHRwOwoKb3AgdGVzdCgpOiB7CiAgQGJvZHkgxAU6IHZvaWQ7Cn07&e=%40typespec%2Fopenapi3&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%7D)
```
op test(): {
  @body body: void;
};
```

RLC generated

```ts
// omit the response body
export interface Test200Response extends HttpResponse {
  status: "200";
}
```

Modular generated

```ts
export async function test(
[...]

PR #2493, opened 2024-05-07
[hold-on]Add the void type support and fix fleet/sphere generation issue

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/2492


qiaozha (commented):
if the only generation error, we found in fleet.management is the void type, I don't think it's worth to put it in the smoke test.

MaryGao:
Pls hold on the review and we need to first review the design https://github.com/Azure/autorest.typescript/issues/2492#issuecomment-2099709887

MaryGao:
Close as duplicate with https://github.com/Azure/autorest.typescript/pull/2521

PR #2503, opened 2024-05-08
Upgrade typespec dep version to v0.56 in May

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/2515


packages/typespec-test/test/contoso/generated/typespec-ts/src/outputModels.ts
```diff
@@ -34,16 +34,6 @@ export interface ResourceOperationStatusOutput {
   result?: WidgetOutput;
 }
 
-/** Provides status details for long running operations. */
-export interface OperationStatusOutput {
```
MaryGao, line 38: The model is failed to set in program map in logic: https://github.com/Azure/autorest.typescript/blob/main/packages/typespec-ts/src/transform/transformSchemas.ts#L138-L146

packages/typespec-ts/src/transform/transformSchemas.ts
```diff
@@ -112,7 +112,7 @@ export function transformSchemas(client: SdkClient, dpgContext: SdkContext) {
     }
   }
   transformHostParameters();
-  program.stateMap(modelKey).forEach((context, tspModel) => {
+  usageMap.forEach((context, tspModel) => {
```
MaryGao, line 115: Confirmed with Timothee, this is abuse of program.stateMap which is only used for saving decorator info: https://github.com/microsoft/typespec/issues/3335#issuecomment-2107933123 

So I replaced it with internal map instead: .

packages/typespec-ts/test/util/testUtil.ts
```diff
@@ -75,23 +76,13 @@ export function createDpgContextTestHelper(
...
-    emitters: [
-      {
-        main: "@azure-tools/typespec-ts",
-        metadata: { name: "@azure-tools/typespec-ts" }
-      }
-    ]
-  };
-  const resolvedOptions = { ...defaultOptions };
-  program.emitters = resolvedOptions.emitters as any;
+  const context = createContextWithDefaultOptions({
```
MaryGao, line 79: We need to explicitly call this to build the internal state for tcgc.

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -1587,21 +1569,26 @@ export function trimUsage(model: any) {
...
       ? "LroType"
       : undefined;
 }
 
-export function isAzureCoreErrorType(t?: Type): boolean {
+export function isAzureCoreErrorType(program: Program, t?: Type): boolean {
+  if (!t || t.kind !== "Model") {
+    return false;
+  }
+  const effective = getEffectiveSchemaType(program, t);
```
MaryGao, line 1584: Bugfixing: we need to consider the anonymous model also.

packages/typespec-test/test/anomalyDetector/generated/typespec-ts/src/rest/responses.ts
```diff
@@ -83,7 +83,7 @@ export interface GetMultivariateBatchDetectionResultDefaultHeaders {
 export interface GetMultivariateBatchDetectionResultDefaultResponse
   extends HttpResponse {
   status: string;
-  body: MultivariateResponseErrorOutput;
+  body: MultivariateErrorResponseOutput;
```
MaryGao, line 86: This is introduced by compiler change https://github.com/microsoft/typespec/pull/2945.

So openapi.json has similar changes.

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/review/eventgrid.api.md
```diff
@@ -50,8 +50,8 @@ export class EventGridClient {
     constructor(endpointParam: string, credential: KeyCredential, options?: EventGridClientOptions);
     acknowledgeCloudEvents(topicName: string, eventSubscriptionName: string, lockTokens: AcknowledgeOptions, options?: AcknowledgeCloudEventsOptionalParams): Promise<AcknowledgeResult>;
     readonly pipeline: Pipeline;
-    publishCloudEvent(topicName: string, event: CloudEvent, options?: PublishCloudEventOptionalParams): Promise<Record<string, any>>;
-    publishCloudEvents(topicName: string, events: CloudEvent[], options?: PublishCloudEventsOptionalParams): Promise<Record<string, any>>;
+    publishCloudEvent(topicName: string, event: CloudEvent, options?: PublishCloudEventOptionalParams): Promise<PublishResult>;
```
MaryGao, line 53: The current generation should be correct.

.vscode/launch.json
```diff
@@ -112,7 +112,7 @@
     {
       "name": "[TypeSpec] Smoke Test Debug",
       "request": "launch",
-      "cwd": "${workspaceFolder}/packages/typespec-test/test/customWrapper",
+      "cwd": "${workspaceFolder}/packages/typespec-test/test/todo_non_branded",
```
MaryGao, line 115:
```suggestion
      "cwd": "${workspaceFolder}/packages/typespec-test/test/customWrapper",

```

.vscode/launch.json
```diff
@@ -113,6 +113,7 @@
       "name": "[TypeSpec] Smoke Test Debug",
       "request": "launch",
       "cwd": "${workspaceFolder}/packages/typespec-test/test/customWrapper",
+
```
MaryGao, line 116:
```suggestion

```

packages/typespec-ts/test/integration/typespec/union-body/main.tsp
```diff
@@ -28,9 +28,9 @@ model ReturnResponse {
 namespace NicepayPlatform {
   @post
   @route("/request-union-body")
-  op requestUnionBody(@body request: RequestRegisterCC | RequestRegisterVA): {};
+  op requestUnionBody(@body request: RequestRegisterCC | RequestRegisterVA): {@bodyRoot _ : {};};
```
qiaozha, line 31: what does the original {} gives us? void?

packages/typespec-test/test/authoring/generated/openapi/2022-05-15-preview/openapi.json
```diff
@@ -1479,23 +1479,6 @@
       "required": [
         "modelLabel"
       ]
-    },
-    "Versions": {
```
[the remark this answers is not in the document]
MaryGao: TBH I don't know why this definition is removed, @timotheeguerin this happened when upgrading to the latest not sure you know why?
https://github.com/Azure/autorest.typescript/blob/7e7d27973daa4ee6ef081e54c47a0c4f2977153b/packages/typespec-test/test/authoring/spec/main.tsp#L28-L32

packages/typespec-test/test/todo_non_branded/tspconfig.yaml
```diff
@@ -9,3 +10,5 @@ options:
     "emitter-output-dir": "{project-root}/generated/typespec-ts"
     packageDetails:
       name: "@notabrand/todo-non-branded"
+  "@azure-tools/typespec-autorest":
```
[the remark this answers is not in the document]
MaryGao: To easily compare the openapi change with RLC changes.

packages/typespec-ts/src/transform/transformSchemas.ts
```diff
@@ -27,9 +27,9 @@ export function transformSchemas(client: SdkClient, dpgContext: SdkContext) {
     SchemaContext[]
   >();
   const schemaMap: Map<any, any> = new Map<any, any>();
+  const usageMap = new Map<Type, SchemaContext[]>();
```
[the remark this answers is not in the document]
MaryGao: Compile error?

packages/typespec-ts/test/integration/typespec/union-body/main.tsp
MaryGao: void.

packages/typespec-test/test/todo_non_branded/tspconfig.yaml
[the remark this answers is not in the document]
timotheeguerin: @marygao I would consider using the openapi3 emitter instead if you want the more accurate representation of the api like what rlc would do

packages/typespec-test/test/authoring/generated/openapi/2022-05-15-preview/openapi.json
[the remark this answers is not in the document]
timotheeguerin: First breaking change in the release notes https://azure.github.io/typespec-azure/docs/release-notes/release-2024-05-07#azure-toolstypespec-autorest

packages/typespec-test/test/openai/generated/openapi/2023-07-01-preview/openapi.json
```diff
@@ -1335,36 +1311,6 @@
         "b64_json"
       ]
     },
-    "ImageSize": {
```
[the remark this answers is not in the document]
MaryGao: @timotheeguerin Regarding the openapi emitter do we have any fix for versioning and do you have any idea of this change? According to tsp `ImageSize` should be added since version `v2023_06_01_Preview`.

https://github.com/Azure/autorest.typescript/blob/7e7d27973daa4ee6ef081e54c47a0c4f2977153b/packages/typespec-test/test/openai/spec/models/images.tsp#L15-L25

packages/typespec-test/test/todo_non_branded/tspconfig.yaml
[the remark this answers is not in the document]
MaryGao: Sure then.

packages/typespec-test/test/openai/generated/openapi/2023-07-01-preview/openapi.json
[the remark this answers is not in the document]
MaryGao: should be a bug https://github.com/Azure/typespec-azure/issues/838

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -242,18 +242,13 @@ export function getEffectiveModelFromType(program: Program, type: Type): Type {
    * set of properties when non-schema properties are excluded.
    */
   if (type.kind === "Model" && type.name === "") {
-    const effective = getEffectiveModelType(program, type, isSchemaProperty);
+    const effective = getEffectiveModelType(program, type, (property) =>
+      isSchemaProperty(program, property)
```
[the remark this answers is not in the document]
MaryGao: no, this part should be just refactoring.

one bug is we leverage our own helper `getEffectiveSchemaType` to consider the nullale case.https://github.com/Azure/autorest.typescript/pull/2503#discussion_r1599420760

packages/typespec-test/test/todo_non_branded/tspconfig.yaml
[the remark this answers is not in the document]
qiaozha: if I remember correctly autorest emitter will respect some of the tcgc customization which RLC should not follow.

packages/typespec-test/test/openai/generated/openapi/2023-07-01-preview/openapi.json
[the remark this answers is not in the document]
timotheeguerin: this was merged you can update to autorest 0.42.1

packages/typespec-test/package.json
```diff
@@ -4,15 +4,15 @@
...
-    "@azure-tools/typespec-autorest": ">=0.41.0 <1.0.0",
-    "@typespec/openapi3": ">=0.55.0 <1.0.0",
-    "@azure-tools/typespec-azure-core": ">=0.41.0 <1.0.0",
-    "@azure-tools/typespec-client-generator-core": ">=0.41.5 <1.0.0",
-    "@typespec/compiler": ">=0.55.0 <1.0.0",
-    "@typespec/http": ">=0.55.0 <1.0.0",
-    "@typespec/rest": ">=0.55.0 <1.0.0",
-    "@typespec/versioning": ">=0.55.0 <1.0.0",
+    "@typespec/openapi": ">=0.56.0 <1.0.0",
+    "@azure-tools/typespec-autorest": ">=0.42.0 <1.0.0",
```
qiaozha, line 8: should we update the version here for the latest bugfixing?

packages/typespec-ts/test/integration/typespec/union-body/main.tsp
timotheeguerin: shoudl be `@body` here if you want actually `{}`, that is actually I think a bug, `@bodyRoot` should just move the body resolution down to that property instead of the root of the response type and not change the meaning

timotheeguerin: filed https://github.com/microsoft/typespec/issues/3372

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/review/eventgrid.api.md
MaryGao: not sure i doubt this would be a bug in geteffectivemodel or somewhere not us, becuase the getAllModels didn't return it yet.

packages/typespec-test/test/todo_non_branded/tspconfig.yaml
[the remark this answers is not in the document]
timotheeguerin: yeah that too, if RLC doesn't respect any of the tcgc decorator but completely use the service, i think openapi3 emitter would be the most accurate

packages/typespec-ts/test/integration/typespec/union-body/main.tsp
qiaozha: can we use `{@bodyRoot _: void }` to express void body? I feel like it's strange if both `{@bodyRoot _: void }` and `{@bodyRoot _: {} }` means response body void? 
Also, I am curious if we can use bodyRoot to express non-object and non-void type body, like string, boolean etc with {}

timotheeguerin: The logic is as follow
1. `@bodyRoot` should just move the implicit body resolution from the root of the request/response to the property you have it. So it should be exactly the same handling to have `op foo(): X` vs `op foo(): {@bodyRoot _: X}` as the over the write meaning.
2. Now the reason why `{}` means no body is because `{@header foo: string}` means no body but `{}` didn't use to. If a property was removed with visibility it wasn't meaning empty body either unless it was marked with `@header` as well. So it was very inconsitent. So to simplify the logic is: after removing the non applicable properties if there is none left it means we have no body. In the rare case where you actually want to send `{}` you can use `@body` which will expect it to be the exact body.

qiaozha: I do understand the intention for bodyRoot as I read about this before when I was working on visibility related design spec, I was just feel like since void can't be used in property type, maybe we should not allow `{@bodyRoot _: void }` too?

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/review/eventgrid.api.md
qiaozha: I suspect it is because we treat anonymous empty model and named empty model differently? I vaguely remember there's a discussion about whether anonymous empty model should be an open dict ?

MaryGao: yeah, i guess so and this is also template argument and this may be the impact also. but i can't reproduce now
https://cadlplayground.z22.web.core.windows.net/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7Cgp1c2luZyBUeXBlU3BlYy5IdHRwOwpAc2VydmljZSh7CiAgdGl0bGU6ICJXaWRnZXQgU8YbIiwKfSkKbmFtZXNwYWNlIERlbW%2FHGzsKCm1vZGVsIEZvbyB7fQpvcCB0ZXN0KCk6IHtAYm9keSBfOsQcO307&e=%40typespec%2Fopenapi3&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40typespec%2Fhttp%2Fall%22%5D%7D%7D

packages/typespec-test/test/eventgrid_modular/spec/main.tsp
```diff
@@ -178,7 +178,7 @@ namespace Azure.Messaging.EventGrid {
     @doc("Single Cloud Event being published.")
     @body 
     event: {event: CloudEvent};
-  }, PublishResult>;
+  }, { @body _: PublishResult; }>;
```
MaryGao, line 181: @timotheeguerin I believe here we should use `@body` not `@bodyRoot` because the PublishResult is an empty model.

tsp in spec repo: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/eventgrid/Azure.Messaging.EventGrid/main.tsp#L250

timotheeguerin: we'll fix it in the spec repo when the issue I linked above is resolved.

lirenhe (approved):

qiaozha (approved):

PR #2511, opened 2024-05-10
Should turn off some rules in esm env for eslint

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/2517


MaryGao (dismissed):
we should only turn off these rules in esm env.

packages/autorest.typescript/test/commands/version-tolerance.ts
```diff
@@ -63,8 +63,8 @@ interface SwaggerInfo {
 const versionToleranceTests: VersionToleranceTest[] = [
   {
     name: "rlcClient",
-    initialSwagger: { input: "dpg_initial.json", output: "rlc-initial" },
-    updatedSwagger: { input: "dpg_update1.json", output: "rlc-updated" }
+    initialSwagger: { input: "dpg-initial.json", output: "rlc-initial" },
```
MaryGao, line 66: do we run this in ci?

packages/rlc-common/src/metadata/buildESLintConfig.ts
```diff
@@ -22,6 +22,16 @@ export function buildEsLintConfig(model: RLCModel) {
   }
   const project = new Project();
   const filePath = ".eslintrc.json";
+  if (model.options?.moduleKind == "esm") {
+    (esLintConfig.rules as any)["@azure/azure-sdk/ts-package-json-module"] =
```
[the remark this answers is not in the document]
MaryGao: pls use the type `const esLintConfig: Record<string, any>` to mitigate this.

packages/autorest.typescript/test/commands/version-tolerance.ts
MaryGao: just confirm we will not run this in ci and so just keep this files up-to-date

packages/rlc-common/src/metadata/buildESLintConfig.ts
```diff
@@ -22,6 +22,12 @@ export function buildEsLintConfig(model: RLCModel) {
   }
   const project = new Project();
   const filePath = ".eslintrc.json";
+  if (model.options?.moduleKind == "esm") {
```
MaryGao, line 25: This should be the only change in this pr.

MaryGao (approved):
LGTM

PR #2521, opened 2024-05-14
Support void type in RLC and Modular

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2492


packages/typespec-ts/test/modularUnit/operations.spec.ts
```diff
@@ -4,6 +4,92 @@ import { assertEqualContent } from "../util/testUtil.js";
...
+          param: void;
+        }
+        op read(...Foo): {};
+          `;
+
+        await emitModularOperationsFromTypeSpec(tspContent);
+        assert.fail("Should throw diagnostic errors");
+      } catch (e: any) {
+        assert.equal(e[0]?.code, "@azure-tools/typespec-ts/invalid-schema");
+        assert.equal(e[0]?.message, "Couldn't get schema for type Intrinsic");
```
qiaozha, line 89: can this error message be more specific?

MaryGao: there are type relevant actually and i could add the assertion here.

MaryGao: l leverage the existing rule `invalid-schema` for this check and let me know if you think we need a new linter for void.

qiaozha (approved):

PR #2544, opened 2024-05-23
Emit serialize util

dgetu:


joheredi:
I believe we can close this PR in favor of #2613

PR #2548, opened 2024-05-24
Failed to build batch for smoke test

v-jiaodi:
Generate from https://github.com/Azure/azure-rest-api-specs/pull/29018

Build error log:
image


qiaozha:
I don't think we need this now, as service team should fix their typespec to not include the client decorator in their main.tsp scope.

PR #2550, opened 2024-05-29
fix duplicate response headers found in appcompliance with tsp

kazrael2119:


qiaozha:
pending on https://github.com/Azure/typespec-azure/issues/938

packages/typespec-ts/test/unit/armTemplate.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+      
+      /** Azure operation completed successfully. */
+      export interface Upload200Response extends HttpResponse {
+          status: "200";
+          body: OnboardResponseOutput;
+      }
+      
+      export interface Upload202Headers {
+          /** The Location header contains the URL where the status of the long running operation can be checked. */
+          "location"?: string;
```
qiaozha, line 92: If we don't have the fix, we will generate two `location` headers.

lirenhe (approved):

Issue #2396, opened 2024-04-02
Analyze breakings introduced by core & accept or mitigate them


MaryGao:
```[tasklist]
### Tasks
- [x] Compare the changes for client-level options `ServiceClientOptions`
- [x] Compare the changes for operation-level options `OperationOptions`
```

kazrael2119:
in hlc ，it imports OperationOptions from @azure/core-client, and it has five parameters:
```
export interface OperationOptions {
  /**
   * The signal which can be used to abort requests.
   */
  abortSignal?: AbortSignalLike;
  /**
   * Options used when creating and sending HTTP requests for this operation.
   */
  requestOptions?: OperationRequestOptions;
  /**
   * Options used when tracing is enabled.
   */
  tracingOptions?: OperationTracingOptions;
  /**
   * Options to override serialization/de-serialization behavior.
   */
  serializerOptions?: SerializerOptions;

  /**
   * A function to be called each time a response is received from the server
   * while performing the requested operation.
   * May be called multiple times.
   */
  onResponse?: RawResponseCallback;
}
```

but in modular, it imports OperationOptions from @azure-rest/core-client ,  and it has four parameters
```
export interface OperationOptions {
  /**
   * The signal which can be used to abort requests.
   */
  abortSignal?: AbortSignalLike;
  /**
   * Options used when creating and sending HTTP requests for this operation.
   */
  requestOptions?: OperationRequestOptions;
  /**
   * Options used when tracing is enabled.
   */
  tracingOptions?: OperationTracingOptions;

  /**
   * A function to be called each time a response is received from the server
   * while performing the requested operation.
   * May be called multiple times.
   */
  onResponse?: RawResponseCallback;
}
```

**breaking** remove serializerOptions

kazrael2119:
in hlc, the optional parameter of client refers **ServiceClientOptions**, but in modular, it refers ClientOptions

kazrael2119:
compare paging breaking between modular and hlc:
1. interface PageSettings no longer has parameter : maxPageSize
2. remove function getContinuationToken
3. generic parameters change:https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/core-paging/src/models.ts#L42 and https://github.com/kazrael2119/azure-sdk-for-js/blob/9fb9ecb7ca9d051f602a65889a06c17af618d37a/sdk/servicenetworking/arm-servicenetworking/src/models/pagingTypes.ts#L55-L60
[...]

PR #2561, opened 2024-06-07
Add api version in ClientOptions and policy inside generated client

qiaozha:
fixes https://github.com/Azure/azure-sdk-for-js/issues/29883 and https://github.com/Azure/autorest.typescript/issues/2396#issuecomment-2147491682 in favor of https://github.com/Azure/azure-sdk-for-js/pull/29904

if there's no client level api version, we will:
1. not generate apiVersion in options
    which means we will not generate a property of apiVersion in the ServiceClientOptions interface.
3. remove ApiVersionPolicy 
    which means, we will generate code `client.pipeline.removePolicy({ name: "ApiVersionPolicy" });` after createClient from core-client-rest.
5. add log warning
    if there's no client level api version, we will add a warning 
   ` logger.warning("This client does not support client api-version, please change it at the operation level");`
   if there's client level api version, but it's required, and it doesn't have a default value, apiVersion will be a positional argument for client factory function. we will ignore api version passed in options, and add a warning 
   `logger.warning("This client does not support to set api-version in options, please change it at positional argument");`
    `ApiVersionPolicy` will also be removed in this case
7. not add client side ClientApiVersionPolicy  
   if there's no client level api version, we don't need to add this policy here, but if there's client level api version in the query position, according to https://github.com/Azure/azure-sdk-for-js/issues/29883#issuecomment-2150188495 we will generate a new `ClientApiVersionPolicy` in the client side, so that we can remove the `ApiVersionPolicy` from core-client-rest in the future.

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>

<div style='direction:ltr'>

  | hasDefault:   true  required: true | hasDefault: true   required: false | hasDefault: false   required: true | hasDefault: false   required: false
-- | -- | -- | -- | --
Path | Required: false  <br> Generate   ApiVersion in Options: true  <br> Remove   ApiVersionPolicy: true <br>  Add   clientApiVersionPolicy: false  <br> { apiVersion =   "default", …options } | Required: false  <br> Generate   ApiVersion in Options: true  <br> Remove   ApiVersionPolicy: true  <br> Add   clientApiVersionPolicy: false <br>  { apiVersion =   "default", …options } | Required: true  <br> Generate   ApiVersion in Options: false  <br> Remove   ApiVersionPolicy: true <br>  Add   clientApiVersionPolicy: false  <br> Options  <br> logWarning: true | Required: false  <br> Generate   apiVersion in options: true  <br> Remove   ApiVersionPolicy: true  <br> Add   clientApiVersionPolicy: false  <br> { apiVersion,   …options }
Query | Required: false  <br> Generate   ApiVersion in Options: true  <br> Remove   ApiVersionPolicy: true  <br> Add   clientApiVersionPolicy: true  <br> { apiVersion =   "default", …options } | Required: false  <br> Generate   ApiVersion in Options: true <br>  Remove   ApiVersionPolicy: true  <br> Add   clientApiVersionPolicy: true <br>  { apiVersion =   "default", …options } | Required: true  <br> Generate   ApiVersion in Options: false  <br> Remove   ApiVersionPolicy: true <br>  Add   clientApiVersionPolicy: true  <br> Options  <br> logWarning: true | Required: false <br>  Generate   apiVersion in options: true <br>  Remove   apiVersionPolicy: true  <br>  Add   clientApiVersionPolicy: true <br>  { apiVersion,   …options }

</div>

</body>

</html>


packages/rlc-common/src/buildClient.ts
```diff
@@ -22,21 +22,38 @@ import {
...
+      };
+    }) ?? [];
 
+  if (
+    model.apiVersionInfo &&
+    !properties.find((p) => p.name === "apiVersion")
+  ) {
+    properties.push({
+      name: "apiVersion",
+      type: "string",
```
qiaozha, line 52: we need to change this type to reflect the actual type.

packages/rlc-common/src/buildClient.ts
```diff
@@ -378,6 +385,21 @@ export function getClientFactoryBody(
...
 
+  let apiVersionStatement: string = "";
+  // Set the default api-version when we have a default AND its position is query
+  if (
+    model.apiVersionInfo?.definedPosition === "query" &&
+    !!model.apiVersionInfo?.defaultValue
+  ) {
+    apiVersionStatement = `
+    const apiVersion = options.apiVersion ?? "${model.apiVersionInfo?.defaultValue}";
+    delete options.apiVersion;
```
joheredi, line 396: I would prefer we don't mutate the options object to avoid side-effects. You can do something like this

```ts
   const {apiVersion, ...otherOptions}  = options;

  // replace the current use of options for `otherOptions` for example:

  const getClient = `const client = getClient(
        endpointUrl, ${credentialsOptions ? "credentials," : ""} otherOptions
      ) as ${clientTypeName};
```

joheredi: Let's try do this, but open to keep the delete if it turns out to be a big effort/risk

joheredi: I think another option is to spread at the createClient signature. For example

```ts
   export function createClient(endpoint: string, {apiVersion, ...options}?: ClientOptions)
```

If you do it this way you don't need to rename to `otherOptions` as the rest spread will contain everything but the properties before it.

packages/typespec-test/test/NetworkAnalytics.Management/generated/typespec-ts/src/rest/networkAnalyticsClient.ts
```diff
@@ -6,18 +6,25 @@ import { logger } from "../logger.js";
...
+
 /**
  * Initialize a new instance of `NetworkAnalyticsContext`
  * @param credentials - uniquely identify client credential
  * @param options - the parameter for all optional parameters
  */
 export default function createClient(
   credentials: TokenCredential,
-  options: ClientOptions = {},
+  options: NetworkAnalyticsContextOptions = {},
```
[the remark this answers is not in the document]
joheredi: With this you won't need to rename options -> otherOptions and don't need to modify all the places where options was used.

packages/rlc-common/src/buildClient.ts
joheredi: [Here is an example of why it is a good practice to avoid functions with side-effects ](https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgGIHt3IN4ChnLogQD8AXMgM5hSgDmANPsmAO7oXW0iPNgAWUCBE416TAjHQBXKKO51cAX1yqY0kAjDAiyKegAU6AA7ailChnQBKCgDd0wACY5mTiABsIkQqZ0hKADoiCGYEc3QvQI90OgMAAwART28UEzMA4OJkUEpnFDg9DS1-eOtlVVx1TQy9TAA5dABlfIBRGBgILQNsEIZkQMH0-0olS0xbZAdnVwJgGAMjP3NkOEpVkABPayyIa1mCFkF0VmRiU9aoKHQoAwAiYZXKfhkPJxAAcjBkfjg7FAEaWyxmuxmgYE2yAARl04NJKChgN9WGtkO5wgEaNItBAXKAjig8nQQHAwLIIHdygQVCpKhjKJEINFYvcALzsjmc5AAdQAkgAVAASVHyyAgHS6YHWnJlrMpqnp30eAWQrIOUlkFDuABY7hIjkIRMg7gJDXq+OwtQAmc0EEJawXQCAAQjuFVw9MZzLiyqCIXK+iWGUo5VwAHow8h+fwUO4vLVcqLAUUav4fqjCnl3GKJVpCNIpaKketKOEweHI76zlgYjxoOn1iFAshed9gOtCnRMC4QYhtEgWFg-o4XMUIJQEXAoJCsyhxZ0tJQPREojEfctMv6FSumWu2bKOTykS8CyLs-PJdKD+z5cvMb5g1bVeqZHJjbr9abhFqvxTP5bjRtfV7WNR0hFdd1PVXFlfStXYAwaZo2lzMAgxGK1QwjZBGjPOcUMofpYKoE83mhAp1lYWNxVAXEcjAO8GWg9dH3g3AgA)

joheredi (approved):
Looks great, left a comment on how to avoid that createClient causes side effects, should be easy to address. Please consider taking the suggestion

packages/rlc-common/src/buildClient.ts
qiaozha: Thanks for the suggestion, I will change it this way.

packages/typespec-test/test/authoring/generated/typespec-ts/src/authoringClient.ts
```diff
@@ -15,11 +19,14 @@ import { AuthoringClient } from "./clientDefinitions.js";
...
   credentials: KeyCredential,
-  options: ClientOptions = {},
+  options: AuthoringClientOptions = {},
 ): AuthoringClient {
   const endpointUrl =
     options.endpoint ?? options.baseUrl ?? `${endpointParam}/language`;
-  options.apiVersion = options.apiVersion ?? "2022-05-15-preview";
+
+  const apiVersion = options.apiVersion ?? "2022-05-15-preview";
+  delete options.apiVersion;
```
qiaozha, line 28:
```suggestion
  options.apiVersion = undefine;
```

packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/bodyComplexRestClient.ts
```diff
@@ -5,13 +5,19 @@ import { getClient, ClientOptions } from "@azure-rest/core-client";
...
 
+export interface BodyComplexRestClientOptions extends ClientOptions {
+  apiVersion?: string;
+}
+
 /**
  * Initialize a new instance of `BodyComplexRestClient`
- * @param options - the parameter for all optional parameters
+ * @param {
+ *     apiVersion = "2016-02-29", ...options} - the parameter for all optional parameters
```
joheredi, line 15: thb I would keep the same @param description we had. This shouldn't impact the public API. The fat that we are destructuring internally is an implementaion detail

joheredi (approved):

packages/autorest.typescript/test/smoke/generated/purview-administration-rest/review/purview-administration-rest.api.md
```diff
@@ -557,11 +557,15 @@ interface ComplexReplacerConfigOutput {
     typeName?: string;
 }
 
+// Warning: (ae-forgotten-export) The symbol "PurviewMetadataPoliciesClientOptions" needs to be exported by the entry point index.d.ts
+//
 // @public
-function createClient(endpoint: string, credentials: KeyCredential, options?: ClientOptions): PurviewMetadataPoliciesClient;
+function createClient(endpoint: string, credentials: KeyCredential, { apiVersion, ...options }?: PurviewMetadataPoliciesClientOptions): PurviewMetadataPoliciesClient;
 
+// Warning: (ae-forgotten-export) The symbol "PurviewAccountClientOptions" needs to be exported by the entry point index.d.ts
```
xirzec, line 565: looks like when we generate the export here, we're only exporting `createClient`:
 ```
export { createClient } from "./purviewAccount";
```

File: packages/rlc-common/src/buildClient.ts [hunk 7]
```diff
@@ -408,15 +437,42 @@ export function getClientFactoryBody(
...
+  if (
+    model.apiVersionInfo?.isCrossedVersion === false &&
+    model.apiVersionInfo?.definedPosition === "query"
+  ) {
+    apiVersionPolicyStatement += `
+      client.pipeline.addPolicy({
+        name: 'ClientApiVersionPolicy',
+        sendRequest: (req, next) => {
+          // Use the apiVesion defined in request url directly
+          // Append one if there is no apiVesion and we have one at client options
```
xirzec, line 465: spelling nit 
```suggestion
          // Use the apiVersion defined in request url directly
          // Append one if there is no apiVersion and we have one at client options
```

xirzec (approved):
Looks good, just two minor remarks. 👍

packages/autorest.typescript/test/smoke/generated/purview-administration-rest/review/purview-administration-rest.api.md
qiaozha: good catch!

packages/autorest.typescript/test/rlcIntegration/generated/bodyComplexRest/src/bodyComplexRestClient.ts
qiaozha: updated.

packages/rlc-common/src/buildClient.ts
qiaozha: should have been resolved.

packages/autorest.typescript/test/smoke/generated/purview-administration-rest/review/purview-administration-rest.api.md
qiaozha: should have been resolved.

packages/autorest.typescript/test/smoke/generated/purview-administration-rest/review/purview-administration-rest.api.md
```diff
@@ -1347,6 +1355,12 @@ type PurviewMetadataPoliciesClient = Client & {
     path: Routes;
 };
 
+// @public (undocumented)
```
qiaozha, line 1358: probably should add document here.

PR #2563, opened 2024-06-11
Add initial serialization refactor through test flag

dgetu:
This PR includes initial work on refactoring the serialization implementation of `typespec-ts` to be a function of Modular clients and TCGC models. Incidentally, this also implements handling of recursive types and extra properties. Some tests have had their inline specs pulled into files for easier review/updating. It also introduces a dependency on `lodash`, which is already pulled in transitively through `rlc-common`. Changes should be easier to reason about when reviewing each commit individually.

Minimum requirements to expose this functionality involves populating the dynamic dispatch table in `serializeUtil` with each TCGC model type's serialization emitter callback and implementing unit tests for serialization scenarios.


joheredi (approved):
This is looking really nice. Thanks for the additional cleanup. Happy to see this also brings us a step closer to fully leveraging TCGC

packages/typespec-ts/test/util/emitUtil.ts
```diff
@@ -440,7 +442,14 @@ export async function emitModularOperationsFromTypeSpec(
         modularCodeModel.clients[0],
         dpgContext,
         modularCodeModel,
-        false
+        false,
+        env["EXPERIMENTAL_TYPESPEC_TS_SERIALIZATION"]
```
qiaozha, line 446: do we have to set the env EXPERIMENTAL_TYPESPEC_TS_SERIALIZATION for the serialize utils unit tests? curious if those tests are really testing something now?

packages/typespec-ts/src/modular/serialization/serializers.ts
```diff
@@ -0,0 +1,270 @@
...
+      >
+    ) => SerializerOutput;
+  };
+  type SerializeHandlerMap = Partial<SerializeHandler<SimpleType>> &
+    SerializeHandler<
+      Exclude<(SdkType | SdkModelPropertyType)["kind"], SimpleType>
+    >;
+
+  const handlers: SerializeHandlerMap = {
+    array: placeholder,
```
qiaozha, line 114: just wonder if we could file issues or have a way to track all those places that we already know it's going to be changed? I notice you added TODO comments in other places.

packages/typespec-ts/test/util/emitUtil.ts
dgetu: The environment does have to be set for unit tests to consume the new code. We have expected test failures in this PR, so as tests start to pass we can look into selectively enabling it for those tests.

packages/typespec-ts/src/modular/serialization/serializers.ts
dgetu: For this in particular, all of the missing functions are contained in this file at the moment. I'll open issues to track remaining work.

PR #2569, opened 2024-06-11
Add unit tests for serializer

joheredi:
Keeping these tests under test-next since they are true unit test. Will consolidate later when re-organizing the tests. Using vitest for new unit tests since it has better support for ESModules


dgetu (approved):
:shipit:

Issue #2354, opened 2024-03-13
Test code and config is not working in sdk repo for modular


MaryGao:
Failed to run test cases with https://github.com/Azure/azure-sdk-for-js/pull/29770
background: have logged with az login

error message:
```
test/public/networkanalytics_operations_test.spec.ts > NetworkAnalytics test > operation list test
[vitest]      → ChainedTokenCredential authentication failed.
[vitest] CredentialUnavailableError: EnvironmentCredential is unavailable. No underlying credential could be used. To troubleshoot, visit https://aka.ms/azsdk/js/identity/environmentcredential/troubleshoot.
[vitest] CredentialUnavailableError: WorkloadIdentityCredential: is unavailable. tenantId, clientId, and federatedTokenFilePath are required parameters.
[vitest]       In DefaultAzureCredential and ManagedIdentityCredential, these can be provided as environment variables -
[vitest]       "AZURE_TENANT_ID",
[vitest]       "AZURE_CLIENT_ID",
[vitest]       "AZURE_FEDERATED_TOKEN_FILE". See the troubleshooting guide for more information: https://aka.ms/azsdk/js/identity/workloadidentitycredential/troubleshoot
[vitest] CredentialUnavailableError: ManagedIdentityCredential: The managed identity endpoint is indicating there's no available identity. Message: unknown Status code: 400
[vitest] More details:
[vitest] Failed to retrieve token from the Azure Active Directory: MSI cert expired: currentTime=2024-05-24 07:01:08.4023368 +0000 UTC, notAfter=2024-02-26 08:37:00 +0000 UTC
[vitest] CredentialUnavailableError: Azure CLI credential expects a 'SystemRoot' environment variable
[vitest] CredentialUnavailableError: The 'Az.Account' module >= 2.2.0 is not installed. Install the Azure Az PowerShell module with: "Install-Module -Name Az -Scope CurrentUser -Repository PSGallery -Force".
[vitest] CredentialUnavailableError: Azure Developer CLI credential expects a 'SystemRoot' environment variable
```

steps:
1. checkout https://github.com/Azure/azure-sdk-for-js/pull/29770
2. run ```rush update && rush build -t @azure/arm-networkanalytics```
3. run ```cd sdk/networkanalytics/arm-networkanalytics```
4. run ```export TEST_MODE=record && rushx test```

joheredi:
@MaryGao can you please provide more details on what is broken?

MaryGao:
[...]

PR #2570, opened 2024-06-11
Bump identity dependency version to ^4.2.1 in generated packages

jeremymeng:


MaryGao (commented):
Thanks for this update! LGTM but we met an identity issue with v4.2.1 and not sure this is the latest version or general one: https://github.com/Azure/autorest.typescript/issues/2354#issuecomment-2162274194.

I will give my approval once I verify it.

MaryGao (approved):
LGTM

PR #2575, opened 2024-06-13
Add array serializer and test

joheredi:
Moving serialize array to its own file and add tests

### DIFF
--- a/packages/typespec-ts/test-next/integration/serialization/serializeArray.test.ts
+++ b/packages/typespec-ts/test-next/integration/serialization/serializeArray.test.ts
@@ -1,0 +1,318 @@
+import { beforeEach, describe, it, expect } from "vitest";
+import { createMyTestRunner, createTcgcContext } from "./test-hots.js";
+import { serializeArray } from "../../../src/modular/serialization/serializeArray.js";
+import {
+  SdkArrayType,
+  SdkContext
+} from "@azure-tools/typespec-client-generator-core";
+import { EmitContext, UsageFlags } from "@typespec/compiler";
+import { BasicTestRunner } from "@typespec/compiler/testing";
+import { SerializeTypeOptions } from "../../../src/modular/serialization/serializers.js";
+import { format } from "prettier";
+
+describe("serializeArray", () => {
+  let runner: BasicTestRunner;
+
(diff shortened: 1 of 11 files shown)


packages/typespec-ts/test-next/integration/serialization/serializeArray.test.ts
```diff
@@ -0,0 +1,318 @@
...
+
+      expect(result).to.equal("myArray.map(Date)");
+    });
+
+    it("should handle a model array", async () => {
+      const { modelArray } = testArrays;
+      const result = serializeArray(modelArray);
+
+      const expected = await format(
+        `myArray.map((e: Foo)=>({...(e),
```
joheredi, line 144: Serializer and deserializer will have (e: Foo) however one refers to the Modular Foo and the other to the RLC foo. @dgetu are we handling this already or this needs a fix

joheredi: I was not setting up the serializerMap that's why I was getting the same name for RLC and Modular references

packages/typespec-ts/test-next/integration/serialization/serializeArray.test.ts
```diff
@@ -0,0 +1,460 @@
...
+      const actual = await format(result, { parser: "typescript" });
+
+      expect(actual).to.equal(expected);
+    });
+
+    it("should handle a model array ", async () => {
+      const { modelArray } = testArrays;
+      const result = serializeArray(modelArray);
+
+      const expected = await format(`myArray.map((e) => serializeFoo(e));;`, {
```
dgetu, line 79: Will the formatter catch this duplicated semicolon? Or is that expected?

dgetu (approved):
Overall looks good! I'm not too familiar with the header requirements, so I'll follow up if we run into validation issues.

packages/typespec-ts/test-next/integration/serialization/serializeArray.test.ts
joheredi: Yeah it is a typo on my side, formatter will remove it. I'll fix it in the next commit

Issue #2572, opened 2024-06-12
[TS] [DPG] BearerAuth generates lower case "bearer <key>"


glharper:
The BearerAuth keyword in TypeSpec seems to generate the following client code in TS:
```
  if (isKeyCredential(credentials)) {
    client.pipeline.addPolicy({
      name: "customKeyCredentialPolicy",
      async sendRequest(request, next) {
        request.headers.set("Authorization", "bearer " + credentials.key);
        return next(request);
      },
    });
  }
  ```
  That fifth line should be (note capitalized Bearer): ``` 
        request.headers.set("Authorization", "Bearer " + credentials.key);
        ```

timovv:
Related: https://github.com/Azure/autorest.typescript/issues/2109

Issue #2109, opened 2023-11-09
Bearer auth scheme is lowercase in request header


timovv:
When making a request to the OpenAI API using the unbranded SDK with this code, authorization fails:

```ts
const client = new OpenAIClient({ key: process.env.OPENAI_API_KEY ?? "" });

try {
  const result = await client.chat.completions.create({
    model: "gpt-3.5-turbo",
    messages: [
      {
        role: "user",
        content: "Say this is a test",
      },
    ],
    temperature: 0.7,
  });
  console.log(result);
} catch (e) {
  console.error(e);
}
```

### Expected behavior

With a valid API key, authorization succeeds because the `Authorization` header has value `Bearer <access-token>`.

### Actual behavior

Authorization fails due to the word "bearer" being lower case in the request: `Authorization: bearer <access-token>`.

timovv:
This can be worked around in the TypeSpec by using

```
@useAuth({
  type: AuthType.http,
  scheme: "Bearer",
})
```

instead of `@useAuth(BearerAuth)`, but we should support `BearerAuth` anyway.

MaryGao:
I notice that this is introduced by `BearerAuth` [definition in typespec](https://github.com/microsoft/typespec/blob/main/packages/http/lib/auth.tsp#L44-L50).  I notice that [the RFC](https://www.rfc-editor.org/rfc/rfc6750#section-6.1.1) didn't mention this should be case sensitive or not but typically it would be `Bearer`.

```
@doc("")
model BearerAuth {
  @doc("Http authentication")
  type: AuthType.http;

  @doc("bearer auth scheme")
  scheme: "bearer";
}
```

archerzz:
According to the [RFC](https://www.rfc-editor.org/rfc/rfc6750#section-1.1):
> Unless otherwise noted, all the protocol parameter names and values
   are case sensitive.

We should fix in tsp.

timotheeguerin:
The value in openapi also is lower case so I think this should be the emitter understanding the known values
https://swagger.io/docs/specification/authentication/bearer-authentication/

Swagger ui understand all cases and sends `Bearer`

We have just kept the authentication inline with openapi3. We should discuss if we want to change and or clarify the intent of that value it but this is not a bug in typespec http right now.

archerzz:
Looks like `scheme` in openapi is always lower case, but the real header value sent is not. Here is another example [Basic authentication](https://swagger.io/docs/specification/authentication/basic-authentication/).
[...]

PR #2576, opened 2024-06-13
Update `bearer` to `Bearer`

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2572 
fixes https://github.com/Azure/autorest.typescript/issues/2109

### DIFF
--- a/packages/typespec-ts/src/transform/transfromRLCOptions.ts
+++ b/packages/typespec-ts/src/transform/transfromRLCOptions.ts
@@ -1,5 +1,5 @@
 import {
-  camelCase,
+  pascalCase,
   NameType,
   normalizeName,
   PackageDetails,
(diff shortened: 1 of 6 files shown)


packages/typespec-ts/src/transform/transfromRLCOptions.ts
```diff
@@ -106,7 +107,13 @@ function processAuth(program: Program) {
         case "http":
           securityInfo.addCredentials = true;
           securityInfo.customHttpAuthHeaderName = "Authorization";
-          securityInfo.customHttpAuthSharedKeyPrefix = auth.scheme;
+          // If it is basic or bearer auth we should generate it as Basic or Bearer
+          securityInfo.customHttpAuthSharedKeyPrefix = [
+            "basic",
```
qiaozha, line 112: where do we set Basic now?

qiaozha (approved):
LGTM

packages/typespec-ts/src/transform/transfromRLCOptions.ts
MaryGao: i believe we don't have this case yet.

Issue #2314, opened 2024-02-27
Support patch lro in modular


MaryGao:
Previously there is some issues with getLroMetadata for PATCH method from TypeSpec. In the latest compiler version these issues are resolved and we need to remove these hard-codes.

deyaaeldeen:
Could you explain what is the issue about, with an example perhaps?

PR #2583, opened 2024-06-14
Remove the hard-code for patch LRO

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2314

Previously there is some issues with getLroMetadata for PATCH method from TypeSpec. In the latest compiler version these issues are resolved and we need to remove these hard-codes.


packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/testRunOperations/api/operations.ts
```diff
@@ -1495,19 +1905,13 @@ export function _stopTestRunSend(
   context: Client,
   testRunId: string,
   options: StopTestRunOptionalParams = { requestOptions: {} },
-): StreamableMethod<
-  LoadTestRunStopTestRun200Response | LoadTestRunStopTestRunDefaultResponse
-> {
+): StreamableMethod<undefined> {
   return context
-    .path("/test-runs/{testRunId}:stop", testRunId)
+    .path("/test-runs/{testRunId}:stopTestRun", testRunId)
```
MaryGao, line 1910: The referred path in RLC and modular is different and we could reproduce this in [playground](https://cadlplayground.z22.web.core.windows.net/cadl-azure/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdNUZdmVyc2lvbmluZ8wfYXp1cmUtdG9vbHMvyCstxhVjb3Jl3yvIK3Jlc291cmNlLW1hbmFnZXLfN2MtY2xpZW50LWdlbmVyYXRvcshtCnVzaW5nIEHENC5Db3JlO9ESLlRyYWl0c8gZVHlwZVNwZWMuUmVzdNEVSHR0cNEVVukBCsgbT3BlbkFQSc9t5QCkR%2BgAo8V9OwoKLy8gLcQBIEhlcmUgaXMgbWFpbi50c3DFFwpAc2VydmljZSh7CiAgdGl0bGU6ICJDb250b3NvIFdpZGdldCBN5wEmLAp9KQpA5wGZZWQoxycuxifHJugAr3MpCm5hbWVzcGFjZSDVKuQAnSoqIFRoySDPcSDnAJwgx3MuICovCmVudW0gyGQg5AC2xEXHESAyMDIyLTA4LTMxxCwgIEB1c2VEZXBlbmRlbmN5KOsBl8hDLnYxXzBfUHJldmlld18yKQogIGDJRjBgLAp9CgpAZG9jKCJGaWxlIGluZm8iKQpA6AJDKCJmaWxlcyIpCm1vZGVsIMQlSW5mb%2BUAmsY4TmFtZSBvZiB0aGUgxC0uIsRlQGtlecQHdmlzaWJpbGl0xA7EH8QvOiBzdHLlAfF9CgppbnRlcmbkAUdGb8xYU3RvcCB0ZXN0IHJ1biBiecoMSWTHZG9w5AINaW9uSWQoIkxvYWRUZXN0UnVuX8Q8xwzFKnPETGlzIFN0YW5kYXJkUucA30%2FIQXPkArHFE0FjxBA86ADpLCB7fSzpAPfEDj7lALbxAnhjdXN0b21pesVPIGluIOYDVesCi8YR5QKK5AI7OiDsALPmAt8iLOQAreYCFzr2AlflAqbqATTHP%2BoAyeUBQuQA8Mca5ACcRm9vLsQT5AC6&e=%40azure-tools%2Ftypespec-autorest&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40azure-tools%2Ftypespec-azure-rulesets%2Fresource-manager%22%5D%7D%7D).

packages/typespec-test/test/widget_dpg/spec/main.tsp
```diff
@@ -176,8 +176,4 @@ interface Budgets {
     // the same name as Widgets.createOrReplace
     @route("/widgets/createOrReplace")
     op createOrReplace is ResourceOperations.LongRunningResourceCreateOrReplace<User>;
-
-    // the patch case for lro
-    @route("/widgets/createOrUpdate")
```
MaryGao, line 181: Patch LRO in MPG is okay but there exists issue in getLroMetadata for data-plane tamplate. But we don't have a real case for PATCH LRO in data-plane yet.

So I tentatively remove this operation. Let's fix this when we have a real case.

packages/typespec-test/test/loadtesting_modular/tspconfig.yaml
```diff
@@ -12,6 +12,7 @@ options:
     enableOperationGroup: true
     isModularLibrary: true
     hierarchyClient: false
+    experimentalExtensibleEnums: true
```
MaryGao, line 15: The extensible union in record has type mismatch issue which is not resolved yet.

I tentatively enable this option to by pass this issue.

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/package.json
```diff
@@ -51,8 +57,6 @@
     "@azure/core-rest-pipeline": "^1.5.0",
     "@azure/logger": "^1.0.0",
     "tslib": "^2.6.2",
-    "@azure/core-lro": "3.0.0",
```
qiaozha, line 54: why is this being removed? Have they removed lro operation ?

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/testRunOperations/testRunOperationsClient.ts
```diff
@@ -150,32 +150,48 @@ export class TestRunOperationsClient {
...
     fileName: string,
     options: GetTestRunFileOptionalParams = { requestOptions: {} },
-  ): Promise<FileInfo> {
+  ): Promise<{
+    url?: string;
+    fileType?: FileType;
+    expireDateTime?: Date;
+    validationStatus?: FileStatus;
+    validationFailureDetails?: string;
+  }> {
```
qiaozha, line 159: Is this expected?

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/testRunOperations/testRunOperationsClient.ts
```diff
@@ -150,32 +150,48 @@ export class TestRunOperationsClient {
...
   /** List the dimension values for the given metric dimension name. */
   listMetricDimensionValues(
     testRunId: string,
     name: string,
+    metricname: string,
     metricNamespace: string,
+    timespan: string,
     options: ListMetricDimensionValuesOptionalParams = { requestOptions: {} },
-  ): PagedAsyncIterableIterator<DimensionValueList> {
+  ): Promise<DimensionValueList> {
```
qiaozha, line 171: Is changing from pagable to non-pagable a breaking? Also why would they have this change, it should be common to see changing from non-pagable to pagable, but should be uncommon in the reverse case?

packages/typespec-test/test/loadtesting_modular/tspconfig.yaml
qiaozha: it should generate as any if it's not enabled, did you see any errors there?

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/testRunOperations/testRunOperationsClient.ts
MaryGao: @weidongxu-microsoft Not sure you have more context. In the latest loadtesting spec,
- there is no LRO now
- listMetricDimensionValues is changed from ResourceList to Operation but becoming non-paging

weidongxu-microsoft: The LRO in Java is handwritten
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/loadtesting/azure-developer-loadtesting/src/main/java/com/azure/developer/loadtesting/LoadTestRunAsyncClient.java#L127-L146

(from what I remember, the status etc. does not match Azure guideline and SDK core lib handling of LRO -- that is the api-version, this version maybe different)

packages/typespec-test/test/loadtesting_modular/tspconfig.yaml
MaryGao: yeah, the case is Record of extensible enum. The modular is built as A | B but the RLC is built as string. So directly assigning would cause compile error.

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/package.json
MaryGao: I believe so, in old spec they defined as customized lro but finally removed out. I checked with service team with below message.

image

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/testRunOperations/testRunOperationsClient.ts
MaryGao: i don't think so, let me investigate there.

MaryGao: should be fixed here: https://github.com/Azure/azure-rest-api-specs/pull/29521

packages/typespec-test/test/loadtesting_modular/generated/typespec-ts/src/testRunOperations/testRunOperationsClient.ts
MaryGao: The LRO change is expected and the paging change is waiting for the service team's reply, but I think we don't need to block this pr. Let me offline confirm this.

qiaozha (approved):

Issue #2476, opened 2024-04-24
Known* definitions mission in TS Generation


sarangan12:
For the following Typespec definition:

```
/** Acs Router Job Status */
union AcsRouterJobStatus {
  /** Router Job Status Pending Classification */
  "PendingClassification",
  /** Router Job Status Queued */
  "Queued",
  /** Router Job Status Assigned */
  "Assigned",
  /** Router Job Status Completed */
  "Completed",
  /** Router Job Status Closed */
  "Closed",
  /** Router Job Status Cancelled */
  "Cancelled",
  /** Router Job Status Classification Failed */
  "ClassificationFailed",
  /** Router Job Status Created */
  "Created",
  /** Router Job Status Pending Schedule */
  "PendingSchedule",
  /** Router Job Status Scheduled */
  "Scheduled",
  /** Router Job Status Schedule Failed */
  "ScheduleFailed",
  /** Router Job Status Waiting For Activation */
  "WaitingForActivation",
  string,
}
```

Now, If you look at the generated SDK, it looks like the following:

```js
// @public
export type AcsRouterJobStatus = string;
```

But, If you look at the similar swagger generated SDK, you can find a definition as:

```js
/** Known values of {@link AcsRouterJobStatus} that the service accepts. */
export const enum KnownAcsRouterJobStatus {
  PendingClassification = "PendingClassification",
  Queued = "Queued",
  Assigned = "Assigned",
  Completed = "Completed",
  Closed = "Closed",
  Cancelled = "Cancelled",
  ClassificationFailed = "ClassificationFailed",
  Created = "Created",
  PendingSchedule = "PendingSchedule",
  Scheduled = "Scheduled",
  ScheduleFailed = "ScheduleFailed",
  WaitingForActivation = "WaitingForActivation"
}

/**
 * Defines values for AcsRouterJobStatus. \
 * {@link KnownAcsRouterJobStatus} can be used interchangeably with AcsRouterJobStatus,
 *  this enum contains the known values that the service supports.
 * ### Known values supported by the service
 * **PendingClassification** \
 * **Queued** \
 * **Assigned** \
 * **Completed** \
 * **Closed** \
 * **Cancelled** \
 * **ClassificationFailed** \
 * **Created** \
 * **PendingSchedule** \
 * **Scheduled** \
 * **ScheduleFailed** \
 * **WaitingForActivation**
 */
export type AcsRouterJobStatus = string;
```

[...]

Issue #2420, opened 2024-04-05
[JS] Redesign Extensible Enum generation


joheredi:
- [x] Finalize Re-design
- [ ] Implementation

# Design
Based on design conversations this is the design for Extensible Enum code gen for the TypeScript Emitter
### Scenario 1: Exhaustive Forever
**TypeSpec**
```ts
union Foo {
   "a",
   "b"
} 
```

**Desired Output**
```ts
export type Foo = "a" | "b"; 
```

#### Remarks:
Represents a closed union of string literals with no expected additional variants, ensuring predictability and type safety.
Direct mapping of union members to a TypeScript union type simplifies understanding and prevents unexpected values.

### Scenario 2:  Eustive for This Service Version

```ts
union Foo {
   "a",
   "b",
   string
} 
```
**Note:** This scenario is under review in TypeSpec side so the above TypeSpec snippet may change, the desired output remains the same.

**Desired Output**
```ts
// Generated
export type Foo = "a" | "b";
declare async function createFoo(foo: Foo): Promise<void>;
declare async getFoo(): Promise<Foo>;  

// Example usage
async function exampleUsage() {
  // Handling unknown string inputs and outputs
  await createFoo("x" as any);  // Casting to `any` when adding unknown variants
  const foo = await getFoo();
  if (foo === "a") {
    console.log("It is a");
  } else if (foo === "b") {
    console.log("It is b");
  } else {
    console.log(`It is ${foo}`);  // Handling unknown case
  }
}
```

#### Remarks

- Accommodates additional string variants that may be introduced in future versions.
- Developers must explicitly handle unknown values to maintain robustness and prevent unexpected behavior at runtime.
- Emphasizes the importance of documentation and examples to guide developers in managing outputs from APIs.

### Scenario 3: Non-exhaustive, Has Unknown Values in the Same API

```ts
@withUnknownValues
union Foo {
  "a",
  "b",
  string
} 
```
**Note:** This scenario is under review in TypeSpec side so the above TypeSpec snippet may change, the desired output remains the same.

**Desired Output**
```ts
// Generated
export enum KnownFooValues {
  a = "a",
  b = "b"
}

export type Foo = string;
declare function createFoo(foo: Foo): void; // Assume this function is implemented elsewhere
declare function getFoo(): Promise<Foo>;  // Assume this function is implemented elsewhere


// Example usage
async function exampleUsage() {
  createFoo(KnownFooValues.a);  // Using enum for known values
[...]

PR #2584, opened 2024-06-14
support-non-exhaustive-enum

qiaozha:
fixes https://github.com/Azure/autorest.typescript/issues/2476
fixes https://github.com/Azure/autorest.typescript/issues/2420


packages/typespec-ts/test/integration/generated/type/union/src/responses.ts
```diff
@@ -26,7 +26,7 @@ export interface StringsOnlySend204Response extends HttpResponse {
 /** The request has succeeded. */
 export interface StringExtensibleGet200Response extends HttpResponse {
   status: "200";
-  body: { prop: string | "b" | "c" };
+  body: { prop: "b" | "c" };
```
qiaozha, line 29: @joheredi for anonymous non exhaustive enum, what do you think should be our behavior here? I feel like either keep the current behavior or just generate the string ?

joheredi: I think we should scope these changes to Modular only. I would prefer if we keep RLC behavior as it. One thing we may want to update in RLC is to generate `string` instead of `"a" | "b" | string` but not a high priority item

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -150,6 +150,10 @@ export function getDeserializePrivateFunction(
     parameters,
     returnType: `Promise<${returnType.type}>`
   };
+
+  if (returnType.type === "DogUnion") {
+    returnType;
+  }
```
qiaozha, line 156:
```suggestion
```

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -1117,6 +1124,9 @@ export function deserializeResponseValue(
   typeStack: Type[] = [],
   format?: string
 ): string {
+  if (type.name === "DogUnion") {
+    type;
+  }
```
qiaozha, line 1129:
```suggestion
```

packages/typespec-ts/test/modularIntegration/extensibleEnums.spec.ts
```diff
@@ -41,7 +41,7 @@ describe("ExtensibleEnums Rest Client", () => {
 
   it("should put unknown value and receives 500", async () => {
     try {
-      const result = await client.string.putUnknownValue("Weekend", {
+      const result = await client.string.putUnknownValue("Weekend" as any, {
```
qiaozha, line 44: In this case, we didn't enable the experimental extensible enums feature. so we will generate literal unions for this case, and unknown value will need to be casted.

packages/typespec-ts/test/modularIntegration/generated/type/enum/extensible/src/api/string/index.ts
```diff
@@ -37,7 +37,7 @@ export async function _getKnownValueDeserialize(
     throw createRestError(result);
   }
 
-  return result.body;
+  return result.body as DaysOfWeekExtensibleEnum;
```
qiaozha, line 40: in this case, it's exaustive enum and we didn't enable the experimental feature, RLC layer is generated as string and Modular layer is generated as literal unions. we have to use the cast here to make compiler happy.

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -1191,6 +1201,25 @@ export function deserializeResponseValue(
...
+        );
+        if (
+          discriminatorProp?.length === 1 &&
+          discriminatorProp[0]?.type.isFixed === false &&
+          discriminatorProp[0].type.isNonExhaustive === false
+        ) {
+          return `${restValue} as ${type.name}`;
+        }
+      }
+      return restValue;
```
qiaozha, line 1222: @joheredi @dgetu this is a new logic related with the deserialization caused by we are not treating exhaustive enum the same in both RLC and Modular layer.

packages/typespec-ts/test/integration/generated/type/union/src/responses.ts
qiaozha: Updated to string.

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
joheredi: Thanks! We'll use this as a guide when porting to the new serializers

joheredi (approved):

Issue #1981, opened 2023-08-18
Split RLC from Modular


qiaozha:

lirenhe:
@qiaozha,  could you add some description for this issue about what the current behavior and the expected output?
If there are doc about the design, feel free to link it there as well.  thanks.

deyaaeldeen:
Would the routes still be generated and exported?

qiaozha:
If you mean in RLC , then yeah.  If you mean in Modular, no, there will be no RLC code in Modular generated code once this task is done.

deyaaeldeen:
But I thought it is valuable for customers to use the RLC companion of the modular client for scenarios where unknown fields in the request or the response are needed. For example, see https://github.com/Azure/azure-sdk-for-js/issues/26579#issuecomment-1687141541

qiaozha:
we will release a public version of RLC for every modular package. so in your example, you could just use @azure-rest/openai when the public RLC of openai is available. 

FYI, your example will only work for ECMAScript customers not for CommonJS customers, that's the main reason we need to split the RLC from Modular.

PR #2616, opened 2024-06-25
Set user agent at different layers for modular

timovv:
~~Update user agent to change depending on which layer is being used (TODO: update these to add -modular)~~
- ~~If the classic client layer is used, `azsdk-js-<packageName>-classic/<version>`~~
- ~~If the modular API layer is used, `azsdk-js-<packageName>-api/<version>`~~
- ~~REST level remains unchanged `azsdk-js-<packagename>-rest/<version>`~~

~~The `azsdk-js-` prefix is omitted for unflavored packages.~~

~~Problem: RLC currently concatenates the passed in `userAgentPrefix` so we may get long user agents like `azsdk-js-<packageName>-api/<version> azsdk-js-<packagename>-rest/<version>`~~

Implement user agent string as defined in [this Loop](https://loop.cloud.microsoft/p/eyJ3Ijp7InUiOiJodHRwczovL21pY3Jvc29mdC5zaGFyZXBvaW50LmNvbS8%2FbmF2PWN6MGxNa1ltWkQxaUlWcElZeko0WVc1ZloxVjFUWE4wUkU5RFIya3pTakkyUkVOUFoxWk9SVVpNYlV3NE9GRkRiR1ZLY0dwSldHbzRjMDFwUlVoVVdWUXRObTAwUmxkaVpEVW1aajB3TVVSSFJqSlBSVFJVUlUxUk5EZFNSa0UzV2tVeVQwWk9RMFEwVmxnMk5ESkZKbU05Sm1ac2RXbGtQVEUlM0QiLCJyIjpmYWxzZX0sInAiOnsidSI6Imh0dHBzOi8vbWljcm9zb2Z0LnNoYXJlcG9pbnQuY29tLzpmbDovci9zaXRlcy82MTU1NGI1Ni1jOTBiLTQ5NjEtYmE3Yy02MWNhYWUyMzUzMzUvU2hhcmVkJTIwRG9jdW1lbnRzL0xvb3BBcHBEYXRhL1VudGl0bGVkLmxvb3A%2FZD13MDI0NjA0ZmMzZmMxNDYzZDlhM2I1YmI2YzE2MWE1OTkmY3NmPTEmd2ViPTEmbmF2PWN6MGxNa1p6YVhSbGN5VXlSall4TlRVMFlqVTJMV001TUdJdE5EazJNUzFpWVRkakxUWXhZMkZoWlRJek5UTXpOU1prUFdJaFdraGpNbmhoYmw5blZYVk5jM1JFVDBOSGFUTktNalpFUTA5blZrNUZSa3h0VERnNFVVTnNaVXB3YWtsWWFqaHpUV2xGU0ZSWlZDMDJiVFJHVjJKa05TWm1QVEF4UkVkR01rOUZOelJCVWtSQlJsRktOMGhXUkVwVlR6SXpWek5CVjBSS1RWb21ZejBsTWtZbVpteDFhV1E5TVNaaFBVeHZiM0JCY0hBbWNEMGxOREJtYkhWcFpIZ2xNa1pzYjI5d0xYQmhaMlV0WTI5dWRHRnBibVZ5Sm5nOUpUZENKVEl5ZHlVeU1pVXpRU1V5TWxRd1VsUlZTSGgwWVZkT2VXSXpUblphYmxGMVl6Sm9hR050Vm5kaU1teDFaRU0xYW1JeU1UaFphVVpoVTBkTmVXVkhSblZZTW1SV1pGVXhlbVJGVWxCUk1HUndUVEJ2ZVU1clVrUlVNbVJYVkd0V1IxUkhNVTFQUkdoU1VUSjRiRk51UW5GVFZtaHhUMGhPVG1GVlZrbFdSbXhWVEZSYWRFNUZXbGhaYlZFeFprUkJlRkpGWkVkTmF6bEdUa1pTUmxSV1JUQk9NVXBIVVZSa1lWSlVTbEJTYXpWRVVrUlNWMWRFV1RCTmExVWxNMFFsTWpJbE1rTWxNakpwSlRJeUpUTkJKVEl5Wm1Ka05tTXdNakV0WVRjeVppMDBNVGs1TFRnMFlXVXRPR1UyT1daa05qUmtaREUwSlRJeUpUZEUiLCJyIjpmYWxzZX0sImkiOnsiaSI6ImZiZDZjMDIxLWE3MmYtNDE5OS04NGFlLThlNjlmZDY0ZGQxNCJ9fQ%3D%3D):

> For a package @azure/my-library version 1.0.0: 
> - Requests from the REST layer will have user agent prefix azsdk-js-my-library/1.0.0 
> - Requests from the API layer will have user agent prefix azsdk-js-api azsdk-js-my-library/1.0.0 
> - Requests from the client layer will have user agent prefix azsdk-js-client azsdk-js-api azsdk-js-my-library/1.0.0


qiaozha:
I think we are missing a modular in the middle ?
Classical client layer: azsdk-js-arm-compute-modular-classic/21.0.1
Api layer:azsdk-js-arm-compute-modular-api/21.0.1
Rest layer: azsdk-js-arm-compute-modular-rest/21.0.1

timovv:
I had a discussion with Jose about this and we settled on the following format for the user agent:

- Modular, initiated at the REST level: `azsdk-js-<packageName>-modular-rest/<version>`
- Initiated at API level: `azsdk-js-modular-api azsdk-js-<packageName>-modular-rest/<version>`
- Initiated at "classic" level: `azsdk-js-modular-classic azsdk-js-modular-api azsdk-js-<packageName>-modular-rest/<version>`

We need to prepend at each layer here since the REST layer __concatenates__ the passed in `userAgentPrefix` to its value (the prefix can't be overridden). Changing this behavior would make it possible for customers to override the user agent value entirely which I believe we want to avoid. So instead of substituting the value based on which layer is being used, we prepend an extra value for each layer instead.

@qiaozha, @MaryGao, any issues with the above proposal? One remaining question I have: for the "classic" level, do we want to call this "classic" or "client"? No strong opinion here.

packages/typespec-test/test/NetworkAnalytics.Management/generated/typespec-ts/src/networkAnalyticsClient.ts
```diff
@@ -37,7 +37,15 @@ export class NetworkAnalyticsClient {
     subscriptionId: string,
     options: NetworkAnalyticsClientOptions = {},
   ) {
-    this._client = createNetworkAnalytics(credential, options);
+    const prefixFromOptions = options?.userAgentOptions?.userAgentPrefix;
+    const userAgentPrefix = prefixFromOptions
+      ? `${prefixFromOptions} azsdk-js-modular-classic`
```
joheredi, line 42: can we name this `classical`, the name classical is after the client being a class :P (as in classical vs prototype inheritance) not so much for it being a classic. I think we'd want to make sure we use classical everywhere of course on a different PR

MaryGao:
> @qiaozha, @MaryGao, any issues with the above proposal? One remaining question I have: for the "classic" level, do we want to call this "classic" or "client"? No strong opinion here.

Before digging into detailed patterns I may ask if we still plan to split the RLC from Modular https://github.com/Azure/autorest.typescript/issues/1981. If yes, I don't think we have the ability to distingush if a RLC is initiated from Modular or normal usage. In that way the RLC would be always like `azsdk-js-<packageName>-rest/<version>`.

qiaozha:
> > @qiaozha, @MaryGao, any issues with the above proposal? One remaining question I have: for the "classic" level, do we want to call this "classic" or "client"? No strong opinion here.
> 
> Before digging into detailed patterns I may ask if we still plan to split the RLC from Modular #1981. If yes, I don't think we have the ability to distingush if a RLC is initiated from Modular or normal usage. In that way the RLC would be always like `azsdk-js-<packageName>-rest/<version>`.

I also noticed this, I left a comment before for the PR description, but then I delete it because of that.

joheredi:
I think it is valid to think about when we split RLC from modular. But at this point we don't have a clear plan on how that is going to happen, and there are a couple of ways it can go. I think we should go with this PR and take a change if needed before GA

qiaozha (approved):
LGTM

joheredi (approved):

PR #2623, opened 2024-06-27
upgrade-core-client-version-2.1.0

qiaozha:
pending https://github.com/Azure/azure-sdk-for-js/pull/30198 release


joheredi (commented):
We also need to update the codegen don't we? I only see changes in the generated tests and in the emitter package.json but none in the actual codegen

PR #2634, opened 2024-06-28
Remove references to @azure/core-util dev dependency

timovv:
Fix broken emitter release.

@azure/core-util is a dev dependency but is being used as a production dependency. This works for development but consumers of the published package are broken since they do not take a dependency on the package. Since it is only one small helper that is being used, I just pulled the helper out of core-util directly instead of changing the dependency.


joheredi (approved):

qiaozha:
I don't quite understand the problem here, we have been using core-util as a dependency for binary bytes conversion since a long time ago
https://github.com/Azure/autorest.typescript/blob/0896ab89e8babb848e062a2f78bf20a8a0c82c24/packages/typespec-ts/src/modular/helpers/operationHelpers.ts#L1377 and in the SDK repo, we have also used core-util as a dependency many SDKs, like openai https://github.com/qiaozha/azure-sdk-for-js/blob/d8a3e9f824ba453b74d1eb8ad41cf11eaa59dc51/sdk/openai/openai/package.json#L136

timovv:
There's no problem with using core-util in the emitted code like in the examples you linked.

In this case we were using helpers from core-util in the emitter itself. But core-util is not a dependency of the emitter; it's only a devDependency. This meant that generating with tsp-client using the released emitter didn't work because core-util doesn't get installed when installing the published emitter package.

qiaozha:
Thanks, Jose has also pointed it out to me that the problem is in the emitter dependency.

PR #2643, opened 2024-07-02
Handle nullable dictionaries in modular

MaryGao:
Hanld the nullable dictionary in modular.

fix the azure fleet build issue.

image

### DIFF
--- a/packages/typespec-test/test/openai_generic/generated/typespec-ts/src/models/models.ts
+++ b/packages/typespec-test/test/openai_generic/generated/typespec-ts/src/models/models.ts
@@ -668,7 +668,7 @@ export interface CreateCompletionRequest {
    * between -1 and 1 should decrease or increase likelihood of selection; values like -100 or 100
    * should result in a ban or exclusive selection of the relevant token.
    */
-  logitBias?: Record<string, number>;
+  logitBias?: Record<string, number> | null;
   /**
    * A unique identifier representing your end-user, which can help OpenAI to monitor and detect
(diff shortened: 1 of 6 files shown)


packages/typespec-test/test/openai_generic/generated/typespec-ts/src/models/models.ts
```diff
@@ -668,7 +668,7 @@ export interface CreateCompletionRequest {
    * between -1 and 1 should decrease or increase likelihood of selection; values like -100 or 100
    * should result in a ban or exclusive selection of the relevant token.
    */
-  logitBias?: Record<string, number>;
+  logitBias?: Record<string, number> | null;
```
MaryGao, line 671: This issue isn't raised up because this happened in input model so converting `Record<string, number>` to `Record<string, number> | null` will have no issue.

But we would have compile error if happened for output models.

qiaozha: Not related to this pr but I found the original typespec is 
` logit_bias?: Record<safeint> | null;` and there's no `@clientName` decorator, we just automatically normalize the name in Modular layer, wonder if that's expected?

qiaozha (approved):
LGTM, a minor comment.

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/models/models.ts
MaryGao: interesting finding, i created an issue for that https://github.com/Azure/autorest.typescript/issues/2645. let's talk there.

Issue #2647, opened 2024-07-03
Need to skip scripts related to samples if enable generateSample: false


kazrael2119:
https://github.com/Azure/azure-sdk-for-js/pull/30102#discussion_r1662407510

PR #2650, opened 2024-07-05
enable sdk-type check, build License and skip scripts if generateSample:false

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/2644
fixes https://github.com/Azure/autorest.typescript/issues/2647
fixes https://github.com/Azure/autorest.typescript/issues/2641


packages/autorest.typescript/test/rlcIntegration/generated/dpgCustomization/package.json
```diff
@@ -21,7 +21,7 @@
     "LICENSE",
     "review/*"
   ],
-  "sdk-type": "client",
+  "sdk-type": "mgmt",
```
qiaozha, line 24: This is expected, autorest will set openapi-type as arm. by default https://github.com/Azure/autorest/blob/fa08f8cc5047c03320be8928d7f62d9235238f0a/packages/libs/configuration/resources/default-configuration.md?plain=1#L10 
and we use both azureArm and openapi-type === arm as isArm condition.

Another reason is for sdk-type related information, it will only get generated when azureSdkForJs is true.

qiaozha (approved):

PR #2653, opened 2024-07-09
New spread support

qiaozha:
fixes https://github.com/Azure/autorest.typescript/issues/2414

In typespec, we have two kinds of spread.
1. the spread of model
```
model Foo {
   prop1: string;
   prop2: boolean;
}
op test(...Foo): void;
```
2. the spread of alias
```
alias Foo = {
   prop1: string;
   prop2: boolean;
};
op test(...Foo): void;
```

Previously, we have decided that we will treat spread of a alias as a payload flatten in this pr https://github.com/Azure/autorest.typescript/pull/2020#issuecomment-1730613201 and spread of a model is not a payload flatten.

With payload flatten, the method signature would look like 
```ts
export function test(prop1: string, prop2: boolean): void;
```
Without payload flatten, the method signature would be:
```ts
export function test(body: {prop1: string, prop2: boolean}): void;
```

And because alias was treated as anonymous model, we used to call `getEffectivePayloadType` from typespec compiler to try to get a valid payload type whenever possible, our previous spread condition becomes we flatten payload for anonymous model. it has a potential draw back which is we don't know if this model is defined as anonymous or become anonymous because of alias spread.

In https://github.com/Azure/typespec-azure/issues/1043, we found that this kind of difference is too complex for customer to understand the client side behavior differently per on model or alias. we have decided to change that spread of a model is also a payload flatten. 

In this PR, we have 
1. integrate with the new spread behavior from tcgc, see  https://github.com/Azure/typespec-azure/pull/886#issue-2307938557 for more details.
2. mitigate the unnecessary flatten in our own test cases. 

now spread is spread, anonymous model is anonymous model.


packages/typespec-test/test/chatApi_modular/generated/typespec-ts/src/rest/clientDefinitions.ts
```diff
@@ -8,10 +8,10 @@ import { Client, StreamableMethod } from "@azure-rest/core-client";
 export interface CreateStreaming {
   /** Creates a new streaming chat completion. */
   post(
-    options?: CreateStreamingParameters,
+    options: CreateStreamingParameters,
```
qiaozha, line 11: this is expected, with the previous tsp 
```
  createStreaming(
    ...StreamingChatCompletionOptions<TState, TContext>,
  ): ChatCompletionChunk<TState, TContext>;
```
the bodyParameter is undefined, and we define the parameter required as 
`parameters?.bodyParameter?.optional === false`

packages/typespec-test/test/eventgrid_modular/generated/typespec-ts/review/eventgrid.api.md
```diff
@@ -50,7 +50,9 @@ export class EventGridClient {
     constructor(endpointParam: string, credential: KeyCredential, options?: EventGridClientOptions);
     acknowledgeCloudEvents(topicName: string, eventSubscriptionName: string, lockTokens: AcknowledgeOptions, options?: AcknowledgeCloudEventsOptionalParams): Promise<AcknowledgeResult>;
     readonly pipeline: Pipeline;
-    publishCloudEvent(topicName: string, event: CloudEvent, options?: PublishCloudEventOptionalParams): Promise<PublishResult>;
+    publishCloudEvent(topicName: string, event: {
+        event: CloudEvent;
+    }, options?: PublishCloudEventOptionalParams): Promise<PublishResult>;
```
qiaozha, line 55: this is expected, as the typespec is defined to have an anonymous body parameter for it.
```
  @post op PublishCloudEvent is Azure.Core.RpcOperation<{
    @doc("content type")
    @header("content-type")
    contentType: "application/cloudevents+json; charset=utf-8";

    @doc("Topic Name.")
    @path
    topicName: string;

    @body 
    event: {event: CloudEvent};
  }, { @body _: PublishResult; }>;
```

packages/typespec-test/test/widget_dpg/generated/typespec-ts/src/classic/widgets/index.ts
```diff
@@ -133,19 +127,17 @@ export function getWidgets(context: WidgetServiceContext) {
     getWidget: (id: string, options?: WidgetsGetWidgetOptionalParams) =>
       getWidget(context, id, options),
     createWidget: (
-      body: CreateWidget,
+      weight: number,
+      color: "red" | "blue",
```
qiaozha, line 131: this is expected as what tsp defined 
```
 @post createWidget(...CreateWidget): {
        @statusCode code: 201;
        @body widget: Widget;
    } | WidgetError;
```

packages/typespec-ts/test/modularIntegration/bodyOptionality.spec.ts
```diff
@@ -42,7 +42,7 @@ describe("Body Optionality Client", () => {
 
   it("should support required-implicit body", async () => {
     try {
-      const result = await client.requiredImplicit({ name: "foo" });
+      const result = await client.requiredImplicit("foo");
```
qiaozha, line 45: should be expected as cadl ranch want to test this.

packages/typespec-ts/test/modularIntegration/spreadParameters.spec.ts
```diff
@@ -12,7 +12,7 @@ describe("SpreadClient Client", () => {
 
   it("should spread named model", async () => {
     try {
-      const result = await client.model.spreadAsRequestBody({ name: "foo" });
+      const result = await client.model.spreadAsRequestBody("foo");
```
qiaozha, line 15: same here, should spread.

packages/typespec-test/test/chatApi_modular/generated/typespec-ts/src/rest/clientDefinitions.ts
qiaozha: issue tracked https://github.com/Azure/typespec-azure/issues/1149

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -841,10 +832,20 @@ function emitBasicOperation(
...
-      bodyParameter.type.properties.length > 0
+      originalBodyType.kind === "Model" &&
+      originalBodyType.name === "" &&
+      [...originalBodyType.properties.keys()].every(
+        (k) =>
+          operation.parameters.properties.has(k) &&
+          (operation.parameters.properties.get(k) ===
+            (originalBodyType as Model).properties.get(k) ||
+            operation.parameters.properties.get(k) ===
+              (originalBodyType as Model).properties.get(k)?.sourceProperty)
```
MaryGao, line 847: tbh i don't want to maintain this logic in modular side. this would introduce new debt for us when adopting to TCGC SdkPackage. i prefer to prioritize the task of integrating with SdkPackage for new changes in this semaster at least for our below goal.

```
we are committed to not accumulating any new debt moving forward.
```

@joheredi how do you think?

qiaozha: https://github.com/Azure/autorest.typescript/issues/2660 issue tracked. I will have another pr to take this as an example.

joheredi (approved):
Let's go with this fix to unblock moving the latest TCGC. In the meantime I'll setup some infrastructure so that we can move to using TCGC types in a more organized and smooth way

packages/typespec-test/test/chatApi_modular/generated/typespec-ts/src/rest/clientDefinitions.ts
qiaozha: Create this https://github.com/Azure/autorest.typescript/issues/2677 to track our codegen work.

Issue #2659, opened 2024-07-11
[RLC] Extensible Enums doesn't have known values in the docs anymore.


joheredi:
Extensible enums don't get the jsdoc with the possible values, it used to include them.

For more details: https://github.com/Azure/azure-sdk-for-js/pull/30373/files#r1674696379

This is the TypeSpec change used in the PR above: https://github.com/Azure/azure-rest-api-specs-pr/pull/17637/files#diff-c3d568193c41dfcf658ca0a92e7012d92b2c77e58e0c70e2e5c66aaf83f806b7

PR #2661, opened 2024-07-12
add-enum-description-for-rlc

qiaozha:
fixes https://github.com/Azure/autorest.typescript/issues/2659

### DIFF
--- a/packages/typespec-ts/src/utils/modelUtils.ts
+++ b/packages/typespec-ts/src/utils/modelUtils.ts
@@ -1446,7 +1446,11 @@ function getPriorityName(schema: Schema, usage?: SchemaContext[]): string {
 }
 
 function getEnumStringDescription(type: any) {
-  if (type.name === "string" && type.enum && type.enum.length > 0) {
+  if (
+    (type.name === "string" || type.alias === "string") &&
+    type.enum &&
+    type.enum.length > 0
+  ) {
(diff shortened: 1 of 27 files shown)


joheredi (approved):

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -1446,7 +1446,11 @@ function getPriorityName(schema: Schema, usage?: SchemaContext[]): string {
 }
 
 function getEnumStringDescription(type: any) {
-  if (type.name === "string" && type.enum && type.enum.length > 0) {
+  if (
+    (type.name === "string" || type.alias === "string") &&
```
MaryGao, line 1450: We also have cases for extensible enum for number, should we consider them also?

MaryGao (approved):
The change looks good to me.

packages/typespec-ts/src/utils/modelUtils.ts
qiaozha: added ut here https://github.com/Azure/autorest.typescript/pull/2661/files#diff-3b7495c06bf1fc41a5e741d3e68d82b771f1052cd2f36773395807e9cb7e4bc4R3638 for boolean type, waiting for typespec team's confirm. Probably won't fix https://github.com/Azure/typespec-azure/issues/1162

PR #2662, opened 2024-07-12
Introduce SdkType context

joheredi:
This PR introduces a context hook that allows to get an SdkType and SdkMethod from a low-level TypeSpec type such as Type or Operation.


packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -754,6 +755,7 @@ function emitBasicOperation(
   rlcModels: RLCModel,
   hierarchies: string[]
 ): HrlcOperation {
+  const getSdkType = useSdkTypes()!;
```
joheredi, line 758: Changes here are an example on how to use this new context.

/cc - @qiaozha, @MaryGao

packages/typespec-ts/src/context/sdkTypes.ts
```diff
@@ -0,0 +1,74 @@
...
+  for (const sdkModel of sdkPackage.models) {
+    if (!sdkModel.__raw) {
+      continue;
+    }
+
+    sdkTypesContext.types.set(sdkModel.__raw, sdkModel);
+  }
+
+  for (const client of sdkPackage.clients) {
+    for (const method of client.methods) {
```
qiaozha, line 64: is every method belong to client now?

packages/typespec-ts/src/context/sdkTypes.ts
```diff
@@ -0,0 +1,74 @@
...
+    operations: new Map<Type, SdkMethod<SdkHttpOperation>>(),
+    types: new Map<Type, SdkType>()
+  };
+
+  for (const sdkEnum of sdkPackage.enums) {
+    if (!sdkEnum.__raw) {
+      continue;
+    }
+
+    sdkTypesContext.types.set(sdkEnum.__raw, sdkEnum);
```
qiaozha, line 52: if sdk packages has enums, should we also have that? for future better adoption?

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -841,6 +843,8 @@ function emitBasicOperation(
   } else {
     bodyParameter = emitBodyParameter(context, httpOperation);
     // Flatten the body parameter if it is an anonymous model
+    const sdkOperation = getSdkType(operation);
```
MaryGao, line 846: Operation is unlikely to type. In SDKType we could split all into different types and adopt into TCGC by need. And operations could be categoried into normal, LRO, paging ect or their generation could be parameters or responses. 

So `no new debt` means if we need something from TCGC, we need to replace relevant parts.

joheredi: We need to find the sweetspot between meaningful refactoring but keeping the changes as scoped as possible

packages/typespec-ts/src/context/sdkTypes.ts
joheredi: In TCGC sdkPackage yes, every method is under a client

packages/typespec-ts/src/context/sdkTypes.ts
joheredi: We are already setting the enums here

packages/typespec-ts/src/modular/helpers/namingHelpers.ts
```diff
@@ -77,6 +76,10 @@ export function getClassicalLayerPrefix(
   return prefix.join(separator);
 }
 
+export function isDefined<T>(thing: T | undefined | null): thing is T {
```
qiaozha, line 79: do we still need this? I remember Daniel's pr introduced this.

common/config/rush/pnpm-lock.yaml
```diff
@@ -602,6 +626,26 @@ packages:
       '@typespec/versioning': 0.57.0_@typespec+compiler@0.57.0
       change-case: 5.4.4
       pluralize: 8.0.0
+    dev: false
+
+  /@azure-tools/typespec-client-generator-core/0.44.0_uovqfdr3tffszkt2tayxtbnshy:
```
qiaozha, line 631: why we have pnpm-lock.yaml change?

packages/typespec-ts/test/modularUnit/anonymousModel.spec.ts
```diff
@@ -364,7 +364,8 @@ describe("anonymous model", () => {
         );
       });
 
-      it("should not flatten model if spread in the payload with required parameters", async () => {
+      // TODO: Re-enable when related work is merged
+      it.skip("should not flatten model if spread in the payload with required parameters", async () => {
```
qiaozha, line 368: I am not quite sure why this is failing now? I didn't see we have done any specific change for it yet?

common/config/rush/pnpm-lock.yaml
joheredi: Removing these changes

packages/typespec-ts/test/modularUnit/anonymousModel.spec.ts
joheredi: yup I was testing the latest version and this was a leftover

packages/typespec-ts/src/modular/helpers/namingHelpers.ts
joheredi: I see it is still used in other places

joheredi: such as in `getModularModelFilePath`

qiaozha: Oh right, I remember I've changed similar thing in modular models pr.

qiaozha (approved):

PR #2669, opened 2024-07-15
Add conditional-request test case

### LINKED ISSUE #2668: Supplement test case for special-headers
image

v-jiaodi:
Add conditional-request test case and remove repeated case.
fixes #2668


v-jiaodi:
Closed it as we don't need to add test for conditional request for now.

PR #2675, opened 2024-07-16
Upgrade dev dependency prettier to version 3.3.3 in generated packages

jeremymeng:
- don't add prettier for azure sdk packages
- update baselines


packages/rlc-common/src/metadata/packageJson/packageCommon.ts
```diff
@@ -43,7 +43,7 @@ export function getCommonPackageDevDependencies(
     "@microsoft/api-extractor": "^7.40.3",
     "@types/node": "^18.0.0",
     eslint: "^8.55.0",
-    prettier: "^3.2.5",
+    prettier: "^3.3.3",
```
qiaozha, line 46: I wonder if we really rely on prettier in non azure?

jeremymeng: I did a search and found that there are some @msinternal tests which use "prettier" but also reference `../../../.prettierrc.json` so I don't know what the design there. For example, https://github.com/Azure/autorest.typescript/blob/1de92fdbe7ed4b830621c85bef6e56d91b06fd47/packages/typespec-ts/test/integration/generated/type/model/flatten/package.json#L48

I didn't see "prettier" being used in other test baselines, so maybe it is safe to remove them, unless we also want to have the ability to format after a non-azure package is generated.

jeremymeng: Since the standalone packages still invoke "prettier" even though the config file path is definitely not valid if not within azure-sdk-for-js.  I logged a new issue to discuss further: https://github.com/Azure/autorest.typescript/issues/2700

jeremymeng: I removed all "prettier" dependencies and formatting scripts from non azure monorepo code generation.

jeremymeng:
Not sure about the current CI failures. I ran the same script locally but didn't see any changes.

MaryGao (approved):

PR #2682, opened 2024-07-19
update openai smoke test

kazrael2119:
Fix a bug in extensible union as enum with nullable and update the spec for openai_modular smoke-test with latest rest api spec repo main branch.


packages/typespec-test/test/openai_modular/generated/typespec-ts/src/rest/outputModels.ts
```diff
@@ -224,7 +258,7 @@ export interface ChoiceOutput {
   /** The log probabilities model for tokens associated with this completions choice. */
   logprobs: CompletionsLogProbabilityModelOutput | null;
   /** Reason for finishing */
-  finish_reason: CompletionsFinishReasonOutput | null;
+  finish_reason;
```
kazrael2119, line 261: image

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -759,6 +767,9 @@ function getSchemaForModel(
   }
   for (const [propName, prop] of model.properties) {
     const restApiName = getWireName(dpgContext, prop);
+    if (restApiName === "finish_reason") {
+      prop;
+    }
```
qiaozha, line 772:
```suggestion
```

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -388,37 +388,45 @@ function getSchemaForUnion(
...
+    });
+    values.push(variantType);
+    if (variantType.typeName) {
+      namedUnionMember = true;
+    }
   }
-  if (asEnum?.open) {
+  if (asEnum?.open && asEnum.members.size > 0) {
+    namedUnionMember = false;
     values = [];
```
MaryGao, line 405: just curious if we need to reset the values and namedUnionMember, why we go to foreach in line 392. Should we use open condition to separate them?

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -388,37 +388,45 @@ function getSchemaForUnion(
...
+        namedUnionMember = true;
+      }
     }
   }
   const schema: any = {};
   if (values.length > 0) {
     schema.enum = values;
     const unionAlias =
-      asEnum?.open && asEnum?.kind
+      asEnum?.open && asEnum?.kind && !namedUnionMember
```
MaryGao, line 421: could you help me understand why the enum member having a name would impact the union alias?

packages/typespec-test/test/openai_modular/spec/examples/2024-05-01-preview/generated_audio_transcription_object.json
```diff
@@ -0,0 +1,17 @@
+{
+  "operationId": "GetAudioTranscriptionAsResponseObject",
```
MaryGao, line 2: i prefer not tracking these examples in smoke testing.

packages/typespec-ts/src/utils/modelUtils.ts
qiaozha:
```
      union ExtensibleStringEnum {
        string;
        "a";
        "b";
      }
     model SimpleModel {
        propNullableRequired: ExtensibleStringEnum | null;
        propNullableFlattened: string | "a" | "b" | null;
     }
```
In the case, `getUnionAsEnum` for `propNullableRequired` and `propNullableFlattened` type will be a valid enum, which means `asEnum.open` is true and `asEnum.kind` has meaningful value. but we can't generate string type for `propNullableRequired` and `propNullableFlattened`. 
```
         propNullableRequired: ExtensibleStringEnum | null;
         propNullableFlattened: string | null;
       }
```
in the `propNullableRequired` case, we should get each value type, therefore, we need to see if the there's any named union in the member. 
in the `propNullableFlattened` case, we should leverage the `asEnum.kind` and `asEnum.nullable` to get the right typescript type.

packages/typespec-ts/src/utils/modelUtils.ts
qiaozha: updated.

packages/typespec-test/test/openai_generic/generated/typespec-ts/src/rest/models.ts
```diff
@@ -461,7 +461,7 @@ export interface CreateFineTuneRequest {
    * "davinci", or a fine-tuned model created after 2022-04-21 and before 2023-08-22. To learn more
    * about these models, see the [Models](/docs/models) documentation.
    */
-  model?: string;
+  model?: string | null;
```
qiaozha, line 464: this is the typespec definition 
``model`?: string | "ada" | "babbage" | "curie" | "davinci" | null;`

packages/typespec-test/test/openai_non_branded/generated/typespec-ts/src/rest/models.ts
```diff
@@ -460,7 +460,7 @@ export interface CreateFineTuneRequest {
    * "davinci", or a fine-tuned model created after 2022-04-21 and before 2023-08-22. To learn more
    * about these models, see the [Models](/docs/models) documentation.
    */
-  model?: string;
+  model?: string | null;
```
qiaozha, line 463: same here, should be expected.

qiaozha (approved):

packages/typespec-ts/src/utils/modelUtils.ts
MaryGao: maybe not a priority, but we may meet case for mixing both.

```
model SimpleModel {
        propNullableRequired: ExtensibleStringEnum | string | "a" | null;
}
```

qiaozha: `string | null` is our answer here.  but I think we should avoid writing typespec like that.

PR #2683, opened 2024-07-19
Fix the nightly build failure in smoke test because of resource issue

MaryGao:
Fix the nightly build failure in smoke test because of resource issue

the build is successful here: https://dev.azure.com/azure-sdk/public/_build/results?buildId=3976274&view=results

### DIFF
--- a/.scripts/nightly.yml
+++ b/.scripts/nightly.yml
@@ -7,6 +7,10 @@ schedules:
         - main
 pr: none
 
+pool:
+  name: azsdk-pool-mms-ubuntu-2004-general
+  vmImage: ubuntu-20.04
+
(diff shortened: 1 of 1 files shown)


.scripts/nightly.yml
```diff
@@ -7,6 +7,10 @@ schedules:
         - main
 pr: none
 
+pool:
+  name: azsdk-pool-mms-ubuntu-2004-general
```
MaryGao, line 11: we should adopt dedicated pool resource to ensure the stablility of build.

lirenhe (approved):

qiaozha (approved):

Issue #2458, opened 2024-04-19
[Bug] RLC doesn't handle the credential generation correctly


qiaozha:
if we have typespec like this `@useAuth(BearerAuth | ApiKeyAuth<ApiKeyLocation.cookie, "session-id">)` https://github.com/bterlson/typespec-todo/blob/main/main.tsp#L12,  the generated rlc client  is missing credential for it, but modular layer has it.
```ts
export default function createClient(
  endpoint: string,
  options: ClientOptions = {},
): TodoClient {
```

MaryGao:
closed as fixed.

Issue #2377, opened 2024-03-25
Fix issues happened when enable modular generation for todo project


kazrael2119:
see pr: https://github.com/Azure/autorest.typescript/pull/2526

```[tasklist]
### Tasks
- [x] Support only one success responses
- [x] Handle unsupported credential
- [ ] Fail to import RLC responses in api layer
- [ ] Form/data in modular is not supported and so generated code can't pass compile
```
### **Support only one success responses**
If we have multiple responses defined without error decorator the code can't compile. We should fix it with assuming only one success response code and others would be skipped with a warning.
### **Handle unsupported credential**
When customers defined credential which we are not supported e.g session api key, we should generate code without compile error but give a warning for not supporting.

### **Fail to import RLC responses in api layer**
After disabling form-data operations and this issue is fixed. So we could fix form data first to check if this is a valid issue.

### **Not support form/data in modular**
We have form/data definition and RLC could handle it correctly but modular can't.

```
    @post
    op createFileAttachment(
      @header contentType: "multipart/form-data",
      @path itemId: TodoItem.id,
      contents: bytes,
    ): OkResponse | NotFoundResponse;
```

And currently the modular would treat it as `Uint8Array` but in RLC it is form data schema. So failed to compile.

```ts
export function _createFileAttachmentSend(
  context: Client,
  itemId: number,
  contents: Uint8Array,
  options: TodoItemsAttachmentsCreateFileAttachmentOptionalParams = {
    requestOptions: {},
  },
): StreamableMethod<
  | TodoItemsAttachmentsCreateFileAttachment200Response
  | TodoItemsAttachmentsCreateFileAttachment404Response
> {
  return context
    .path("/items/{itemId}/attachments", itemId)
    .post({
      ...operationOptionsToRequestParameters(options),
      contentType: (options.contentType as any) ?? "multipart/form-data",
      body: { contents: uint8ArrayToString(contents, "base64") },
    }) as StreamableMethod<
    | TodoItemsAttachmentsCreateFileAttachment200Response
    | TodoItemsAttachmentsCreateFileAttachment404Response
  >;
}
```

qiaozha:
@kazrael2119 can you retry this with the latest code?

joheredi:
Is this still reproducing? @qiaozha @MaryGao can you please update the title with information about the issue and add details on what's causing this?

MaryGao:
Should have dependency with https://github.com/Azure/autorest.typescript/issues/2279

PR #2704, opened 2024-07-25
Fix the todo generation issues 

MaryGao:
fixes first two issues in https://github.com/Azure/autorest.typescript/issues/2377 and https://github.com/Azure/autorest.typescript/issues/2458

- [x] Support only one success responses
- [x] Handle unsupported credential


packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -189,14 +190,8 @@ export function getDeserializePrivateFunction(
...
-    const validStatus = [
-      ...new Set(
-        operation.responses
-          .flatMap((r) => r.statusCodes)
-          .filter((s) => s !== "default")
-      )
-    ];
-
+    // Directly using the status code of first response
+    const validStatus = response.statusCodes;
```
MaryGao, line 194: Directly use the first success status codes.

packages/typespec-ts/src/utils/credentialUtils.ts
```diff
@@ -0,0 +1,41 @@
...
+  authorization: Authentication
+): HttpAuth[] {
+  const authList = [];
+  for (const option of authorization.options) {
+    for (const auth of option.schemes) {
+      switch (auth.type) {
+        case "apiKey":
+          if (auth.in === "cookie" || auth.in === "query") {
+            // just skip un-supported authentication and report a warning
+            reportDiagnostic(program, {
```
MaryGao, line 22: Skip un-supported authentication and report a warning

qiaozha: What exactly does cookie mean? I feel like it's just set a header with the name cookie, similar like custom key credential?

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
qiaozha: is this because there's is no real case that has multiple responses in typespec?

MaryGao: yes.

packages/typespec-ts/src/utils/credentialUtils.ts
MaryGao: https://swagger.io/docs/specification/authentication/cookie-authentication/

Simply speaking I don't think cookie auth is just setting a header with customers' input.

Generally browser would automatically maitain the cookie information and this is transparent for end users when trying to communicate with service side. So the server side could save information like session id in Set-Cookie header and when sending any requests these cookies would be sent by Cookie header properly(with scope or some policy) by browser. 

As for node environemnt, if we try to monitor this process we would have a policy which could do something like BearereAuth, 1. go to somewhere to authenticate the users; 2. if successful we will have Set-Cookie header in response. 3. read this info and maintain cookies info in client; 4. append these cookie info to send real requests.

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -1706,15 +1707,14 @@ function emitCredentialParam(
   const auth = getAuthentication(context.program, namespace);
```
MaryGao, line 1707: @joheredi I was wondering if we need a pr to update the credential type first with the idea of your commit: https://github.com/joheredi/autorest.typescript/commit/dfa6948c7ebf939490cb649263c758475a5f0e52#diff-74f5c341faa7623f0c46c5a1c030d16afc796ec149b852c862b68160ea63c448.

Then my fix could directly be based on that change.

packages/typespec-ts/src/lib.ts
```diff
@@ -237,6 +237,12 @@ const libDef = {
       messages: {
         default: paramMessage`Please note the body type of default response for operation - ${"operationName"} is not a model type.`
       }
+    },
+    "un-supported-credential": {
+      severity: "warning",
+      messages: {
+        default: paramMessage`The defined authentication - ${"credentialType"} is not supported yet and it may require customizations.`
```
joheredi, line 244:
```suggestion
        default: paramMessage`Authentication type ${"credentialType"} is not supported.`
```

packages/typespec-ts/src/modular/buildCodeModel.ts
joheredi: Unfortunately we are not ready to use that yet.

MaryGao: Sure let's have this fix now.

qiaozha (approved):

PR #2711, opened 2024-07-29
Upgrade to the latest tcgc to v.44.3

MaryGao:


packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -1458,20 +1456,10 @@ function emitEnumMember(context: SdkContext, member: any): Record<string, any> {
   };
 }
 
-function emitSimpleType(
-  context: SdkContext,
-  type: Scalar | IntrinsicType | SdkBuiltInType
-): Record<string, any> {
-  let sdkType: SdkBuiltInType;
-  if (type.kind === "Scalar" || type.kind === "Intrinsic") {
```
qiaozha, line 1466: is this because of Dapeng's change?

qiaozha (approved):
LGTM

packages/typespec-ts/src/modular/buildCodeModel.ts
MaryGao: yeah, his change would cause our compile failure but in our side we only use SdkBuiltInType so I just remove this logic.

PR #2712, opened 2024-07-30
Add CADL ranch d.ts generation

dgetu:
This change replaces all ts files in the CADL ranch output with an API summary. Some tests were configured not to emit metadata, or to emit to a subfolder of the project root. All tspconfigs must be configured to emit to the root of the project and to emit metadata (specifically package.json).


packages/typespec-ts/test/commands/run.js
```diff
@@ -1,67 +1,181 @@
...
+                emitDeclarationOnly: true,
+                declarationMap: true,
+                removeComments: true,
+                rootDir: outputPath()
+            },
+            rootNames: [joinPath(outputPath(), "src/index.ts")]
+        });
+        // side effect loads source files into memory
+        // nothing will be emitted if this is omitted
+        program.getSourceFiles();
```
qiaozha, line 70: I wonder how we run the integration test with ts-morph program to load all the generated files? 
Also, want to confirm if we are not running the test parallelly ?

packages/typespec-ts/test/commands/run.ts
```diff
@@ -0,0 +1,224 @@
+import { fileURLToPath } from "url";
```
qiaozha, line 1: what's the difference between this ts file and the js file? I don't think we need both?

qiaozha (commented):
I notice we didn't add any api view file in https://github.com/Azure/autorest.typescript/tree/dd505de1900770cb747f3fb73bfc11500c58ae5d/packages/typespec-ts/test/integration/generated/hello, wonder why? 
Also, we don't need to add .d.ts file with ts morph program loads all the files internally ?

I am also wondering, what would the error messages look like in this way ?

packages/typespec-ts/test/commands/run.ts
```diff
@@ -0,0 +1,224 @@
...
+  IExtractorConfigPrepareOptions
+} from "@microsoft/api-extractor";
+import { npxCommand } from "./runCommand.js";
+import * as fs from "fs/promises";
+import { Dirent } from "fs";
+import { createTaskLogger } from "./logger.js";
+import { createProgram } from "typescript";
+
+export async function runTypespec(config: any, mode: any) {
+  await run(createContext());
```
joheredi, line 16: nit: I was getting confused thinking that run was trying to run a task to create the context, may be clearer if you create the context first and then pass it to the run function

packages/typespec-ts/test/commands/run.ts
```diff
@@ -0,0 +1,224 @@
...
+      logger: () => logger,
+      mode: () => mode,
+      sourceTypespec: () => sourceTypespec,
+      targetFolder: () => targetFolder,
+      testRoot: () => testRoot
+    };
+  }
+}
+
+async function run(ctx: any) {
```
joheredi, line 36: maybe we could give run a more descriptive name :)

packages/typespec-ts/test/integration/generated/authentication/api-key/auth-apikey.api.md
```diff
@@ -0,0 +1,105 @@
+## API Report File for "@msinternal/auth-apikey"
```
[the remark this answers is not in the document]
joheredi: Instead of the api.md we should keep the rollup.d.ts which will allow us to also reference the client in tests without the implementation being present. tspconfig.yaml needs to be present as well as things are organized today.

dgetu:
> I notice we didn't add any api view file in https://github.com/Azure/autorest.typescript/tree/dd505de1900770cb747f3fb73bfc11500c58ae5d/packages/typespec-ts/test/integration/generated/hello

This test is stale - there's no corresponding tsp file in the cadl ranch repo.

>we don't need to add .d.ts file with ts morph program loads all the files internally ?

This is running tsp compile by spawning a new process, so we don't have access to the ts-morph instance. I don't know of a way to use in-memory files for API Extractor either, so I think we have to write .d.ts files to disk. If there's an easier way that doesn't involve using the compiler API, I'm all ears.

> what would the error messages look like in this way ?

Error messages should be similar. Both the output of `tsc` and API Extractor are written to stdout.

packages/typespec-ts/test/integration/generated/payload/pageable/.gitignore
```diff
@@ -0,0 +1,6 @@
+/**
```
joheredi, line 1: It would be cool if we had a way to automate adding new cadl ranch specs so that .gitignore and config are automatically added :) just an idea for later

joheredi (approved):
This is very excitin :) thanks for the hard work

packages/typespec-ts/test/commands/runCommand.js
```diff
@@ -1,7 +1,7 @@
 import { spawn } from "child_process";
 
 export async function runCommand(command, args = [], workingDirectory, logger) {
-  return new Promise((resolve, reject) => {
+  return await new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
dgetu: There are a few things we can do in this scenario, in order of most to least preferred:

1. `return await` or otherwise await the promise before returning
2. Return the promise and keep the function as async. The runtime [won't wrap the promise in another promise.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#rewriting_a_promise_chain_with_an_async_function)
3. Change `runCommand` to a synchronous function (i.e. remove `async` from line 3). This is identical to 2, but I prefer the style of 2 over this. `typescript-eslint` [favors 1 over 2 and 3](https://typescript-eslint.io/rules/return-await/) for error handling and perf reasons.

When I wrote this, I didn't know that the runtime only implicitly wraps non-promise values when they're returned from async functions, so I figured that there were potential data race issues. I'd suggest we take on the [`no-floating-promises` typescript-eslint rule](https://typescript-eslint.io/rules/no-floating-promises/) since failing to await a promise is so rarely correct.

PR #2713, opened 2024-07-30
disable sampleUrl generation if generateSample is false

kazrael2119:


packages/rlc-common/src/metadata/buildReadmeFile.ts
```diff
@@ -376,7 +376,11 @@ function createMetadata(model: RLCModel): Metadata | undefined {
     serviceDocURL: productDocLink,
     packageSourceURL: packageSourceURL,
     packageNPMURL: `https://www.npmjs.com/package/${clientPackageName}`,
-    samplesURL: packageSourceURL && `${packageSourceURL}/samples`,
+    samplesURL: packageSourceURL
```
MaryGao, line 379: could be like this
```ts
samplesURL: model.options.generateSample && packageSourceURL && `${packageSourceURL}/samples`,
```

MaryGao (approved):

PR #2717, opened 2024-07-31
fix resolve conflicts bug

kazrael2119:
separate fix from https://github.com/Azure/autorest.typescript/pull/2697/commits/17dafe13eb539d52d4545cf8843bb939e0293146

### DIFF
--- a/packages/typespec-ts/src/modular/buildCodeModel.ts
+++ b/packages/typespec-ts/src/modular/buildCodeModel.ts
@@ -1561,7 +1561,7 @@ function emitOperationGroups(
     context.rlcOptions?.hierarchyClient === false &&
     context.rlcOptions?.enableOperationGroup
   ) {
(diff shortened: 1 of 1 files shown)


packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -1589,25 +1589,10 @@ function addHierarchyOperationGroup(
   return [];
 }
 
-function resolveConflictIfExist(operationGroups: OperationGroup[]) {
+function resolveConflict(operationGroups: OperationGroup[]) {
```
MaryGao, line 1592: i was wondering if we could rename this as sth e.g appendOperationGroupPrefix to make it clearer because this is not mainly for resolving conflict now.

MaryGao (approved):

qiaozha (approved):

Issue #2506, opened 2024-05-08
support serialize and deserialize for additional properties in compatible mode 


qiaozha:
Currently, we won't be able to do any serialize and deserialize for additional properties, what pr https://github.com/Azure/autorest.typescript/pull/2469 supports is there's no any serialization work between modular and rlc,

but typespec as the below 
```
model Foo {
  ...Record<utcDateTime>
  prop: utcDateTIme;
}
```
In RLC layer, we will generate
```
interface Foo extends Record<string, string>{
  prop: string;
}
```
but in Modular, we will get
```
interface Foo extends Record<string, Date>{
  prop: Date;
}
```
serialization and deserialization will be required in this case, regardless if it's compatible mode or not.

qiaozha:
I will take some time to verify it, before we can close it.

Issue #2471, opened 2024-04-23
Additional properties non-legacy client implementation


qiaozha:
According to https://github.com/Azure/autorest.typescript/issues/2218, for additional properties in legacy client, we generate model with additionalProperties bag to accept additional properties in the modular layer, it will pending on the serialize and deserialize refactor.

qiaozha:
close it as plane changed.

qiaozha:
we will keep using the additional properties bag for non legacy clients.

PR #2720, opened 2024-08-01
use-record-of-any-for-additional-properties-in-modular

qiaozha:
As @joheredi confirmed with @bterlson, we should just generate record of any for additional properties like what we used to do for mgmt plane. 

Therefore revert the compatibilityMode flag related logic. and since both Modular layer and RLC layer are using the same Record of any, we don't need to have any special serialize and deserialize logic for them. Customers will have to pass the rest api layer value datetime string like the below example by themselves. 

```
model Foo {
  ...Record<utcDateTime>
  prop: utcDateTIme;
}

```
closes https://github.com/Azure/autorest.typescript/issues/2506
closes https://github.com/Azure/autorest.typescript/issues/2471

### DIFF
--- a/packages/rlc-common/src/interfaces.ts
+++ b/packages/rlc-common/src/interfaces.ts
@@ -244,7 +244,6 @@ export interface RLCOptions {
   flavor?: PackageFlavor;
   enableModelNamespace?: boolean;
   hierarchyClient?: boolean;
-  compatibilityMode?: boolean;
   experimentalExtensibleEnums?: boolean;
 }
 
--- a/packages/typespec-ts/src/lib.ts
+++ b/packages/typespec-ts/src/lib.ts
@@ -11,7 +11,6 @@ import { Options } from "prettier";
 
 export interface EmitterOptions extends RLCOptions {
   branded?: boolean;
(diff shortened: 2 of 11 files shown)


joheredi:
We still need to do serialization and deserialization. Even though we are generating this as a `Record<string, unknown>` the emitter still has the information about the type that is going to be coming from/to the wire. For example,

```ts
model SpreadModelRecord {
  knownProp: string;
  ...Record<ModelForRecord>;
}
```

Will emit

```ts
export interface SpreadModelRecord extends Record<string, any> {
  knownProp: string;
}

export function spreadModelRecordSerializer(item: SpreadModelRecord) {
  const { knownProp, ...additionalProperties } = item;
  return {
    ...serializeRecord(additionalProperties, modelForRecordSerializer),
    knownProp: modelForRecordSerializer(item.knownProp)
  };
}
```

When we are generating the serializer, we do have the information to know the type of the Record so we should be able to deserialize accordingly.

The current serializer is missing this part and we'll need to implement it.

qiaozha:
close it as we will keep support additional properties bag for non legacy client.

PR #2722, opened 2024-08-01
Fix build error

v-jiaodi:
Monitor @typespec/compiler dev version build
https://dev.azure.com/azure-sdk/public/_build/results?buildId=4011943&view=logs&j=d0d2a462-287f-51d5-e186-98b6b49f577f

### DIFF
--- a/packages/typespec-ts/src/modular/buildCodeModel.ts
+++ b/packages/typespec-ts/src/modular/buildCodeModel.ts
@@ -1231,17 +1231,32 @@ function applyEncoding(
 
 function mergeFormatAndEncoding(
   format: string | undefined,
-  encoding: string,
+  encoding: string | undefined,
   encodeAsFormat: string | undefined
-): string {
+): string | undefined {
   switch (format) {
     case undefined:
-      return encodeAsFormat ?? encoding;
+      return encodeAsFormat ?? encoding ?? format;
     case "date-time":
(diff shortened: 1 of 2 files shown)


MaryGao (approved):

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -1231,17 +1231,32 @@ function applyEncoding(
 
 function mergeFormatAndEncoding(
   format: string | undefined,
-  encoding: string,
+  encoding: string | undefined,
```
lirenhe, line 1234: Do we need any test case for this?

MaryGao: currently we can't, but we could include test cases in https://github.com/Azure/autorest.typescript/issues/2676.

the case could be prepared when encode int as string is released.

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -298,17 +298,32 @@ function applyEncoding(
 
 function mergeFormatAndEncoding(
   format: string | undefined,
```
MaryGao, line 300: it seems we duplicate the same function two different places. i prefer to keep it considering we have plan to split Modualr from RLC.

Issue #2049, opened 2023-10-09
Client side default value for server host parameters handling


qiaozha:
let's say we have a definition of @server like 
```
@server(
  "{endpoint}/client/structure/{client}",
  "",
  {
    @doc("Need to be set as 'http://localhost:3000' in client.")
    endpoint: url = "http://localhost:3000",

    @doc("Need to be set as 'default', 'multi-client', 'renamed-operation', 'two-operation-group' in client.")
    client: ClientType,
  }
)
```

In current Modular implemetation, we will simply ignore all kinds of default value even it's a client level parameter.
so the createClient function will be like 

```typescript
export interface ServiceClientOptions extends ClientOptions {}

export function createService(
  endpoint: string,
  client: ClientType,
  options: ServiceClientOptions = {}
): ServiceContext {
 // other logic
  const clientContext = getClient(endpoint, client, options); // this is where we call the rlc layer client factory function
  return clientContext;
}
```
but in RLC, we will have a clientFactory function like 
``` typescript
export interface ServiceClientOptions extends ClientOptions {
  endpoint?: string;
}

export default function createClient(
  client: ClientType,
  options: ServiceClientOptions = {}
): ServiceClient;
```
there's a mismatch between the Modular layer and RLC layer. The reasons that we have current Modular implementation to ignore default value are:
 1. We have a guideline that says we should not respect any kinds of client side default value they are just for documentation purpose. 
 2. If there's a parameter that just adds a default value without any other changes, it would not be a breaking in the API level, but will become a breaking change for the client side if we respect the default value.

But I wonder what should be the best practice here: 

Option 1: We keep the current implementation of both Modular and RLC, and change the way we call the RLC layer client factory function.
```typescript
export function createService(
  endpoint: string,
  client: ClientType,
  options: ServiceClientOptions = {}
): ServiceContext {
 // other logic
  const clientContext = getClient(client, { endpoint,  ...options }); // this is where we call the rlc layer client factory function
  return clientContext;
}
```

As mgmt plane is mostly following the same pattern as we do in RLC, keep the current implementation of Modular i.e. Option 1 will be a breaking change for the mgmt plane. 

Option 2: we could have a flag in the codegen based on Option 1 to decide whether respect the default value or not. for mgmt plane, we can just respect the client level default value and the generated code will look like:
```typescript
export interface ServiceClientOptions extends ClientOptions {
  endpoint: string
}

export function createService(
  client: ClientType,
[...]

PR #2728, opened 2024-08-02
Remove internal RLC layer from modular

### LINKED ISSUE #2725: Remove `rest/` layer from modular output
Remove everything under the `rest/` folder from the modular output. This work includes:
- Coming up with an alternate approach for `isUnexpected`
- Removing dependency on RLC models
- Removing dependency on any utils in RLC

timovv:
This PR removes the `rest/` directory from the modular generator. There is no significant change in the public API and the runtime serialization logic remains the same.

Key points:
- Removed references to RLC models in the internal operation send and deserialize functions. The send functions now return a `PathUncheckedResponse`, and the deserialize functions accept a `PathUncheckedResponse` instead of the RLC model types. The actual API methods still accept and return the API layer models.
- In the deserialize functions, instead of using `isUnexpected`, the status code of the response is now checked inline against a list of expected status codes.
- The context factory function now calls `getClient` from `@azure-rest/core-client` directly instead of creating an RLC client from the REST layer.
  - Things that the RLC `getClient` used to do, such as adding the custom API version policy and/or key credential policy, are now done inline in the context factory function.
  - The ClientContext interface is now declared in the API layer as an interface extending `Client` from `@azure-rest/core-client`, instead of being a re-export of the RLC definition. 
- Extracted and redefined some RLC utilities (e.g. the logger and collection builders) that are being used by the Modular code.

After this PR, the actual emitted Modular code will not contain anything generated by the RLC emitter, excluding metadata the README and package.json. However, the emitter still relies on the RLC code model to construct the Modular code model. I plan to follow up on decoupling from the RLC code model and metadata generation shortly after this PR.

Fixes #2725.


packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -143,17 +85,18 @@ export function getSendPrivateFunction(
 
 export function getDeserializePrivateFunction(
   operation: Operation,
-  needSubClient: boolean,
-  needUnexpectedHelper: boolean,
+  _needSubClient: boolean,
```
joheredi, line 88: I think we should be able to remove this safely

joheredi (commented):
I think this is looking really good

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
timovv: Removed

qiaozha (commented):
thanks for taking care of this

packages/typespec-ts/src/modular/buildClientContext.ts
```diff
@@ -63,77 +69,100 @@ export function buildClientContext(
...
+    isExported: true
+  });
 
-    factoryFunction = clientContextFile.addFunction({
-      docs: getDocsFromDescription(description),
-      name: `create${name}`,
-      returnType: `${rlcClientName}`,
-      parameters: params,
-      isExported: true
+  const endpointParam = buildGetClientEndpointParam(factoryFunction, client);
```
qiaozha, line 91: feel like we might have missed the default logic handling for other parameters

qiaozha: https://github.com/timovv/autorest.typescript/blob/62159e3252799f32b10629a79dcb3d2552fd219d/packages/rlc-common/src/buildClient.ts#L319-L347

packages/typespec-ts/src/modular/buildLroFiles.ts
```diff
@@ -376,10 +358,6 @@ export function buildGetPollerHelper(
           options.abortSignal?.removeEventListener("abort", abortListener);
           pollOptions?.abortSignal?.removeEventListener("abort", abortListener);
         }
-        if (options.initialRequestUrl || initialResponse) {
-          response.headers["x-ms-original-url"] =
-            options.initialRequestUrl ?? initialResponse!.request.url;
-        }
```
qiaozha, line 382: why this is removed?

packages/typespec-ts/src/modular/buildOperations.ts
```diff
@@ -170,6 +143,11 @@ export function buildOperationFiles(
 
     addImportsToFiles(codeModel.runtimeImports, operationGroupFile);
     addImportBySymbol("serializeRecord", operationGroupFile);
+    addImportBySymbol("buildMultiCollection", operationGroupFile);
```
qiaozha, line 146: does this mean we will generate those helpers regardless if it's applicable or not?

packages/typespec-ts/src/modular/buildHelperSerializers.ts
```diff
@@ -64,8 +110,33 @@ export function emitSerializerHelpersFile(
     symbolMap.set(isRecordElementSupportedFunction.symbol, sourceFile);
   }
 
+  if (!symbolMap.has(buildMultiCollection.symbol)) {
+    symbolMap.set(buildMultiCollection.symbol, sourceFile);
```
qiaozha, line 114: I wonder if we can make this internal helper? modular customer doesn't need this, it's just a RLC helper.

packages/typespec-ts/src/modular/buildPagingFiles.ts
```diff
@@ -124,37 +122,6 @@ export function buildPagingHelpers(
...
   }
 
-  const checkingPagingRequestContent = needUnexpectedHelper
-    ? `if (isUnexpected(response)) {
-      throw createRestError(
-        \`Pagination failed with unexpected statusCode \${response.status}\`,
-        response
-      );
-    }`
-    : `const Http2xxStatusCodes = [
```
qiaozha, line 134: wonder if it could be an issue to change from the enumerate status codes to 200-299 range?

packages/typespec-ts/test/modularIntegration/generated/azure/core/lro/rpc/generated/src/api/operations.ts
```diff
@@ -22,11 +17,7 @@ export function _longRunningRpcSend(
...
   body: GenerationOptions,
   options: LongRunningRpcOptionalParams = { requestOptions: {} },
-): StreamableMethod<
-  | LongRunningRpc202Response
-  | LongRunningRpcDefaultResponse
-  | LongRunningRpcLogicalResponse
-> {
+): StreamableMethod {
   return context
     .path("/azure/core/lro/rpc/generations:submit")
```
qiaozha, line 22: pathUnchecked ?

packages/typespec-ts/src/modular/buildLroFiles.ts
timovv: The reason this header value was being set seems to be because the `isUnexpected` helper was using it. Since we're not using `isUnexpected` anymore I think it's ok to remove. Unless there's another place where we're using this value I don't know about?

packages/typespec-ts/src/modular/buildPagingFiles.ts
timovv: Gut feeling is this would be ok, seems unlikely that any 2xx status code wouldn't be expected (or any non-2xx status wouldn't be unexpected). @joheredi you're touching these helpers in your PR too, right? IIRC you were doing the same thing (ie accepting 2xx status as ok)? Do you think it's OK?

packages/typespec-ts/src/modular/buildClientContext.ts
timovv: I think this is covered here but let me know if there's something missing in that logic:

https://github.com/timovv/autorest.typescript/blob/62159e3252799f32b10629a79dcb3d2552fd219d/packages/typespec-ts/src/modular/helpers/clientHelpers.ts#L79-L93

packages/typespec-ts/src/modular/buildHelperSerializers.ts
timovv: Agreed. These helpers are just used by the modular generated code and aren't exported for customers.

packages/typespec-ts/src/modular/buildOperations.ts
timovv: Yeah, looks like we do generate regardless for now. Ideally we shouldn't be doing this, but I think it's OK for now. I don't want to touch this area more than I have to in this PR since I know Jose is working in this space with his Binder PR. (@joheredi FYI)

packages/typespec-ts/test/modularIntegration/generated/azure/core/lro/rpc/generated/src/api/operations.ts
timovv: Good catch, but unfortunately using `pathUnchecked` doesn't work due to this bug I discovered while trying to apply your suggestion: https://github.com/Azure/azure-sdk-for-js/issues/30654. Using `path` directly bypasses the bug as by default `path` has type `Function`, which is less strict than `pathUnchecked`. I would like to proceed with using `path` for now and once we fix the Core bug we can revisit and change this to `pathUnchecked`.

packages/typespec-ts/src/modular/buildOperations.ts
joheredi: Yeah the binder will make sure helpers that are not used are not generated, so I'm okay keeping this for now and my PR on the Binder will address it

packages/typespec-ts/src/modular/buildClientContext.ts
qiaozha: I did not see the default value was handled in https://github.com/Azure/autorest.typescript/blob/dd505de1900770cb747f3fb73bfc11500c58ae5d/packages/typespec-test/test/parametrizedHost/generated/typespec-ts/src/rest/parametrizedHostClient.ts#L30-L32 and currently it's missing in https://github.com/timovv/autorest.typescript/blob/62159e3252799f32b10629a79dcb3d2552fd219d/packages/typespec-test/test/parametrizedHost/generated/typespec-ts/src/api/parametrizedHostContext.ts#L22-L25

packages/typespec-test/test/NetworkAnalytics.Management/generated/typespec-ts/src/api/networkAnalyticsContext.ts
```diff
@@ -2,30 +2,53 @@
...
     ...options,
     userAgentOptions: { userAgentPrefix },
+    loggingOptions: { logger: options.loggingOptions?.logger ?? logger.info },
+    credentials: {
+      scopes: options.credentials?.scopes ?? [`${endpointUrl}/.default`],
+    },
+  };
+  const clientContext = getClient(endpointUrl, credential, updatedOptions);
+  clientContext.pipeline.removePolicy({ name: "ApiVersionPolicy" });
+  const apiVersion = options.apiVersion ?? "2023-11-15";
```
qiaozha, line 37: @wanlwanl I thnk we need to change the version detection logic accordingly, after this pr get merged from `src/rest/XXXClient.ts` to `src/api/XXXContext.ts`

packages/typespec-ts/src/modular/buildClientContext.ts
qiaozha: I also think this kind of default value change can not be catched by the api view.md @joheredi this is a potential risk if we just check in the api view.md file.

timovv: Thanks. This was happening because we weren't populating the `clientDefaultValue` for these params in the code model, like I was expecting. I updated the code model to set the `clientDefaultValue` to the default value for endpoint path params only, and this fixed the issue. I think this is the right behavior, it doesn't make sense to have a server-side default value for the endpoint.

> I also think this kind of default value change can not be catched by the api view.md

This is true, but it also would have been caught by an integration test (or a scenario test like I was showing yesterday), so maybe we can add coverage there.

joheredi (approved):

packages/typespec-ts/src/modular/buildClientContext.ts
qiaozha: I don't think this is correct, we should respect the parameter default in the `@server` decorators, that's why we should respect the endpoint default value, also in this case, those parameters are parts of the endpoint https://github.com/Azure/autorest.typescript/blob/dd505de1900770cb747f3fb73bfc11500c58ae5d/packages/typespec-test/test/parametrizedHost/generated/typespec-ts/src/rest/parametrizedHostClient.ts#L30-L36 

In this case, if we don't populating the clientDefaultValue, those parameters should be required and go into the getClient method signature instead of options.

timovv: Can you give an example of what this example should look like?

timovv: Is the issue here that `apiVersion` isn't getting a default value when it should be? Or is there something else that is also not being considered?

qiaozha: This is the typespec definition https://github.com/Azure/autorest.typescript/blob/dd505de1900770cb747f3fb73bfc11500c58ae5d/packages/typespec-test/test/parametrizedHost/spec/main.tsp#L20-L32 , we ignore clientDefaultValue for everywhere else but this `@server` decorator. 

if we choose to ignore the default value for it, then we should not handle the default value for endpoint or apiVersion either. the `createParameterizedHost` function should be like 
```ts
export function createParametrizedHost(
  credential: TokenCredential,
  host: string,
  domain: string,
  suffix: string,
  apiVersion: string,
  options: ParametrizedHostClientOptionalParams = {},
): 
```

There're some more context here for how to handle endpoint and how to handle the clientDefaultValue 
https://github.com/Azure/autorest.typescript/pull/2032#discussion_r1349875580 
https://github.com/Azure/autorest.typescript/issues/2049

qiaozha: > Is the issue here that `apiVersion` isn't getting a default value when it should be? Or is there something else that is also not being considered?

we have the default value for apiVersion https://github.com/Azure/autorest.typescript/pull/2561/files#diff-c381779e571bb42446df22900b2c16f1f7036a24f71042ac0bd83e4ef47ff630R28

Issue #2363, opened 2024-03-19
How to update exsting metadata without losing manual config code?


MaryGao:
How to update exsting metadata without losing manual code in generated code?

MaryGao:
We have a customer impact for this issue https://github.com/Azure/azure-rest-api-specs/pull/28383/checks?check_run_id=23571639343, so I marked this as P0.

PR #2732, opened 2024-08-07
Update the existing package.json to correct lro or paging dependencies

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2363

Automatically updates the package.json with correct paging and LRO dependencies for Azure SDK:
- non LRO -> LRO;
- v2 LRO -> v3 LRO;
- non Paging -> Paging;

Other two small changes:
- Add option clearOutputFolder which will be useful in pipeline we could delete output folder
- Change the customization folder to generated


packages/typespec-ts/README.md
```diff
@@ -145,6 +145,14 @@ addCredentials: true
 credentialKeyHeaderName: Your-Subscription-Key
 ```
 
+### clearOutputFolder
+
+If we enable this option `clearOutputFolder` we would empty the whole output folder. By default we only empty the sources folder which means any metadata files will not be removed if it is at project root. This would be useful in pipeline.
+
+```yaml
+clearOutputFolder: true
```
MaryGao, line 153: @wanlwanl Qiaoqiao and I discussed that we could always enable `clearOutputFolder: true` to empty project folder to avoid in-compitible customization code issue.

/cc @qiaozha

packages/rlc-common/src/metadata/buildPackageFile.ts
```diff
@@ -92,6 +133,10 @@ function isAzureMonorepoPackage(model: RLCModel): boolean {
   );
 }
 
+function isAzurePackage(model: RLCModel): boolean {
```
qiaozha, line 136: I remember we have this logic already?

MaryGao: let me check

MaryGao: do you mean these two sperated ones: https://github.com/Azure/autorest.typescript/blob/main/packages/rlc-common/src/metadata/buildPackageFile.ts#L88-L99

packages/typespec-ts/README.md
qiaozha: If automation side could delete the output folder by default, then I guess this option can be omit in the codegen automation config? But I agree this option is necessary for codegen.

packages/rlc-common/src/metadata/buildPackageFile.ts
qiaozha: Can you search all the other places that are using flavor === "azure" now? I wonder if we should make them consistent to leverage the helper here ?

packages/typespec-ts/README.md
wanlwanl: FYI, automation side always clear output folder by default for now.

MaryGao: thanks for this info and i agreed with @qiaozha as codegen we should have this option.

packages/rlc-common/src/metadata/buildPackageFile.ts
MaryGao: resolved.

packages/typespec-ts/test/integration/generated/authentication/api-key/tspconfig.yaml
```diff
@@ -1,12 +0,0 @@
-emit:
```
qiaozha, line 1: why remove this?

qiaozha (approved):

packages/typespec-ts/test/integration/generated/authentication/api-key/tspconfig.yaml
MaryGao: reverted

PR #2743, opened 2024-08-12
set-core-util-dependency-with-binder

### LINKED ISSUE #2733: OpenAI generation missing core-util depedency
See https://github.com/Azure/azure-rest-api-specs/pull/30078/checks?check_run_id=28343657468, we have one automation failure for openai, which is because we have imported `stringToUint8Array` for bytes from core-util, but we didn't add that dependency in package.json 

This is pending on the work for the new binder management @joheredi

qiaozha:
fixes #2733


packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -645,14 +649,16 @@ function buildBodyParameter(
     bodyParameter.type.type === "byte-array" &&
     !bodyParameter.isBinaryPayload
   ) {
-    addImportToSpecifier("coreUtil", runtimeImports, "uint8ArrayToString");
+    const uint8ArrayToStringReference = resolveReference(
+      AzureCoreDependencies.uint8ArrayToString
```
joheredi, line 653: you should be able to do this

```ts
const dependencies = useDependencies();
resolveReference(dependencies.uint8ArrayToString)
```

Which will resolve the appropriate dependency. @azure/core-util if flavor is azure or @typespec/ts-http-runtime otherwise

qiaozha: it works, thanks

packages/typespec-ts/test/modularUnit/modelsGenerator.spec.ts
```diff
@@ -744,7 +744,7 @@ describe("modular encode test for property type bytes", () => {
       `
       export function fooSerializer(item: Foo): Record<string, unknown> {
         return {
-          prop1: uint8ArrayToString(item["prop1"], "base64"),
+          prop1: _PLACEHOLDER_o13_(item["prop1"], "base64"),
```
qiaozha, line 747: need to fix this.

packages/typespec-test/test/NetworkAnalytics.Management/generated/typespec-ts/package.json
```diff
@@ -66,7 +67,6 @@
     "mkdirp": "^3.0.1",
     "typescript": "~5.5.3",
     "tshy": "^2.0.0",
-    "@azure/core-util": "^1.0.0",
```
MaryGao, line 69: just curios why this is in dev dependency?

packages/autorest.typescript/test/smoke/generated/agrifood-data-plane/package.json
```diff
@@ -24,6 +24,7 @@
   ],
   "dependencies": {
     "@azure-rest/core-client": "^1.4.0",
+    "@azure/core-util": "^1.9.2",
```
MaryGao, line 27: is this temp? i remember there is no dependency from swagger RLC.

MaryGao (approved):

PR #2746, opened 2024-08-12
[WIP]Uri template experienmental

MaryGao:


packages/typespec-test/test/faceai/generated/typespec-ts/src/clientDefinitions.ts
```diff
@@ -1637,8 +1637,10 @@ export interface GetLivenessWithVerifySessionAuditEntries {
 export interface Routes {
   /** Resource for '/operations/\{operationId\}' has methods for the following verbs: get */
   (path: "/operations/{operationId}", operationId: string): GetOperationResult;
-  /** Resource for '/detect' has methods for the following verbs: post */
-  (path: "/detect"): DetectFromUrl;
+  /** Resource for '/detect\{?returnFaceId,returnFaceLandmarks,returnFaceAttributes,recognitionModel,returnRecognitionModel,detectionModel,faceIdTimeToLive\}' has methods for the following verbs: post */
+  (
+    path: "/detect{?returnFaceId,returnFaceLandmarks,returnFaceAttributes,recognitionModel,returnRecognitionModel,detectionModel,faceIdTimeToLive}",
+  ): DetectFromUrl;
```
MaryGao, line 1643: @timotheeguerin Just confirmed with you offline, the uriTemplate should be a fully template, as said in the design we should be able to call that in the SDK to form the the template as below.
 
```
expandUriTemplate(route.uriTemplate, {...pathParams, ...queryParams});
```

But when I switched to uriTemplate in JS I found only the query parameter of routed operation would append the query parameter. 

For auto routed operation e.g https://github.com/Azure/autorest.typescript/blob/main/packages/typespec-test/test/faceai/spec/routes.persondirectory.tsp#L162-L168 in which ListRequestOptions would contain query also, the path has no change. Is this a bug?

timotheeguerin: filed bug https://github.com/microsoft/typespec/issues/4153 probably will do a hotfix for it

PR #2754, opened 2024-08-14
Use ESLint flat config files in generated package

jeremymeng:
The legacy config files (.eslintrc.json) are deprecated in latest version of ESLint. This PR updates codegen to use flat config.

- delete .eslintrc.json from azure monorepo generated packages and add eslint.config.mjs for them
- remove `--ext` arguments passed to eslint via command line as they are no longer valid when using flat config


packages/autorest.typescript/package.json
```diff
@@ -36,7 +36,7 @@
     "validate-version-tolerance": "npm run start-test-server:v2 & npm run generate-version-tolerance && npm run generate-version-tolerance:tests && npm run test-version-tolerance && npm run stop-test-server",
     "generate-version-tolerance": "ts-node ./test/commands/version-tolerance.ts",
     "generate-version-tolerance:tests": "ts-node ./test/commands/version-tolerance.ts --test-gen",
-    "regen-test-baselines": "npm run generate-swaggers && npm run rlc-generate-swaggers && npm run smoke-test",
+    "regen-test-baselines": "npm run generate-swaggers && npm run rlc-generate-swaggers && npm run generate-version-tolerance && npm run smoke-test",
```
qiaozha, line 39: I am not sure why we remove the generate-version-tolerance for this case?

jeremymeng: This is manually edited.  It probably was never added to the `regen-test-baselines` script.  I found some baselines were not re-generated thus CI failed.  So I am adding it.

qiaozha: oh okay, thanks.

qiaozha (approved):

PR #2767, opened 2024-08-26
fix-special-union-within-anonymous-model-issue

qiaozha:
fixes https://github.com/Azure/autorest.typescript/issues/2768
fixes https://github.com/Azure/autorest.typescript/issues/2772


packages/typespec-ts/test/modularUnit/scenarios.spec.tmp.ts
qiaozha, line 1: @timovv @joheredi  it seems like we can't load the test correctly with this scenarios.spec.ts, I am renaming it for temporary. let me know if you have any concerns or any thoughts on how to resolve this problem?

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -1072,10 +1072,7 @@ function intOrFloat(value: number): string {
 }
 
 function enumName(name: string): string {
-  if (name.toUpperCase() === name) {
-    return name;
-  }
-  return applyCasing(name, { casing: CASING }).toUpperCase();
```
qiaozha, line 1078: @joheredi do you have any idea why we need to use upper case for enum member name ?

packages/typespec-test/test/NetworkAnalytics.Management/generated/typespec-ts/src/models/models.ts
```diff
@@ -389,8 +389,8 @@ export enum KnownManagedServiceIdentityType {
   SystemAssigned = "SystemAssigned",
   /** UserAssigned */
   UserAssigned = "UserAssigned",
-  /** SystemAssigned, UserAssigned */
-  "SystemAssigned, UserAssigned" = "SystemAssigned, UserAssigned",
+  /** SystemAndUserAssigned */
+  SystemAndUserAssigned = "SystemAssigned, UserAssigned",
```
qiaozha, line 393: I remember we generate the previous way in HLC, @kazrael2119 could you help double check?

qiaozha: see issue https://github.com/Azure/autorest.typescript/issues/2772

packages/typespec-ts/test/modularUnit/serializeUtil.spec.ts
```diff
@@ -1035,6 +1035,198 @@ describe("modular special union serialization", () => {
       true
     );
   });
+
+  it("should generate serialize util if there're special discriminated union within a anonymous model", async () => {
```
MaryGao, line 1039: Could you simplify this case with the root issue? Does the anomymous model happen in body type?

wanlwanl (approved):

packages/typespec-ts/test/modularUnit/serializeUtil.spec.ts
MaryGao: I don't like the idea of directly copying the spec into our UTs because this would introduce a lot of un-necessary changes and also not good for review. we may have plan to remove smoke tests and add more IT or UTs. if these test cases are not small, we may not be effective also.

wanlwanl: I found all other tests in this file test in the same way, since it's quite urget to meet the MPG automation deadline, we may need to keep the same method to test. @raych1 , what do you think?
Would you drive the test improvement in the future? @MaryGao

PR #2775, opened 2024-08-27
Standardize OSS copyright header

jeremymeng:
We should follow the guidance at https://docs.opensource.microsoft.com/releasing/general-guidance/copyright-headers/

This PR updates all non-consistent headers and changes code gen to use the correct casing.


packages/autorest.typescript/src/restLevelClient/generateRestLevel.ts
```diff
@@ -135,7 +135,7 @@ export async function generateRestLevelClient() {
     const isJson = /\.json$/gi.test(filePath);
     const isSourceCode = /\.(ts|js)$/gi.test(filePath);
     let fileContents = fs.readFileSync(filePath);
-    const licenseHeader = `// Copyright (c) Microsoft Corporation.\n// Licensed under the MIT license.\n`;
+    const licenseHeader = `// Copyright (c) Microsoft Corporation.\n// Licensed under the MIT License.\n`;
```
jeremymeng, line 138: codegen was using lower case so I am not 100% sure why some test baselines have upper case and some have lower case.

packages/typespec-ts/src/utils/emitUtil.ts
```diff
@@ -61,7 +61,7 @@ async function emitFile(
   const microsoftHeader = isAzureFlavor
     ? `// Copyright (c) Microsoft Corporation.\n`
     : "";
-  const licenseHeader = `${microsoftHeader}// Licensed under the MIT license.\n`;
+  const licenseHeader = `${microsoftHeader}// Licensed under the MIT License.\n`;
```
jeremymeng, line 64: codegen was using lower case so I am not 100% sure why some test baselines have upper case and some have lower case.

qiaozha (approved):

packages/typespec-ts/src/utils/emitUtil.ts
qiaozha: Could you point out to me where's the upper case baselines test?

jeremymeng: @qiaozha maybe those are outdated? https://github.com/Azure/autorest.typescript/blob/669662f259c957f71474e4e2cd611f9a200dc749/packages/autorest.typescript/test/smoke/generated/storage-resource-manager/samples-dev/operationsListSample.ts#L10

qiaozha: https://github.com/Azure/autorest.typescript/blob/669662f259c957f71474e4e2cd611f9a200dc749/packages/autorest.typescript/src/generators/samples/hlcSamples.ts.hbs#L3 is this the reason?

qiaozha: for sample generation, we use handle bars template.

qiaozha: I think the operation interface is because of this https://github.com/Azure/autorest.typescript/blob/669662f259c957f71474e4e2cd611f9a200dc749/packages/autorest.typescript/src/typescriptGenerator.ts#L120

PR #2776, opened 2024-08-28
Boolean literal type

qiaozha:
fixes https://github.com/Azure/autorest.typescript/issues/2777

### DIFF
--- a/packages/typespec-ts/test/modularUnit/modelsGenerator.spec.ts
+++ b/packages/typespec-ts/test/modularUnit/modelsGenerator.spec.ts
@@ -109,7 +109,7 @@ describe("model property type", () => {
   });
 
 
-  it.only("should handle boolean literal type", async () => {
+  it("should handle boolean literal type", async () => {
     const tspContent = `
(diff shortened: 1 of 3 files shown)


packages/typespec-ts/test/modularUnit/modelsGenerator.spec.ts
```diff
@@ -107,6 +107,65 @@ describe("model property type", () => {
...
     });
   });
+
+
+  it("should handle boolean literal type", async () => {
+    const tspContent = `
+    @doc("The configuration for a streaming chat completion request.")
+    model StreamingChatCompletionOptions {
+      @doc("Indicates whether the completion is a streaming or non-streaming completion.")
+      stream: true;
```
MaryGao, line 117: it seems changing from `true` to `boolean` would become a breaking.

MaryGao (approved):
Approved for this small change.

But the case is interesting that single value becomes a default, iirc this is aligned with hlc and would not introduce breakings between modular and hlc? also any change from single to non-single would introduce sdk breakings?

Issue #2758, opened 2024-08-19
Remove the core-paging dependencies in modular


MaryGao:
See [dicussion](https://github.com/Azure/autorest.typescript/issues/2696#issuecomment-2295709968), since modular has no core-paging dependency and no need to add it into package.json.

PR #2781, opened 2024-08-29
Remove the paging dependency in Modular

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2758


packages/rlc-common/src/metadata/packageJson/azurePackageCommon.ts
```diff
@@ -51,7 +52,7 @@ export function getAzurePackageDependencies({
     };
   }
 
-  if (hasPaging) {
+  if (hasPaging && !isModularLibrary) {
```
MaryGao, line 55:
```suggestion
  if (hasPaging && isModularLibrary !== true) {
```

packages/rlc-common/src/metadata/buildPackageFile.ts
```diff
@@ -100,7 +101,8 @@ export function updatePackageFile(model: RLCModel, existingFilePath: string) {
   }
   const packageInfo = JSON.parse(packageFile.getFullText());
 
-  if (hasPaging) {
+  // no need to update paging dependency for modular libraries
+  if (hasPaging && model.options?.isModularLibrary !== true) {
```
qiaozha, line 105: Can we just set hasPaging as false before generate package.json if it's modular library? Or set it as false from the begining if there's no other places require this hasPaging?

MaryGao: yeah, i also think of this way. i didn't do this because i hope the paging metadata is still correct. overriding it as false can't correctly reflect the package info. so the intention of this change is that `this package has paging operations but we don't add paging dependency in modular`.

not sure you have any strong preference here?

qiaozha (approved):

PR #2803, opened 2024-09-05
Emit orphan models in RLC

joheredi:
This PR enables RLC generating orphan models

### DIFF
--- a/packages/typespec-test/test/contoso/spec/client.tsp
+++ b/packages/typespec-test/test/contoso/spec/client.tsp
@@ -1,0 +1,7 @@
+import "./main.tsp";
+import "@azure-tools/typespec-client-generator-core";
+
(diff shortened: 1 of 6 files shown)


packages/typespec-test/test/contoso/spec/client.tsp
```diff
@@ -0,0 +1,7 @@
+import "./main.tsp";
+import "@azure-tools/typespec-client-generator-core";
+
+using Azure.ClientGenerator.Core;
+
+@@access( Azure.Contoso.WidgetManager.MyModel, Access.public);
+@@usage( Azure.Contoso.WidgetManager.MyModel, Usage.input);
```
MaryGao, line 7: @qiaozha should RLC respect usage and access in client.tsp? does we ignore it by design?

PR #2807, opened 2024-09-09
Support sample.env generation in rlc-common for typespec-ts

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/2806

### DIFF
--- a/packages/rlc-common/src/metadata/buildSampleEnvFile.ts
+++ b/packages/rlc-common/src/metadata/buildSampleEnvFile.ts
@@ -1,0 +1,23 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+import { RLCModel } from "../interfaces.js";
+
+const sampleEnvText = `
+# App registration secret for AAD authentication
+AZURE_CLIENT_SECRET=
+AZURE_CLIENT_ID=
+AZURE_TENANT_ID= 
+`;
+
+export function buildSampleEnvFile(model: RLCModel) {
+  if (
+    model.options?.generateMetadata === true ||
+    model.options?.generateSample === true
+  ) {
+    let filePath = "sample.env";
+    return {
+      path: filePath,
+      content: sampleEnvText.trim()
+    };
+  }
+}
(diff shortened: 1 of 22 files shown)


packages/rlc-common/src/metadata/buildSampleEnvFile.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+import { RLCModel } from "../interfaces.js";
+
+const sampleEnvText = `
+# App registration secret for AAD authentication
+AZURE_CLIENT_SECRET=
```
MaryGao, line 7: Could you check if this is still valid?

packages/rlc-common/src/metadata/buildSampleEnvFile.ts
```diff
@@ -0,0 +1,23 @@
...
+AZURE_CLIENT_ID=
+AZURE_TENANT_ID= 
+`;
+
+export function buildSampleEnvFile(model: RLCModel) {
+  if (
+    model.options?.generateMetadata === true ||
+    model.options?.generateSample === true
```
MaryGao, line 15: add one condition: only generated this file in azure packages

packages/rlc-common/src/metadata/buildSampleEnvFile.ts
```diff
@@ -0,0 +1,23 @@
...
+import { RLCModel } from "../interfaces.js";
+
+const sampleEnvText = `
+# App registration secret for AAD authentication
+AZURE_CLIENT_SECRET=
+AZURE_CLIENT_ID=
+AZURE_TENANT_ID= 
+`;
+
+export function buildSampleEnvFile(model: RLCModel) {
```
MaryGao, line 12: Could you share the failure message with me if no sample.env provided?

kazrael2119: image

MaryGao (changes requested):
Need to verify if the env var works!

packages/rlc-common/src/metadata/buildSampleEnvFile.ts
qiaozha: @HarshaNalluru could you hep check if tthis is expected as of the security concern of spring groove

Issue #2793, opened 2024-09-03
Support explode for query in RLC


MaryGao:

MaryGao:
[...]

PR #2814, opened 2024-09-11
URI template support in RLC

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2793
fixes https://github.com/Azure/autorest.typescript/issues/2792

### Breakings

CollectionFormat with csv and multi are most commonly used cases, for `csv` there is no breaking. And for `multi` there will be a breaking and enabling the option `compatibilityQueryMultiFormat: true` would mitigate this breaking.

CollectionFormat with tsv is no longer supported and we are not aware of any real cases yet.

### Possible values
- allowReserved: true/false
- value: primitive/array/object
- explode: true/false
- style: form/spaceDelimited/pipeDelimited

#### Path allowReserved 
Support to define allowReserved for path parameter but the style and explode support for path parameter is not in-scope of this pr. The default for allowReserved would be 
- allowReserved = false

| allowReserved |      Supported |  
|----------|:-------------:|
| true |  y |
| false |    y|  

#### Query expansion
Please note the allowReserved for query paremeter is not in scope of this pr. Except the form style we would also include the spaceDelimited and pipeDelimited mainly considering backward-compatibility. The default setting would be:
- style = form & explode = false

| Style  | Explode | Uri Template   | Primitive value id = 5 | Array id = [3, 4, 5]    | Object id = {"role": "admin", "firstName": "Alex"} | Supported |
| ------ | ------- | -------------- | ---------------------- | ----------------------- | -------------------------------------------------- |:-------------:|
| form| false   | `/users{?id}`  | `/users?id=5`          | `/users?id=3,4,5`       | `/users?id=role,admin,firstName,Alex`              |  y |
| form*| true*    | `/users{?id*}` | `/users?id=5`          | `/users?id=3&id=4&id=5` | `/users?role=admin&firstName=Alex`                 |  y |
| spaceDelimited| false | n/a | n/a         | /users?id=3%204%205 | n/a            |  y |
| pipeDelimited| false | n/a | n/a         | `/users?id=3\|4\|5` | n/a            |  y |

### Out-of-scope cases
- tsv tab separated format(no real case yet.)
- Header expansion(header design is not changed and any support in header is out-of this pr's scope.)
- Path with explode
- Query with allowReserved


packages/typespec-ts/test/integration/routes.spec.ts
```diff
@@ -32,13 +32,94 @@ describe("RoutesClient Rest Client", () => {
...
-      .path("/routes/path/simple/standard/primitive{param}", "a")
+      .path("/routes/path/reserved-expansion/template/{param}", {
+        value: "foo/bar baz",
+        allowReserved: true
+      })
       .get();
     assert.strictEqual(result.status, "204");
   });
 
+  it("should have explode: true array", async () => {
```
MaryGao, line 45: This case failed with 400 issue and I think we build correct url but failed in server side: https://github.com/Azure/cadl-ranch/issues/731.

packages/typespec-ts/test/integration/routes.spec.ts
```diff
@@ -32,13 +32,94 @@ describe("RoutesClient Rest Client", () => {
...
+            value: ["a", "b"],
+            explode: true,
+            style: "form"
+          }
+        }
+      });
+    assert.strictEqual(result.status, "204");
+  });
+
+  it("should have explode: true record", async () => {
```
MaryGao, line 60: @timovv this case failed with 400 because we didn't prepare the object correctly.

image
image

packages/typespec-ts/test/integration/routes.spec.ts
```diff
@@ -32,13 +32,94 @@ describe("RoutesClient Rest Client", () => {
...
+            value: { a: 1, b: 2 },
+            explode: true,
+            style: "form"
+          }
+        }
+      });
+    assert.strictEqual(result.status, "204");
+  });
+
+  it("should have explode: true primitive", async () => {
```
MaryGao, line 75: This case failed with same reason, I plan to not generate wrapper type for primitive because there explode or style information will have no effect for the url. But @timovv I think you could decide in core side if we need to support primitive value type.

image

packages/typespec-ts/test/integration/routes.spec.ts
```diff
@@ -32,13 +32,94 @@ describe("RoutesClient Rest Client", () => {
     assert.strictEqual(result.status, "204");
   });
 
-  it.skip("should have PathParameters SimpleExpansion Standard primitive", async () => {
+  it("should have allowReserved: true", async () => {
```
MaryGao, line 35: pass.

packages/typespec-ts/test/integration/routes.spec.ts
```diff
@@ -32,13 +32,94 @@ describe("RoutesClient Rest Client", () => {
...
+            value: "a",
+            explode: true,
+            style: "form"
+          }
+        }
+      });
+    assert.strictEqual(result.status, "204");
+  });
+
+  it("should have explode: false array", async () => {
```
MaryGao, line 90: pass.

packages/typespec-ts/test/integration/routes.spec.ts
```diff
@@ -32,13 +32,94 @@ describe("RoutesClient Rest Client", () => {
...
+      .path("/routes/query/query-expansion/standard/record")
+      .get({
+        queryParameters: {
+          param: { a: 1, b: 2 }
+        }
+      });
+    assert.strictEqual(result.status, "204");
+  });
+
+  it("should have explode: false primitive", async () => {
```
MaryGao, line 112: passed.

packages/typespec-ts/test/integration/routes.spec.ts
```diff
@@ -32,13 +32,94 @@ describe("RoutesClient Rest Client", () => {
...
+      .path("/routes/query/query-expansion/standard/array")
+      .get({
+        queryParameters: {
+          param: ["a", "b"]
+        }
+      });
+    assert.strictEqual(result.status, "204");
+  });
+
+  it("should have explode: false record", async () => {
```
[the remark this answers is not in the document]
MaryGao: I re-thinked this a little and it would be problematic if we convert any object to format like a,1,b,2 in core e.g if this is a Date, toString would be date string but we may not know how to format as form style. maybe here we should always generate it explictly in codegen.
As interface Param { value: Record<string, string>; explode: false; style: "form" }.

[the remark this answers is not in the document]
timovv: Honestly it feels strange to me that this is considered a valid case. I don't really understand why we would serialize an object in this way. Is there a reference (specification or something) that describes this particular behavior?

packages/typespec-ts/test/integration/routes.spec.ts
timovv: We _can_ not throw here (in that case we would just ignore `explode` if we don't get an array), but I agree with your plan and think it would be better if we just didn't generate the wrapper in this case, since the combination of primitive and `explode: true` doesn't really make sense.

packages/typespec-ts/test/integration/routes.spec.ts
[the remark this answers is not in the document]
timovv: Never mind, I see it described here: https://swagger.io/docs/specification/v3_0/describing-parameters/. We can implement that, but like you say there might be a problem for objects where `toString` actually provides a meaningful value like Date. Do you know if there are any existing cases where RLC generates a type like that for a query parameter?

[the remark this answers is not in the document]
timovv: How about, if the object has a custom `toString` implementation, use that, otherwise expand it?

packages/autorest.typescript/src/restLevelClient/transforms/transformParameterTypes.ts
```diff
@@ -331,4 +331,4 @@ function getAllOperations(model: CodeModel): Operation[] {
   }
 
   return operations;
-}
+}
```
MaryGao, line 334:
```suggestion
}

```

packages/typespec-ts/test/unit/uriTemplate.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { assert } from "chai";
+import { emitParameterFromTypeSpec } from "../util/emitUtil.js";
+import { assertEqualContent } from "../util/testUtil.js";
+
+describe.only("Client definition generation", () => {
```
MaryGao, line 5:
```suggestion
describe("Client definition generation", () => {
```

packages/typespec-test/test/faceai/generated/typespec-ts/src/serializeHelper.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+export function buildUnexplodedFormStyleValue<ValueType>(value: ValueType) {
```
timovv, line 4: In the azure sdk repo this will probably fail the lint check since exported functions need to have a return type specified. So we might want to do something like:
```suggestion
export function buildUnexplodedFormStyleValue<ValueType>(value: ValueType): { explode: false, style: "form", value: ValueType } {
```
If we do that we should also be able to drop `as const` where we return.

timovv: Also, I don't recall the outcome of the design discussion around this, but I do wonder if the helpers are truly necessary? Seeing them being used in practice I'm not sure if `buildUnexplodedFormStyleValue(value)` is that much clearer than `{ explode: false, style: "form", value }`. The customer still has to specify that they want `explode: false` and `style: "form"` -- they just specify it in the name of the helper function instead of explicitly in the properties on the wrapper.

Maybe we could get away without the additional API surface of the helper function.

packages/typespec-test/test/translator/tspconfig.yaml
```diff
@@ -6,6 +6,7 @@ options:
     generateMetadata: true
     generateTest: true
     azureSdkForJs: false
+    compatibilityMode: true
```
timovv, line 9: `compatabilityMode` is used outside of this area too, right? Do we have documentation somewhere on what exactly `compatabilityMode` does that I can look at? If not, I think a centralized doc which explains what changes when you enable the flag would be quite helpful.

packages/typespec-ts/test/integration/generated/routes/src/index.d.ts
```diff
@@ -363,12 +390,18 @@ export declare interface QueryParametersQueryContinuationExplodeArray204Response
...
+    style: "form";
+}
+
 export declare interface QueryParametersQueryContinuationExplodeArrayQueryParam {
     queryParameters: QueryParametersQueryContinuationExplodeArrayQueryParamProperties;
 }
 
 export declare interface QueryParametersQueryContinuationExplodeArrayQueryParamProperties {
-    param: string[];
+    param: QueryParametersQueryContinuationExplodeArrayParamQueryParam;
```
timovv, line 404: This looks like a breaking change, but it isn't, right? We didn't support this scenario until now?

packages/typespec-test/test/translator/generated/typespec-ts/review/cognitiveservices-translator.api.md
```diff
@@ -97,6 +97,13 @@ export interface BreakSentenceQueryParamProperties {
     script?: string;
 }
 
+// @public (undocumented)
+export function buildExplodedFormStyleValue<ValueType>(value: ValueType): {
```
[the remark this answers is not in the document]
MaryGao: updated.

packages/typespec-test/test/faceai/generated/typespec-ts/src/serializeHelper.ts
MaryGao: > Also, I don't recall the outcome of the design discussion around this, but I do wonder if the helpers are truly necessary?

Yeah we discussed the helper function in original design. Here is the [original playground](https://www.typescriptlang.org/play/?ssl=36&ssc=10&pln=36&pc=30#code/C4TwDgpgBAyqA2EAq5oF4oCIBmB7ATgLaZQA+WAzmAIYDGEAIhPAJaEvAQAmJ5mYLSE1btOPANwAoSSwB2nfNjrQACtXzVCEBQDVq8AK4QAPDpSQAfFADekqPagA3fUYBcUM6ikBfaXIVK9LAQ+CysAF7UwCy4sgCy2tRcUdTGAKLm0BAAHpyyXBRQAEa4uIjUsgA0sJlQOXkFsAjIqFa2DnXZYPC4XBDuGV52DhTN7jCZPn7yIYHQAILwPQDuAEoQFCGO3AnASSk2w-b6K+ub+Ntc7sD4RpK+kqCQUABipQCKRvggL2EKahotAooBgAZptCE9IYTKNQrIAOYAbQAulYAGTBUIRKIxeKJZJ7Yw3IzVHAEYgWKRPaBvXCfEIgABC6jBQJCIKgrIh+ChRmMsLkSNRUAxMBCYRYkWisV2+0JxIgpLwREw6Kgi1OGy2O3xKSkkhyYAIwCg-lmyleHy+IFZKnwuEg+GiG0OHQAjtauQoKO52h0HNg-iF3LT6d9fvB-upwQopP6HEV1D7LXTrcz8F6QnGHL4HmbFBbVrgDJxXQ4aMAABYACgrlfcmCKPSKmAAlO4AOLaKYyGYFoJdk1++zw7TVh3S2Qhq0M232x3Oijtpy4FhcHt9WjwdTQWixUZQLcsCDydxFksQfXYAyyWiT4oGMJcCNRwHc6vOaHuAUIlHL0PWi+ISZvgZb2Pg2gGPgshgf6ho9H01y3IqRz+qMICIA2yrEJUqEdJ+dwdLm0jXre95FI+8BcOmIEfi4-RQD+Qr-jO3w0dGbKgcOUAQcAUEwdxHTwb0DEKrh8b2OhmFYNhmDiRJBEoRJUAnLgaxahc3BIYROb3NIR4nsAAB0dbVo2zZtkZo7ANW3EerOHHcsmgkBkG+DuBRT5ARmjkKNWiKYNQaItqStByVgXBQNkmDIq28nxom+DJp5VHsW+fkBZwowxa2qG5q2+oGfIJlRDW5m4C2rZWWOdmer5ITOXhrmRsGsHKcJiFQGJTXxop7gBUFIVYGFpKRdFyLxcpjFjDJ5KYD1UDeJNHSJY1U0OB1onIctElSQxZIqjtvX0f1mBZcAMVHf6qnqeclzaRAC2+ER9y5UAA). But there is a discussion if we should build a helper per parameter or build the helper per different condition(e.g explode or style). I decided to generate per condition. The main reason is the pattern in one RP is limited and may apply to most parameters.

```
Per one SDK, the explode and style pattern could be limited to one pattern. For example the pattern may fix to explode: false & style: "form".
```

packages/typespec-test/test/translator/tspconfig.yaml
MaryGao: sure, i will add the doc to express that.

packages/typespec-ts/test/integration/generated/routes/src/index.d.ts
```diff
@@ -453,12 +498,18 @@ export declare interface QueryParametersQueryContinuationStandardRecord204Respon
...
+    style: "form";
+}
+
 export declare interface QueryParametersQueryContinuationStandardRecordQueryParam {
     queryParameters: QueryParametersQueryContinuationStandardRecordQueryParamProperties;
 }
 
 export declare interface QueryParametersQueryContinuationStandardRecordQueryParamProperties {
-    param: Record<string, number>;
+    param: QueryParametersQueryContinuationStandardRecordParamQueryParam;
```
qiaozha, line 512: is there a case where value type is array of objects?

packages/typespec-ts/test/integration/azurecore.spec.ts
```diff
@@ -92,42 +92,37 @@ describe("Azure Core Rest Client", () => {
...
+          id: 2,
+          name: "John",
+          etag: "11bdc430-65e8-45ad-81d9-8ffa60d55b5a",
+          orders: [{ id: 2, userId: 2, detail: "a TV" }]
+        }
+      ]
+    };
+    assert.strictEqual(result.status, "200");
+    assert.deepEqual(result.body, responseBody);
   });
```
qiaozha, line 127: why remove the catch error part in this case? what about other cases?

packages/typespec-ts/test/integration/generated/parameters/collection-format/src/index.d.ts
```diff
@@ -87,14 +117,20 @@ export declare interface QueryPipes204Response extends HttpResponse {
...
+
 export declare type QueryPipesParameters = QueryPipesQueryParam & RequestParameters;
 
 export declare interface QueryPipesQueryParam {
     queryParameters: QueryPipesQueryParamProperties;
 }
 
 export declare interface QueryPipesQueryParamProperties {
-    colors: string;
+    colors: QueryPipesColorsQueryParam;
```
qiaozha, line 133: why this is not show as compatibleMode?

qiaozha, line 133: why this is not show as compatibleMode?

packages/typespec-ts/test/integration/generated/parameters/collection-format/src/index.d.ts
```diff
@@ -105,14 +141,20 @@ export declare interface QuerySsv204Response extends HttpResponse {
...
+
 export declare type QuerySsvParameters = QuerySsvQueryParam & RequestParameters;
 
 export declare interface QuerySsvQueryParam {
     queryParameters: QuerySsvQueryParamProperties;
 }
 
 export declare interface QuerySsvQueryParamProperties {
-    colors: string;
+    colors: QuerySsvColorsQueryParam;
```
qiaozha, line 157: same here.

packages/typespec-ts/test/integration/collectionFormat.spec.ts
```diff
@@ -113,18 +103,4 @@ describe("Collection Format Rest Client", () => {
...
-  //           colors: buildMultiCollection(colors, 'colors')
-  //         },
-  //         skipUrlEncoding: true
-  //       });
-  //       assert.strictEqual(result.status, "200");
-  //     } catch (err) {
-  //       assert.fail(err as string);
-  //     }
-  //   });
 });
```
qiaozha, line 106: why is the last case removed?

packages/typespec-ts/test/integration/routes.spec.ts
```diff
@@ -32,13 +35,100 @@ describe("RoutesClient Rest Client", () => {
...
+    assert.strictEqual(result.status, "204");
+  });
+
+  it("should have explode: true record", async () => {
+    const result = await client
+      .path("/routes/query/query-expansion/explode/record")
+      .get({
+        queryParameters: {
+          param: {
+            value: { a: 1, b: 2 },
```
qiaozha, line 79: what would the serialize result would be for record of objects?

packages/typespec-ts/test/modularUnit/operations.spec.ts
```diff
@@ -121,8 +121,12 @@ describe("operations", () => {
           ...Bar): OkResponse;
           `;
 
-      const operationFiles =
-        await emitModularOperationsFromTypeSpec(tspContent);
+      const operationFiles = await emitModularOperationsFromTypeSpec(
+        tspContent,
+        {
+          mustEmptyDiagnostic: false
```
qiaozha, line 127: what diagnostic warning will we get? is it the query format deprecation warning? but why it's not reported before query format should be deprecated a long time ago ?

packages/typespec-ts/test/unit/uriTemplate.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { assert } from "chai";
+import { emitParameterFromTypeSpec } from "../util/emitUtil.js";
+import { assertEqualContent } from "../util/testUtil.js";
+
+describe("Client definition generation", () => {
+  it("should generate method-level parameter", async () => {
+    const clientDef = await emitParameterFromTypeSpec(
+      `
+    @route("template/{+param}")
+    op template(param: string): void;
```
qiaozha, line 10: can you add tests to test against the generated Routes and generated helpers?

packages/typespec-ts/test/util/emitUtil.ts
```diff
@@ -207,9 +210,13 @@ export async function emitParameterFromTypeSpec(
     libraryName: "test",
     schemas: [],
     parameters,
+    helperDetails,
     importInfo: {
       internalImports: importSet,
       runtimeImports: buildRuntimeImports("azure")
+    },
+    options: {
+      sourceFrom: "TypeSpec"
```
qiaozha, line 219: is the default not TypeSpec?

packages/typespec-ts/README.md
```diff
@@ -153,6 +153,16 @@ If we enable this option `clearOutputFolder` we would empty the whole output fol
 clearOutputFolder: true
 ```
 
+### compatibilityMode
```
qiaozha, line 156: can you use a new option, in the url template case, compatibilityMode is not fully backward compatible, you still drop the support for ssv and tsv, pipe etc. While in additionaly properties case, it is really backward compatible.

packages/typespec-ts/test/integration/generated/parameters/collection-format/src/index.d.ts
```diff
@@ -7,13 +7,31 @@ import { StreamableMethod } from '@azure-rest/core-client';
...
+export declare function buildExplodedFormStyle<ValueType>(value: ValueType): {
+    explode: true;
+    style: "form";
+    value: ValueType;
+};
+
 export declare function buildMultiCollection(items: string[], parameterName: string): string;
 
-export declare function buildPipeCollection(items: string[] | number[]): string;
+export declare function buildUnexplodedFormStyle<ValueType>(value: ValueType): {
```
qiaozha, line 18: is it possible to pass the explode and style as parameter in the helper function and build one helper function? I find it quite confusing if there's multiple different helpers. Beside I think typespec/openapi docs do have those information, customers are not stranger to those if they are authoring RLC code from the typespec/openapi references ?

packages/typespec-test/test/faceai/generated/typespec-ts/src/serializeHelper.ts
joheredi: I think @timovv makes a good point. Since these are helper functions, I'd suggest we hold off on implementing them for now. We can gather feedback from users to understand their use cases better and then develop the helper functions based on their input. This approach ensures that we create helpers that are genuinely useful, rather than committing to something we’re not fully certain about.

MaryGao: Thanks all and I will remove this helper for now.

packages/rlc-common/src/buildIndexFile.ts
```diff
@@ -312,7 +313,8 @@ function generateRLCIndex(file: SourceFile, model: RLCModel) {
     hasSsvCollection(model) ||
     hasPipeCollection(model) ||
     hasTsvCollection(model) ||
-    hasCsvCollection(model)
+    hasCsvCollection(model) ||
+    (model.helperDetails?.parameterBuilders ?? []).length > 0
```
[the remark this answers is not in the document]
MaryGao: i plan to remove all logic in modular pr, currently modular still have this dependency.

packages/typespec-ts/test/integration/collectionFormat.spec.ts
MaryGao: no this case anymore

packages/typespec-ts/test/integration/generated/routes/src/index.d.ts
MaryGao: for complicated nested cases e.g objected nested array or array nested objects i don't think we have agreement on how to serialize them correctly. our code is not ready either.

i think this would be a cross-language issue and prefer to defer any discussion for them for now considering no real usage.

packages/typespec-ts/test/integration/routes.spec.ts
MaryGao: as above mentioned.

packages/typespec-ts/test/util/emitUtil.ts
MaryGao: if we didn't set this this would be `undefined`. i just set it explicitly in code.

in our code we would set it as TypeSpec.

packages/typespec-ts/README.md
MaryGao: updated as `compatibilityQueryMultiFormat` option.

packages/typespec-ts/test/integration/generated/parameters/collection-format/src/index.d.ts
```diff
@@ -130,7 +172,7 @@ export declare interface QueryTsvQueryParam {
 }
 
 export declare interface QueryTsvQueryParamProperties {
-    colors: string;
+    colors: string[];
```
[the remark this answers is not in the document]
MaryGao: yes, we only keep original code for multi cases and which is only happened in reality.

packages/typespec-ts/test/integration/azurecore.spec.ts
MaryGao: removed useless codes here. we don't need to catch any error to fail. if exception happens the case would fail directly.

i updated cases i met in this pr. but i remember i told jiaodi to remove for any new cases.

packages/typespec-ts/test/integration/generated/parameters/collection-format/src/index.d.ts
MaryGao: changed the option to compatibilityQueryMultiFormat.

MaryGao: changed the option to compatibilityQueryMultiFormat.

packages/typespec-ts/test/integration/generated/parameters/collection-format/src/index.d.ts
MaryGao: changed the option to compatibilityQueryMultiFormat.

packages/typespec-ts/test/modularUnit/operations.spec.ts
MaryGao: this is a bugfix and here we should report diagnostic of `unable-serialized-type` for header but not: https://github.com/marygao/autorest.typescript/blob/uri-template-support/packages/typespec-ts/src/utils/parameterUtils.ts#L152-L156.

so the new warning is introduced by this pr's fix.

packages/typespec-ts/test/unit/uriTemplate.spec.ts
MaryGao: updated the Routes generation.

packages/typespec-ts/src/utils/parameterUtils.ts
```diff
@@ -0,0 +1,234 @@
...
+          code: "un-supported-format-cases",
+          format: {
+            paramName: parameter.name,
+            explode: String(parameter.explode),
+            format: parameter.format
+          },
+          target: NoTarget
+        });
+      }
+      return buildSerializationInfo("string");
```
MaryGao, line 108: If this is tsv formt and add description for this string type.

packages/typespec-ts/test/integration/generated/routes/src/index.d.ts
MaryGao: create an issue for complicated nested cases.

qiaozha (approved):

packages/typespec-ts/test/integration/generated/routes/src/index.d.ts
MaryGao: https://github.com/Azure/autorest.typescript/issues/2906

PR #2837, opened 2024-09-24
Remove the import logic of paging

kazrael2119:


packages/rlc-common/src/buildSchemaType.ts
```diff
@@ -63,15 +63,6 @@ export function generateModelFiles(
     modelsFile.addInterfaces(objectsDefinitions);
     modelsFile.addTypeAliases(objectTypeAliases);
     modelsFile.addTypeAliases(objectAliases);
-    if (importedModels.has("Paged")) {
```
MaryGao, line 66: Could help double check any other places including Paged, paging or core-paging?

MaryGao (approved):
Approved for removing useless code.

PR #2838, opened 2024-09-24
Support encode int as string

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2676


packages/typespec-ts/src/modular/helpers/typeHelpers.ts
```diff
@@ -37,6 +37,17 @@ const simpleTypeMap: Record<string, TypeMetadata> = {
   unknown: { name: "any" }
 };
 
+function handleSimpleType(
+  type: Type,
+  format?: string
+): TypeMetadata | undefined {
+  if (type.type === "integer") {
+    // Treat integer as number if format is specified as string
```
MaryGao, line 45: Update the comments

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -1260,7 +1261,7 @@ function emitStdScalar(
         format: {
           propertyName: newScalar?.name ?? ""
         },
-        target: newScalar ?? scalar
+        target: NoTarget
```
MaryGao, line 1264: The scalar type is a raw compiler scalar type and logging this type in our side would occupy the terminal but without any value.

packages/typespec-ts/test/modularUnit/scenarios.spec.ts
```diff
@@ -48,9 +48,9 @@ const OUTPUT_CODE_BLOCK_TYPES: Record<string, EmitterFunction> = {
   },
 
   // Snapshot of the entire models file
-  "(ts|typescript) models": async (tsp, {}, namedUnknownArgs) => {
-    const configs = namedUnknownArgs ? (namedUnknownArgs["configs"] as Record<string, string>): {};
-    const result = await emitModularModelsFromTypeSpec(tsp, configs);
+  "(ts|typescript) models": async (tsp, { }, namedUnknownArgs) => {
+    const configs = namedUnknownArgs ? (namedUnknownArgs["configs"] as Record<string, string>) : {};
+    const result = await emitModularModelsFromTypeSpec(tsp, { mustEmptyDiagnostic: false, ...configs });
```
qiaozha, line 53: I wonder if we could add another set of test to allow mustEmptyDiagnostic: false, instead of changing the original test set?

packages/typespec-ts/test/unit/modelsGenerator.spec.ts
```diff
@@ -584,16 +584,16 @@ describe("Input/output model type", () => {
       `
       );
     });
-
-    // TODO: pending with typespec definition https://github.com/microsoft/typespec/issues/2762
-    it.skip("should handle decimal/decimal128 with encode `string`", async () => {
+    it("should handle int/decimal/decimal128 with encode `string`", async () => {
```
qiaozha, line 587:
```suggestion
    it("should handle int/decimal/decimal128/int8 with encode `string`", async () => {
```

packages/typespec-ts/test/unit/modelsGenerator.spec.ts
```diff
@@ -611,6 +611,7 @@ describe("Input/output model type", () => {
       export interface SimpleModel { 
         "prop1": string;
         "prop2": string;
+        "x": string;
```
qiaozha, line 614: why the quote here is not removed? I think it should be removed if no special character in between?

MaryGao: I reduce quote here, actually both are accepted because we would compare after prettier so they are the same syntax.

packages/typespec-ts/test/modularUnit/scenarios.spec.ts
MaryGao: I may not prefer to have another set of test cases and for now I move this to markdown file to feed the config.

qiaozha (approved):

PR #2843, opened 2024-09-25
[engsys] Use dev-tool command in tshy for monorepo

maorleger:
Resolves https://github.com/Azure/azure-sdk-for-js/issues/31211

This PR updates the codegen for ESM packages within the monorepo to use the `build-package` command instead of using tshy directly.

The changes have been made in the azure-sdk-for-js repo, this PR ensures codegen behaves.

## callout

this is my first PR in the autorest.typescript repo so please do assume I forgot / missed something and let me know what was missed. I was surprised that smoke tests did not fail, but maybe there's a different command to run them

### DIFF
--- a/packages/rlc-common/src/metadata/packageJson/buildAzureMonorepoPackage.ts
+++ b/packages/rlc-common/src/metadata/packageJson/buildAzureMonorepoPackage.ts
@@ -130,7 +130,7 @@ function getEsmDevDependencies({
   }
 
   return {
-    tshy: "^1.11.1",
+    tshy: "^2.0.0",
     ...testDevDependencies
   };
 }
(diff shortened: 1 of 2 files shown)


maorleger:
See:

https://github.com/Azure/azure-sdk-for-js/commit/bfefb073bb30f7c393b9c3d86771d93e6e45c782
https://github.com/Azure/azure-sdk-for-js/commit/64c72d5ddac126f8464e226470afd07fe9f90af0

timovv:
I think the lack of changes to the smoke tests is because we don't overwrite the `package.json` if it's already present, we just make updates to certain parts of it. It would be good to keep the test baselines up to date though. You should be able to get the smoke tests up to date by going into the `typespec-test` package and running `rush build -t . && rushx smoke-test`

timovv (approved):

PR #2860, opened 2024-10-17
Test for workloads

kazrael2119:
A draft pr to investigate sample gen


MaryGao:
@kazrael2119 could you try with this pr: https://github.com/Azure/autorest.typescript/pull/2868?

Issue #2830, opened 2024-09-19
Unexpected files under static-helpers are generated


kazrael2119:
https://github.com/Azure/azure-sdk-for-js/pull/30948#discussion_r1766168177

kazrael2119:
another issue: 
if all the operationss in a rp are not lro, codegen will still generate pollingHelpers.ts under static-helpers

if a rp doesn't has paging operations, codegen will still generate pagingHelpers.ts under static-helpers:
https://github.com/Azure/azure-sdk-for-js/pull/31172

PR #2873, opened 2024-10-22
Fix static-helper relative path

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/2830


packages/typespec-ts/src/framework/sample.ts
```diff
@@ -67,7 +67,7 @@ sourceFile2.addStatements(`${functionReference}();`);
 sourceFile2.addStatements(`let obj: ${modelReference} = { id: 1 };`);
 
 // Apply imports to ensure correct references
-binder.resolveAllReferences();
+binder.resolveAllReferences("");
```
qiaozha, line 70:
```suggestion
binder.resolveAllReferences("/modularPackageFolder/src");
```

qiaozha (approved):

PR #2895, opened 2024-11-06
Move to use dev-tool vendored version of cross-env for JS repository

jeremymeng:


jeremymeng:
Thanks for help fixing the CI. For unknown reason, when I run `smoke-test` locally no files were changed.

packages/rlc-common/src/metadata/packageJson/buildAzureStandalonePackage.ts
```diff
@@ -68,6 +68,8 @@ function getStandaloneCjsDevDependencies(config: AzurePackageInfoConfig) {
     "@rollup/plugin-json": "^6.0.0",
     "@rollup/plugin-multi-entry": "^6.0.0",
     "@rollup/plugin-node-resolve": "^13.1.3",
+    ...(config.moduleKind === "cjs" &&
```
MaryGao, line 71: just confirm this dependency `cross-env` is only useful for cjs clients?

MaryGao (approved):
LGTM and small confirmation.

packages/rlc-common/src/metadata/packageJson/buildAzureStandalonePackage.ts
jeremymeng: Yes, ESM uses vitest and won't need cross-env to setup the TS_NODE_OPTIONS

kazrael2119:
need use "dev-tool vendored" for minify, unit-test:node, clean scripts, I will push a commit to update this soon

kazrael2119:
@jeremymeng Do I need to submit a pr to fix https://github.com/Azure/autorest.typescript/issues/2910? looks we should remove uglify-js in devdependency but now it still exists

jeremymeng:
> @jeremymeng Do I need to submit a pr to fix #2910? looks we should remove uglify-js in devdependency but now it still exists

yes please

PR #2908, opened 2024-11-14
check-no-emit-flag

qiaozha:
This PR is trying to resolve the issue where we just open a tsp file and the typespec vscode extension would compile it with noEmit flag, which we should not emit files.


packages/typespec-ts/src/index.ts
```diff
@@ -304,6 +304,9 @@ export async function $onEmit(context: EmitContext) {
     }
 
     binder.resolveAllReferences(modularSourcesRoot);
+    if (program.compilerOptions.noEmit || program.hasError()) {
```
MaryGao, line 307: is this only applied to modular? do we have issue in rlc?

MaryGao, line 307: `program.hasError()` is for what? does it mean compiler error or sth? i just notice if there is error thrown we would fail the generation directly.

MaryGao (approved):

Issue #2882, opened 2024-10-25
Not recommend users to use client_secret


v-jiaodi:
https://github.com/Azure/autorest.typescript/pull/2871#discussion_r1816179552

![Image](https://github.com/user-attachments/assets/1b5b8d89-92eb-48ab-a1c3-8b5201ad6a79)

PR #2922, opened 2024-11-21
Fix readme issues and duplicated licence lines issue

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/2882
fixes https://github.com/Azure/autorest.typescript/issues/2865
fixes https://github.com/Azure/autorest.typescript/issues/2864


qiaozha (approved):

packages/typespec-ts/static/static-helpers/pagingHelpers.ts
```diff
@@ -1,6 +1,3 @@
-// Copyright (c) Microsoft Corporation.
```
MaryGao, line 1: These codes are till in our emitter repo, as files I think we still need these headers.

kazrael2119: emitter will add these two line from https://github.com/kazrael2119/autorest.typescript/blob/b61ee35905401ede30f1597ce49b410dd4a4d732/packages/typespec-ts/src/utils/emitUtil.ts#L61-L65

MaryGao: @lirenhe do you know the guidance here? do we need to add license headers for these static resources? these codes are files not under `src` folder but `static` [folder ](https://github.com/Azure/autorest.typescript/tree/main/packages/typespec-ts/static). And we would copy these files into publish artifacts as static resources also.

FYI - the generated code would always have lisence header.

MaryGao: image

PR #2925, opened 2024-11-22
[autorest.typescript] Allow generating ESM-style imports

maorleger:
This PR adds the `module-kind` flag to autorest options, allowing the generated code to have ESM-style imports/exports. Those may be generated with `.js` extension and/or `index.js` for directory-style imports.

The default remains CJS in autorest, but the idea is to then add this option to each HLC package that migrates to ESM in our repo.

The yaml would look like:
```yaml
module-kind: esm
```

For any package that has already migrated

TODOs:

- [x] Document this flag
- [x] Changelog


packages/autorest.typescript/src/autorestSession.ts
```diff
@@ -44,6 +44,7 @@ export interface AutorestOptions {
   lenientModelDeduplication?: boolean;
   useLegacyLro?: boolean;
   flavor?: PackageFlavor;
+  moduleKind?: "cjs" | "esm";
```
maorleger, line 47: Alternatively this can be `esm?: boolean` since we can only have esm or cjs and cjs is the default

packages/autorest.typescript/src/generators/clientFileGenerator.ts
```diff
@@ -142,7 +144,7 @@ export function generateClient(clientDetails: ClientDetails, project: Project) {
     });
     clientFile.addImportDeclaration({
       namedImports: ["createLroSpec"],
-      moduleSpecifier: `./lroImpl`
+      moduleSpecifier: getImportModuleName(`./lroImpl`, moduleKind)
```
maorleger, line 147: Introduced some convenience for file import (as opposed to a directory import) where you can just pass the cjs name and it'll append `.js` if moduleKind is esm

packages/autorest.typescript/src/generators/clientFileGenerator.ts
```diff
@@ -156,7 +158,7 @@ export function generateClient(clientDetails: ClientDetails, project: Project) {
             true /* shouldGuard */
           )}Impl`
       ),
-      moduleSpecifier: "./operations"
+      moduleSpecifier: getImportModuleName({ cjsName: "./operations", esModulesName: "./operations/index.js" }, moduleKind)
```
maorleger, line 161: For directory imports we can use the original function parameters

packages/autorest.typescript/src/generators/clientFileGenerator.ts
```diff
@@ -168,23 +170,23 @@ export function generateClient(clientDetails: ClientDetails, project: Project) {
...
+      moduleSpecifier: getImportModuleName({ cjsName: "./operationsInterfaces", esModulesName: "./operationsInterfaces/index.js" }, moduleKind)
     });
   }
 
   if (hasInlineOperations && shouldImportParameters(clientDetails)) {
     addTracingOperationImports(clientFile, ".");
     clientFile.addImportDeclaration({
       namespaceImport: "Parameters",
-      moduleSpecifier: "./models/parameters"
+      moduleSpecifier: getImportModuleName("./models/parameters", moduleKind)
```
maorleger, line 181: Another approach that does not require adding this logic on every `addImportDeclaration` / `addExportDeclaration` instance is to update relative imports and exports when [walking the AST for all source code](https://github.com/maorleger/autorest.typescript/blob/autorest-swagger-esm/packages/autorest.typescript/src/typescriptGenerator.ts#L139)

it may be much simpler, but will not be as explicit - curious if y'all have a preference!

maorleger: I created a POC for how that would look in #2926 , would love to hear if you have a preference or a better alternative!

maorleger: Didn't get any feedback so going with this option (the explicit `getImportModuleName` call)

packages/autorest.typescript/src/autorestSession.ts
maorleger: Decided to stick to moduleKind to match existing verbiage in rlc-comment and typespec-ts

maorleger:
@joheredi @sarangan12 @qiaozha @MaryGao  friendly ping on this - could y'all review and provide your feedback and/or approval? 🙏

packages/autorest.typescript/src/utils/nameConstructors.ts
```diff
@@ -0,0 +1,35 @@
...
+ * library being generated.
+ *
+ * @param name The name of the module to import - this can be a string representing the base name of the module or an object with the cjsName and esModulesName properties. *
+ * @param moduleKind The type of module being generated. This can be either "cjs" or "esm".
+ *
+ * @example
+ *
+ * ```ts
+ * getImportModuleName("myModule", "cjs") // returns "myModule"
+ * getImportModuleName("myModule", "esm") // returns "myModule.js"
```
dgetu, line 19: q: Does this handle `getImportModuleName("myModule.js", "esm")`? Or `getImportModuleName("myModule.js", "cjs")`?

dgetu (approved):
LGTM, just some questions out of curiosity :)

packages/autorest.typescript/src/utils/nameConstructors.ts
maorleger: Good Q! Right now it does not. It expects a baseName as a parameter which is what we provide in all usages. If we need to enhance this we can assuming there's a good usecase. Otherwise, we can wait for the need to arise since this is opt-in

qiaozha:
@kazrael2119 can you help verify if the generated code for this works fine?

PR #2926, opened 2024-11-23
WIP - [autorest.typescript] Allow generating ESM imports (ts-morph option)

maorleger:
Another option for adding extensions when the user opts-in to ESM imports. This alternative navigates all source files after codegen and applies the transformation generically

An alternative to #2925

### DIFF
--- a/packages/autorest.typescript/src/typescriptGenerator.ts
+++ b/packages/autorest.typescript/src/typescriptGenerator.ts
@@ -5,8 +5,8 @@ import * as prettier from "prettier";
 import * as fsextra from "fs-extra";
 import * as path from "path";
 import { CodeModel } from "@autorest/codemodel";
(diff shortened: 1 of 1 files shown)


packages/autorest.typescript/src/typescriptGenerator.ts
```diff
@@ -135,21 +135,53 @@ export async function generateTypeScriptLibrary(
     const filePath = file.getFilePath();
     const isJson = /\.json$/gi.test(filePath);
     const isSourceCode = /\.(ts|js)$/gi.test(filePath);
-    let fileContents = fs.readFileSync(filePath);
 
+
+    if (isSourceCode /* && getAutorestOptions().moduleKind === "esm"*/) {
```
maorleger, line 140: We'd have to do the same thing for `export from` declarations

Issue #2917, opened 2024-11-19
Do not generate error response model in MPG if it is not referenced


kazrael2119:
We have two cases for ErrorResponse ErrorDeail ErrorAdditionalInfo models:
1: these three models are all orphan models, in this case we need to remove these three models manually in src

2: ErrorDeail and ErrorAdditionalInfo are referenced from any other models but ErrorResponse not. In this case we only remove ErrorResponse in src

PR #2934, opened 2024-11-29
Upgrade tcgc to ignore error model generation in MPG if it is not referenced

v-jiaodi:
fixes https://github.com/Azure/autorest.typescript/issues/2917


packages/typespec-ts/test/modularUnit/scenarios/models/serialization/errorModels.md
```diff
@@ -0,0 +1,39 @@
...
+
+## TypeSpec
+
+This is tsp definition.
+
+```tsp
+model Foo {
+  name: string;
+}
+model ErrorDetail {
```
MaryGao, line 11: could you add a UT for error detail is referenced?

MaryGao (approved):

qiaozha:
@MaryGao have we reviewed the breaking between HLC and Modular for this?

MaryGao:
@qiaozha The pr is to remove the orphan error models which we agreed to not generate them as a decision during [MPG release](https://github.com/Azure/autorest.typescript/issues/2828). But let me know if we need to revisit this decision.

Issue #2355, opened 2024-03-13
title in tspconfig.yaml is not working in modular


MaryGao:
failed to add title in tspconfig.yaml to set the package client name
![image](https://github.com/Azure/autorest.typescript/assets/98569699/2865dbbd-e438-4910-a432-4001a7137e00)
generated client name:
![image](https://github.com/Azure/autorest.typescript/assets/98569699/b52e121c-a537-496a-8ffd-a9f8b4a3deab)

joheredi:
@MaryGao can you please add details about what's incorrect on the config file?

joheredi:
Following the categorization checklist, I think we should add support for this. I would suggest we do this, making sure that this is an optional configuration option. We should be able to generate even when the option is not provided.


**Breaking Change Categorization**
Bug (Unwanted Breaking Changes):

- [ ] Missing data from the spec in the new implementation.
- [ ] Incorrect results produced by the new implementation.
- [ ] Slight differences in the new implementation that do not improve user experience. If uncertain, categorize as unwanted.
- [x] Taking away a capability that was previously supported

Improvements (Acceptable Breaking Changes):

- [ ] Fixes broken behavior in the new implementation.
- [ ] Offers clear enhancements to user experience.
- [ ] Eliminates or discourages bad patterns.
- [ ] Provides additional functionality.
- [ ] Aligns more closely with TypeSpec standards.

MaryGao:
may be relevant to issue https://github.com/Azure/azure-sdk-for-js/pull/30534#discussion_r1692267688

Issue #2698, opened 2024-07-24
can't set expected client name for readme.md if set @@clientName in client.tsp


kazrael2119:
https://github.com/Azure/autorest.typescript/pull/2697/files#diff-f204b0d6971e33587c7caa3083ebbd3d65060ab2bc38a140336a94b519c3c4e3R7
set client name in client.tsp but this client name doesn't work in readme.md
https://github.com/Azure/autorest.typescript/pull/2697/files#diff-9c8451c7dabaf8ee5e1ccc587fdae3ee0d7df8bb7a49efdfe9d59aa42abc2ee4R31

qiaozha:
In this issue, the client name we used in readme.md should come from Modular layer, however what we are using is the RLC client name. 
The root cause is currently we put the package.json readme.md generation into the rlc-common library, and we are leveraging the RLCModel to get all the information we need. See details from [here](https://github.com/Azure/autorest.typescript/blob/main/packages/rlc-common/src/metadata/buildReadmeFile.ts#L333)

we could have three options to resolve this.
**Option 1**: we add the modularClientName in RLCModel and set this value in [buildCodeModel emitClients](https://github.com/Azure/autorest.typescript/blob/main/packages/typespec-ts/src/modular/buildCodeModel.ts#L1778) 

Pros:
1. the implementation cost is not high.

Cons:  
1. we will probably need more information from modular model in the future. for example, generate some samples in the readme. it's not easy to extend this change to other scenarios.
2. in multi client scenario, we have multiple modular clients, but we should still generate one readme.md file. we need more information and design about how the readme file should be look like.

**Option 2**: it doesn't really quite make sense to put the Modular metadata generation into rlc-common library. we could split the modular metadata generation out from rlc-common, and we could try to pass modular model to buildMetadata. 

Pros:
1. Our code will be more clearer.

Cons:
1. we already have three different scenarios for metadata generation, Azure monorepo, azure standalone package, flavorless scenario, and each of them has to consider esm and commonjs config. it will probably duplicate a lot of effort if we split the modular metadata generation from rlc-common.

**Option 3**: we could change the Modular model interface to link it with RLC model. and try to pass modular model to createMetadata. 

Pros:
1. it might be a little more easy to extend for other scenarios.

Cons:
1. Since we are trying to get rid of Modular code model to use SdkPackage completely, not sure if we still want to add modular and rlc model mapping. and we will probably add new more technical debt to the SdkPackage adoption.

PR #2941, opened 2024-12-05
support title config for typespec generation

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/2355


qiaozha:
we should add some ut for it.

qiaozha:
@kazrael2119 could you resolve the conflicts here?

packages/typespec-test/test/NetworkAnalytics.Management/tspconfig.yaml
```diff
@@ -1,6 +1,8 @@
 emit: ["@azure-tools/typespec-ts"]
 options:
   "@azure-tools/typespec-ts":
+    typespecTitleMap: 
+        NetworkAnalyticsClient: NetworkAnalyticsManagementClient
```
qiaozha, line 5: @joheredi do you have any concerns about this kind of mapping that we provide to customers to modify the client name ? Which we have to in my point of view, because in the case of multiple clients, we can't use a single string option to know which client name we are trying to override.

qiaozha (approved):
LGTM, but let's wait for @joheredi's feedback.

packages/rlc-common/src/interfaces.ts
```diff
@@ -251,6 +251,7 @@ export interface RLCOptions {
   clearOutputFolder?: boolean;
   ignorePropertyNameNormalize?: boolean;
   compatibilityQueryMultiFormat?: boolean;
+  typespecTitleMap?: Record<string, string>;
```
MaryGao, line 254: could you update the readme with this map? i guess the key is typespec name and the value is the name we'd like to rename?

packages/typespec-test/test/NetworkAnalytics.Management/tspconfig.yaml
MaryGao: is the config `title` still working? https://github.com/Azure/autorest.typescript/blob/3454e579e3e312c944055351c4fde0ed0e9fe67b/packages/typespec-ts/README.md#title

packages/rlc-common/src/interfaces.ts
qiaozha: we can't because of this issue https://github.com/Azure/autorest.typescript/issues/2698

MaryGao (approved):
Approved and pls add docs and some confirmation!

packages/typespec-ts/README.md
```diff
@@ -78,12 +78,13 @@ Provide the metadata for `package.json`, `README.md` and user-agent information.
...
-### title
+### typespecTitleMap
 
-Generally the codegen will leverage the title defined in `@client` and `@service` decorator in TypeSpec to name our client. But if you'd like to override it you could config the `title` info.
+Generally the codegen will leverage the title defined in `@client` and `@service` decorator in TypeSpec to name our client. But if you'd like to override it you could config the `typespecTitleMap` info. The key is the client name from typespec, and the value is the client name we'd like to rename
 
 ```yaml
-title: AnomalyDetectorRest
+typespecTitleMap: 
+  AnomalyDetectorClient: AnomalyDetectorRest
```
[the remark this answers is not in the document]
kazrael2119: this only works for modular

packages/typespec-test/test/NetworkAnalytics.Management/tspconfig.yaml
qiaozha, line 5:
```suggestion
        NetworkAnalyticsClient: NetworkAnalyticsManagementApi
```
Change it into a non Client suffix so that it can better reflect when typespecTitleMap is a better option.

PR #2948, opened 2024-12-11
Disable generation if enabling noEmit flag

MaryGao:
Our emitter honored `--no-emit` flag but in the middle of generation, this would lead to some actions taken accidentally like clean up the generation folder etc. It's better to disable generation if enabling noEmit flag at the begining of emitter no matter it is RLC or modular.

This could help fix the files automatical deletion issue in smoke test when typespec code extension enabled.


qiaozha:
⁠Can we special case the clearSrcFolder only to resolve this ? I feel like no-emit doesn't mean to do nothing.

MaryGao:
> ⁠Can we special case the clearSrcFolder only to resolve this ? I feel like no-emit doesn't mean to do nothing.

@qiaozha could you help me understand what is the value we could get from if we don't plan to emit codes? 

I learned from emitter template and other emitters that to skip generation at [the begining](https://github.com/microsoft/typespec/blob/main/packages/compiler/templates/emitter-ts/src/emitter.ts#L4).

Another concern is for performance, during enabling code extension the typespec server side would automatically emit all configed specs, since our emitter has performance issue it would occupy the local machine CPUs. 

{FE4CEAC3-FD4D-45D0-BE24-974D6A7E02AF}

qiaozha:
did you also notice that the above template link is putting no-emit before emitFile?  the same for openapi3 emitter https://github.com/microsoft/typespec/blob/main/packages/openapi3/src/openapi.ts#L264.

MaryGao:
> did you also notice that the above template link is putting no-emit before emitFile? the same for openapi3 emitter https://github.com/microsoft/typespec/blob/main/packages/openapi3/src/openapi.ts#L264.

template may not be a good example because it has limited logic to emitFile only. not sure openapi3 has performance issue but our moduler emitter does have. i'd like to ensure we understand the benefit during this cost paid.

qiaozha (approved):
LGTM, since tcgc doesn't prepare the sdkPackage when it's noEmit, I don't think our own code logic would work as expected with the current implementation either.

PR #2952, opened 2024-12-13
Update todo tsp

v-jiaodi:


packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -501,6 +501,10 @@ function emitBodyParameter(
       isBinaryPayload: isBinaryPayload(context, body.type, contentTypes),
       optional: !isBodyRequired(httpOperation.parameters)
     };
+  } else if (body.bodyKind === "multipart") {
```
qiaozha, line 504: I am not sure I follow the fixes here? we already filtered out the bodyKind for single type which is what we can handle in line 485.

packages/typespec-ts/src/modular/buildCodeModel.ts
```diff
@@ -913,7 +918,10 @@ function emitBasicOperation(
     // Flatten the body parameter if it is an anonymous model
     const originalBodyType = httpOperation.parameters.body.type;
 
-    if (
+    // skip body parameter if it is not a body parameter
+    if (!bodyParameter) {
```
MaryGao, line 922: we crashed with undefined here.

packages/typespec-ts/src/modular/buildCodeModel.ts
MaryGao: This is more for comment purpose. because this is where to handle mfd.

qiaozha:
Can you verify if it works in latest main? if it has been resolved, can you close this pr?

v-jiaodi:
Closed as it works in latest main.

PR #2964, opened 2024-12-26
Add ai chat smoke

v-jiaodi:
Smoke test aichait:
image

Integration test generate code error:
image


qiaozha:
close it as there's dependency inconsistent here.

Issue #2962, opened 2024-12-24
template as model attribute


alfie80:
Hi team,

I met a problem with typespec-ts emitter when using template as model's attribute.

Here is my TypeSpec

```TypeSpec
model FruitRequest<FruitType> {
    order: Order<FruitType>[];
}

model Order<FruitType> {
    kind: FruitType;
    price: float64;
    amount: int32;
}

model Apple {
...
}

model Banana {
...
}

model AppleRequest is FruitRequest<Apple>;
model BananaRequest is FruitRequest<Banana>;

```

So the problem is in generated TypeScript models, it generated `Order` interface twice.
```TypeScript
interface Order {
    kind: Apple;
    price: float64;
    amount: int32;
}

interface Order {
    kind: Banana;
    price: float64;
    amount: int32;
}

interface AppleRequest {
order: Array<Order>;
}
interface BananaRequest {
order: Array<Order>;
}
```

Is it expected results? Or should it have two separate order interfaces, e.g. `AppleOrder` and `BananaOrder`? 
Thanks.

MaryGao:
For template I would always recommand you to give a friendly name with `@friendlyName` decorator([doc](https://typespec.io/docs/standard-library/built-in-decorators/#@friendlyName)).

[playground](https://cadlplayground.z22.web.core.windows.net/?c=QGZyaWVuZGx5TmFtZSgie25hbWV9RnJ1aXRSZXF1ZXN0IiwgxQ9UeXBlKQ0KbW9kZWzGEschPMkfPiB7DQogIG9yZGVyOiBPxAfLHVtdOw0KfQ0KDQrVdMUv1W3QScZma2luZDrKLjvEFHByaWNlOiBmbG9hdDY0xRNhbW91bnQ6IGludDMy6ACBxlhBcHBsZSB7yxJCYW5hbmHNE8Ul5wDqIGlz7gD6xR0%2BO84%2F2C3GHsQu&e=%40typespec%2Fopenapi3&options=%7B%7D)

```tsp
@friendlyName("{name}Order", FruitType)
model Order<FruitType> {
    kind: FruitType;
    price: float64;
    amount: int32;
}
```

So that the client emitter could generate proper names e.g AppleOrder and BananaOrder. But you are right, even without a friendely name emitter should ensure not generate conflicting names with a default naming rule, maybe OrderApple.

alfie80:
[...]

Issue #2958, opened 2024-12-19
ExternalError: Emitter "@azure-tools/typespec-ts" crashed


GraceBrigham:
Tried running `tsp-client update` with this tsp-location.yaml:
```
directory: specification/ai/Azure.AI.Projects
commit: 9071d12cee337c024170f727f1ce53e56f54c54a
repo: Azure/azure-rest-api-specs
```

Got this error:
```
ExternalError: Emitter "@azure-tools/typespec-ts" crashed! This is a bug.
Please file an issue at https://github.com/Azure/autorest.typescript/issues

Error: Unexpected! Inserting syntax kind of InterfaceDeclaration, but ExpressionStatement was inserted.
    at getRangeWithoutCommentsFromArray (C:\Users\nbrigham\Repos\azure-sdk-for-js\sdk\ai\ai-projects\newGenerations\TempTypeSpecFiles\node_modules\ts-morph\dist\ts-morph.js:1355:23)
    at insertIntoBracesOrSourceFileWithGetChildren (C:\Users\nbrigham\Repos\azure-sdk-for-js\sdk\ai\ai-projects\newGenerations\TempTypeSpecFiles\node_modules\ts-morph\dist\ts-morph.js:2601:12)
    at SourceFile._insertChildren (C:\Users\nbrigham\Repos\azure-sdk-for-js\sdk\ai\ai-projects\newGenerations\TempTypeSpecFiles\node_modules\ts-morph\dist\ts-morph.js:11674:20)
    at SourceFile.insertInterfaces (C:\Users\nbrigham\Repos\azure-sdk-for-js\sdk\ai\ai-projects\newGenerations\TempTypeSpecFiles\node_modules\ts-morph\dist\ts-morph.js:11487:25)
    at SourceFile.addInterfaces (C:\Users\nbrigham\Repos\azure-sdk-for-js\sdk\ai\ai-projects\newGenerations\TempTypeSpecFiles\node_modules\ts-morph\dist\ts-morph.js:11481:25)
    at buildParameterTypes (file:///C:/Users/nbrigham/Repos/azure-sdk-for-js/sdk/ai/ai-projects/newGenerations/TempTypeSpecFiles/node_modules/@azure-tools/rlc-common/dist-esm/buildParameterTypes.js:48:28)
[...]

PR #2976, opened 2025-01-02
Fix the duplicated template name without friendlyName in RLC

v-jiaodi:
This issue found in https://github.com/Azure/autorest.typescript/issues/2962 and https://github.com/Azure/autorest.typescript/issues/2958.


packages/typespec-ts/test/unit/modelsGenerator.spec.ts
```diff
@@ -2894,6 +2894,94 @@ describe("Input/output model type", () => {
         `
       });
     });
+
+    it("should generate template model successfully even without @friendlyName", async () => {
```
MaryGao, line 2898: @jiaodi could you add this case into modular also under `modularUnit/scenarios/models/template` folder?

packages/typespec-ts/test/azureIntegration/generated/payload/multipart/src/index.d.ts
```diff
@@ -82,21 +86,27 @@ export declare type FileWithHttpPartOptionalContentTypeRequest = FormData | Arra
...
 }
 
 export declare type FileWithHttpPartSpecificContentTypeRequest = FormData | Array<FileWithHttpPartSpecificContentTypeRequestProfileImagePartDescriptor>;
 
 export declare interface FileWithHttpPartSpecificContentTypeRequestProfileImagePartDescriptor {
     name: "profileImage";
-    body: HttpPart;
+    body: HttpPartFileSpecificContentType;
+    filename?: string;
+    contentType?: string;
```
MaryGao, line 109: @v-jiaodi could you check if the newly-added properties are expected?

packages/typespec-ts/test/azureIntegration/generated/azure/core/lro/standard/src/index.d.ts
```diff
@@ -66,7 +66,7 @@ export declare interface Delete202Headers {
 
 export declare interface Delete202Response extends HttpResponse {
     status: "202";
-    body: OperationStatusOutput;
+    body: OperationStatusErrorOutput;
```
MaryGao, line 69: need to create an issue to track the friendly name issue.

packages/typespec-ts/test/unit/modelsGenerator.spec.ts
v-jiaodi: resolved

packages/typespec-ts/test/azureIntegration/generated/payload/multipart/src/index.d.ts
MaryGao: These properties are added when the type is File. So the change is expected. But the relevant condition has issue https://github.com/azure/autorest.typescript/blob/c21658459be2b17cf9ca226004f7c39e12036c4b/packages/rlc-common/src/buildObjectTypes.ts#L115. I think full support in RLC would be covered by https://github.com/Azure/autorest.typescript/issues/2412.

MaryGao (approved):
LGTM

packages/typespec-ts/test/azureIntegration/generated/payload/multipart/src/index.d.ts
qiaozha: why we need to hack in rlc-common for this? I think we should infer whether it's mfd from it's content type not the generated type name?

qiaozha (approved):
Can you leave a comment in RLC mfd support issue about the hack part?

packages/typespec-ts/test/azureIntegration/generated/payload/multipart/src/index.d.ts
MaryGao: Commented here: https://github.com/Azure/autorest.typescript/issues/2412#issuecomment-2586391463.

MaryGao:
> Can you leave a comment in RLC mfd support issue about the hack part?

Commented here: https://github.com/Azure/autorest.typescript/issues/2412#issuecomment-2586391463.

PR #2983, opened 2025-01-09
Rename the inner method name with `_methodName`

MaryGao:
fixes https://github.com/Azure/autorest.typescript/issues/2985
When building operation interfaces the method `get{Namespace}` is only for inner usage and we should not export this method and it's better to add the underscore prefix to resolve pontential conflicts.


packages/typespec-test/test/widget_dpg/spec/main.tsp
```diff
@@ -177,4 +177,8 @@ interface Budgets {
     // the same name as Widgets.createOrReplace
     @route("/widgets/createOrReplace")
     op createOrReplace is ResourceOperations.LongRunningResourceCreateOrReplace<User>;
+
+    @get getBudgets(
```
MaryGao, line 181: @jiaodi could you add this case into our UT?

qiaozha (approved):

PR #2989, opened 2025-01-10
Generate nullable as type alias and resolve recursive reference in union

v-jiaodi:
fixes https://github.com/Azure/autorest.typescript/issues/2993


packages/typespec-test/test/keyVault.Administration/generated/typespec-ts/review/keyvault-admin.api.md
```diff
@@ -0,0 +1,415 @@
...
+
+// @public
+export type ErrorModel = {
+    code?: string;
+    message?: string;
+    innerError?: ErrorModel_1;
+} | null;
+
+// @public
+export type ErrorModel_1 = {
```
qiaozha, line 32: @maorleger this PR can fix the generation error, but it will have duplicate models issue because of this https://github.com/Azure/typespec-azure/issues/1677#issuecomment-2612003284 And @tadelesh will fix that after the holiday in tcgc. will you be okay with this fix for now and remove the duplicated models while you prepare the PR ? And we will fix this in emitter side after the holiday.
If you agree with this, I will try to release a new emitter with the current fix before the holiday.

packages/typespec-ts/src/framework/hooks/binder.ts
```diff
@@ -286,6 +286,9 @@ class BinderImp implements Binder {
         const importDec = this.addImport(file, declarationSourceFile, name);
         name = importDec.alias ?? name;
       }
+      if (placeholderKey === "__PLACEHOLDER_o16__") {
+        declaration;
+      }
```
qiaozha, line 291:
```suggestion
```

packages/typespec-ts/src/modular/emitModels.ts
```diff
@@ -495,6 +516,9 @@ export function normalizeModelName(
       nameType
     )}>`;
   }
+  if (type.kind === "nullable") {
```
qiaozha, line 519:
```suggestion
  // TODO see https://github.com/Azure/typespec-azure/issues/2125
  if (type.kind === "nullable") {
```

packages/typespec-test/test/keyVault.Administration/generated/typespec-ts/review/keyvault-admin.api.md
maorleger: @qiaozha thanks for the heads up - this will be fine for me as KeyVault's generated code is not directly exported. We have a convenience layer and can smooth over any duplication as this. Thank you!

packages/typespec-test/test/openai_non_branded/generated/typespec-ts/review/openai-non-branded.api.md
```diff
@@ -797,10 +797,16 @@ export interface OpenAIFile {
 }
 
 // @public
-export type Prompt = string | string[] | number[] | number[][];
+export type Prompt = Prompt_1 | null;
 
 // @public
-export type Stop = string | string[];
+export type Prompt_1 = string | string[] | number[] | number[][];
```
MaryGao, line 803: i was thinking if we should delievy an alpha verion for keyvalut first. the correct fix without xxx_1 should be non-breaking, but this pr would introduce a breaking.

packages/typespec-test/test/openai_generic/generated/typespec-ts/review/openai-generic.api.md
```diff
@@ -797,10 +797,16 @@ export interface OpenAIFile {
 }
 
 // @public
-export type Prompt = string | string[] | number[] | number[][];
+export type Prompt = Prompt_1 | null;
 
 // @public
-export type Stop = string | string[];
+export type Prompt_1 = string | string[] | number[] | number[][];
```
MaryGao, line 803: is this because tcgc wrongly tagged this union type as name type but it should be an anoymous one?

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -390,30 +390,33 @@ function getSchemaForUnion(
...
-      });
-      values.push(variantType);
-      if (variantType.typeName) {
-        namedUnionMember = true;
+    } else {
+      for (const variant of variants) {
+        // We already know it's not a model type
+        const variantType = getSchemaForType(dpgContext, variant.type, {
+          ...options,
+          needRef: isAnonymousModelType(variant.type) ? false : true
```
MaryGao, line 410: Can we verify it? Is it because we change the needRef condition so we fix the max stack issue in RLC?
```
union A {
          null,
          A
        }
```

MaryGao (approved):
I don't want to block this but did left some comments, please take a look!

packages/typespec-ts/src/utils/modelUtils.ts
qiaozha: tcgc has a problem handle the case too https://github.com/Azure/typespec-azure/issues/2132

packages/typespec-test/test/openai_generic/generated/typespec-ts/review/openai-generic.api.md
qiaozha: yeah, see issue here https://github.com/Azure/typespec-azure/issues/2133

packages/typespec-test/test/openai_non_branded/generated/typespec-ts/review/openai-non-branded.api.md
qiaozha: breaking only exists when there's a real release, and as far as I know keyvault is the only one that has this issue and is going to release, and @maorleger already said they will not release with this duplicate _N suffix. And it seems like an easy fix and can be easily caught during the review stage.
However, any services that are using recursive self reference union would have the same generation crash issue in pipeline. Compare with the generation crash there, we then spend resources to investigate during this period, I prefer to have a new release out with the latest tag.
/cc @joheredi @xirzec Could you kindly keep that in mind while you review the servics onboarding PR? Thanks.

PR #2995, opened 2025-01-14
Support `@bodyRoot` nested with `@body` cases in Modular

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/3036

Support `@bodyRoot` nested with `@body` cases in modular.

```tsp
op read(@bodyRoot body: {
  @body param?: string;
  @query optionalQuery?: string;
  @header optionalHeader?: string;
}): void;
```


MaryGao:
Closed as root cause resolved: https://github.com/Azure/autorest.typescript/issues/3036#issuecomment-2658346596.

Issue #2988, opened 2025-01-10
[HLC & Modular] Update samples with dotenv changes


MaryGao:
Applied change in https://github.com/Azure/azure-sdk-for-js/pull/32598/commits/6c4c8bba4d756b72de78e2581b408e0d48cbc933

PR #2998, opened 2025-01-14
Update sample template for Modular/RLC/HLC

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/2988


MaryGao:
@kazrael2119 Could you adopt this change for RLC and Modular also?

packages/typespec-ts/src/modular/emitSamples.ts
```diff
@@ -216,6 +216,7 @@ function emitMethodSamples(
     const normalizedDescription =
       description.charAt(0).toLowerCase() + description.slice(1);
     const functionDeclaration: FunctionDeclarationStructure = {
+      returnType: "Promise<void> ",
```
MaryGao, line 219: As I said use exampleFunctionType.retureType instead.

packages/rlc-common/src/static/sampleTemplate.ts
```diff
@@ -48,7 +46,7 @@ async function {{name}}() {
 
 {{/each}}
 
-async function main() {
+async function main(): Promise<void>  {
```
MaryGao, line 49: Remove useless spaces.

MaryGao (commented):
Left comments for implementation part but I don't think we need refreshing all samples.

MaryGao (approved):

PR #3008, opened 2025-01-17
fix-windows-env-issue-in-model-namespace-hierarchy

qiaozha:
As title, there's some issues in building the model namespaces hierarchies in windows environment, 
which cause some of the index.ts files fail to generate and it reports issue like 
image

this pr fixes this and add the todo smoke test into windows environment ci check.


lirenhe (approved):

packages/typespec-ts/src/modular/emitModels.ts
```diff
@@ -249,7 +249,8 @@ export function getModelNamespaces(
   ) {
     if (
       model.clientNamespace.startsWith("Azure.ResourceManager") ||
-      model.clientNamespace.startsWith("Azure.Core")
+      model.clientNamespace.startsWith("Azure.Core") ||
+      model.crossLanguageDefinitionId === "TypeSpec.Http.File"
```
qiaozha, line 253:
```suggestion
      model.crossLanguageDefinitionId === "TypeSpec.Http.File" // filter out the TypeSpec.Http.File model similar like what java does here https://github.com/microsoft/typespec/blob/main/packages/http-client-java/emitter/src/code-model-builder.ts#L2589
```

PR #3014, opened 2025-01-21
Add await for sync function

### LINKED ISSUE #3013: Sample generation is missing `await` for async function
https://github.com/Azure/azure-sdk-for-js/pull/32606

Image

v-jiaodi:
fixes #3013


packages/rlc-common/src/static/sampleTemplate.ts
```diff
@@ -50,7 +50,7 @@ async function {{name}}() {
 
 async function main() {
 {{#each samples}}
-  {{this.name}}();
+  await {{this.name}}();
```
MaryGao, line 53: Could you check if we need this fix for HLC samples?

MaryGao (approved):

PR #3023, opened 2025-01-23
fix operation hierarchies info in sample gen

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/3035

unexpected sample generated if ```hierarchyClient: false```
image


packages/typespec-ts/test/modularUnit/scenarios/samples/operations/multipleExamplesInOneFile.md
```diff
@@ -189,19 +185,15 @@ async function employeesCreateOrUpdateMinage(): Promise<void> {
...
-    "9KF-f-8b",
-    {
-      properties: {
-        age: 1,
-        city: "gydhnntudughbmxlkyzrskcdkotrxn",
-        profile: "ms",
-      },
-      tags: { key2913: "urperxmkkhhkp" },
-      location: "itajgxyqozseoygnl",
+  const result = await client.createOrUpdate("rgopenapi", "9KF-f-8b", {
```
qiaozha, line 188: @MaryGao Could you help check if we need to manual set hierarchyClient true or enableOperationGroup true for those unit test?

MaryGao: yes, this is not only for sample but also for all UTs because we didn't adopt the transformRLCOption which have any default setting logic.

not the current priority but i think we could refactor this in future.

qiaozha (approved):

MaryGao (approved):
I found the root cause here:
During our sample UT, we didn't call `getMethodHierarchiesMap` explicitly which will change the method name in some condition. So all UTs passed.

But during real release generation we will always call `getMethodHierarchiesMap` so the method name would be like `Employees_delete`. 

To reproduce it we could update nothing but add a call `getMethodHierarchiesMap` before any sample gen starts. 

But I think the fix is correct and so I approved.

PR #3042, opened 2025-01-31
remove references of api-extractor-json-types and package-json-module rules

jeremymeng:
As these rules are removed in JS repo PR https://github.com/Azure/azure-sdk-for-js/pull/32820


MaryGao (approved):

packages/typespec-test/test/ai/generated/typespec-ts/eslint.config.mjs
```diff
@@ -4,10 +4,8 @@ export default azsdkEslint.config([
   {
     rules: {
       "@azure/azure-sdk/ts-modules-only-named": "warn",
-      "@azure/azure-sdk/ts-apiextractor-json-types": "warn",
       "@azure/azure-sdk/ts-package-json-types": "warn",
       "@azure/azure-sdk/ts-package-json-engine-is-present": "warn",
-      "@azure/azure-sdk/ts-package-json-module": "off",
```
MaryGao, line 10: just curious why we remove this rule? not change to following

```
{
  "module": "./dist/esm/index.js"
}
```

https://github.com/azure/autorest.typescript/blob/main/packages/typespec-test/test/NetworkAnalytics.Management/generated/typespec-ts/sdk/test/arm-test/package.json#L205

jeremymeng: We are moving away from using eslint to check for simple things like certain field value in package.json etc. in favor of the dev-tool check. https://github.com/Azure/azure-sdk-for-js/blob/2b05e1f81a81095709c64529d830f6cf4495cca4/common/tools/dev-tool/src/checks/packageJson.ts#L25

We should add check there instead.

jeremymeng: > And I notice we have the guideline - https://azure.github.io/azure-sdk/typescript_implementation.html#ts-package-json-module, do we need to update it also?

Yeah some of the guidelines are outdated after migrating to ESM. We should address that as well.

MaryGao: Got it and thanks for sharing!

PR #3044, opened 2025-02-03
Use recommended boilerplate LICENSE

jeremymeng:
per https://docs.opensource.microsoft.com/releasing/general-guidance/boilerplate-license-files/

we don't need the year in there.


xirzec (approved):

packages/autorest.typescript/src/generators/static/licenseFileGenerator.ts
```diff
@@ -5,9 +5,9 @@ import { Project } from "ts-morph";
 import { getAutorestOptions } from "../../autorestSession";
 
 const mitLicenseText = `
-The MIT License (MIT)
+Copyright (c) Microsoft Corporation.
 
-Copyright (c) ${new Date().getFullYear()} Microsoft
```
qiaozha, line 10: I have a minor question regarding License, @jeremymeng @xirzec do you think if it still makes sense to generate MIT license of Microsoft Corporation for unbranding scenarios?

jeremymeng: @qiaozha Great question! It probably should be up to customers to choose licenses for their generated code.  This is the same across languages. Better to confirm with codegen crew.

PR #3047, opened 2025-02-06
[EngSys] Use agents from 1ES pool

mikeharder:


.scripts/smoke-test.yml
```diff
@@ -9,9 +9,12 @@ jobs:
     strategy:
       matrix:
         Linux_Node18:
-          Pool: ${{ parameters.LinuxPool }}
+          Pool: azsdk-pool-mms-ubuntu-2004-general
```
weshaggard, line 12: Do you not want to use to the new pool?

mikeharder: I'm only touching this pipeline because I noticed it was using the "Azure Pipelines" pool, which should be reserved for Mac agents.  I'd rather wait to convert this whole repo to the new pool.

weshaggard (approved):

PR #3058, opened 2025-02-18
Remove @microsoft/api-extractor from common dev dependencies of generated packages

jeremymeng:
Flavorless already adds it https://github.com/Azure/autorest.typescript/blob/0c74c64758e76ecfd5baaa798ea8198e50bc7a17/packages/rlc-common/src/metadata/packageJson/buildFlavorlessPackage.ts#L21

This PR removes it from common dev dependencies and ensure Azure standalone has it.

Azure js mono repo relies on dev-tool to extract api so this dependency is not needed.


packages/typespec-test/test/hierarchy_generic/generated/typespec-ts/package.json
```diff
@@ -70,6 +69,7 @@
     "vitest": "^3.0.3",
     "@azure-tools/test-credential": "^2.0.0",
     "@azure-tools/test-recorder": "^4.0.0",
+    "@microsoft/api-extractor": "^7.40.3",
```
qiaozha, line 72: I don't think we are leveraging dev tool in non azure generation?

jeremymeng: That's right. We need this dependency and the script below runs `api-extractor run --local`.  Only azure mono repo uses dev-tool and won't need this dependency.

qiaozha:
@v-jiaodi please help us verify if the api-extractor will be removed in azure-sdk-for-js is true case.

v-jiaodi:
> @v-jiaodi please help us verify if the api-extractor will be removed in azure-sdk-for-js is true case.

@qiaozha I tried to set `azureSdkForJs: true` for smoke test, but the api-extractor was not removed.

jeremymeng:
> @qiaozha @jeremymeng I tried to set `azureSdkForJs: true` for smoke test, but the api-extractor was not removed.

Interesting. Did you notice any package.json files that contain both @azure/dev-tool and api-extractor?  I also couldn't find any removal of api-extractor in my changes.

qiaozha (approved):

qiaozha:
Going to merge it as offline confirmed with @v-jiaodi this is working expected in sdk repo.

PR #3108, opened 2025-03-14
Fix incorrect references to discriminated model serializers

joheredi:
Fix issue where the emitter tries to reference serializers with the wrong name.

### DIFF
--- a/packages/typespec-ts/test/modularUnit/scenarios/models/deserialization/discriminatedUnion.md
+++ b/packages/typespec-ts/test/modularUnit/scenarios/models/deserialization/discriminatedUnion.md
@@ -1,0 +1,91 @@
+# Should generate serializers for discriminated model
+
+Verify that the serializers are correctly referenced within the switch statement of the base serializers.
+
+## TypeSpec
+
+This is tsp definition.
+
+```tsp
+@discriminator("kind")
+model AWidgetData {
+    kind: string;
+}
+
+model AOAIModelConfig extends AWidgetData {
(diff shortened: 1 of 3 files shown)


packages/typespec-ts/test/modularUnit/scenarios/models/deserialization/discriminatedUnion.md
```diff
@@ -0,0 +1,91 @@
...
+  };
+}
+
+/** Alias for AWidgetDataUnion */
+export type AWidgetDataUnion = AoaiModelConfig | MaasModelConfig | AWidgetData;
+
+export function aWidgetDataUnionDeserializer(item: any): AWidgetDataUnion {
+  switch (item.kind) {
+    case "kind0":
+      return aoaiModelConfigDeserializer(item as AoaiModelConfig);
```
MaryGao, line 54: I would assume this is already fixed in my pr: https://github.com/Azure/autorest.typescript/pull/3102.

MaryGao (approved):

packages/typespec-ts/test/modularUnit/scenarios/models/deserialization/discriminatedUnion.md
joheredi: I think the issue is still present there, normalizing twice the subTypeName seems to be the cause of the inconsistency. The long term fix though should be using the binder and resolveReference. But wanted to do the quick fix to unblock our partners.

PR #3111, opened 2025-03-14
Fix some warnings for next version

v-jiaodi:


packages/typespec-test/test/translator/spec/text-translation/models-translate.tsp
```diff
@@ -12,9 +12,7 @@ namespace Translator.TextTranslation;
 model TranslateParameters {
     ...CommonParameters;
 
-    @query({
-      format: "multi"
```
qiaozha, line 16: this is multi collection format, we should set explode as true

qiaozha:
close it as no need.

PR #3124, opened 2025-03-20
Basic HttpPart support for RLC

timovv:
Basic support for `HttpPart` in RLC plus new cadl-ranch cases


packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -171,6 +174,40 @@ export function getSchemaForType(
   }
 
   if (type.kind === "Model") {
+    const httpPart = getHttpPart(program, type);
```
MaryGao, line 177: Should we double check the bodyKind also?  https://github.com/witemple-msft/typespec/blob/witemple-msft/http-file-body/packages/http/src/types.ts#L486-L492

See [issue ](https://github.com/microsoft/typespec/issues/6530)here for httppart used in common content type.

packages/typespec-ts/test/integration/generated/payload/multipart/src/index.d.ts
```diff
@@ -26,31 +26,31 @@ export declare type ComplexHttpPartsModelRequest = FormData | Array<ComplexHttpP
 
 export declare interface ComplexHttpPartsModelRequestAddressPartDescriptor {
     name: "address";
-    body: HttpPartAddress;
+    body: Address;
```
MaryGao, line 29: Another spector change would happen soon - https://github.com/microsoft/typespec/pull/6533.

MaryGao:
I know the pr is in draft and just share my early feedbacks.

joheredi (commented):
The changes look reasonable to me. Let me know when you feel this is ready for another review

packages/typespec-ts/src/utils/modelUtils.ts
timovv: I had a look at this and it's a bit complicated to check the bodyKind since we don't have access to the operation definition here. I'd like to get this basic implementation checked in first to unblock the main use case and will open an issue to revisit this. Hopefully people won't be using HttpPart outside of MFD very often.

MaryGao (approved):

packages/typespec-ts/src/utils/modelUtils.ts
timovv: https://github.com/Azure/autorest.typescript/issues/3144

Issue #3077, opened 2025-02-26
Improve the generateTest  config when it is absent


MaryGao:
Improve the generateTest  config when it is absent

kazrael2119:
if  azure-sdk-for-js and generateTest or generate-test is undefined:
-  if there is no test folder & undefined, enable generateTest as true;
-  if there is test foler, do nothing

PR #3125, opened 2025-03-21
Improve generateTest config 

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/3077

1： if we manually set generateTest as false
image
emitter will not generate test folder and ralated dependencies and configs
image

2： if we don't set generate-test in tsconfig
image
it could generate test folder
image

### DIFF
--- a/packages/typespec-ts/src/index.ts
+++ b/packages/typespec-ts/src/index.ts
@@ -140,10 +140,10 @@ export async function $onEmit(context: EmitContext) {
   console.timeEnd("onEmit: load static helpers");
   const extraDependencies = isAzurePackage({ options: rlcOptions })
     ? {
-        ...AzurePollingDependencies,
-        ...AzureCoreDependencies,
-        ...AzureIdentityDependencies
-      }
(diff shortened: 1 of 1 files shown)


packages/typespec-ts/src/index.ts
```diff
@@ -390,7 +390,18 @@ export async function $onEmit(context: EmitContext) {
     const shouldGenerateMetadata =
       option.generateMetadata === true ||
       (option.generateMetadata === undefined && !hasPackageFile);
-    // TODO detect whether test folder exists. If yes generateTest should be false.
+    const existingTestFolderPath = join(
+      dpgContext.generationPathDetail?.metadataDir ?? "",
+      "test"
+    );
+    const hasTestFolder = await existsSync(existingTestFolderPath);
+    if (option.azureSdkForJs) {
```
MaryGao, line 398: if(option.generateTest === undefined) {
    option.generateTest = option.azureSdkForJs && hasTestFolder ? false: true;
}

MaryGao (approved):

PR #3133, opened 2025-03-25
ai chat protocol smoke test

### LINKED ISSUE #3134: ai chat protocol generation code build error
Image

v-jiaodi:
fixes #3134 
Build error:
image


packages/typespec-ts/test/modularUnit/scenarios/models/models.md
```diff
@@ -0,0 +1,35 @@
...
+  index: 123;
+}
+op read(@path id: string): {
+  @bodyRoot result: StreamingChatCompletionOptions
+};
+```
+
+## Models
+
+```ts models
```
qiaozha, line 18: can you also add ts operations for it?

packages/typespec-ts/test/modularUnit/scenarios/models/models.md
```diff
@@ -0,0 +1,96 @@
...
+## TypeSpec
+
+```tsp
+model StreamingChatCompletionOptions {
+  stream: true;
+  messages: "aaaaa";
+  index: 123;
+}
+op read(@path id: string): {
+  @bodyRoot result: StreamingChatCompletionOptions
```
qiaozha, line 12: Please also set request body for this so that we can test both request and response. thanks

qiaozha (approved):

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -1350,7 +1357,10 @@ export function serializeRequestValue(
     case "model": // this is to build serialization logic for spread model types
       return `{${getRequestModelMapping(context, type, "").join(",")}}`;
     case "constant":
-      return `${nullOrUndefinedPrefix}${getConstantValue(type)}`;
+      if (isTopLevel) {
```
MaryGao, line 1360: not quite understand this change, why top-level is different that other levels?

MaryGao (approved):

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
MaryGao: because for top-level we skip these params in signiture so we would fallback to constant, but other levels we could get that property directly.

Issue #3195, opened 2025-05-14
Need to verify whether we still need to upgrade dependency versions in package.json after pnpm migration


kazrael2119:
Migration tool will update some dependencies to the latest version, so we need verify whether we still need to upgrade dependency versions in package.json.

qiaozha:
I don't think this is after pnpm migration. We should call out if we found any dependency are not in the workspace or catalog.

PR #3200, opened 2025-05-16
Support option for rlc swagger and fix missing dependencies

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/3195

### DIFF
--- a/packages/autorest.typescript/test/rlcIntegration/generated/multipleUrlParameters/package.json
+++ b/packages/autorest.typescript/test/rlcIntegration/generated/multipleUrlParameters/package.json
@@ -57,6 +57,8 @@
     "@azure/eslint-plugin-azure-sdk": "^3.0.0",
     "@azure/identity": "^4.6.0",
     "@types/node": "^18.0.0",
+    "@eslint": "^9.9.0",
+    "autorest": "latest",
     "typescript": "~5.8.2"
   },
   "scripts": {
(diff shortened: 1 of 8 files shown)


packages/rlc-common/src/metadata/packageJson/buildAzureMonorepoPackage.ts
```diff
@@ -86,6 +86,8 @@ export function getAzureMonorepoDependencies(config: AzureMonorepoInfoConfig) {
         : "workspace:*",
       "@azure/identity": !shouldUsePnpmDep ? "^4.6.0" : "catalog:internal",
       "@types/node": !shouldUsePnpmDep ? "^18.0.0" : "catalog:",
+      "@eslint": !shouldUsePnpmDep ? "^9.9.0" : "catalog:",
+      ...(config.specSource === "Swagger" && { autorest: "latest" }),
```
MaryGao, line 90: why we have this?

jeremymeng: @MaryGao It may be historic convention that we just install `autorest` globally.  I don't see any issue with adding it to the catalog.  Should we use "latest" mostly?

MaryGao: Yes, for autorest it is mainly for legacy SDKs which is generated from swagger path. I think it would become less and less in that path. 

BTW do you think if we should manage all deps into catalog/workspace in SDK repo?

kazrael2119: @jeremymeng We plan to use ``` autorest: "catalog:" ``` and merge this pr. So we need to add this dependency into catalog in SDK repo. Do you have any concerns?

MaryGao (approved):

packages/autorest.typescript/test/rlcIntegration/generated/multipleUrlParameters/package.json
```diff
@@ -57,6 +57,8 @@
     "@azure/eslint-plugin-azure-sdk": "^3.0.0",
     "@azure/identity": "^4.6.0",
     "@types/node": "^18.0.0",
+    "@eslint": "^9.9.0",
```
MaryGao, line 60: eslint

PR #3210, opened 2025-05-23
Remove rush dependency

v-jiaodi:


MaryGao:
@v-jiaodi could you work with @wanlwanl to verify if the floating versions could also build successfully with pnpm commands?

If yes, I think we don't need to upgrade emitter versions with @jeremymeng's migration pr. we could upgrade emitters after that.

MaryGao:
@jeremymeng Thanks for the clarification! What is the plan to switch to catalog or workspace versions?

MaryGao:
@v-jiaodi Could you fix the ci issue for this pr? SDK repo have contributed this change: https://github.com/Azure/autorest.typescript/pull/3210.

v-jiaodi:
> @jiaodi Could we focus on the dev part first https://github.com/Azure/azure-sdk-for-js/blob/f6e317aec3cc878925002e648712fd278ec2ea32/sdk/quota/arm-quota/package.json#L85-L99?

The current status is just like this.

packages/autorest.typescript/src/autorestSession.ts
```diff
@@ -44,8 +44,6 @@ export interface AutorestOptions {
   lenientModelDeduplication?: boolean;
   useLegacyLro?: boolean;
   flavor?: PackageFlavor;
-  //TODO should remove this after finish the release tool test
-  shouldUsePnpmDep?: boolean;
```
MaryGao, line 48: i remember automation has used this option, @skywing918 do we need any update there?

packages/rlc-common/test/integration/packageJson.spec.ts
```diff
@@ -346,7 +346,7 @@ describe("Package file generation", () => {
         "test:browser",
         "dev-tool run build-test && dev-tool run test:vitest --browser"
       );
-      expect(packageFile.scripts).to.have.property("pack", "npm pack 2>&1");
+      expect(packageFile.scripts).to.have.property("pack", "pnpm pack 2>&1");
```
MaryGao, line 349: why we have this diff?

packages/autorest.typescript/src/autorestSession.ts
v-jiaodi: `shouldUsePnpmDep` is not used in automation.

packages/autorest.typescript/test/integration/generated/additionalProperties/package.json
```diff
@@ -6,8 +6,8 @@
   "version": "1.0.0-preview1",
   "engines": { "node": ">=20.0.0" },
   "dependencies": {
-    "@azure/core-client": "^1.9.3",
-    "@azure/core-rest-pipeline": "^1.19.1",
+    "@azure/core-client": "^1.9.2",
```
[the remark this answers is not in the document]
v-jiaodi: This was an incorrect change. It has already been reverted.

Issue #3228, opened 2025-05-29
[sample generation] Fail to generate sample for `unknown` type with object values


v-jiaodi:
Fail to generate sample for `unknown` type with object values

tsp: 
```
model Test {
  baseProfile?: Record<unknown>;
}
```

generate sample fail:

![Image](https://github.com/user-attachments/assets/08359fd5-51cf-4fcc-9194-164aca97934c)

PR #3229, opened 2025-05-29
Support `unknown` type in sample generation

v-jiaodi:
fixes https://github.com/Azure/autorest.typescript/issues/3228


packages/typespec-ts/test/modularUnit/scenarios/samples/parameters/parameterTypesCheck.md
```diff
@@ -11,6 +11,7 @@ model Foo {
 }
 
 model Widget {
+  unknownValue: unknown;
```
MaryGao, line 14: @jiaodi add more cases with unknown type here:
for array, string, number, null and boolean.

MaryGao (approved):

PR #3246, opened 2025-06-10
[sample generation] contoso smoke for ```Type 'string' is not assignable to type 'Uint8Array'```

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/3247


qiaozha:
close it please fix it in test gen pr.

PR #3285, opened 2025-06-13
Fix property normalization inconsistancy among sample and client generation

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/3286


packages/typespec-ts/test/modularUnit/scenarios/modelsGenerator/modelsGenerator.md
```diff
@@ -2637,3 +2637,43 @@ export function baseSerializer(item: Base): any {
   return { foo: item["foo"] };
 }
 ```
+
+# only: should handle model property name with capitalized first letter
```
MaryGao, line 2641: @kazrael2119 this is an sample generation failure, pls move this to sample UT folder. And we need to generate two parts for 1) options and 2) relevant samples.

Issue #3155, opened 2025-04-09
Ensure no compile issue for query parameter re-injection in Paging


MaryGao:
Ensure there is no compile issue for this case for both RLC and Modular

MaryGao:
https://azure.github.io/typespec-azure/docs/howtos/generate-client-libraries/13pagingoperations/#parameterized-next-links-against-guidelines

MaryGao:
AppConfig(RLC)/KeyVault(Modular) has this requirement.

qiaozha:
I don't think this is a p0 item, if you mean to make sure the compile works, then please change the title as appropriate or create another issue to distinguish the real support of this feature.

MaryGao:
@v-jiaodi Please help ensure there is no compile issue for this case for both RLC and Modular. We only need to support the first paging call could pass our UTs.

PR #3289, opened 2025-06-16
Ensure no compile issue for query parameter re-injection in Paging

v-jiaodi:
fixes https://github.com/Azure/autorest.typescript/issues/3155


packages/typespec-ts/test/azureModularIntegration/azureCorePage.spec.ts
```diff
@@ -71,4 +71,12 @@ describe("Page Client", () => {
     assert.strictEqual(items[0]?.id, validUser.id);
     assert.strictEqual(items[0]?.etag, validUser.etag);
   });
+
+  it("should list core page withParameterizedNextLink", async () => {
+    // To Do: This test is failing because the mock API does not support the next link parameterization.
```
MaryGao, line 76:
```suggestion
    // TODO: We can only get the first page data because the parameter re-injection is not implemented yet.
```

packages/typespec-ts/test/azureIntegration/azureCorePage.spec.ts
```diff
@@ -114,4 +114,24 @@ describe("Azure Core Page Rest Client", () => {
     assert.strictEqual(result[0]?.name, "Madge");
     assert.strictEqual(result[0]?.etag, "11bdc430-65e8-45ad-81d9-8ffa60d55b59");
   });
+
+  it("should list core page withParameterizedNextLink", async () => {
+    // Expected query parameters on initial request: includePending=true, select=name
+    // To Do: Implement the pagination logic to handle the next link with parameters
```
MaryGao, line 120:
```suggestion
    // TODO: We can only get the first page data because the parameter re-injection is not implemented yet.
```
```

MaryGao (approved):

PR #3297, opened 2025-06-17
Update tcgc to latest version

v-jiaodi:


lirenhe:
Can we use copilot to handle the upgrade work?

packages/typespec-ts/package.json
```diff
@@ -114,7 +114,7 @@
   },
   "peerDependencies": {
     "@azure-tools/typespec-azure-core": "^0.57.0",
-    "@azure-tools/typespec-client-generator-core": "^0.57.0",
+    "@azure-tools/typespec-client-generator-core": "^0.57.1",
```
MaryGao, line 117: In the latest tcgc version the current RLC iterations would break among functions listOperationGroups and listOperationsInOperationGroup due to cache issues. So all RLC generation would fail.

MaryGao:
Close due to duplicated with https://github.com/Azure/autorest.typescript/pull/3302

PR #3317, opened 2025-06-23
improve generate-metadata

kazrael2119:


packages/typespec-ts/src/index.ts
```diff
@@ -403,8 +403,7 @@ export async function $onEmit(context: EmitContext) {
     );
     const hasPackageFile = await existsSync(existingPackageFilePath);
     const shouldGenerateMetadata =
-      option.generateMetadata === true ||
-      (option.generateMetadata === undefined && !hasPackageFile);
+      option.generateMetadata === true || !hasPackageFile;
```
MaryGao, line 406: What if data-plane SDKs don't want to generate metadata so they turn off the option "generate-medadata=false"?

This may happen in local experience where service team needs to refresh their SDKs.

MaryGao (changes requested):
Left one comment.

MaryGao (approved):

packages/typespec-ts/src/index.ts
maorleger, line 406: I think this broke the scenario where a user explicitly passes `false` to generate-metadata which should be honored. Keyvault is a prime example of this https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/keyvault/keyvault-admin/package.json#L45

PR #3319, opened 2025-06-23
Add smoke test for BusinessCases

v-jiaodi:
image


packages/typespec-ts/test/modularUnit/scenarios/models/azureCoreErrorModels.md
```diff
@@ -0,0 +1,515 @@
...
+}
+
+/** model interface ErrorDetail */
+export interface ErrorDetail_1 {
+  code: string;
+  message: string;
+  details?: string;
+}
+
+export function errorDetailSerializer(item: ErrorDetail_1): any {
```
qiaozha, line 274: This serializer function name doesn't have a _1 because there's no azure core error response in the input. but the input param type has _1, which should be correct.

PR #3346, opened 2025-07-02
Upgrade azure specs and update case

v-jiaodi:


packages/autorest.typescript/test/smoke/generated/web-resource-manager/review/web-resource-manager.api.md
```diff
@@ -2369,11 +2369,10 @@ export interface DiagnosticCategoryCollection {
 }
 
 // @public
-interface DiagnosticData_2 {
+export interface DiagnosticData {
```
v-jiaodi, line 2372: expected?

MaryGao (approved):
Approved this since I think api view issue should not block us.

PR #3401, opened 2025-07-29
Patch LRO generation would fail with void 202 response

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/issues/3402

https://dev.azure.com/azure-sdk/public/_build/results?buildId=5142119&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692&l=324
image

### DIFF
--- a/packages/typespec-ts/src/lib.ts
+++ b/packages/typespec-ts/src/lib.ts
@@ -454,7 +454,13 @@ const libDef = {
     "no-paging-items-defined": {
       severity: "warning",
       messages: {
-        default: paramMessage`Please specify @items property for the paging operation - ${"operationName"}.`
+        default: paramMessage`Operation '${"operationName"}' is marked with @list but does not have @pageItems defined.`
+      }
+    },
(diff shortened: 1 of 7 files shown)


packages/typespec-ts/test/modularUnit/scenarios/apiOperations/apiOperations.md
```diff
@@ -1161,3 +1161,316 @@ export class TestingClient {
...
+  @doc("The initiator of the FileShareSnapshot. This is a user-defined value.")
+  initiatorId?: string;
+
+  @doc("The metadata")
+  metadata?: Record<string>;
+}
+
+@armResourceOperations
+interface FileShareSnapshots {
+ updateFileShareSnapshot is ArmCustomPatchAsync<
```
MaryGao, line 1300: could you simply this case? there are too many properties and comments. i don't think we need these noises.

PR #3437, opened 2025-08-13
Verify that the library option is set to true for @azure-rest/purview-datamap

v-jiaodi:
image


packages/typespec-test/test/Azure.Analytics.Purview.DataMap/spec/routes.tsp
```diff
@@ -0,0 +1,2135 @@
...
+@doc("The ApiVersion query parameter for Atlas API.")
+model AtlasApiVersionParameter {
+  @doc("The API version to use for this operation.")
+  @query("api-version")
+  @minLength(1)
+  apiVersion?: string;
+}
+
+alias AtlasApiVersionParameterTraits = Azure.Core.Traits.VersionParameterTrait<AtlasApiVersionParameter>;
+
```
v-jiaodi: A custom optional `apiVersion` was defined, but since all operations use `AtlasApiVersionParameterTraits`, `optional` is set to `true`.

packages/typespec-ts/src/modular/helpers/clientHelpers.ts
```diff
@@ -73,14 +73,15 @@ export function getClientParameters(
   const hasDefaultValue = (p: SdkParameter) =>
     p.clientDefaultValue || p.__raw?.defaultValue || p.type.kind === "constant";
   const isRequired = (p: SdkParameter) =>
-    !p.optional &&
-    ((!hasDefaultValue(p) &&
+    // Special case: when apiVersionAsRequired is true, apiVersion should always be considered required
```
MaryGao: I don't think this fix is correct. We should trust always the defined optionality for api-version. If it is optional in spec, why we set it as required?

packages/typespec-ts/test/modularUnit/scenarios/clientContext/optionalApiVersion.md
```diff
@@ -0,0 +1,142 @@
+# handle with optional api-version parameter via custom VersionParameterTrait
+
+## Description
+
+This test reproduces the issue found in Azure.Analytics.Purview.DataMap where:
+
+- TypeSpec defines an optional api-version parameter: `apiVersion?: string`
+- TCGC correctly sets `optional: true` for this parameter
+- However, the generated client code incorrectly adds a `ClientApiVersionPolicy` instead of treating it as truly optional
```
MaryGao: when leveraging AI to generate content pls ensure you understand what AI is said, do you understand what following mentioned?

```
However, the generated client code incorrectly adds a `ClientApiVersionPolicy` instead of treating it as truly optional
```

MaryGao (approved):

PR #3457, opened 2025-08-20
[sample gen] fail to generate samples-dev as tcgc doesn't give the samples

kazrael2119:
from this spec pr https://github.com/Azure/azure-rest-api-specs/pull/35588
We found that it doesn't generate the samples. Then I debugged in our emitter, I found that this is because tcgc doesn't match the sample and operation, so we couldn't generate them. 

but I don't know the root cause of the tcgc mismatch


qiaozha:
@kazrael2119 please verify if the fix in tcgc works or not

kazrael2119:
close as should be fixed
image

PR #3463, opened 2025-08-22
Fix: Remove unwanted serializer/deserializer function exports from index.ts files

### LINKED ISSUE #3460: function deserializer and serializer should not be exported in `index.ts`
function deserializer and serializer should not be exported in `index.ts`.
Image

v-jiaodi:
fixes #3460


packages/typespec-test/test/Keyvault.Management/generated/typespec-ts/src/models/models.ts
```diff
@@ -0,0 +1,3637 @@
...
+  readonly name?: string;
+  /** Resource type of the key vault resource. */
+  readonly type?: string;
+  /** Azure location of the key vault resource. */
+  readonly location?: string;
+  /** Tags assigned to the key vault resource. */
+  readonly tags?: Record<string, string>;
+}
+
+export function resourceDeserializer_1(item: any): Resource_1 {
```
v-jiaodi, line 1344: Because the model `Resource` is duplicated, `Resource_1` and the corresponding function Deserializer_1 are generated. Update the codegen logic to not export such similar functions.

MaryGao: Is it becuase the Serializer_1 is not matching our original condition so it is wronly exposed?

v-jiaodi: yes

MaryGao (approved):

PR #3502, opened 2025-09-18
Fix the missing parameters in operation options

### LINKED ISSUE #3501: Emitter Incorrectly uses original operation for implementation when generating an overridden (client.tsp) operation
The JS emitter is having issues generating an overridden TypeSpec operation in client.tsp.

**Repro**
Here’s a minimal repro:
https://github.com/joheredi/operation-override-issue/blob/main/client.tsp

Note how the `getSecret` operation overrides the original operation defined in https://github.com/joheredi/operation-override-issue/blob/main/routes.tsp by:
   - Making secretVersion a required parameter
   - Removing outContentType

To reproduce:

1. tsp install
2. tsp compile client.tsp
3.  open tsp-output/@azure-tools/typespec-ts/src/api/operations.ts
4. Incorrect generation can be found at line 26

**Problem**
The TypeScript emitter correctly uses the overridden operation to generate the operation signature. However, it appears to use the original operation when generating the implementation. As a result, the generated implementation tries to access a parameter (`outContentType`) that no longer exists in the overridden operation.

**Expected Behavior**
The emitter should consistently use the overridden operation for both signature and implementation generation.

v-jiaodi:
fixes #3501


packages/typespec-ts/src/modular/buildOperations.ts
```diff
@@ -130,7 +130,7 @@ export function buildOperationOptions(
         p.onClient === false &&
         !(
           p.isGeneratedName &&
-          (p.name === "contentType" || p.name !== "accept")
+          (p.name === "contentType" || p.name === "accept")
```
MaryGao, line 133: i am not sure why you have this fix but it is not related to accept parameter, and it is a question on how tcgc gives data to us.

MaryGao (approved):
Approved. We have spector cases for that, could you add it and remove the smoke test?

packages/typespec-test/test/testOperation/spec/client.tsp
```diff
@@ -0,0 +1,34 @@
...
+        /**
+         * The version of the secret. This URI fragment is optional. If not specified, the latest version of the secret is returned.
+         */
+        @path("secret-version")
+        @clientName("version", "go")
+        secretVersion: string;
+    },
+    {}
+>;
+@@override(KeyVault.getSecret, getSecret);
```
MaryGao, line 34: override should not update the path or urltemplate, i think tcgc should report errors if the path/urltemplate is changed.

Issue #3425, opened 2025-08-11
The code generator crashes when processing response types with undefined for `application/cose`


v-jiaodi:
Generate code error:
<img width="1006" height="517" alt="Image" src="https://github.com/user-attachments/assets/de66ffc9-6851-48f1-9777-758f39cac361" />

Please note the followings:
- @azure-rest/codetransparency is not in our smoke testing and we put their spec in draft pr: https://github.com/Azure/autorest.typescript/pull/3424
- to reproduce this issue, you need to first add their spec in our pr
- i take a look at the spec and I guess this issue may be from https://github.com/Azure/autorest.typescript/blob/52c97f63a54a1f2276e8f0d9558fb06fdc5d4543/packages/typespec-test/test/Microsoft.CodeTransparency/spec/main.tsp#L109-L123.
- i think for "application/cose" content type the body bytes should be generated as "binary" thing
- it's better to add this case in our modular unit test

MaryGao:
We need to verify if our implementation works with spector cases - https://github.com/microsoft/typespec/issues/8405.

PR #3572, opened 2025-11-04
Fix binary response handling when response body is not valid UTF-8

timovv:
This works around an issue with Core where it converts all response bodies to text unless `streamingResponseStatusCodes` is set to the appropriate value ([code here](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/core/ts-http-runtime/src/nodeHttpClient.ts#L184)). This is problematic since if the body is binary, it may not be valid UTF-8. Attempting to decode invalid UTF-8 corrupts the response.

Since this is done at the HTTP client level, before any pipeline policies or the Core REST client layer has a chance to look at the response status codes, a Core fix here is not easy and would likely involve new API surface. We are looking at possible Core designs to fix the issue but this work is not currently prioritized.

As a workaround, calling `asNodeStream` or `asBrowserStream` on the response will set `streamingResponseStatusCodes` to true, so that the generated client can access the raw stream. This PR does that for the specific case of binary response bodies and creates a new static helper which reads the stream into a Uint8Array directly. As part of this, I had to update the static helper logic to support browser-mapped files. I also fixed the tests for the bytes download scenario so that we are checking the response body against the actual file bytes instead of the UTF-8 decoded value.


maorleger (approved):

MaryGao:
@timovv Thanks for the fix, we recently received issue for below case - https://github.com/Azure/autorest.typescript/issues/3425, this may be unknown content type with bytes and we would generate `Uint8Array` response body for us, is this supported in core?

```
@doc("The MIME content type a Cose body is application/cose, containing a CoseSign1 signature.") 
 @header("Content-Type") 
 contentType: "application/cose"; 
  
 @doc("Location of the receipt") 
 @header("Location") 
 location?: string; 
  
 @doc("Receipt body in COSE format") 
 @bodyRoot 
 body: bytes; 
```

PR #3580, opened 2025-11-06
Fix TypeScript emitter crash when processing binary responses with application/cose content type

### LINKED ISSUE #3425: The code generator crashes when processing response types with undefined for `application/cose`

v-jiaodi:
fixes #3425


qiaozha:
Pending on spector test case.

PR #3583, opened 2025-11-07
 Fix URI template variable name decoding for better service compatibility

joheredi:
### Background

TypeSpec gives us operation URIs in the format of URI Template (RFC 6570). The RFC 6570 specification doesn't allow hyphens (`-`) in variable names and requires them to be percent-encoded (e.g., `/pools{?api%2Dversion,timeOut}`).

According to the URI Template specification, `/pools{?api%2Dversion,timeOut}` must be interpreted the same as `/pools{?api-version,timeOut}`.

While most frameworks properly decode URIs before handling them, some services don't follow this practice correctly (e.g. https://github.com/Azure/azure-sdk-for-js/issues/36367). To guard our customers against these service malpractices, we should proactively decode percent-encoded characters that are illegal in URI Template syntax but legal in URI before returning the expanded URI.

### Changes

This PR introduces a `normalizeUnreserved` function that decodes percent-encoded unreserved characters in expanded URIs:

- **Unreserved characters decoded**: `-` (hyphen), `.` (dot), `~` (tilde)
- **RFC 3986 compliant**: Only decodes characters that are explicitly allowed to be unencoded
- **Safe transformation**: Leaves other percent-encoded characters intact to preserve URL integrity

### Implementation Details

1. **New function**: `normalizeUnreserved(uri: string)` that selectively decodes only RFC 3986 unreserved characters
2. **Integration**: Applied to the result of `expandUrlTemplate` before returning the final URI
3. **Pattern matching**: Uses regex to identify and decode only `%2D` (hyphen), `%2E` (dot), and `%7E` (tilde) in both uppercase and lowercase

### Test Coverage

Added comprehensive test cases covering:
- Percent-encoded hyphens (`%2D` → `-`)
- Percent-encoded dots (`%2E` → `.`) 
- Percent-encoded tildes (`%7E` → `~`)
- Both uppercase and lowercase hex encoding
- Verification that other encoded characters remain untouched

### Impact

This change improves compatibility with services that don't properly handle percent-encoded URI template variables, while maintaining full RFC compliance and not breaking any existing functionality.


MaryGao (approved):
Approved but one thing to worry about is if this change would impact other parts which wants to be encoded but not, should we do this revert encode when we get the urlTemplate from typespec compiler e.g change api%2DVersion back to api-version. Will this be safe?
 
btw is this fix urgent? do we need to bump a patch version?

Issue #3588, opened 2025-11-10
Modular: Inconsistent Serializer Function Names in Discriminated Unions


maorleger:
## Summary

The TypeSpec TypeScript emitter generates discriminated union serializers that reference non-existent short-form function names instead of the actual fully-qualified function names that are defined. This results in code that fails to compile due to undefined function references.

## Environment

- **TypeSpec Compiler**: 1.5.0
- **@azure-tools/typespec-ts**: 0.46.0
- **@azure-tools/typespec-client-generator-core**: 0.61.3
- **Repository**: azure-sdk-for-js
- **Package**: @azure/search-documents

## Expected Behavior

When a discriminated union type has a `@@clientNamespace` decorator, the union serializer should call the correctly-scoped serializer function names that match the actual function definitions.

**Expected generated code:**
```typescript
export function azureSearchDocumentsKnowledgeBaseKnowledgeRetrievalReasoningEffortUnionSerializer(
  item: KnowledgeRetrievalReasoningEffortUnion,
): any {
  switch (item.kind) {
    case "minimal":
      return azureSearchDocumentsKnowledgeBaseKnowledgeRetrievalMinimalReasoningEffortSerializer(
        item as KnowledgeRetrievalMinimalReasoningEffort,
      );
    default:
      return azureSearchDocumentsKnowledgeBaseKnowledgeRetrievalReasoningEffortSerializer(item);
  }
}
```

## Actual Behavior

The union serializer calls a short-form function name that doesn't exist:

**Actual generated code:**
```typescript
export function azureSearchDocumentsKnowledgeBaseKnowledgeRetrievalReasoningEffortUnionSerializer(
  item: KnowledgeRetrievalReasoningEffortUnion,
): any {
  switch (item.kind) {
    case "minimal":
      return knowledgeRetrievalMinimalReasoningEffortSerializer(  // ❌ Function doesn't exist
        item as KnowledgeRetrievalMinimalReasoningEffort,
      );
    default:
      return azureSearchDocumentsKnowledgeBaseKnowledgeRetrievalReasoningEffortSerializer(item);
  }
}

// Later in the file, the actual function is defined with the full name:
export function azureSearchDocumentsKnowledgeBaseKnowledgeRetrievalMinimalReasoningEffortSerializer(
  item: AzureSearchDocumentsKnowledgeBaseKnowledgeRetrievalMinimalReasoningEffort,
): any {
  return { kind: item["kind"] };
}
```

[...]

PR #3589, opened 2025-11-11
Fix inconsistent Serializer Function Names in Discriminated Unions cross namesapce

v-jiaodi:
fixes https://github.com/Azure/autorest.typescript/issues/3588
- serializer/deserializer name will not include namespace name when conflict happens
- the import would be missing when sub types are in different namespaces in discrinminated union case

### DIFF
--- a/packages/typespec-ts/src/modular/emitSamples.ts
+++ b/packages/typespec-ts/src/modular/emitSamples.ts
@@ -274,6 +274,7 @@ function buildParameterValueMap(example: SdkHttpOperationExample) {
 }
 
 function prepareExampleValue(
+  context: SdkContext,
   name: string,
   value: SdkExampleValue | string,
   isOptional?: boolean,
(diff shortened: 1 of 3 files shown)


packages/typespec-ts/test/modularUnit/scenarios/models/deserialization/extends.md
```diff
@@ -719,3 +719,201 @@ export enum KnownVersions {
...
+model CustomNormalizer extends LexicalNormalizer {
+
+  tokenFilters?: TokenFilterName[];
+
+  charFilters?: CharFilterName[];
+
+  #suppress "@azure-tools/typespec-azure-core/casing-style" "Maintain compatibility with existing models."
+  `@odata.type`: "#Microsoft.Azure.Search.CustomNormalizer";
+}
+@@clientName(LexicalNormalizer.`@odata.type`, "odataType");
```
v-jiaodi: using `@@clientName` affected the inheritance of properties with the same name, causing duplication.

PR #3643, opened 2025-12-15
Prevent flatten property decorator from being applied to non-model types

### LINKED ISSUE #3631: Flatten property should be ignored if tagged in an non-model type like `unknown`
Case: 

- https://github.com/Azure/azure-rest-api-specs/tree/main/specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub
- https://github.com/Azure/azure-rest-api-specs/tree/main/specification/storagesync/resource-manager/Microsoft.StorageSync/StorageSync

swagger link: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/preview/2025-05-01-preview/openapi.json#L6099
```
    "Operation": {
      "type": "object",
      "description": "A Event Hub REST API operation",
      "properties": {
        "name": {
          "type": "string",
          "description": "Operation name: {provider}/{resource}/{operation}",
          "readOnly": true
        },
        "properties": {
          "description": "Properties of the operation",
          "x-ms-client-flatten": true
        }
      }
    },
```
typespec : 
```
model Operation {

  @visibility(Lifecycle.Read)
  name?: string;

  #suppress "@azure-tools/typespec-azure-core/no-unknown" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
  properties?: unknown;
}

@@Azure.ClientGenerator.Core.Legacy.flattenProperty(Operation.properties);
```
Currently, using flatten will cause `Operation.properties` to be lost.

JialinHuang803:
fixes #3631

This PR fixes an issue where the `@flattenProperty` decorator could be applied to non-model types (primitives, arrays, unions, etc.), which would cause compile errors on generated code.


MaryGao (changes requested):
Could we update the fix to not change any state from TCGC?

MaryGao (approved):

PR #3690, opened 2026-01-20
Fix API version parameter name handling

joheredi:
### Problem

The TypeScript code generator had an issue when generating client code:

1. **API Version Parameter Name**: The generator was hardcoding `apiVersion` as the parameter name in several places, but TypeSpec specifications can use different names (e.g., `version` in Azure Blob Storage). This caused:
   - Incorrect destructuring: `const { apiVersion: _, ...updatedOptions }` when the actual property was `version`
   - Undefined variable references in the generated code
   - Type mismatches between the interface definitions and the implementation

### Solution

1. **API Version Parameter Name Fix**:
   - Updated `buildGetClientOptionsParam` to accept an optional `apiVersionParamName` parameter
   - Modified buildClientContext.ts to extract the API version parameter early and pass its actual name
   - The destructuring and variable declarations now use the correct parameter name from the TypeSpec


JialinHuang803 (approved):
Do we need to have a UT for it?

PR #3705, opened 2026-01-27
Test whether format changes can be formatted correctly

v-jiaodi:
Update a single line to multiple lines, the expected result is that the code remains unchanged after formatting.

### DIFF
--- a/packages/typespec-test/package.json
+++ b/packages/typespec-test/package.json
@@ -23,7 +23,7 @@
     "@types/mocha": "^5.2.7"
   },
   "scripts": {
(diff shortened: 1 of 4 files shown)


JialinHuang803:
The difference is introduced by Prettier. When you provide multi-line input, Prettier may format it differently than if you provide single-line input.

Bottom line: While Prettier does format the output, it **doesn't produce identical output** for semantically equivalent but syntactically different inputs. The indentation, line breaks, and chaining style you provide as input can influence how Prettier decides to format the final result.

If you want consistent output regardless of how you write the template string, you should write the code in **a single line** and let Prettier handle all formatting decisions.

Issue #3709, opened 2026-01-28
[TSP Migration] Placeholder not resolved for polymorphic base type serializer/deserializer when no subtypes exist


kazrael2119:
issue found in https://github.com/Azure/azure-rest-api-specs/pull/39844
When a TypeSpec model is decorated with `@discriminator` but has no subtypes extending from it, the generated TypeScript code contains unresolved placeholders like 
```
export function servicePlacementPolicyDescriptionArraySerializer(
  result: Array<ServicePlacementPolicyDescription>,
): any[] {
  return result.map((item) => {
    return __PLACEHOLDER_o163_sserializer__(item);  // ❌ Unresolved
  });
}

export function servicePlacementPolicyDescriptionArrayDeserializer(
  result: Array<ServicePlacementPolicyDescription>,
): any[] {
  return result.map((item) => {
    return __PLACEHOLDER_o163_sdeserializer__(item);  // ❌ Unresolved
  });
}
```

typespec code:
https://github.com/Azure/azure-rest-api-specs/pull/39844/changes#diff-c8343c1bffdd125cfe5149265a5090ebaf075f0502bbb65886eb6e30375696fcR222-R229

PR #3710, opened 2026-01-28
Fix the generation of discriminate models without Subtypes

kazrael2119:
For the discriminate models without Subtypes, generate as a normal model
https://github.com/Azure/autorest.typescript/issues/3709

### DIFF
--- a/packages/typespec-ts/src/utils/modelUtils.ts
+++ b/packages/typespec-ts/src/utils/modelUtils.ts
@@ -1187,7 +1187,7 @@ function getSchemaForArrayModel(
               .join(" | ");
           }
         }
-      } else if (schema.items.type.includes("|")) {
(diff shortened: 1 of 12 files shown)


packages/typespec-ts/src/modular/serialization/buildSerializerFunction.ts
```diff
@@ -184,7 +184,11 @@ function buildPolymorphicSerializer(
 
   const subTypes = type.discriminatedSubtypes;
   if (!subTypes) {
-    return;
+    // No subtypes - treat as a regular model and generate simple serializer
+    return buildModelTypeSerializer(context, type, {
+      nameOnly,
+      skipDiscriminatedUnionSuffix: false
+    });
```
v-jiaodi: @MaryGao for without subTypes Discriminated model, What is the expected generated code?

PR #3713, opened 2026-01-30
[Modular] Support multiple services in one client

JialinHuang803:
This pull request introduces support for multi-service scenarios in the TypeSpec TypeScript emitter, enabling the generation of SDKs that combine multiple services into a single package.

### Changes

**Multi-service support in emitter and code generation:**

* Added an `isMultiService` flag to `RLCOptions` and updated emitter logic to use this flag for handling multi-service scenarios, including skipping API version enums and adjusting namespace calculations.
* Used full namespace segments for multiple services when generating model hierarchy. This ensures models from different services are in separate folders.
* Removed the `apiVersion` parameter from the client and clientContext for multi-service, and operations use their default apiVersion value directly. `KnownVersons` enums are not exported.
* Updated project file and export generation logic to correctly handle exports and subpaths for multi-service packages, skipping api subpaths.

**RLC dependency code updates:**
* Updated the `getRLCClients()` method to preserve multi-service structure for single Modular clients instead of splitting them.
* Updated the `getDefaultService()` method to accept an `isModularLibrary` parameter (defaults to true). The "more-than-one-service" diagnostic is now only reported for RLC libraries, since Modular libraries support multiple services by design.

### DIFF
--- a/packages/autorest.typescript/CHANGELOG.md
+++ b/packages/autorest.typescript/CHANGELOG.md
@@ -1,3 +1,14 @@
+## 6.0.57 (2026-02-02)
+
+- [Feature] Support @disablePageable. Please refer to [#3715](https://github.com/Azure/autorest.typescript/pull/3715)
+- [Feature] Add Azure Alternate Type spector case and fix numeric type generation bug. Please refer to [#3558](https://github.com/Azure/autorest.typescript/pull/3558)
+- [Feature] Remove apiVersion pipeline policy and move default apiVersion value to operation level. Please refer to [#3702](https://github.com/Azure/autorest.typescript/pull/3702)
+- [Feature] Generate `PlainDate` as a Date instead of as a string. Please refer to [#3675](https://github.com/Azure/autorest.typescript/pull/3675)
+- [Feature] Add missing multipart test coverage for wire names and optional parts. Please refer to [#3692](https://github.com/Azure/autorest.typescript/pull/3692)
+- [Feature] Bump TypeSpec dependencies to latest stable. Please refer to [#3698](https://github.com/Azure/autorest.typescript/pull/3698)
+- [Bugfix] Fix the generation of discriminate models without Subtypes. Please refer to [#3710](https://github.com/Azure/autorest.typescript/pull/3710)
+- [Bugfix] Filtering out parameters with default values when generating sample code. Please refer to [#3619](https://github.com/Azure/autorest.typescript/pull/3619)
+
 ## 6.0.56 (2026-01-26)
 
 - [Feature] Support LRO + Paging operations in Modular SDK. Please refer to [#3676](https://github.com/Azure/autorest.typescript/pull/3676)
--- a/packages/autorest.typescript/package.json
+++ b/packages/autorest.typescript/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@autorest/typescript",
-  "version": "6.0.56",
+  "version": "6.0.57",
   "scripts": {
     "build": "tsc -p . && npm run copyFiles",
     "build:test:browser:rlc": "tsc -p tsconfig.browser-test.json && ts-node test/commands/prepare-deps.ts --browser && webpack --config webpack.config.test.js --env mode=rlc",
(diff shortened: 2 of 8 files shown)


packages/typespec-ts/src/modular/buildProjectFiles.ts
```diff
@@ -35,23 +35,26 @@ function buildExportsForMultiClient(
...
   if (emitterOptions.options.hierarchyClient) {
-    for (const flattenedClient of clientMap) {
-      const { subfolder } = getModularClientOptions(flattenedClient);
-      const client = flattenedClient[1];
-      const methodMap = getMethodHierarchiesMap(context, client);
-      for (const [prefixKey, _] of methodMap) {
-        const prefixes = prefixKey.split("/");
-        if (prefixKey === "") {
-          continue;
+    // Skip generating api subpath for multi-service
```
joheredi: Can you file an issue to track the work to design and enable the api subpath for multiservice and reference it in this comment. This is to give more context on why it is skipped

packages/typespec-ts/src/modular/emitModels.ts
```diff
@@ -312,6 +317,14 @@ export function getModelNamespaces(
       return [];
     }
     const segments = model.namespace.split(".");
+    // Keep full namespace segments if multiple services are present
```
joheredi: Is this for name deduplication?

packages/typespec-ts/src/modular/buildProjectFiles.ts
JialinHuang803: Updated the comment, and it's tracked in https://github.com/Azure/autorest.typescript/issues/3717.

packages/typespec-ts/src/transform/transfromRLCOptions.ts
```diff
@@ -33,10 +33,7 @@ export function transformRLCOptions(
     emitterOptions,
     dpgContext.generationPathDetail?.rootDir ?? ""
   );
-  if (emitterOptions["is-modular-library"] !== false) {
-    options.isModularLibrary = true;
-  }
```
qiaozha: Is it safe to remove this? I think this implies that if it's undefined, it will set isModularLibrary as true.

packages/typespec-ts/src/utils/clientUtils.ts
```diff
@@ -18,23 +18,64 @@ import { SdkContext } from "./interfaces.js";
...
-    if (Array.isArray(clientService)) {
-      clientService.forEach((ns) => services.add(ns));
-    } else {
-      services.add(clientService);
+export function getRLCClients(
+  dpgContext: SdkContext,
+  isModularLibrary?: boolean
+): SdkClient[] {
+  const modular =
+    isModularLibrary ?? dpgContext.rlcOptions?.isModularLibrary ?? false;
```
qiaozha: Not sure why we need to explicitly pass isModularLibrary if we have dpgContext.rlcOptions.isModularLibrary?

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -1672,15 +1675,18 @@ export function predictDefaultValue(
   return;
 }
 
-export function getDefaultService(program: Program): Service | undefined {
+export function getDefaultService(
+  program: Program,
+  isModularLibrary: boolean = true
```
qiaozha: wonder if we should leverage dpgContext for both program and isModularLibrary parameters instead of adding new parameter isModularLibrary maybe more other parameters in the future all the way down the road?

packages/typespec-ts/src/utils/modelUtils.ts
```diff
@@ -1672,15 +1675,18 @@ export function predictDefaultValue(
...
+): Service | undefined {
   const services = listServices(program);
   if (!services || services.length === 0) {
     reportDiagnostic(program, {
       code: "no-service-defined",
       target: NoTarget
     });
   }
-  if (services.length > 1) {
+  if (services.length > 1 && !isModularLibrary) {
```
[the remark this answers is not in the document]
JialinHuang803: If isModularLibrary is false, that means the program is trying to generate an RLC package where multiple services in one client is not supported. Then the "more-than-one-service" diagnostic will be reported.

PR #3714, opened 2026-01-30
Response Header Deserialization

joheredi:
## Executive Summary

This design document outlines a feature enhancement to the TypeScript SDK emitter that enables automatic deserialization of HTTP response headers into strongly-typed model interfaces. This aligns SDK behavior with HTTP semantics where response headers carry important metadata alongside the response body, and provides developers with a unified, type-safe API for accessing all response data.

### Current State
In HTTP, responses consist of two distinct parts:
- **Headers**: Metadata about the response (e.g., `x-user-id`, `content-type`, `etag`)
- **Body**: The main response payload

TypeScript SDK-generated clients currently deserialize only the response body into a model type, leaving headers as untyped data available only through raw `result.headers` access. This creates several issues:

1. **Type Safety Loss**: Header values are strings that developers must manually parse and type-cast
2. **Incomplete Models**: Response models don't reflect the complete response contract
3. **Developer Friction**: Developers must know which headers exist and manually check `result.headers` for critical values
4. **API Inconsistency**: Response body is fully typed, but important header values are not

### User Perspective
When a service contract specifies that a response includes both body data AND header metadata (e.g., a versioning header, request ID, or custom business data):

```tsp
model UserResponse {
  id: string;
  name: string;
  
  @header("x-user-version")
  version: string;
  
  @header("x-request-id")
  requestId: string;
}

op getUser(): UserResponse;
```

Users expect to access all response data through a single, typed interface:
```typescript
const user = await getUser();
console.log(user.id);           // ✓ Easy, typed
console.log(user.name);         // ✓ Easy, typed
console.log(user.version);      // Currently: ✗ Not available directly
console.log(user.requestId);    // Currently: ✗ Not available directly
```

**Desired behavior**: Headers should be seamlessly integrated into the response model, just like body properties.

### Design Approach
Transform the deserialization pipeline to:
1. Make header properties part of the response model interface
2. Mark header properties as optional (since headers may not always be present)
3. During deserialization, extract both body and header values and merge them into the response object
4. Maintain backward compatibility with existing clients that don't use headers

#### Change 1.1: Include Header Properties in Model Interfaces

**Before:**
```typescript
export interface UserResponse {
  id: string;
  name: string;
}
```

**After:**
```typescript
export interface UserResponse {
  id: string;
  name: string;
  version?: string;        // From @header("x-user-version")
  requestId?: string;      // From @header("x-request-id")
}
```

**Rationale:**
- Headers are part of the HTTP response contract and should be visible in the model
- Optional modifier (`?`) reflects HTTP reality: headers may not always be present
- Developers get full type information without breaking existing code that doesn't use headers

#### Change 1.2: Header Properties Always Optional

Header properties are always marked as optional (with `?` modifier) regardless of their TypeSpec definition because:

1. **HTTP Semantics**: HTTP headers are fundamentally optional; clients must handle cases where headers are absent
2. **Network Reliability**: Response headers can be lost or stripped by intermediaries (proxies, gateways, CDNs)
3. **Defensive Programming**: Even if a service promises to send a header, defensive clients should handle its absence
4. **API Evolution**: Services may remove or conditionally omit headers in future versions

```typescript
export interface UserResponse {
  // Body properties keep their original optionality
  id: string;           // Required
  email?: string;       // Optional
  
  // Header properties are always optional
  version?: string;     // Always optional, even if TypeSpec declares it required
  requestId?: string;   // Always optional, even if TypeSpec declares it required
}
```

#### Change 2.1: Deserializer Signature Enhancement

**Before:**
```typescript
export function userResponseDeserializer(item: any): UserResponse {
  return {
    id: item["id"],
    name: item["name"]
  };
}
```

**After:**
```typescript
export function userResponseDeserializer(item: any, headers?: any): UserResponse {
  return {
    id: item["id"],
    name: item["name"],
    version: headers?.["x-user-version"],
    requestId: headers?.["x-request-id"]
  };
}
```

**Key Points:**
- `headers` parameter is optional (has `?` modifier), maintaining backward compatibility
- Uses optional chaining (`?.`) when accessing headers to safely handle `undefined` values
- Header values are read using their wire names (e.g., lowercase with hyphens: `"x-user-version"`)
- Straightforward object property assignment; no complex transformation logic

#### Change 2.2: Selective Header Parameter Passing

To avoid unnecessary parameter passing and maintain clean signatures:

**Only pass headers when the model has header properties:**

```typescript
// Model WITH header properties: pass headers
return userResponseDeserializer(result.body, result.headers);

// Model WITHOUT header properties: don't pass headers
return basicMetadataDeserializer(result.body);
```

**Benefits:**
- Generated code remains clean and readable
- Minimal overhead for models that don't use headers
- Type-safe: TypeScript validates correct number of arguments

#### Change 2.3: Handling Inheritance

For models with inheritance, header properties from base models are also extracted:

```typescript
model Pet {
  name: string;
  
  @header("x-pet-version")
  version: string;
}

model Cat extends Pet {
  meow: boolean;
  
  @header("x-cat-color")
  color?: string;
}

// Deserializer handles headers from both Cat and Pet
export function catDeserializer(item: any, headers?: any): Cat {
  return {
    // Pet properties
    name: item["name"],
    version: headers?.["x-pet-version"],
    // Cat properties
    meow: item["meow"],
    color: headers?.["x-cat-color"]
  };
}
```

### 3. Operation Deserialization Integration

Operations that return models with headers automatically pass headers during deserialization:

```typescript
export async function _getUserDeserialize(
  result: PathUncheckedResponse
): Promise<UserResponse> {
  const expectedStatuses = ["200"];
  if (!expectedStatuses.includes(result.status)) {
  }
  
  // Automatically includes headers in deserialization
  return userResponseDeserializer(result.body, result.headers);
}

export async function getUser(
  context: Client,
  options: GetUserOptionalParams = { requestOptions: {} }
): Promise<UserResponse> {
  const result = await _getUserSend(context, options);
  return _getUserDeserialize(result);
}
```

## HTTP Alignment

This design aligns with HTTP specifications and best practices:

### RFC 7230/7231: HTTP Semantics and Header Fields
- Recognizes headers as a first-class part of response semantics
- Treats headers as metadata that may be important to the application

### REST API Design Principles
- **Semantic Completeness**: Response objects now contain complete response semantics
- **Defensive Programming**: Always-optional headers follow HTTP robustness principle
- **Layered Abstraction**: Hides HTTP plumbing (headers vs. body distinction) from SDK users

## Backward Compatibility

✅ **Fully Backward Compatible**

- Existing code that doesn't access header properties continues to work without changes
- New header properties are optional and default to `undefined`
- Deserializers can be called with or without the `headers` parameter
- Models that don't have header properties are unaffected

### 1. Type Safety
```typescript
// Before: Manual string access, no type checking
const version = result.headers?.["x-user-version"];  // type: string | undefined

// After: Full type checking through model interface
const user = await getUser();
const version = user.version;  // type: string | undefined, caught by IDE/linter
```

### 2. Unified Response Access
```typescript
// Before: Mixed access patterns
const user = {
  id: deserializedObject.id,
  version: rawHeaders["x-user-version"]
};

// After: Consistent model-based access
const user = await getUser();  // All properties in one place
```

### 3. IntelliSense Support
```typescript
// IDE provides full autocomplete for all response properties
const user = await getUser();
user.  // ← IDE shows: id, name, version, requestId
```

### 4. Documentation
Types serve as inline documentation:
```typescript
// Type definition shows all available response data
interface UserResponse {
  id: string;           // User ID
  name: string;         // User display name
  version?: string;     // API version from x-user-version header
  requestId?: string;   // Request tracking ID from x-request-id header
}
```

### Scope of Changes
- **Model Interface Generation** (`emitModels.ts`): Include header properties, mark as optional
- **Deserializer Generation** (`buildDeserializerFunction.ts`): Add optional `headers` parameter, extract header values
- **Operation Deserialization** (`operationHelpers.ts`): Conditionally pass headers based on model composition

### Detection Logic
- Uses `@header` TypeSpec decorator to identify header properties
- Scans model and all ancestor models for header properties
- Only passes headers parameter when model contains header properties

### DIFF
--- a/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
+++ b/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
@@ -298,7 +298,7 @@ export function getDeserializePrivateFunction(
         statements.push(
           `const responseContentType = result.headers?.["content-type"] ?? "";
           if (${isXmlContentTypeRef}(responseContentType)) {
-            return ${xmlDeserializerName}(${deserializedRoot});
+            return ${xmlDeserializerName}(${deserializedArgs});
           }
           return ${jsonDeserializerName}(${deserializedArgs});`
         );
@@ -330,7 +330,7 @@ export function getDeserializePrivateFunction(
       ) as string | undefined;
 
       if (xmlDeserializerName) {
-        statements.push(`return ${xmlDeserializerName}(${deserializedRoot})`);
+        statements.push(`return ${xmlDeserializerName}(${deserializedArgs});`);
       } else {
         // Fall back to JSON deserializer if XML deserializer is not available
         const deserializeFunctionName = buildModelDeserializer(
@@ -1707,8 +1707,17 @@ export function getResponseMapping(
   const allParents = type.kind === "model" ? getAllAncestors(type) : [];
   // Include header properties when deserializing responses without changing
   // the existing selection of non-metadata properties.
-  const properties =
+  let properties =
     type.kind === "model" ? getAllProperties(context, type, allParents) : [];
+
+  if (
+    overrides?.propertyFilter &&
+    typeof overrides.propertyFilter === "function"
+  ) {
+    properties = properties.filter(
+      (p) => !overrides.propertyFilter || overrides.propertyFilter(p)
+    );
+  }
   if (type.kind === "model") {
     const headerProps: SdkModelPropertyType[] = [];
     const addHeaderProps = (model: SdkModelType) => {
@@ -1826,7 +1835,7 @@ function areAllResponseHeadersInModel(
 ): boolean {
   // Collect all header properties from the model and its ancestors
   const modelHeaderNames = new Set<string>();
-  
+
   const addHeaderPropsFromModel = (model: SdkModelType) => {
     model.properties?.forEach((p) => {
       if (isHeader(context.program, p.__raw!)) {
@@ -1839,10 +1848,10 @@ function areAllResponseHeadersInModel(
       }
     });
   };
-  
+
   // Add headers from the model itself
   addHeaderPropsFromModel(modelType);
-  
+
   // Add headers from ancestor models
   const allParents = getAllAncestors(modelType);
   allParents.forEach((parent) => {
--- a/packages/typespec-ts/src/modular/serialization/buildXmlSerializerFunction.ts
+++ b/packages/typespec-ts/src/modular/serialization/buildXmlSerializerFunction.ts
@@ -1,7 +1,11 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import { FunctionDeclarationStructure, StructureKind } from "ts-morph";
+import {
+  FunctionDeclarationStructure,
+  ParameterDeclarationStructure,
+  StructureKind
+} from "ts-morph";
 import {
   SdkModelPropertyType,
   SdkModelType,
@@ -26,11 +30,12 @@ import { resolveReference } from "../../framework/reference.js";
 import { refkey } from "../../framework/refkey.js";
 import { reportDiagnostic } from "../../lib.js";
 import { NoTarget } from "@typespec/compiler";
-import { isMetadata } from "@typespec/http";
+import { isMetadata, isHeader } from "@typespec/http";
 import { normalizeModelPropertyName } from "../type-expressions/get-type-expression.js";
 import { isReadOnly } from "@azure-tools/typespec-client-generator-core";
 import { buildModelSerializer } from "./buildSerializerFunction.js";
 import { buildModelDeserializer } from "./buildDeserializerFunction.js";
+import { getResponseMapping } from "../helpers/operationHelpers.js";
 
 /**
  * Checks if a model type has XML serialization options defined
@@ -189,6 +194,10 @@ function buildPropertyMetadataArray(
       continue;
     }
     if (isMetadata(context.program, property.__raw!)) {
+      continue;
+    }
+    // Skip header properties as they are handled separately
+    if (isHeader(context.program, property.__raw!)) {
       continue;
     }
 
@@ -375,6 +384,20 @@ export function buildXmlModelDeserializer(
     return resolveReference(refkey(type, "xmlDeserializer"));
   }
 
+  // Check if model or its ancestors have header properties that need special handling.
+  // If so, the deserializer function will accept an optional 'headers' parameter
+  // to deserialize header values into the model properties.
+  const allProps = type.kind === "model" ? (type.properties ?? []) : [];
+  const ancestorProps =
+    type.kind === "model"
+      ? getAllAncestors(type)
+          .filter((p) => p.kind === "model")
+          .flatMap((p) => (p as SdkModelType).properties ?? [])
+      : [];
+  const hasHeaderProperties = [...allProps, ...ancestorProps].some((p) =>
+    isHeader(context.program, p.__raw!)
+  );
+
   const deserializeFromXmlRef = resolveReference(XmlHelpers.deserializeFromXml);
   const xmlPropertyDeserializeMetadataRef = resolveReference(
     XmlHelpers.XmlPropertyDeserializeMetadata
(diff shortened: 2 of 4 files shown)


packages/typespec-ts/src/modular/helpers/operationHelpers.ts
```diff
@@ -1648,21 +1654,54 @@ export function getResponseMapping(
...
   const allParents = type.kind === "model" ? getAllAncestors(type) : [];
+  // Include header properties when deserializing responses without changing
+  // the existing selection of non-metadata properties.
   const properties =
     type.kind === "model" ? getAllProperties(context, type, allParents) : [];
+  if (type.kind === "model") {
+    const headerProps: SdkModelPropertyType[] = [];
+    const addHeaderProps = (model: SdkModelType) => {
+      model.properties?.forEach((p) => {
+        if (isHeader(context.program, p.__raw!)) {
```
timovv: Nit, could TCGC help here? I see there is a [`SdkServiceResponseHeader`](https://azure.github.io/typespec-azure/docs/libraries/typespec-client-generator-core/reference/js-api/interfaces/sdkserviceresponseheader/) type that looks promising but couldn't tell at a glance whether we have access to that info here

timovv (approved):
Looks good! A couple of questions but nothing major

qiaozha (changes requested):
Could you hold this PR until we get @MaryGao 's confirm? 

/cc @lirenhe we need to prioritize this response model type related breaking discussion as soon as possible.

MaryGao:
> Could you hold this PR until we get @MaryGao 's confirm?
> 
> /cc @lirenhe we need to prioritize this response model type related breaking discussion as soon as possible.

We(Jose and I) have some offline discussions and here are the agreements with Maor and Jeff. In short:
- No generation in Modular FooResponse = Foo & FooResponseHeaders
- Treat non-model as special case to mitigate breaking
    - non-model type without headers e.g {body: string; }
    - non-model type with headers e.g {body: string; fooHeader: string;  barHeader: string; }
- To include headers in model, Jose's PR would adopt bodyRoot nested with header ideas to mitigate storage breaking.
    - To un-block storage we plan to have this design firstly, we could review this again once any conclusion reached out cross team in TCGC.

Overall the decision looks good to me and I will review this PR for the adoption and verify the impact in our side.

packages/typespec-ts/test/modularUnit/scenarios/models/response/headerInResponse.md
```diff
@@ -0,0 +1,185 @@
...
+  const expectedStatuses = ["200"];
+  if (!expectedStatuses.includes(result.status)) {
+    throw createRestError(result);
+  }
+
+  return userDeserializer(result.body, result.headers);
+}
+```
+
+# [void] Header properties included in the response model when there is no response body
```
MaryGao: As I mentioned this is the case I have concerns. I wonder if we have real case in storeage yet. If no, I prefer we figure out how to express following two scenarios in TypeSpec side first.
- headers are included in response return type
- headers are not included in return type by intention

Currently we would include all headers in responses and this is common that ARM SDKs would introduce new changes especially DELETE with void return type. 

If we have changed the design in future, this would be a breaking for ARM libraries.

packages/typespec-ts/test/modularUnit/scenarios/models/response/headerInResponse.md
```diff
@@ -0,0 +1,185 @@
+# Header properties included in the response model interface by default
```
MaryGao: Have we implemented other cases in gist?

I may wonder if the bodyRoot case is supported: https://gist.github.com/joheredi/d1e310b6d1464ae89fcaefb00601693d#file-3_bodyroot_decorator-md.

```ts
model User {
  name: string;
  email: string;

  @header("x-user-id")
  userId?: string;

  @header
  @encode("rfc7231")
  createdAt?: utcDateTime;
}

op getUser(): { @bodyRoot user: User} & { @header fooHeader: string};
```

packages/typespec-ts/src/modular/helpers/operationHelpers.ts
MaryGao: As I mentioned, TCGC isn't fully ready to support this case. So currently the runtime in TCGC has some limitation that we need to touch the raw data I think.

joheredi:
Closing in favor of https://github.com/Azure/autorest.typescript/pull/3728 which takes a more scoped and simple approach

PR #3753, opened 2026-02-12
Support inferring accept header

joheredi:
### Summary

This PR adds support for automatically inferring the `Accept` request header from the response content types defined in the TypeSpec. The feature is opt-in and controlled by the `inferAcceptHeader` client option to prevent breaking changes to existing generated libraries.

### Changes

**Implementation (operationHelpers.ts):**
- Added `isAccept()` helper function to identify Accept header parameters
- Added `getAcceptHeaderValue()` function that:
  - Uses explicit constant Accept parameter values when available
  - Infers from response content types when `inferAcceptHeader` client option is enabled
- Added `getResponseContentTypes()` helper to collect and deduplicate content types from operation responses
- Modified `getHeaderAndBodyParameters()` to accept a `client` parameter and include the Accept header in generated code

**Test Scenarios (inferAcceptHeader.md):**
- Test case for single content type (`application/json`)
- Test case for multiple content types (`application/json, application/xml`)
- Test case verifying no Accept header inference when client option is not set

### Usage

To enable Accept header inference, use the `@@clientOption` decorator in your TypeSpec:

```tsp
@@clientOption(MyService, "inferAcceptHeader", true, "javascript");
```

### Behavior

| Scenario | Accept Header Behavior |
|----------|----------------------|
| Explicit constant Accept param | Uses the explicit value |
| `inferAcceptHeader: true` | Infers from response content types |
| No option set | No Accept header added (preserves backward compatibility) |

### Why This Change

When services define response content types in their TypeSpec, the generated client should properly communicate what content types it accepts. This improves interoperability and allows servers to respond with the appropriate content type.

---

### DIFF
--- a/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
+++ b/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
@@ -151,7 +151,8 @@ export function getSendPrivateFunction(
     `return context.path(${pathStr}).${operationMethod}({...${resolveReference(dependencies.operationOptionsToRequestParameters)}(${optionalParamName}), ${getHeaderAndBodyParameters(
       dpgContext,
       operation,
-      optionalParamName
+      optionalParamName,
+      client
     )}});`
   );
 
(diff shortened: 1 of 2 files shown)


joheredi:
> Just wondering if this is something TCGC should take care of? Doesn't seem like this behavior should vary across languages

This is a good point, I'll bring it up with the TypeSpec/Emitter group

PR #3754, opened 2026-02-12
 Use content-type aware deserialization for error responses

joheredi:
### Problem

When an operation returns an XML error response (e.g., Azure Blob Storage's `StorageError`), the generated deserialize function always used the JSON deserializer (`storageErrorDeserializer(result.body)`), ignoring the response's content type. This caused deserialization failures for services that return XML error bodies.

### Solution

Updated the error deserialization code generation in operationHelpers.ts to detect whether exception types support XML serialization and generate the appropriate deserializer:

- **XML-only exceptions**: Directly uses the XML deserializer (e.g., `storageErrorXmlDeserializer(result.body)`)
- **Dual-format exceptions** (both XML and JSON): Generates a runtime content-type check using `isXmlContentType` to select the correct deserializer
- **JSON-only exceptions** (default): Behavior unchanged — uses the JSON deserializer

### Example generated output

Before:
```typescript
const error = createRestError(result);
error.details = storageErrorDeserializer(result.body);
throw error;
```

After (dual-format):
```typescript
const error = createRestError(result);
const responseContentType = result.headers?.["content-type"] ?? "";
error.details = isXmlContentType(responseContentType)
  ? storageErrorXmlDeserializer(result.body)
  : storageErrorDeserializer(result.body);
throw error;
```

### Changes

- **operationHelpers.ts**
  - Added `xmlDeserializer` and `isXmlOnly` fields to `ExceptionThrowDetail`
  - Added `defaultXmlDeserializer` and `defaultIsXmlOnly` fields to `OperationExceptionDetails`
  - Updated `getExceptionDetails` to check exception `contentTypes` and model XML serialization metadata, building XML deserializers when applicable
  - Added `getExceptionDeserializeExpr` helper to generate the correct deserialization expression based on format support
  - Updated `getExceptionThrowStatement` to emit content-type–aware deserialization for exceptions with XML support

### DIFF
--- a/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
+++ b/packages/typespec-ts/src/modular/helpers/operationHelpers.ts
@@ -151,8 +151,7 @@ export function getSendPrivateFunction(
     `return context.path(${pathStr}).${operationMethod}({...${resolveReference(dependencies.operationOptionsToRequestParameters)}(${optionalParamName}), ${getHeaderAndBodyParameters(
       dpgContext,
       operation,
-      optionalParamName,
-      client
+      optionalParamName
     )}});`
   );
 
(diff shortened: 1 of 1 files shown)


timovv:
Seems ok in principle. The code seems to be combined with the accept header PR work, could you separate it out?

timovv (approved):
LGTM

Issue #3750, opened 2026-02-11
the metadata files should be update when re-generation


skywing918:
After the initial generation, metadata files (such as `metadata.json`, `README.md`, etc.) are created, 
but when re-generation, **these metadata files are not updated**. This leads to inconsistencies 

**The expected behavior** is that repeated generation should consistently refresh and update relevant fields in these metadata files, to ensure they always reflect the latest generated state. 
- client name

kazrael2119:
update `client name` will impact:

1. readme https://github.com/Azure/azure-sdk-for-js/pull/37048/changes/45f4b2f86cbe019b95ad6a13322f66300645a591#diff-9dd24c0d2d44c0adbdd50e4b24c4aec46981f010533a0796a4bbcefb15dcdcca
2. snippets test file https://github.com/Azure/azure-sdk-for-js/pull/37048/changes/45f4b2f86cbe019b95ad6a13322f66300645a591#diff-d3f158ec5599aa958d5142cf42421faca6b7ab1a0ee5023fa787662406849219
3. constantPaths in package.json https://github.com/Azure/azure-sdk-for-js/pull/37295

PR #3764, opened 2026-02-13
re-generation, the metadata files should be update

skywing918:
fixes https://github.com/Azure/autorest.typescript/issues/3750
image

image
image


packages/typespec-ts/src/index.ts
```diff
@@ -577,6 +576,19 @@ export async function $onEmit(context: EmitContext) {
           dpgContext.generationPathDetail?.metadataDir
         );
       }
+
+      // Regenerate snippets.spec.ts if test folder exists
+      if (option.generateTest && isAzureFlavor && hasTestFolder) {
```
kazrael2119: I remember that the value of generateTest should be conflict with the value of hasTestFolder, could you check about this

PR #3780, opened 2026-02-20
Disable parseAttributeValue in XML 

### LINKED ISSUE #3775: XML: "1.0" is coerced to "1"
"<Version>1.0</Version>" currently is parsed into an `Version = 1`.

Using the options in core-xml fixes it

```ts
    parseAttributeValue: false,
    parseTagValue: false,
```

Hope they won't break anything?

joheredi:
Fixes #3775

   ### Problem

   `fast-xml-parser` was configured with `parseTagValue: true` (default) and `parseAttributeValue: true`, which automatically coerces values that look numeric. For a TypeSpec `string` property like `version`, the XML value `<Version>1.0</Version>`
  was parsed as the number `1`, stripping the `.0`.

   ### Solution

   Disable `fast-xml-parser`'s automatic type coercion globally (`parseTagValue: false`, `parseAttributeValue: false`) so all XML values arrive as raw strings. A new `primitiveSubtype` field (`"string" | "number" | "boolean"`) is added to the
  deserialization metadata, allowing the deserialization layer to handle type conversions explicitly based on the TypeSpec type information.

   ### Changes

   **Runtime helper** (`static/static-helpers/serialization/xml-helpers.ts`):
   - Added `primitiveSubtype` to `XmlPropertyDeserializeMetadata` interface
   - Disabled `parseTagValue` and `parseAttributeValue` in parser defaults
   - Updated `deserializePrimitiveValue` to convert raw strings based on subtype: `"boolean"` → `String(value).toLowerCase() === "true"`, `"number"` → `Number(value)`, `"string"` → as-is
   - Fixed `unixTimestamp` date handling to work with string input
   - Threaded `primitiveSubtype` through `deserializeXmlObject` and `deserializeArrayProperty`

   **Code generator** (`src/modular/serialization/buildXmlSerializerFunction.ts`):
   - Extended `getPropertyTypeInfo` to map `SdkType.kind` to `primitiveSubtype`:
     - `"string"` / `"url"` → `"string"`
     - `"boolean"` → `"boolean"`
     - All numeric kinds (`int8`, `int32`, `float64`, etc.) → `"number"`
   - Emits `primitiveSubtype` in generated deserialization metadata arrays

   **Tests** (`test-next/unit/static-helpers/xml-helpers.test.ts`):
   - Updated 10 existing tests to include `primitiveSubtype`
   - Added new test verifying that `<Version>1.0</Version>` correctly deserializes to string `"1.0"` (the original bug scenario)

   ### Safety

   - When `primitiveSubtype` is not provided, `deserializePrimitiveValue` returns the value as-is (raw string), which is safer than accidental numeric coercion
   - `undefined`/missing XML elements are guarded by existing `rawValue === undefined` checks before any conversion runs
   - Empty elements (`<Tag/>`) return `""` which the early-return guard converts to `undefined`, never reaching boolean/number conversion

### DIFF
--- a/packages/typespec-ts/static/static-helpers/serialization/xml-helpers.ts
+++ b/packages/typespec-ts/static/static-helpers/serialization/xml-helpers.ts
@@ -413,6 +413,14 @@ function deserializePrimitiveValue(
     return stringToUint8Array(value, bytesEncoding ?? "base64");
   }
 
+  // Convert raw XML string to the expected JS type
+  if (primitiveSubtype === "boolean") {
+    return String(value).toLowerCase() === "true";
+  }
+  if (primitiveSubtype === "number") {
+    return Number(value);
+  }
+
   return value;
 }
 


timovv (approved):
One comment, not blocking though

PR #3795, opened 2026-02-27
extract and export buildProject for flight-instructor on-demand gener…

qiaozha:
…ation

### DIFF
--- a/packages/typespec-ts/src/utils/modelUtils.ts
+++ b/packages/typespec-ts/src/utils/modelUtils.ts
@@ -1619,7 +1619,7 @@ export function getValueTypeValue(
           getValueTypeValue(value.properties.get(x)!.value)
         ])
       );
-    case "ScalarValue":
+    default:
       // TODO: handle scalar value
       return undefined;
   }


packages/typespec-ts/src/index.ts
```diff
@@ -657,3 +602,176 @@ export async function renameClientName(
...
+  // Wire up the shared contexts used by the builder functions.
+  provideContext("rlcMetaTree", new Map());
+  provideContext("symbolMap", new Map());
+  provideContext("outputProject", outputProject);
+  provideContext("emitContext", {
+    compilerContext: fakeEmitContext,
+    tcgcContext: dpgContext
+  });
+
+  const staticHelpers = await loadStaticHelpers(
```
JialinHuang803: I may have some concerns here - if a new helper is added, we'd need to update both places, which is easy to forget and hard to contain.

packages/typespec-ts/src/index.ts
```diff
@@ -657,3 +602,176 @@ export async function renameClientName(
...
+        ...AzureCoreDependencies,
+        ...AzureIdentityDependencies
+      }
+    : { ...DefaultCoreDependencies };
+
+  const binder = provideBinder(outputProject, {
+    staticHelpers,
+    dependencies: { ...extraDependencies }
+  });
+  provideSdkTypes(dpgContext);
```
JialinHuang803: I am thinking about aggregating the code from line 729 (provideContext) to this line as a new shared method called something like `initializeModularPipeline()` which returns the binder, and both `$onEmit` and `buildProject` can call it. Not sure if it would be better.

PR #3845, opened 2026-03-16
Fix: Exclude required headerCollectionPrefix params from operation signature

joheredi:
### 1. headerCollectionPrefix required parameter fix

**Problem:** When a **required** header parameter is decorated with `@@clientOption("headerCollectionPrefix", ...)`, it was incorrectly included as a positional parameter in the operation method signature. However, the serialization code in the operation body already correctly skips this parameter — resulting in a required parameter that is never used.

**Root Cause:** In `getOperationSignatureParameters()` (`operationHelpers.ts`), the filter that builds the method signature excluded constants, client params, cookies, and auto-generated headers — but did **not** exclude headers with `headerCollectionPrefix`.

**Fix:** Added a single filter condition in `getOperationSignatureParameters()`:

```typescript
getClientOptions(p, "headerCollectionPrefix") === undefined
```

This works because `SdkMethodParameter` extends `DecoratedType`, so `getClientOptions()` can read the `@clientOption` decorator directly from the method parameter.

### Testing

- Added new unit test scenario: `headerCollectionPrefixRequiredSkipSignature.md` — verifies a **required** `metadata` header with `headerCollectionPrefix` is excluded from both the send function and public operation signatures.
- All 581 existing unit tests pass.

### DIFF
--- a/packages/typespec-ts/src/index.ts
+++ b/packages/typespec-ts/src/index.ts
@@ -478,9 +478,7 @@ export async function $onEmit(context: EmitContext) {
       }
     }
 
-    if (option.generateMetadata === false) {
-      // Explicit opt-out: skip all metadata generation and updates
-    } else if (shouldGenerateMetadata) {
+    if (shouldGenerateMetadata) {
       const commonBuilders = [
         buildRollupConfig,
         buildApiExtractorConfig,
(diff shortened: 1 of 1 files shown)


jeremymeng (approved):

jeremymeng (approved):

packages/typespec-ts/test/modularUnit/scenarios/operations/headerParam/headerParamWithClientInitialization.md
```diff
@@ -1,4 +1,4 @@
-# Optional header parameter with clientInitialization
+# only:Optional header parameter with clientInitialization
```
JialinHuang803: The only prefix is not removed. Will create a PR to remove that.

PR #3852, opened 2026-03-18
Fix: Restore full error.details handling for binary wrap-non-model-return deserializers

joheredi:
### Problem

Commit `49500efaa` introduced the `wrap-non-model-return` feature which defaults to `true` for all Azure-flavored clients. The **binary wrap** path had two regressions:

1. **Deserializer skipped `error.details` extraction** — `if (!(shouldWrap && isBinary))` completely bypassed `getExceptionThrowStatement()`, the only function that generates per-status-code `error.details`, exception headers, `@clientOption` headers, and `restErrorCodeHeader` assignment.
2. **Operation function bypassed `getBinaryResponse`/`parsedBody`/`parsedHeaders`** — an early return skipped all normal response processing.

### Root Cause

The binary wrap feature introduced an early-return block in `getOperationFunction` that bypassed the entire normal flow (storage-compat setup, binary response helper, header deserialization, response assembly). The deserializer also took `StreamableMethod` directly and called `getBinaryStream` internally, which consumed the `HttpResponse` inside the deserializer scope — making it unavailable for `_downloadDeserializeHeaders()` in the operation function.

### Fix (Approach: Split helper)

Created a new `getBinaryStreamResponse` static helper that **only** handles platform-specific stream extraction (Node: `asNodeStream()`, Browser: `asBrowserStream()`), then let the generated deserializer handle error checking using the **same `getExceptionThrowStatement` code path** as all other operations.

 The key architectural change: moved `getBinaryStreamResponse` from inside the deserializer to the operation function level, so the `HttpResponse` result is in scope for all downstream consumers:

   - `_downloadDeserialize(result)` — status check + error enrichment via `getExceptionThrowStatement`
   - `_downloadDeserializeHeaders(result)` — header parsing
   - `addStorageCompatResponse(...)` — storage-compat assembly

  The deserializer parameter uses an intersection type (`PathUncheckedResponse` & `XXXResponse`) so it can access both error-handling properties and stream properties without type casts.

#### Changes

**New files:**
- `static/static-helpers/serialization/get-binary-stream-response.ts` — Node variant
- `static/static-helpers/serialization/get-binary-stream-response-browser.mts` — Browser variant

**Modified files:**
- `src/modular/static-helpers-metadata.ts` — Registered `getBinaryStreamResponse` helper
- `src/modular/helpers/operationHelpers.ts`:
  - Removed `if (!(shouldWrap && isBinary))` guard that skipped the status check in the deserializer
  - Deserializer parameter uses intersection type `PathUncheckedResponse & XXXResponse` for binary wrap
  - Removed early-return block in operation function — binary-wrap now flows through the normal path
  - Operation function calls `getBinaryStreamResponse` for binary-wrap, `getBinaryResponse` for non-wrapped binary
  - Fixed bodyType to use wrapped response type name for `StorageCompatResponseInfo` generics
- `test/modularUnit/scenarios/operations/wrapNonModelReturn.md` — Updated binary scenario + added error enrichment scenarios
- `test/modularUnit/scenarios/operations/storageCompat/storageCompatResponse.md` — Added void+headers+error enrichment scenario

### Generated code before vs after

**Before (no error handling):**
```typescript
export async function _getLogsDeserialize(result: StreamableMethod): Promise<GetLogsResponse> {
  return getBinaryResponseBody(result, ["200"]);
}
```

**After (full error handling):**
```typescript
export async function _getLogsDeserialize(
  result: PathUncheckedResponse & GetLogsResponse,
): Promise<GetLogsResponse> {
  const expectedStatuses = ["200"];
  if (!expectedStatuses.includes(result.status)) {
    const error = createRestError(result);
    error.details = apiErrorDeserializer(result.body);
    error.details = { ...(error.details as any), ..._getLogsDeserializeExceptionHeaders(result) };
    throw error;
  }

  return { blobBody: result.blobBody, readableStreamBody: result.readableStreamBody };
}
```

```typescript
export async function getLogs(
  context: Client,
  options: GetLogsOptionalParams = { requestOptions: {} },
): Promise<GetLogsResponse> {
  const streamableMethod = _getLogsSend(context, options);
  const result = await getBinaryStreamResponse(streamableMethod);
  return _getLogsDeserialize(result);
}
```

### Validation

- ✅ `pnpm build` — success
- ✅ `npm run unit-test` — 585 passing, 2 pending
- ✅ `pnpm format` — clean
- ✅ `npm run lint` — clean


jeremymeng:
So currently `_downloadDeserialize()` sends the request, gets the response, returns the streaming body shape `{ blobBody, readableStreamBody,}` but it doesn't return the http response, so I have nothing to feed `_downloadDeserializeHeaders()`

jeremymeng:
I pushed a branch to show the issue and my manual fix.  Not sure if it is easy to update the code gen to have similar fix, when the operation is returning stream body AND storage compat option is enabled.

- branch: https://github.com/Azure/azure-sdk-for-js/tree/storage/codegen-issue-and-manual-fix
- changes from codegen: https://github.com/Azure/azure-sdk-for-js/commit/604f94b13f233fe5a9f08f1d478e69863280f669
- manual fix which now allows us to use the generated method without any workaround: https://github.com/Azure/azure-sdk-for-js/commit/80194bba9a06c08c4aa6b0e4ec99c6917644469b

my manual fix includes
- move `await getBinaryStream(streamableMethod)` into `_downloadDeserialize` so that its result can be passed to `_downloadDeserialize` and `_downloadDeserializeHeaders`
- update the parameter of `_downloadDeserialize` so that it takes the return value from `getBinaryStream()`
- update return types of streaming methods to not have the incorrect `Uint8Array` return type

PR #3888, opened 2026-04-02
Add new xml tests

timovv:
Copilot PR

### DIFF
--- a/packages/typespec-ts/test/azureModularIntegration/payloadXml.spec.ts
+++ b/packages/typespec-ts/test/azureModularIntegration/payloadXml.spec.ts
@@ -141,7 +141,8 @@ describe("Payload XML Client", () => {
     const expected = { tags: ["fiction", "classic"] };
 
     it("should get model with wrapped primitive custom item names", async () => {
-      const result = await client.modelWithWrappedPrimitiveCustomItemNamesValue.get();
+      const result =
+        await client.modelWithWrappedPrimitiveCustomItemNamesValue.get();
       assert.deepEqual(result, expected);
     });
 
@@ -215,7 +216,8 @@ describe("Payload XML Client", () => {
     };
 
     it("should get model with renamed unwrapped model array", async () => {
-      const result = await client.modelWithRenamedUnwrappedModelArrayValue.get();
+      const result =
+        await client.modelWithRenamedUnwrappedModelArrayValue.get();
       assert.deepEqual(result, expected);
     });
 
@@ -230,7 +232,8 @@ describe("Payload XML Client", () => {
     };
 
     it("should get model with renamed wrapped and item model array", async () => {
-      const result = await client.modelWithRenamedWrappedAndItemModelArrayValue.get();
+      const result =
+        await client.modelWithRenamedWrappedAndItemModelArrayValue.get();
       assert.deepEqual(result, expected);
     });
 


packages/typespec-ts/test/azureModularIntegration/payloadXml.spec.ts
```diff
@@ -25,6 +30,65 @@ describe("Payload XML Client", () => {
...
+  describe("ModelWithRenamedProperty", () => {
+    const expected = { title: "foo", author: "bar" };
+
+    it("should get model with renamed property", async () => {
+      const result = await client.modelWithRenamedPropertyValue.get();
+      assert.deepEqual(result, expected);
+    });
+
+    it("should put model with renamed property", async () => {
+      await client.modelWithRenamedPropertyValue.put(expected);
```
xirzec, line 42: is this expected to throw if the shape is wrong? Maybe we should check that?

packages/typespec-ts/test/azureModularIntegration/payloadXml.spec.ts
```diff
@@ -25,6 +30,65 @@ describe("Payload XML Client", () => {
...
+  });
+
+  describe("ModelWithRenamedFields", () => {
+    const expected = {
+      inputData: { name: "foo", age: 123 },
+      outputData: { name: "bar", age: 456 }
+    };
+
+    it("should get model with renamed fields", async () => {
+      const result = await client.modelWithRenamedFieldsValue.get();
```
xirzec, line 53: it's difficult to understand these assertions since the wire and model shape isn't obvious without looking at the client implementation... I feel like it would be more useful if we used a pipeline policy to show the wire string and validate that instead

packages/typespec-ts/test/azureModularIntegration/payloadXml.spec.ts
timovv: You mean the shape on the wire? Yes, the Spector test server would return a 400 if it doesn't get what it expects. The generated client then throws, causing the test to fail. I don't know what we'd be checking beyond "is Spector doing its job here"? Which I'm not sure is really our problem?

I think there are already some Spector test cases somewhere which are for ensuring the emitter does throw on an unexpected status code. I don't know if we would want to add a test case for that in every instance.

packages/typespec-ts/test/azureModularIntegration/payloadXml.spec.ts
timovv: This test is exercising the emitted models and deserializer specifically. For this case the Spector test service returns a well-known shape and we are validating two things:
- Are the generated models right, that is, was the property renamed in the public API surface? If not, then the test should fail to compile, at least in theory... and even if it did, I would expect the assertion would fail
- Is the mapping from the wire to the model correct at runtime? If not, again, the assertion fails

Maybe I am misunderstanding the comment... as I understand it, more or less all of our Spector tests follow a similar pattern. If we just validate what comes back from Spector on the wire, we're not testing the emitter, we are testing the Spector test server.

JialinHuang803:
@timovv this is the PR created by the weekly scheduled workflow: https://github.com/Azure/autorest.typescript/pull/3909/changes#diff-d2dad8f4c10062d316a32e2d0df40379607a8fabd0c7cf64a79fe0bcb7c84f73 which includes the spector tests here (as we discussed before), and the implementation is almost the same. Feel free to have a review on it.

qiaozha:
Can we close this pr in favor of #3909?

timovv:
Yeah, that PR has this coverage in there already.

PR #3911, opened 2026-04-13
Fix arm resource model property filtering for specific cases

### LINKED ISSUE #3910: Need to filter out specific cases for ARM when dealing with metadata model properties
There are cases from Compute where some metadata properties should be filtered out for some reasons. Previously we only filtered out "name" property from ARM resources extending from CommonTypes.Resource and it's not enough.

Case 1: https://github.com/Azure/azure-rest-api-specs/blob/b7c7ac179ff8c30149b84e5f7811a8ee1fd812d0/specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/VirtualMachineScaleSetExtension.tsp#L37-L41.
The ARM resource doesn't extend from CommonTypes.Resource, and the name of the property isn't "name".
Image
The reason why this property should be removed from the model is that the `@key` decorator adds an additional method parameter for all the ARM operations that use this resource model. And `@path` + `@visibility(Lifecycle.Read)` means this property should not exist in the http body payload (the `@Http.Private.includeInapplicableMetadataInPayload(false)` decorator in https://github.com/Azure/azure-rest-api-specs/blob/d578ca07528799866e3c30db70d9b193b3a27a83/specification/compute/resource-manager/Microsoft.Compute/Compute/common/models.tsp#L149-L150 makes it happen for Autorest, but TCGC doesn't support it.

Case 2: https://github.com/Azure/azure-rest-api-specs/blob/d578ca07528799866e3c30db70d9b193b3a27a83/specification/compute/resource-manager/Microsoft.Compute/Compute/common/models.tsp#L159-L162
Similar with Case 1, but without `@key`.
Image
Although this is a workaround for TypeSpec conversion and will be removed, but we also want to remove this kind of property from the ARM model. This property is removed from the model in Autorest because of the `@Http.Private.includeInapplicableMetadataInPayload(false)` decorator.

[Playground link](https://azure.github.io/typespec-azure/playground/?options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40azure-tools%2Ftypespec-azure-rulesets%2Fresource-manager%22%5D%7D%7D&c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7CtIZcmVzdNUZdmVyc2lvbmluZ8wfYXp1cmUtdG9vbHMvyCstxhVjb3Jl3yvIK3Jlc291cmNlLW1hbmFnZXIiOwoKdXNpbmcgSHR0cDvHDFJlc3TIDFbpAI7IEkHESi5Db3JlzhJSx1xNxls7CgovKiogQ29udG9zb8RUxR4gUHJvdmlkZXIg5gCDbWVudCBBUEkuICovCkBhcm3IIE5hbWVzcGFjZQpAc2VydmljZSgjeyB0aXRsZTogIsdXyC1IdWJDbGllbnQiIH0pCkDnAUNlZCjnAL9zKQpuyFAgTWljcm9zb2Z0LtJG7wC2QVBJIMdNc%2BQAoWVudW3oARNzIHsKICDELjIwMjEtMTAtMDEtcHJldmlld8g1xDQgIOQA10NvbW1vblTkAY%2FHQCj1ATcuyykuyGoudjUpxEjHXscXCiAgYNJ7YCwKfQoKI3N1cHByZXNz%2FwI7b3JlL2RvY3XkAYdhdGlvbi1yZXF1aXJlZCIgIkZvciBiYWNrd2FyZCBjb21wYXRpYmlsaXR5Igptb2RlbCBTdWLoAMBSZWFkT25seeUBFuUBGMUIICrqAftJZMcT5AEVICBAdmlzx1AoTGlmZWN5Y2zkARBhZOQA7SAgaWQ%2FOusCbi5hcm3IS0lkZW50aWZp5AJn%2FwDu8ADubm8tcHJpdmF0ZS11c2FnZf8A6N9b0FtsZWdhY3nfWsZaQGRvYyjkAsljcmV0ZSBwcm94eSDoA6EgxWAgY2FuIGJlIGNyZWF0ZWQgYnkgYWxpYeUDdXRoaXPFJiDmA4VhIOQAj2lmaWPES3BlcnR5xR8uIikKQOQD3i5Q5gD3LmluY2x1ZGVJbmFwcGxpY2FibGVNZXRhZGF0YUluUGF5bG9hZChmYWxzZSkKQPYCpUzlAOcuY3VzdG9txSPIIucCGVDkAM7zAh48UOYAnGllcyBleHRlbmRzIHt9PuUB8MgQ0zfnAlXmATNUaGXpASgt8AD1xFlmb3LmAR%2FIJuQBBMRAyiQ%2FOuQAo8cNOwrwArPEY%2BQEMSBvZiB0aGUgcGFy5ASoyVDuAshwYXRo%2FwLS5gCAQHNlZ%2BQDaigixlFz6ACYxQ7kBPI6IHN0cuUFbX3lBIoKICogRGVzY3JpYmVzIGEgVmlydHVhbCBNYWNoaW5lIFNjYWxlIFNldCBF5AE%2B5AQXLgrkAJn%2FArj4BhEvYXJtLeYB18oc5QLOLWRpc2NvdXL%2FAtnpBBvnAKrnAKnlAKhTZXTpAKbmA%2BJz%2BgIp30DqAb4%2B%2BwRy5ADu%2FwR07wGiZW5jb2RlZOQBjCgi5wMh5AUYL2pzb24iLCAixFPnAbd2beQFpz%2FqAbT5AIjkAkztAIjEEds58AKAVk0gc%2BUB6nNldOYDKOcB6v8Ch%2F8ChyLJS%2BgCikBrZXkoInZtc3PpAWrkAQvHHtEY%2FgKv5QM66gOXb2b%2FAsHtAsH%2FAkbwAgbyAgVJxG92YWx1ZeQCaHDmB%2BRkIGFuZMQQZGlmZmXlA9Vmcm9t5wCrZXZpb3VzxjUsxRTpASEgaGFuZGxlciB3aWxs5AWUZm9yY2VkIHRvIHVwZGF0ZSBldmVuIGnmAbDKN2NvbmZpZ3Vy5QJFIGhhcyBub3QgY2hhbmdlZO4BwcVSVcVNVGFn6wIefQrlCRroAk5PcGXGT3MKaW50ZXJm5AjV%2FwEp8wEg5AJk5AE%2BxUvkAMhnZXTuAL3uAJ%2FEH2lzIEHqAInkAlI8xiQgIP8Agm4sySlQYXJhbWV0ZXJzID3nAJfIAekAn8gB5gCnZXhw5AG3ZXjlBNLnAK%2FkA515IMQM7ADJ5wC0yUDmALzIAUBxdWVyeSgiJMZS5wLXyAHHFusBesgafcUKPvYDqcl85ACW5gerIG9y6AIXYW74AVHGK09y5gHz7wFcQ8cdUmVwbGFjZUFzeW5j%2FwFs9AFsTHJvSGVhZOYBbEFybUxyb0xv5gTSxhg8RmluYWxSZXN1bHQgPd9Zbj4gJslbxAHrCZ5Gb3VuZMVZcy5SZXRyeUFmdGVyxmXoAT59&e=%40azure-tools%2Ftypespec-autorest&vs=%7B%7D)

JialinHuang803:
See issue description in #3910.

To fix the gap of `@Http.Private.includeInapplicableMetadataInPayload(false)` and `@key` handling in TypeSpec for ARM resource models after supporting generating metadata parameters in models (#3848 ), the workaround here is to filter out metadata properties that are **required** and has **visibility.read** for ARM.

### DIFF
--- a/packages/typespec-ts/src/modular/emitModels.ts
+++ b/packages/typespec-ts/src/modular/emitModels.ts
@@ -33,6 +33,7 @@ import {
   isReadOnly,
   listAllServiceNamespaces
 } from "@azure-tools/typespec-client-generator-core";
+// import { isKey } from "@typespec/compiler";
 import {
   getExternalModel,
   getModelExpression,
@@ -59,7 +60,7 @@ import {
 import path from "path";
 import { refkey } from "../framework/refkey.js";
 import { useContext } from "../contextManager.js";
-import { isMetadata, isOrExtendsHttpFile } from "@typespec/http";
+import { isMetadata, isOrExtendsHttpFile, Visibility } from "@typespec/http";
 import { isAzureCoreErrorType } from "../utils/modelUtils.js";
 import { getHeaderClientOptions } from "./helpers/clientOptionHelpers.js";
 import { isExtensibleEnum } from "./type-expressions/get-enum-expression.js";
@@ -736,7 +737,6 @@ function buildModelInterface(
   // properties (@header, @query, @path) since they are deserialized separately.
   // For input models, keep metadata properties — users need to pass them.
   const hasInputUsage = (type.usage & UsageFlags.Input) === UsageFlags.Input;
-  const isArmResource = isArmResourceModel(type);
   const interfaceStructure = {
     kind: StructureKind.Interface,
     name: normalizeModelName(context, type, NameType.Interface, true),
@@ -746,15 +746,8 @@ function buildModelInterface(
         if (!hasInputUsage && p.__raw && isMetadata(context.program, p.__raw)) {
           return false;
         }
-        // Skip the "name" metadata property on ARM resource models.
-        // ARM resource "name" is a @path property inherited from the base Resource type
-        // and is handled by the ARM infrastructure, not set by the user directly.
-        if (
-          isArmResource &&
-          p.name === "name" &&
-          p.__raw &&
-          isMetadata(context.program, p.__raw)
-        ) {
+        // Skip required metadata properties with Read visibility for ARM as they are not intended to be in the model
+        if (context.arm && p.__raw && isMetadata(context.program, p.__raw) && !p.optional && p.visibility?.includes(Visibility.Read)) {
           return false;
         }
         // filter out the flatten property to be processed later
@@ -948,20 +941,6 @@ export function normalizeModelName(
   const internalModelPrefix =
     isPagedResultModel(context, type) || type.isGeneratedName ? "_" : "";
   return `${internalModelPrefix}${normalizeName(namespacePrefix + type.name, nameType, true)}${unionSuffix}`;
-}
-
-/**
- * Checks if a model descends from the ARM common-types Resource base type
- * (TrackedResource, ProxyResource, etc.) by walking the ancestor chain.
- */
-function isArmResourceModel(type: SdkModelType): boolean {
-  const ancestors = getAllAncestors(type);
-  return ancestors.some(
-    (ancestor) =>
-      ancestor.kind === "model" &&
-      ancestor.crossLanguageDefinitionId ===
-        "Azure.ResourceManager.CommonTypes.Resource"
-  );
 }
 
 function buildModelPolymorphicType(context: SdkContext, type: SdkModelType) {
--- a/packages/typespec-ts/src/modular/emitModels.ts
+++ b/packages/typespec-ts/src/modular/emitModels.ts
@@ -747,11 +747,11 @@ function buildModelInterface(
           return false;
         }
         // Skip required metadata properties with Read visibility for ARM as they are not intended to be in the model
+        // These properties are not be generated no matter they are in input or output models in most cases in HLC
         if (
           context.arm &&
           p.__raw &&
           isMetadata(context.program, p.__raw) &&
-          !p.optional &&
           p.visibility?.includes(Visibility.Read)
         ) {
           return false;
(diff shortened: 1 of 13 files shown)


packages/typespec-ts/src/modular/emitModels.ts
```diff
@@ -746,14 +746,13 @@ function buildModelInterface(
         if (!hasInputUsage && p.__raw && isMetadata(context.program, p.__raw)) {
           return false;
         }
-        // Skip the "name" metadata property on ARM resource models.
-        // ARM resource "name" is a @path property inherited from the base Resource type
-        // and is handled by the ARM infrastructure, not set by the user directly.
+        // Skip required metadata properties with Read visibility for ARM as they are not intended to be in the model
```
qiaozha, line 749: wonder why not also filter out optional but visibility read metadata properties?

qiaozha: Could you add some description about why this filter won't introduce issues to response?

qiaozha (approved):
LGTM, just some minor comments.

packages/typespec-ts/src/modular/emitModels.ts
JialinHuang803: Seems the optionality doesn't matter here. Removed the optional check.

Regarding response, I just tested in the playground, and the metadata & visibility.read properties are removed from the whole model if `@Http.Private.includeInapplicableMetadataInPayload(false)` is set or `ResourceNameParameter` is used (not sure about the root cause for the second case). Therefore, both request and response should not have this type of properties. I didn't add such long description in the code, so explain a bit more here.

PR #3926, opened 2026-04-20
Decouple typespec-ts from rlc-common

timovv:
In preparation for move to the TypeSpec repo. This PR removes the dependency `typespec-ts` has on `rlc-common` so that we don't need to migrate `rlc-common` to the new repo. Since autorest is going away, there's no point in having a separate rlc-common package.

This was achieved by:
- Copying the rlc-common source code into `typespec-ts/src/rlc-common/`
- Updating all rlc-common references in `typespec-ts` to refer to the new subfolder instead of the external package
- Fixing lint errors 
- Removing (some) now-unused code in the rlc-common subfolder.

This PR makes the migration easier and opens the door for a more thorough refactor/integration of the rlc-common code in the future.


qiaozha:
@JialinHuang803 could you help take a look at this PR? Thanks

packages/typespec-ts/src/framework/hooks/sdkTypes.ts
```diff
@@ -18,7 +18,7 @@ import {
 } from "../../modular/helpers/operationHelpers.js";
 import { normalizeModelPropertyName } from "../../modular/type-expressions/get-type-expression.js";
 import { reportDiagnostic } from "../../lib.js";
-import { NameType, normalizeName } from "@azure-tools/rlc-common";
+import { NameType, normalizeName } from "../../rlc-common/index.js";
```
JialinHuang803, line 21: Could we avoid the 'rlc' related naming?

timovv: Yeah, we could. The reason I left it in for now is because (a) I don't know a better name and (b) it makes it clear which parts of the codebase _used_ to be part of rlc-common, which might aid us if we do a more thorough refactor in future. That said I don't feel too strongly. Do you have any thoughts on a better name?

JialinHuang803: If we don't plan to fully decouple modular from rlc in this PR, like the metadata generation and rlc options, I agree that we keep the name as it is for now.

JialinHuang803 (approved):

maorleger (approved):
Seems like the right approach with autorest going away, and makes moving to typespec-azure much easier

PR #3967, opened 2026-05-15
[typespec-ts] Fix missing config/ tsconfig files when updating existing packages

maorleger:
## Problem

Since the 0.53.0 release (which included #3938), SDK Validation is broken for all TypeSpec-based PRs in azure-rest-api-specs. The error looks like:

```
[warp] Invalid config in warp.config.yml: target "browser" references tsconfig
"./config/tsconfig.src.browser.json" which does not exist at
.../azure-sdk-for-js/sdk/cognitiveservices/arm-cognitiveservices/config/tsconfig.src.browser.json
```

## Root Cause

There are two code paths in `packages/typespec-ts/src/index.ts` for metadata generation:

1. **New package** (`shouldGenerateMetadata = true`) — generates `warp.config.yml` AND the `config/` tsconfig files ✅
2. **Update existing package** (`shouldGenerateMetadata = false`, i.e. `package.json` already exists and `generate-metadata` is not explicitly `true`) — generates `warp.config.yml` but **not** the `config/` tsconfig files or `tsconfig.json` ❌

SDK automation always hits path (2) since the packages already exist in azure-sdk-for-js. So `warp.config.yml` gets written with `./config/tsconfig.src.browser.json` references, but the `config/` directory is never created. Maybe it assumes all existing packages in the repo would have been migrated already but that ended up not being the case?

## Fix

I added the missing `buildTsConfig`, `buildTsSrcEsmConfig`, `buildTsSrcBrowserConfig`, `buildTsSrcReactNativeConfig`, and `buildTsSrcCjsConfig` builders to the update path, matching what the new-package path already does. Also added a consistency test to make sure every tsconfig path referenced in `warp.config.yml` has a corresponding builder.

## Validation

I repro'd the bug e2e with a minimal ARM TypeSpec spec with `azure-sdk-for-js: true` and a pre-existing `package.json` (to trigger the update path), then verified the fix by applying the same process:

**Before fix:**
| File | Status |
|------|--------|
| `warp.config.yml` | ✅ Generated (references `./config/tsconfig.src.*.json`) |
| `config/tsconfig.src.browser.json` | ❌ Missing |
| `config/tsconfig.src.esm.json` | ❌ Missing |
| `config/tsconfig.src.cjs.json` | ❌ Missing |
| `config/tsconfig.src.react-native.json` | ❌ Missing |
| `tsconfig.json` | ❌ Not updated |

**After fix:**
| File | Status |
|------|--------|
| `warp.config.yml` | ✅ Generated |
| `config/tsconfig.src.browser.json` | ✅ Generated |
| `config/tsconfig.src.esm.json` | ✅ Generated |
| `config/tsconfig.src.cjs.json` | ✅ Generated |
| `config/tsconfig.src.react-native.json` | ✅ Generated |
| `tsconfig.json` | ✅ Updated with `./config/` references |

All existing tests pass (rlc-common 103/103, typespec-ts unit tests unchanged). I also confirmed the autorest.typescript generator does not have this issue — it uses a single linear code path with no new/update split.

### DIFF
--- a/packages/rlc-common/test/integration/warpConfig.spec.ts
+++ b/packages/rlc-common/test/integration/warpConfig.spec.ts
@@ -4,6 +4,12 @@
 import { describe, it, expect } from "vitest";
 
 import { buildWarpConfig } from "../../src/metadata/buildWarpConfig.js";
+import {
+  buildTsSrcEsmConfig,
+  buildTsSrcBrowserConfig,
+  buildTsSrcReactNativeConfig,
+  buildTsSrcCjsConfig
+} from "../../src/metadata/buildTsConfig.js";
 import { createMockModel } from "./mockHelper.js";
 
 describe("warp.config.yml generation", () => {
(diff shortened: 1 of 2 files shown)


qiaozha:
close it as we don't need to fix in this way. let me know if any concerns @maorleger

PR #3970, opened 2026-05-15
[typespec-ts] POC: Separate rendering and data manipulation concerns

maorleger:
## What and Why

I wanted to address a few concerns and provide a POC that shows a different approach. As I was designing it, I focused on:

1. Separation of concerns: separate what to render from how we render it
2. Ease of maintenance: simplifying the logic at the "view" or render layer is a common strategy to get around how hard it can be to test views. By having a separate transformation between TCGC and what we end up using as input for rendering we can test that logic well and in isolation
3. Preparation for Alloy.js - we keep trying to do a big migration right into alloy and it doesn't seem to pan out. There's a common saying "for each desired change, make the change easy (warning: this may be hard), then make the easy change." and so a goal here is to make it easier to switch to alloy in the future while not bundling that switch into this changeset.

As I was thrashing through the idea, I checked what other emitters are doing and realized I landed on almost the same architecture go and rust emitters use today 😄 

At that point I added goal 4: be similar to other emitters. The more we can align the easier it'll be to jump between emitters or just grok the overall architecture.

So I then updated some names and concepts, and tried to align with the existing emitters. Landing on this desired architecture:

```
TCGC Adapter  →  Code Model (language IR)  →  Codegen (string rendering)
```

Today our `src/modular/` directory has ~36 files that mix TCGC interpretation with ts-morph rendering in the same functions. This makes the code hard to test in isolation, hard to swap rendering backends (e.g. Alloy.js), and structurally different from our sibling emitters.

### The three layers

**1. `src/tcgcadapter/`** — The _only_ layer that imports TCGC types. Single-pass transformation from `SdkContext` → `TSCodeModel`. This is analogous to:
- Go: `typespec-go/src/tcgcadapter/` → `tcgcToGoCodeModel()`
- Rust: `typespec-rust/src/tcgcadapter/` → `tcgcToCrate()`

**2. `src/codemodel/`** — Pure TypeScript-specific IR types (`TSCodeModel`, `TSClient`, `TSMethod`, `TSEndpointConfig`, etc.). No dependencies on TCGC or ts-morph. This is analogous to:
- Go: `codemodel.go/` (`CodeModel`, `Client`, `Method`, `Package`)
- Rust: `codemodel/` (`Crate`, `Client`, `Method`, `Model`)

**3. `src/codegen/`** — Walks the code model tree and emits source files. Zero TCGC imports. This is analogous to:
- Go: `codegen.go/` (`Emitter.emit()` → `generateClientFactory()`, `generateOperations()`, etc.)
- Rust: `codegen/` (`CodeGenerator.emitContent()` → `emitClients()`, `emitModels()`, etc.)

### POC scope

This PR proves the pattern end-to-end with **client context files** only (`api/fooContext.ts`). Everything else still uses the existing path — I am using a feature-flag-style routing in `src/index.ts` so the old and new paths coexist cleanly.

The idea is that we can incrementally migrate each file category (operations, classical client, models, serialization, etc.) by:
1. Adding adapter functions to populate the code model
4. Adding codegen functions to walk and emit those files
5. Removing the old builder call

### What I would love feedback on

- Does the `TSCodeModel` shape feel right? I tried to match Go/Rust's patterns (client-centric tree with methods, parameters, operation groups) while keeping it TypeScript-idiomatic
- The adapter currently reuses some existing helpers from `src/modular/helpers/` — is that pragmatic enough or should we be more aggressive about extracting clean adapter-only logic?
- The codegen layer uses `resolveReference` / `useDependencies` hooks for import management — this is a rendering concern that stays in codegen, but I wonder if there's a cleaner way to handle it

### Validation

- Unit tests: 639 passed ✅
- Smoke tests: all specs succeeded ✅
- Integration: pre-existing api-extractor bug (same on `main`, not introduced by this PR)

Not for merging yet — just want to get eyes on the direction before expanding scope :)


maorleger:
Closing in favor of a more complete rewrite coming up

PR #3979, opened 2026-05-18
[typespec-ts] Revert #3938 (migrate generated configs to eng/tsconfigs pattern) and its two dependents

### LINKED ISSUE #3964: newly generated JS repo flavor packages fail to build
## issue: 

The code generator should also generate the new `./config` and relevant config files but didn't

## analysis 
SDK Validation - JS is broken for all TypeSpec-based PRs in azure-rest-api-specs

Error:
 

[warp] Invalid config in warp.config.yml: target "browser" references tsconfig 
"./config/tsconfig.src.browser.json" which does not exist at 
.../azure-sdk-for-js/sdk/cognitiveservices/arm-cognitiveservices/config/tsconfig.src.browser.json
 

Root cause:
 

The JS emitter @azure-tools/typespec-ts version 0.53.0/0.53.1 (published today, May 14) generates warp.config.yml with
new-style local paths (./config/tsconfig.src.browser.json) per Azure/autorest.typescript#3938 (
https://github.com/Azure/autorest.typescript/pull/3938) ("feat: migrate generated configs to eng/tsconfigs pattern").
However, the emitter is generating warp.config.yml with these new paths without also generating the config/ directory
containing the thin tsconfig files that extend from eng/tsconfigs/.

## fix

We should include the config/tsconfig.src.*.json files that are referenced in warp.config.yml under `targets` section.  They should be similar to files under https://github.com/Azure/azure-sdk-for-js/tree/e27a67644732133358031641557731736e0b4f4b/sdk/core/core-rest-pipeline/config

maorleger:
Resolves #3964

The TL;DR is that since v0.53.0, SDK Validation has been broken for any TypeSpec-based PR in `azure-rest-api-specs` because the generated `warp.config.yml` references `./config/tsconfig.src.*.json` files that don't actually get emitted on the "update existing package" code path. That's the path SDK automation usually hits, so this affects everyone.

I went back and forth on whether to take #3967 (which patches the symptom by adding the missing builders to the update path) or to just revert the whole thing. I landed on revert for two reasons:

1. The pre-#3938 design was strictly simpler — one builder (`buildTsSrcConfig`), one template, repo-root `../../../tsconfig.src.*.json` references, single linear code path. No new-vs-update fork to keep in sync, and known-good for both new and existing packages.
2. The eng/tsconfigs alignment goal of #3938 is worth doing, but I think it should be reattempted with a cleaner design (single linear path + the consistency test #3967 added). That feels safer as a follow-up than as a fast-follow on top of an already-broken release.

I'm open to hearing if there's a reason to keep the #3938 design — but for now this gets us back to a healthy place.

### What this PR reverts

Three commits, in dependency order:

| Commit | PR | Why reverted |
|---|---|---|
| `679311a5` | #3943 — Do not generate React Native build targets by default | Depends on #3938's split warp targets |
| `813fd759` | **#3938 — migrate generated configs to eng/tsconfigs pattern** | Root cause |
| `ab712bd0` | #3941 — stop generating vitest.esm.config.ts | Depends on #3938's removal of `buildVitestConfig("esm")` |

All 11 unrelated commits between #3938 and HEAD (version bumps, real bug fixes, spector tests, the RestError re-export, etc.) are kept as-is.

### Conflict resolution notes

Reverting #3938 surfaced 5 conflicts in regenerated test fixtures under `packages/typespec-test/test/.../generated/`. They were all about `NodeReadableStream` (added by #3938) vs `RestError`/`isRestError` (added later by unrelated #3930). I kept the `RestError`/`isRestError` exports and dropped the `NodeReadableStream` type alias — that matches what the reverted emitter would actually produce.

### Validation

I built the emitter from this branch and ran an end-to-end pipeline (`tsp compile` → `pnpm turbo build` against a local `azure-sdk-for-js`) for both code paths:

| Spec | Path | Result |
|---|---|---|
| `cognitiveservices/CognitiveServices.Management` (Azure/azure-rest-api-specs#41922) | Update existing package — the case that's failing on `main` | ✅ 19/19 turbo tasks |
| `scvmm/ScVmm.Management` (deleted the existing SDK package first to force the new path) | New package | ✅ 19/19 turbo tasks |

I also spot-checked the generated `warp.config.yml` and it's back to `../../../tsconfig.src.*.json` — no `./config/` indirection.

### Follow-up

- Re-attempt the eng/tsconfigs alignment with a single linear builder path (no new-vs-update fork) and the consistency test from #3967.
- I noticed two other regression classes while testing this branch (orphan subpath exports in `warp.config.yml` for modular operation groups, and raw `NodeJS.*` types leaking into RLC browser code). Those are unrelated to #3938 and I'll file separate issues.


maorleger:
[Doing the migration for all packages moves 3000+ files of unnecessary config changes](https://github.com/Azure/azure-sdk-for-js/pull/38557). That's not really great, and we want to simplify all this if we can. I think reverting this change and revisiting with a fresh perspective is the right choice but feel free to disagree!

JialinHuang803:
"https://github.com/Azure/autorest.typescript/pull/3943 — Do not generate React Native build targets by default" may not relate to #3938 because the react-native build target has been existing in the warp.config since it was introduced in the codegen based on https://github.com/Azure/autorest.typescript/pull/3802.

And we need to revert the changes on the packages in the sdk repo. Please correct me if my understanding is wrong.

maorleger:
> "#3943 — Do not generate React Native build targets by default" may not relate to #3938 because the react-native build target has been existing in the warp.config since it was introduced in the codegen based on #3802.
> 
> And we need to revert the changes on the packages in the sdk repo. Please correct me if my understanding is wrong.

Sorry I forgot to close this, you're quite right on both counts - and Jeremy did end up migrating the entire repo so this PR is not even needed. Closing 👍

Issue #3907, opened 2026-04-13
[JavaScript][Perf] Optimize speed to generate sdk


JialinHuang803:

lirenhe:
Do you have any data about how much time it takes to build some large RP, e.g.  Compute, Network, AppService, etc.?

JialinHuang803:
The recorded time for the Generate SDK job is as follows:
Network - 46m28s
SQL - 29m59s
Compute - 12m57s

lirenhe:
Let's target to reduce the total time within 10 min for those RPs,  looks there is a big gap especially for Network.

PR #4004, opened 2026-05-26
Improve resolveReferences performance

JialinHuang803:
## Summary

Two complementary changes to `Binder` that reduce the cost of the `resolve references` emit phase:

1. **Batch placeholder replacement** in `resolveAllReferences` so each source file gets **one** bulk `text.replace` + **one** `replaceWithText`, instead of one cycle per placeholder. Each `replaceWithText` discards ts-morph's cached AST and re-parses the file, so the previous per-placeholder loop grew as O(placeholders per file × file size).
2. **Bulk-add import declarations** with a single `addImportDeclarations` call per file, instead of looping `addImportDeclaration` once per import. Each individual call triggers a full file re-parse in ts-morph.

## What changed

For each source file the binder now:

1. Scans the text once via `collectPlaceholders` to collect placeholders actually present in the file.
2. Iterates the precomputed `declarationByPlaceholder` / `dependencyByPlaceholder` maps (preserving the original insertion order, so name-collision aliasing in `addImport` is unchanged).
3. Accumulates a single placeholder → local-name map.
4. Applies all replacements via `applyReplacements` in one bulk `text.replace` + one `replaceWithText`.
5. Emits all collected imports for the file in a single `addImportDeclarations` call (sort order unchanged).

Only `packages/typespec-ts/src/framework/hooks/binder.ts` is touched.

## Performance results

Measured by running `npx tsp compile client.tsp --emit=@azure-tools/typespec-ts` against the real Azure spec files (`specification/network/.../Network/client.tsp` and `specification/compute/.../Compute/client.tsp`), with per-phase `console.time` instrumentation around `$onEmit`. Single run per side.

| Spec    | `resolve references` (before) | `resolve references` (after) | Total `onEmit` (before) | Total `onEmit` (after) |
|---------|------------------------------:|-----------------------------:|------------------------:|-----------------------:|
| Network |                     22:08.539 |                       4.397s |              31:25.022  |              7:27.367  |
| Compute |                      3:20.601 |                       2.082s |               5:17.883  |              1:47.209  |

- **Network: ~302× speedup** on the binder phase, **~24 min saved** per SDK generation (31:25 → 7:27 end-to-end).
- **Compute: ~96× speedup** on the binder phase, **~3.5 min saved** per SDK generation (5:18 → 1:47 end-to-end).

Incremental gain from the follow-up `addImportDeclarations` commit, on top of the placeholder-batching commit:

| Spec    | `resolve references` before | after  | Δ                  |
|---------|----------------------------:|-------:|-------------------:|
| Network |                      8.664s | 4.397s | −49% (−4.27s)      |
| Compute |                      2.942s | 2.082s | −29% (−0.86s)      |

See the baseline timing breakdown in [#3907](https://github.com/Azure/typespec-azure/issues/4749) — `resolve references` accounted for 70.5% of total emit time on Network and 63.1% on Compute before this change.


JialinHuang803:
There is an existing test file for binder https://github.com/Azure/autorest.typescript/blob/main/packages/typespec-ts/test-next/integration/hooks/binder.test.ts and all cases pass.

packages/typespec-ts/src/framework/hooks/binder.ts
```diff
@@ -249,9 +249,54 @@ class BinderImp implements Binder {
    * Applies all tracked imports to their respective source files.
    */
   resolveAllReferences(sourceRoot: string, testRoot?: string): void {
+    // Pre-compute placeholder -> declaration/dependency maps
+    const declarationByPlaceholder = new Map<
+      string,
+      [unknown, DeclarationInfo | StaticHelperMetadata]
```
qiaozha, line 255: why the first element of the tuple is unknown? is it the return type of this.serializePlaceholder?

qiaozha: oh wait it should be the input of this.serializePlaceholder.

qiaozha (approved):

packages/typespec-ts/src/framework/hooks/binder.ts
maorleger: not blocking for this PR but I wonder if we can get some more wins by batching operations. This pattern is explicitly called out in https://ts-morph.com/manipulation/performance#performance-tip-batch-operations and may be an easy replacement to make

Have you seen the [performance ts-morph docs](https://ts-morph.com/manipulation/performance#performance-tip-batch-operations) ? Could be useful to point an agent to them and find low-hanging fruit to increase perf as well

maorleger (approved):
I'm far from an expert here but the perf numbers are definitely exciting! I linked to the ts-morph performance documentation in a comment and I wonder if there are other places where we can batch operations for quick wins, not blocking of course

packages/typespec-ts/src/framework/hooks/binder.ts
JialinHuang803: Yes, the `unknown` here is consistent with the key type of this.declarations: `private declarations = new Map<unknown, DeclarationInfo>();`.

packages/typespec-ts/src/framework/hooks/binder.ts
JialinHuang803: Great call, thanks for the pointer, Maor! I replaced the loop with a single `addImportDeclarations` bulk call in the latest commit.

PR #4015, opened 2026-06-01
Batch ts-morph mutations across modular emit phases

JialinHuang803:
## Summary

Generalise the ts-morph batching pattern shipped in #4004 (which targeted `resolveReferences`) into a reusable `framework/sourceFileBatch.ts` module, and apply it to the modular emit phases that still walked the `add*` hotpath one statement at a time.

In ts-morph 23, every individual `sourceFile.addX(...)` call re-parses the entire source file via `insertChildText` -> `doManipulation`. Loops that call `addExportDeclaration` / `addInterface` / etc. against the same file therefore grow as O(N x file_size). The bulk variants (`addExportDeclarations([...])`, `addInterfaces([...])`, ...) collapse N re-parses into 1 per kind.

## Design

A single module-level reference counter (`batchDepth`) tracks how many `beginSourceFileBatch` scopes are open, with one shared `Map<SourceFile, StatementStructures[]>` collecting pending writes. Each `enqueueStatement` appends to the queue while `batchDepth > 0` and writes through to ts-morph immediately otherwise -- so call sites like `addDeclaration` can use it unconditionally.

Only the outermost `flushSourceFileBatch` drains the queue; inner flushes just decrement the counter. That keeps nested begin/flush pairs composable: an inner caller can open its own batch without prematurely flushing what an outer caller queued. On drain, structures are grouped by `StructureKind` and dispatched through the per-kind bulk APIs (`addExportDeclarations` etc.), collapsing N re-parses into one per kind per file.

**Why the binder is not migrated.** The binder already collects imports per file in `this.imports` and issues a single `file.addImportDeclarations(arr)` per file at the end of `resolveReferences`. That is the same end-state the new framework produces (one re-parse per file per kind), so routing it through `enqueueStatement` would add per-call overhead without changing the flush cost, and would push binder-specific ordering policy (sort by module specifier) into a generic module. The framework targets the other pattern: many independent callers each writing one statement to a shared file, or a single caller that interleaves writes with AST reads.

## What changed

- **New** `packages/typespec-ts/src/framework/sourceFileBatch.ts`
  - `beginSourceFileBatch` / `flushSourceFileBatch` -- reference-counted scope; nested begin/flush pairs compose. The flush is wrapped in `try/finally` so a thrown writer never leaks queue state into the next emit.
  - `enqueueStatement(file, structure)` -- queues while a batch is open, writes immediately otherwise (single entry point for both modes).
  - `getEffectiveExportedNames` / `getQueuedExportNames` -- batch-aware views of the export set, so dedup against still-pending writes stays correct.
  - On flush, structures are grouped by `StructureKind` and dispatched through the per-kind bulk APIs (works around a ts-morph 23 `RangeParentHandler` bug in mixed-kind `addStatements`).

- `framework/declaration.ts` -- drops its own ad-hoc batch and routes `addDeclaration` through `enqueueStatement`. Single code path for batched/unbatched.

- `modular/emitModels.ts` -- wraps `emitTypes` main loop with `beginSourceFileBatch` / `flushSourceFileBatch` (try/finally).

- `modular/buildSubpathIndex.ts` -- wraps `buildSubpathIndexFile`; `partitionAndEmitExports` enqueues both the type-only and value exports.

- `modular/buildRootIndex.ts` -- wraps `buildRootIndex` and `buildSubClientIndexFile`; all four `addExportDeclaration` call sites use `enqueueStatement`. `getExistingExports` swapped to `getEffectiveExportedNames` (one-line body change); `getExportedDeclarations().keys()` dedup sets union with `getQueuedExportNames`.

All queues in the index-builder call sites are single-kind (`ExportDeclaration`), so emitted output is byte-identical. The `emitTypes` queue mixes kinds and groups by kind on flush -- same behaviour shipped in the prior emitTypes batching work.

## Performance results

Measured by `tsp compile` of the Network and Compute ARM specs, against the post-#4004 baseline.

### Network (`client.tsp`, total onEmit `7:27.367` -> `2:20.474`, **-69%**)

| phase | before | after | delta |
| --- | --- | --- | --- |
| emit models - emitTypes | 4:32.239 | 1.588s | -99.4% |
| build api subpath index | 58.813s | 50.054s | -15% |
| build classic subpath index | 17.015s | 0.452s | -97% |
| buildRootIndex | 27.414s | 0.372s | -98.6% |

### Compute (`client.tsp`, total onEmit `1:47.209` -> `0:56.485`, **-47%**)

| phase | before | after | delta |
| --- | --- | --- | --- |
| emit models - emitTypes | (slow) | 0.66s | huge |
| build api subpath index | 17.482s | 13.500s | -23% |
| build classic subpath index | 3.856s | 0.291s | -92% |
| buildRootIndex | 6.204s | 0.168s | -97% |


packages/typespec-test/test/ai/generated/typespec-ts/src/models/agents/index.ts
JialinHuang803, line 1: All the diff on the generated files are caused by the reorder of interface, functions, etc.

JialinHuang803:
Closing this PR and will proceed with https://github.com/Azure/typespec-azure/pull/4613

